PR #7, opened 2018-08-16
Updated update.sh - by joanbono from iceman1001/proxmark3

TomHarkness:
The update.sh is not my code, just had the idea. Authored by joanbono

install.sh is working tested under OS X High Sierra. I'll do a PR for iceman fork too.


RfidResearchGroup:
Nice one.  
even for install.sh...    
Now just an entry in changelog.md  and what about proxmark3.sh ?

TomHarkness:
Working on other files too. I'll push them as soon as I've had time for a bit more testing :)

PR #11, opened 2018-08-16
Add libcanberra-gtk-module to install.sh

reidrankin:
Under Ubuntu 18.04, the proxmark3 client throws an error that the canberra-gtk-module is missing; this makes that message go away.

(I'm not sure if it's a problem on Homebrew as well, because I don't have a Mac for testing.)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@ set -x
 function installProxmark_Linux {
   # install dependencies for Proxmark3 source code.
   sudo apt-get install p7zip git build-essential libreadline5 libreadline-dev gcc-arm-none-eabi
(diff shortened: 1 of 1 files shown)


RfidResearchGroup:
This feels like specific version related problem...
so ppl who use 14.04, 16,04, 17.04  (or which ever version before 18.04)  might download something unneeded.  
Is there a way to detect os-version (uname?) and a a row with  git install libcanberra instead?

TomHarkness:
No issues without this on Mac.

RfidResearchGroup:
Well, if it doesn't break anything I can just merge it :)
Thanks!

TomHarkness:
Yes same thing as for udev. I'll do a PR latter.

Issue #6, opened 2018-08-15
Bug and partial Fix - update.sh for OS X users


TomHarkness:
**Update.sh does not find pm3 ion osx due to the "?" switch not working**

Along with the above we have no symlink in place and I think it may be good to put one in.

**Original:**

function wait4proxmark {
	echo >&2 "Waiting for Proxmark to appear..."
	while [ ! -c /dev/ttyACM? -a ! -L /dev/pm3-? ]; do
		sleep .1
	done
	local PM3=`ls -1 /dev/pm3-? /dev/ttyACM? 2>/dev/null | head -1`
	echo >&2 -e "Found proxmark on ${PM3}\n"
	echo $PM3
}

**The following works fine but line 5 also throws a missing ']' error.** Not sure about linux yet. Maybe it could be formatted better?

function wait4proxmark {
	echo >&2 "Waiting for Proxmark to appear..."
	while [ ! -c /dev/ttyACM? || /dev/tty.usbmodem* -a ! -L /dev/pm3-? ||  /dev/pm3* ]; do
		sleep .1
	done
	local PM3=`ls -1 /dev/pm3-? /dev/ttyACM? dev/tty.usbmodem* /dev/pm3*  2>/dev/null | head -1`
	echo >&2 -e "Found proxmark on ${PM3}\n"
	echo $PM3
}

Also I think somewhere it would be good to check OS, detect OS X and put some symlinks in place. Working code:

 	if [[ $OSTYPE == darwin* ]]; then
		echo "I am a Mac - creating symlink to /dev/pm3"
			ln -s $PM3 /dev/pm3
		;fi

TomHarkness:
Seems like the best structure to make this cross compatible with Linux and OS X without having issues with conditional OR is something like:

	while [ ! -c /dev/ttyACM? || /dev/tty.usbmodem* ] -a [ ! -L  /dev/pm3-? || /dev/pm3* ] ; do


This works but still throws missing '[' error.

RfidResearchGroup:
@joanbono   just made solution for one script over at iceman fork.   Its been merged.

No need to over complicate things.   I would also suggest implementing the same dual idea into the proxmark.sh  script since its uses the look up comport aswell inside.

and don't forget to add a entry in changelog.md   for eternal glory.

@TomHarkness  said something about making a symbolic link  to /dev/pm3 as it is done for linux ppl with udev rules.  (sudo make udev)    making things a bit easier for future support questions about my comport is not showing.

[...]

PR #13, opened 2018-08-17
Modified install.sh

joanbono:
Added Ubuntu 18.04 detection, which needs `libcanberra-gtk-module` to compile the code as mentioned in #6  :

```bash
  # Install libcanberragtk in Ubuntu 18.04
  if [[ $(cat /etc/issue | awk '{print $2}') = *"18.04"* ]]; then 
    apt-get install libcanberra-gtk-module
  fi
```

Also, improved compilation times using max number of processors available on each machine:

+ Linux: `make all -j$(nproc)`
+ macOS: `make -j$(sysctl -n hw.physicalcpu)`


RfidResearchGroup:
Great.   Now,  is someone able to test it out??

PR #15, opened 2018-08-17
Add some mifare keys. 

TomHarkness:


RfidResearchGroup:
Thats a lot of commits for two keys...    Would you mind cleaning that up?

TomHarkness:
Sorry about that - I made a mess! Still a bit new to using Git in this way :P

PR #17, opened 2018-08-23
Populate emulator keys from `hf mf fchk`

gkelly:
Just like `hf mf chk`, when the 't' option is provided make fchk put the
found keys into emulator memory.


gkelly:
This is already in the iceman fork (https://github.com/iceman1001/proxmark3/pull/236#issuecomment-415307189).

RfidResearchGroup:
Excellent!

PR #24, opened 2018-09-05
lfops.c CmdEM410xdemod empty TAG ID fix.

Defensor7:
CmdEM410xdemod returns "EM TAG ID: 0000000000 - (00000_000_00000000)" without card because considers a negative numbers as true.

https://stackoverflow.com/questions/14646718/if-statement-integer


RfidResearchGroup:
Great catch!

RfidResearchGroup:
I also pushed another place which needs this check.

PR #27, opened 2018-09-07
Changes in ListenReaderField and some debug messages

Defensor7:
Added to appmain.c ListenReaderField:
`while(!usb_poll_validate_length()) { ... } DbpString("Stopped"); break;`

Added debug messages to let client understand current status of operations.


RfidResearchGroup:
Why two loops ?

Issue #31, opened 2018-09-15
Compile Errors with latest pull


TomHarkness:
lfops.c: In function 'setT55xxConfig':
lfops.c:83:5: error: this 'if' clause does not guard... [-Werror=misleading-indentation]
     if (!FlashInit())
     ^~
lfops.c:86:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
  Flash_CheckBusy(BUSY_TIMEOUT);
  ^~~~~~~~~~~~~~~
lfops.c: In function 'loadT55xxConfig':
lfops.c:103:5: error: this 'if' clause does not guard... [-Werror=misleading-indentation]
     if (!FlashInit())
     ^~
lfops.c:106:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
  Flash_CheckBusy(BUSY_TIMEOUT);
  ^~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [obj/lfops.o] Error 1
make: *** [armsrc/all] Error 2

TomHarkness:
Fixed with PR #32

Issue #30, opened 2018-09-14
No USB connection after update


Peterthegreat:
Hi everyone,

I was running my proxmark3 rdv 4 for the second time and decided to follow the instructions for this repo, so I proceeded with the image upgrade.

```
[ 6552.518021] cdc_acm 1-3:1.0: ttyACM1: USB ACM device
[ 6570.767352] usb 1-3: USB disconnect, device number 77
[ 6573.085096] usb 1-3: new full-speed USB device number 78 using xhci_hcd
[ 6578.237065] usb 1-3: device descriptor read/64, error -110
[ 6583.493272] usb 1-3: New USB device found, idVendor=9ac4, idProduct=4b8f
[ 6583.493274] usb 1-3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[ 6583.493276] usb 1-3: Product: PM3 Device
[ 6583.493277] usb 1-3: Manufacturer: proxmark.org
[ 6583.493278] usb 1-3: SerialNumber: 888
[ 6583.493814] cdc_acm 1-3:1.0: ttyACM1: USB ACM device
[ 6689.121873] usb 1-3: USB disconnect, device number 78
[ 6691.348203] usb 1-3: new full-speed USB device number 79 using xhci_hcd
[ 6696.712160] usb 1-3: new full-speed USB device number 80 using xhci_hcd
[ 6702.020367] usb 1-3: New USB device found, idVendor=9ac4, idProduct=4b8f
[ 6702.020370] usb 1-3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[ 6702.020371] usb 1-3: Product: PM3 Device
[ 6702.020372] usb 1-3: Manufacturer: proxmark.org
[ 6702.020373] usb 1-3: SerialNumber: 888
[ 6702.020907] cdc_acm 1-3:1.0: ttyACM1: USB ACM device
root@kali:~/proxmark3RDV4# client/flasher /dev/ttyACM1 -b bootrom/obj/bootrom.elf armsrc/obj/fullimage.elf
Loading ELF file bootrom/obj/bootrom.elf
Loading usable ELF segments:
0: V 0x00100000 P 0x00100000 (0x00000200->0x00000200) [R X] @0x94
1: V 0x00200000 P 0x00100200 (0x00000d60->0x00000d60) [R X] @0x298

Loading ELF file armsrc/obj/fullimage.elf
Loading usable ELF segments:
[...]

PR #32, opened 2018-09-15
Fix syntax

TomHarkness:
Added missing braces. Code now compiles. Should help issues #31 & #30


RfidResearchGroup:
Funny thing,  on mingw gcc 6.4  it doesn't do a warning,  but on gcc7.1  it does.
Thanks for the fix!

PR #33, opened 2018-10-11
Auth14443-4

merlokk:
here :

1. move argtable from pm3 repo
2. adapt revevg to argtable
3. add https://github.com/merlokk/proxmark3/commit/7dadcc959fb7009b6e8bbde4a644aa2f7f1b7a98

@iceman1001 look please into code. maybe it needs to change something

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -110,6 +110,8 @@ CMDSRCS =	crapto1/crapto1.c \
 			polarssl/sha1.c \
 			polarssl/sha256.c \
 			polarssl/base64.c \
+			cliparser/argtable3.c\
+			cliparser/cliparser.c\
 			loclass/cipher.c \
 			loclass/cipherutils.c \
 			loclass/ikeys.c \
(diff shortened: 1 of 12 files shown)


RfidResearchGroup:
Cool.
In the long run I see no use of polluting  **hf  mf**  with Mifare plus stuff.  
I started  **hf mfdes** long time ago in order to have Desfire support there.   It relies also on ISO7816-4..

Where do you suggest we put it all in?    a new  ISO7618 place?  Since RDV40 isn't contactless only..  The smart card commands will benifit also from this.

merlokk:
auth differs in plus and desfire as i see. auth4 because it works in "classic" mode of plus.
I do a new "plus" series (`hf mfp`) but....   by now there is too small info about reading/writing Plus cards
but there will be some commands from lua script)

as for desfire - it have huge hill of commands and modes and maybe if it needs to implement it - it needs to implement some small parts of commands...
there are 3 types of desfire and 4 command modes (types + ISO)
lets think about desfire later....   because all this commands completely useless without datasheet.... needs to know too much things.... and if someone have datasheet - he have support and have a reader...

PR #36, opened 2018-10-21
Added full key set of russian Troika card and removed duplicates.

zhovner:
also added missed commas. Is comma needed in the last key?


RfidResearchGroup:
Actually,  I don't think any of the commas is every used in the code.  It only read 12 hexadecimal from start of a new line.   The rest is ignored

RfidResearchGroup:
Great! Thanks!

PR #39, opened 2018-10-22
Added new keys from MifareClassicTool project. Remove two dups

zhovner:
I found that MifareClassicTool have some uniq keys and added it here.  Also remove two duplicates.


RfidResearchGroup:
Thanks!
And here I thought I had them all :)

PR #48, opened 2018-11-07
include Mbedtls

merlokk:
crypto tests passed
emv works
aes works
sha1, sha256 works

should be ok.


merlokk:
maybe needs to check `cmdflashmem.h`
and needs to get rid of `polarssl/des.c` but in next PR

merlokk:
as for `cmdflashmem.h`
it makes a signature and send it to arm. then arm dont write it to flash(
so )
good part: crypto works)
bad part: arm side flash write dont work(

RfidResearchGroup:
no,  that signature is reserved for RRG,   its our detection of a geniue RDV40.  So we don't encourage the writing of own signatures.   hence not enabled.

Did you pull in the whole crypto-lib?   Nice!

merlokk:
here is mbedtls with some parts is cut. but. they easily can be added with add a files and uncomment lines in config.h
I have added all the parts that are needed for our area (as i see). 
i tried to add less code, but here is cmac and error modules that includes everything)
p.s. ecdsa and asn.1 certificates added

PR #52, opened 2018-11-11
Update Legic.lua script

CumpsD:
Initial PR to update `legic.lua` to work with the new commands. Feedback appreciated!

Hacked this up last night since some of the PM commands don't exist anymore, and the PM client has some extra checks in `legic write` now.

### DIFF
--- a/client/scripts/legic.lua
+++ b/client/scripts/legic.lua
@@ -513,6 +513,14 @@ function readFromPM3()
     return tag
 end
 
+function padString(str)
+  if (str:len() == 1) then
+    return '0'..str
+  end
+
+  return str
+end
+
 ---
 -- write virtual Tag to real Tag
 function writeToTag(tag)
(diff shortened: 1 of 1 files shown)


RfidResearchGroup:
Great!  

The only thing I miss with this script is the ability to detect if its a binary file / eml file.   
I actually want the dump to save both formats,  and the restore should also be able to do both.

RfidResearchGroup:
And the mosci-patch,   there should be a param in the usbcmd to enable writing those bytes...

CumpsD:
I've just updated it to not crash ;) It was still doing `load` while that command didn't exist.

As for the mosci-patch, I looked around, it seems it already made it in here: https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/legicrf.c#L471 (This lower limit is 4, an it used to be 6)

The client has these bits:
 * https://github.com/RfidResearchGroup/proxmark3/blob/master/client/cmdhflegic.c#L600-L603 Blocks 0-4 in any case
 * https://github.com/RfidResearchGroup/proxmark3/blob/master/client/cmdhflegic.c#L610-L621 A warning for the DCF

I have to test my newly written card tommorow, bit it wrote it without errors. The lua script properly writes offset 6 first and then 5, which is the "magic" in the legic DCF.

One gotcha which bit me yesterday, DCF can only be lowered :) New cards are fine `ff ff`, but if you mistakenly write some low values to it (`ea 01`), changes are high you bricked it, since it's unlikely your new DCF will be lower then `01` :)

RfidResearchGroup:
Yes,   the DCF acts like a onetime counter, only deceasing.   Not sure if it only decrease  ie: 10,9,8,7..
or if it only allows flipping bits to zero like the onetimepad in UL/NTAG card..

RfidResearchGroup:
Would you mind adapting the script to save  both BIN and EML?   It currently saves BIN.

PR #53, opened 2018-11-14
get rid of polarssl's Des

merlokk:


RfidResearchGroup:
Shall I merge?

merlokk:
yes. it works)

PR #67, opened 2018-12-06
Emv commands works with smartcard interface

merlokk:
`emv scan` not works with sc at all. It needs to rebuild some parts.
`emv exec` works, but for me only up to GPO
the rest commands works (at least put data to interface and receive data)


merlokk:
OK. It works.
But not so stable as needs.

RfidResearchGroup:
Excellent PR!

PR #70, opened 2018-12-14
Add script to test for MADs

mazodude:
I have added a lua script to test for MADs (Mifare Application Directory) v1 (for now)
The plan is to update this later for MADs v2 and v3.

### DIFF
--- a/client/scripts/mifare_ad.lua
+++ b/client/scripts/mifare_ad.lua
@@ -1,0 +1,186 @@
+
+-- Ability to read what card is there
+local getopt = require('getopt')
+local cmds = require('commands')
+local taglib = require('taglib')
+
+local desc = 
+[[This script will automatically check Mifare cards for MADs
+(Mifare Application Directory)
+
+Arguments:
+	-d 				debug logging on
+	-h 				this help
+
+]]
+local example = "script run xxx"
+local author = "Mazodude"
+---
+-- PrintAndLog
+function prlog(...)
+	print(...)
+end
+--- 
+-- This is only meant to be used when errors occur
+function oops(err)
+	prlog("ERROR: ",err)
+	return nil,err
+end
+
+--- 
+-- Usage help
+function help()
+	prlog(desc)
+	prlog("Example usage")
+	prlog(example)
+end
+
+function debug(...)
+	if DEBUG then 
+		prlog("debug:", ...)
+	end
+end
+
+
+local function show(data)
+	if DEBUG then
+	    local formatString = ("H%d"):format(string.len(data))
+	    local _,hexdata = bin.unpack(formatString, data)
+	    debug("Hexdata" , hexdata)
+	end
+end
+--- Fire up a connection with a tag, return uid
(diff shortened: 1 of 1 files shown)


client/scripts/mifare_ad.lua
```diff
@@ -0,0 +1,186 @@
...
+	if DEBUG then
+	    local formatString = ("H%d"):format(string.len(data))
+	    local _,hexdata = bin.unpack(formatString, data)
+	    debug("Hexdata" , hexdata)
+	end
+end
+--- Fire up a connection with a tag, return uid
+-- @return UID if successfull
+-- @return nil, errormessage if unsuccessfull
+local function open()
```
iceman1001: How about using the lualibs functions;

	-- GET TAG UID
	tag, err = lib14a.read(false, true)
	if not tag then return oops(err) end
core.clearCommandBuffer()

client/scripts/mifare_ad.lua
```diff
@@ -0,0 +1,186 @@
...
+local example = "script run xxx"
+local author = "Mazodude"
+---
+-- PrintAndLog
+function prlog(...)
+	print(...)
+end
+--- 
+-- This is only meant to be used when errors occur
+function oops(err)
```
iceman1001: add "local" to all functions, please.

client/scripts/mifare_ad.lua
```diff
@@ -0,0 +1,186 @@
...
+Arguments:
+	-d 				debug logging on
+	-h 				this help
+
+]]
+local example = "script run xxx"
+local author = "Mazodude"
+---
+-- PrintAndLog
+function prlog(...)
```
iceman1001: prlog vs print??   just don't.

iceman1001:
Good start!  Just some adjustments

mazodude:
I was just following the code from ndef_dump.lua. I will go ahead and make the changes.
It's my first time ever using lua

iceman1001:
I know,  The scripts still has some old quirks in them. 
lua is much simpler than you think,  but the interaction with pm3 client can be scary.

client/scripts/mifare_ad.lua
iceman1001: you will need 
local lib14a = require('read14a')

mazodude:
@iceman1001 Thank for your input. The script now reflects your suggestions

client/scripts/mifare_ad.lua
```diff
@@ -168,7 +167,7 @@ local function main( args)
 
 	local uid = open()
 	if (uid==nil) then oops("No card present") return end
-	print(("UID: %s"):format(uid))
+	print(("UID: %s"):format(uid.uid))
```
iceman1001: I here I would suggest you move the code from "open()"  instead...    Since you basically are doing the same thing.

iceman1001:
Much better,   just one more thing.

mazodude:
Change has been made. Thanks

PR #73, opened 2018-12-19
Fix detection of MADs so that is checks the correct byte

mazodude:
Should also be able to check for v2


iceman1001:
I suggest you add your stuff this this script instead, which was there before

https://github.com/RfidResearchGroup/proxmark3/blob/master/client/scripts/mifare_access.lua

mazodude:
Ok no worries. Should I close this one then?

iceman1001:
yes,  that would be a good way.

PR #87, opened 2019-02-04
Add pwd and counters to mfu dump format

bogiton:


client/cmdhfmfu.h
```diff
@@ -21,7 +21,8 @@ typedef struct {
 	uint8_t	pack[2];
 	uint8_t tbo1[1];
 	uint8_t	signature[32];	
-	//uint8_t	counter[3];
+	uint8_t	counter[3][3];
```
iceman1001: Isn't counters 24bits?  uint32_t[3] ?
Or keep it even simpler,   Counter1,  Counter2, Counter3  ?

I guess this messes up old dumps though when reading them.   One of the main reasons I didn't fiddle with this part ...

client/cmdhfmfu.h
```diff
@@ -21,7 +21,8 @@ typedef struct {
...
+	uint8_t pwd[4];
```
iceman1001: And authentication key could be  4bytes or 16 (triple des  ULC)

client/cmdhfmfu.h
bogiton: Yes, 3 counters of 3 bytes each. Thus the double array.

client/cmdhfmfu.h
bogiton: You are right. I was only thinking of ev1...

bogiton:
I actually changed the mfu_dump_t struct because of the json output. I didn't know that other locations are depending on that.
Maybe I should close this one and make separate ones.

PR #93, opened 2019-02-15
Use Variable for MIFARE_XXX_MAXSECTOR

vratiskol:
- Use Variable for MIFARE_XXX_MAXSECTOR
- Modify type uint16_t for NumOfBlocks => 256


iceman1001:
I notice you didn't pull before you did this PR.  The uint16 was already fixed.

PR #107, opened 2019-03-04
Mad read

merlokk:


iceman1001:
cool, but why add a parameter for key, when ndef standard states it must be one key?
if optional key, I suggest a other example key than original default key.  This way users don't confuse themselfs that they need to give default key when running the command.

merlokk:
I have added it for not standard cards and for reading with key B.
maybe ill add some examples there

iceman1001:
yup,  maybe some helptext saying why and the examples shouldnt point to default key which we should try to use regardless

PR #108, opened 2019-03-04
Add missing header

doegox:
Latest merge (hf mf* mad) didn't compile due to a missing header.


iceman1001:
you are fast, @doegox

PR #109, opened 2019-03-04
Mifare mad - do not exit if checksum is shitty

sh7d:
Well, in my not-so-big town, public transport card have mad sector with broken checksum (or unlikely proxmark client have broken crc implementation, but for now i'm unable to determine that).
Beside of that, sector is perfectly fine, and have information that are consistent with the actual usage of memory in the card, so i just make some small changes in proxmark client code to read it anyway

My proposal is - just write warning about bad crc and don't stop mad decoding procedure

### DIFF
--- a/client/mifare/mad.c
+++ b/client/mifare/mad.c
@@ -112,14 +112,14 @@ int madCRCCheck(uint8_t *sector, bool verbose, int MADver) {
 		uint8_t crc = CRC8Mad(&sector[16 + 1], 15 + 16);
 		if (crc != sector[16]) {
 			if (verbose)
-				PrintAndLogEx(ERR, "Wrong MAD%d CRC. Calculated: 0x%02x, from card: 0x%02x", MADver, crc, sector[16]);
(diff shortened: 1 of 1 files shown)


merlokk:
there is an app from nxp
https://play.google.com/store/apps/details?id=com.nxp.taginfolite
just check there.
btw) there is an error in MAD datasheet. so some implementations have wrong crc calc....

sh7d:
Unfortunately my phone don't support mifare cards :(

Anyway, speaking stricte of functionality, i suppose if sector is broken proxmark should try to extract as much information as possible

iceman1001:
I agree, if failed crc,  print that crc failed and go on.

merlokk:
maybe add a command line parameter to suppress this error?
like  -s / -skip

iceman1001:
that would not follow all other situations when crc sums fails.  
I rather have a warning message,  and continue operations than break/exit

merlokk:
ok

PR #110, opened 2019-03-04
Delete MAD script as merlokk has implemented it

mazodude:


iceman1001:
it doesnt hurt that this script exist still after its been incoorporated.   It can still inspire users to learn how to use proxmark lua scripting.   orr?

mazodude:
I'll leave that up to you @iceman1001. I'm not sure that I will get time to finish this script and merge it into the mifare_access.lua script

PR #113, opened 2019-03-05
Add missing i2c init in sc upgrade

doegox:
Running sc upgrade was failing if we didn't run other sc command first because some init was missing


iceman1001:
great catch!

PR #114, opened 2019-03-05
Look for aidlist.json in the client dir

doegox:
when proxmark3 client is called outside its dir (e.g. when using proxmark3.sh for upper dir), "sc brute" didn  find aidlist.json


iceman1001:
better solution indeed!

PR #115, opened 2019-03-05
cmds history: don't push to history if last command is replayed

doegox:
we don't like long history full of repeated commands :)


iceman1001:
This one is a good one!

PR #142, opened 2019-03-26
Start key file for Mifare Ultralight

mazodude:


iceman1001:
now,  will there be a **hf mfu chk**  command with this file?   Or should I just merge it ?

mazodude:
It is only the key. No chk command added.
Maybe next time

PR #143, opened 2019-03-27
Git attribute

vratiskol:
- Force LF for .c and .h
- Ignore JSON file


iceman1001:
@doegox   those force LF,  it doesn't inflict anything on the new idea of **make style** command?

doegox:
It seems to do the job, thanks.
I'll remove it from "make style"

PR #145, opened 2019-04-04
'hf mf sim' - bugs fix, refactoring

McEloff:
Method Mifare1ksim().
1. Bug. After succeed authenticate to sector, traffic is encrypted. But trace log contains decrypted data and 'trace list mf' decrypts already decrypted sequence twice. Fixed: called mf_crypto1_decryptEx() instead of mf_crypto1_decrypt() to use separate buffers. Calls of LogTrace() and EmSendXXX() saves source encrypted sequence.
2. Bug for Android smartphones. Nested authentication preparing 'nT' must calculate parity bits by non-encrypted data like 'aT'-response, see commit 459835b5. Fixed: using new method mf_crypto1_encryptEx() with init-keystream instead of crypto1_word(..., cuid ^ nonce, ...).
3. Case block 'MFEMUL_WORK' can operate on non-encrypted command sequence. So 'HALT', 'RATS' or unknown commands may be received after 'SELECT' and before 'AUTH'. Fixed: using flag 'encrypted_data' in EmSend4bit() to check what  answer need - encrypted or not.
4. Added some integrity checks


armsrc/mifareutil.h
```diff
@@ -57,6 +57,10 @@
 
 #define cardSTATE_TO_IDLE() cardSTATE = MFEMUL_IDLE; LED_B_OFF(); LED_C_OFF();
 
+#ifndef MifareBlockToSector
+#define MifareBlockToSector(block) (block < 128 ? block / 4 : (block - 128) / 16 + 32)
```
merlokk: https://github.com/RfidResearchGroup/proxmark3/blob/master/client/mifare/mifare4.h#L64

iceman1001:
Nice! However I am waiting for @vratiskol  to push his rewamped **hf mf sim** changes.
After that this fixes should be adapted to it.

armsrc/mifareutil.h
McEloff: 'mifare4.h' is a part of client, Not good idea to include this part into 'armsrc'

vratiskol:
@McEloff Nice work, happy to know someone else needs Mifare sim ;)
@iceman1001 I will pull request [my work](https://github.com/vratiskol/proxmark3/tree/hf_mf_sim) really soon (even if some instability persists).
Some enhancements have already be include as I merge official proxmark and RRG version of hf mf sim (CRC for ex.)

iceman1001:
yup,  go for your PR,   then  this one can be adapted on to it.  And from there we solve the rest of the problems :)  Go Go Go!

armsrc/mifareutil.h
merlokk: ok

doegox:
@vratiskol PR merged, now you can adapt yours @McEloff , thanks :)

iceman1001:
@McEloff   need help?

McEloff:
I think, new release of Mifare1ksim() method is not better than previous. Anyone tests that? For example, in selection procedures used EmSendPrecompiledCmd() calls, that sends answer as raw-data without precalculated parity. I see in trace log incorrected parity bits within responses 'uid+bcc' on tag select. My Android smartphone could not perform select procedure completed.
Also, 'NR AR ATTACK' is limited up to 8 found keys. At end of simulation whole array of 'ar_nr_resp' unconditionally sends back to client.  But client in method CmdHF14AMf1kSim() waits only one element of array to interactive attack.

I can apply minimal changes of my patch and push it. Or else need more time to fix all of those bugs.

iceman1001:
with regards to your fixes, no the current mf sim is not better.   However it added support for 1k,2k,4k size.
Your pr addresses other weaknesses but both of you were changing the same file.  Always a mess to merge that.

vratiskol:
@McEloff I try to add some functionnality from both Proxmark repo and RRG repo (like CRC, precalculate reply) and handling of different Mifare size card.
About test and card selection, I will take a look at it. 
I did loads of tests and read with mf 1k dump and MFC.
All was ok but not really stable. 
As you are working on this part too, we decide to comit my work to take avantage the creation of mifaresim.c file.

Maybe the fact I put all debug message off is influencing Mifare selection (time to reply)
Sorry If the situation seems really confusing to you.
Hope we could work all on the subject to improve this functionality.

iceman1001:
aha,  now I get it.  This is a part ..   *doh*  Sorry,  I will merge it!
Looking forward to next part.

McEloff:
I commit some little changes to new release of Mifare1ksim() method. Unfortunately, current method is very unstable for further use.

iceman1001:
Great,  lets get your fixes merged!

McEloff:
Foreign others commits have many... many... bugs! For example, now all incoming reader sequences are doubled in trace log, because  in method EmGetCmd() of 'iso14443a.c' module was added EmLogTraceReader() call. Why?

iceman1001:
just clean it out.

iceman1001:
you are trying to add commits to this closed PR.

vratiskol:
> EmLogTraceReader 
This function is coming from official Proxmark repo as I added some extra function from it.
I don't remember which function used it and why, anyway, feel free to clean it out if it's not needed :)

McEloff:
Thanks for official Proxmark repo! Now not working both simulators '14a' (ultralight/NTAG) and 'mf' (Classic)

PR #146, opened 2019-04-05
fix: compile format warning

gtpy:
warning: format ‘%u’ expects argument of type ‘unsigned int’, but argument 2 has type ‘size_t {aka long unsigned int}’ [-Wformat=]

### DIFF
--- a/client/scripting.c
+++ b/client/scripting.c
@@ -797,7 +797,7 @@ static int l_T55xx_detect(lua_State *L) {
             if (size != 1)  return returnToLuaWithError(L, "Wrong size of useGB, got %d , expected 1", (int) size);
             sscanf(p_gb, "%u", &gb);
             useGB = ( gb ) ? true : false;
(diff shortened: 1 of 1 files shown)


iceman1001:
yeah... that one..   Have you tested that on mixed dev environments like linux / mingw / osx?   Different environments defines it differently.

gtpy:
Only tested on Linux.

doegox:
Read https://stackoverflow.com/questions/174612/cross-platform-format-string-for-variables-of-type-size-t and make your mind :)

I guess the safest option is to use `"%zu"` as we are already compiling the client with `-std=C99`, but maybe test it on Windows first to be sure...

iceman1001:
yeah,   I agree.  Use` %zu`

doegox:
Done in 84753b91c041643b02dbba15948982d0df38aeed.

gtpy:
@doegox @iceman1001 yes you are right to use `%zu` thanks :) for the hint

PR #147, opened 2019-04-05
[hf mf sim] Rework - extract to mifaresim.c 

vratiskol:
- extract function from iso14443a.c to mifaresim.c 
- Add **hf mf sim t** on client to select card type (mifare mini, 1k, 2k, 4k)

/!\ Code needs to be reviewed / enhanced to be more stable.


armsrc/mifareutil.c
```diff
@@ -592,14 +592,27 @@ void emlClearMem(void) {
     memset(emCARD, 0, CARD_MEMORY_SIZE);
 
     // fill sectors trailer data
-    for (uint16_t b = 3; b < 256; ((b < 127) ? (b += 4) : (b += 16)))
+    for (uint16_t b = 3; b <= MIFARE_4K_MAXBLOCK; ((b <= MIFARE_2K_MAXBLOCK) ? (b += 4) : (b += 16)))
```
doegox: I think this goes now out of the buffer when b==MIFARE_4K_MAXBLOCK

doegox: hmm no it's fine as we're jumping over these values and b will never be exactly == MIFARE_4K_MAXBLOCK, still the condition looks a bit weird ;)

doegox:
merged manually to get rid of the client/cmddata.c regression

iceman1001:
Thanks!

PR #150, opened 2019-04-08
[Hf mf sim] Add comments

vratiskol:
- Add comment about emulation problem <=> printing debug message
- Change log level


doegox:
ack, thanks

iceman1001:
Feels that those large repeating switches could be in a function instead.

PR #153, opened 2019-04-09
'hf mf sim' + 'hf 14a sim' now back to stable

McEloff:
Deleted unused methods, fix bugs


vratiskol:
I have tested the mifare part and everything is OK with my android phone and MCT. 
@McEloff Thanks for you work

McEloff:
Not at all :)
Print 'trace list mf' after simulation classic tag. Are there in log requests 'HALT' before 'WUPA' ? I do not understand, why 'HALT' not logs in this release of method. Previous release of Mifare1ksim logs HALT requests successful. This is the bug still :)

PR #154, opened 2019-04-10
'hf mf sim', 'hf 14a sim' online and offline ar/nr nonces collection

McEloff:
1) Added interactive mode for 'hf 14a sim'
2) Optimized online (interactive mode) and offline process of ar/nr nonces collections for 'hf 14a sim' and 'hf mf sim'
3) Fix AR/NR nonces collection algorithm  in previously merged Mifare1ksim() method
4) Performs postprocess key attack after tag simulation finished
5) Extends offline array up to 32 ar/nr collection per tag simulation

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -822,9 +822,6 @@ bool prepare_allocated_tag_modulation(tag_response_info_t *response_info, uint8_
 // 'hf 14a sim'
 //-----------------------------------------------------------------------------
 void SimulateIso14443aTag(int tagType, int flags, uint8_t *data) {
-
-#define ATTACK_KEY_COUNT 8 // keep same as define in cmdhfmf.c -> readerAttack()
-
     uint8_t sak = 0;
     uint32_t cuid = 0;
     uint32_t nonce = 0;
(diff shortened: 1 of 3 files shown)


armsrc/iso14443a.c
```diff
@@ -1390,7 +1389,31 @@ void SimulateIso14443aTag(int tagType, int flags, uint8_t *data) {
...
+                     ar_nr_collected[i].ar,    //AR1
+                     ar_nr_collected[i].nonce2,//NT2
+                     ar_nr_collected[i].nr2,   //NR2
+                     ar_nr_collected[i].ar2    //AR2
+                    );
+        }
+    }
+    // interrupt interactive client
+    if ((flags & FLAG_INTERACTIVE) == FLAG_INTERACTIVE)
+        cmd_send(CMD_ACK, 1, 0, 0, 0, 0);
```
iceman1001: I removed this printing years ago,   I see no reason to bring it back. 
hence the need for flag_interactive is also useless.

iceman1001:
A interesting approach would to drop all those prints,  
and save those nonces in the flashmem,   and download at a given command to run the keyrecovery.

McEloff:
You are right! Flashmem is better for this purpose. My PM V3 without extended flash memory, then I can not release given features now. Let close pull request.

iceman1001:
@McEloff  I thought you had a RDV4.   My bad.

pwpiwi:
No need to use flashmem. There is RAM on all revisions of PM3.

McEloff:
Only for RDV4 platform developing compiler has definition 'WITH_FLASH'. This condition provides access to library for reading and writing individual data variables, structs and arrays from and to various locations. This condition cannot be applied to developing RDV3 and below platforms (V3 devices crashes on startup). There was an idea to use flashmem as non-volatile storage when device working offline instead of printing results back to client after reconnecting.

iceman1001:
@McEloff  correct,     use the hardware advantages of rdv4.

pwpiwi:
Or support all platforms. Flash-Memory is on all devices.

iceman1001:
Wrong repo for that.  This is the dedicated rdv4 repo.  There is too many different  flashmem sizes for the MCU to be useful.  Feel free to add that offical repo.

PR #155, opened 2019-04-11
Add 36bit HID format, extend calcWiegand() to include oem bits

davidbeauchamp:
As requested by @iceman1001 here https://github.com/iceman1001/proxmark3/pull/285, notes below copied for convenience.

I came across a bag full of HID tags using this format that I couldn't find documented anywhere. After some poking and prodding at the bits I worked out the format and figured I would contribute the addition. I added support to the demoudulator, tested reading, cloning, wiegand output, and simulation. I did not add support for this format to the brute forcer as it would require refactoring to support the OEM segment, and just hard coded a 0 into the OEM parameter added to `clacWiegand` inside the brute forcing code.

This format is backwards in that the parity logic is swapped, so I duplicated the existing function to do this. Plenty of head scratching to finally figure that out as none of the other HID formats seem to be encoded this way.

Format Details:
```
FC     1  - 16  - 16 bit
cardno 17 - 33  - 16 bit
oem    34 - 35  -  2 bit
Odd  Parity  0th bit  1-17
Even Parity 36th bit 18-35
```
Some example bitstreams, card number is printed on tag along with sales order number from original purchase:
```
1 1011000000010000 0101101110100000 00 1 FC = 45072, Card = 23456, OEM = 0
1 1011000000010000 0101101111000001 00 0 FC = 45072, Card = 23489, OEM = 0
1 1011000000010000 0101101111100111 00 1 FC = 45072, Card = 23527, OEM = 0
1 1011000000010000 0110000100000100 00 0 FC = 45072, Card = 24836, OEM = 0
1 1011000000010000 0110000000110001 00 1 FC = 45072, Card = 24625, OEM = 0
```

### DIFF
--- a/client/cmdlfhid.c
+++ b/client/cmdlfhid.c
@@ -108,7 +108,7 @@ static bool sendTry(uint8_t fmtlen, uint32_t fc, uint32_t cn, uint32_t delay, ui
     if (verbose)
         PrintAndLogEx(INFO, "Trying FC: %u; CN: %u", fc, cn);
 
-    calcWiegand(fmtlen, fc, cn, bits);
+    calcWiegand(fmtlen, fc, cn, bits, 0);
 
     uint64_t arg1 = bytebits_to_byte(bits, 32);
     uint64_t arg2 = bytebits_to_byte(bits + 32, 32);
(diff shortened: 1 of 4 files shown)


iceman1001:
Thanks bro!

iceman1001:
I also suggest you add one row in the changelog file,  :)

iceman1001:
Btw,  I like your codestyle.  its precise and clear.  I have an idea,  the offical repo got a refactoring of lf hid / paradoc clone a few months ago (january) ,  I never liked the refactoring,  so I never added it here. 
The function would be great to have but not like its implemented.

However I get the feeling you may be the right man for the task.  Are you up to it?

davidbeauchamp:
I am still getting accustomed to the code base but I will certainly give it a shot. I will ping you on the forum if I come up with questions --

iceman1001:
yup,  send an email also :)

PR #158, opened 2019-04-15
Jablotron ID length fix

merlokk:


merlokk:
it because bytebits_to_byte compiles in arm...

iceman1001:
Good catch!
it was because bytebits_to_byte returns a uint32_t  :)

PR #161, opened 2019-04-16
appveyor fix

merlokk:


doegox:
BTW @merlokk you can do more offline tests now, no need for fake ``comx`` anymore.
```
./proxmark3 -c 'emv test'
```

PR #163, opened 2019-04-16
'hf mf sim' bugs fix, RATS support, etc

McEloff:
1. Bug. Tag type missing if 'u <uid>' option place at the end of command line 'hf mf sim...'
2. Bug. Wrong ATQA used for single/double/triple uid sizes
3. Add. Possible to extract 4/7 uid, sak, atqa from emulator memory to exactly simulate loaded tag dump - 'hf mf sim' without parameters
4. Add. Support RATS+DESELECT responses for mifare NXP 2K tags simulation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Fix/Add 'hf mf sim' bugs fix, RATS support, etc (@mceloff)
  - Fix serial of FPC. (@ryan)
  - Fix 'data shiftgraphzero' corrupting end of GraphBuffer (@doegox)
  - Fix 'hf legic info' - unsegmented card now uses card size to calc remaining length (@iceman)
--- a/armsrc/mifaresim.c
+++ b/armsrc/mifaresim.c
@@ -34,8 +34,6 @@
 #include "protocols.h"
 #include "apps.h"
 
-uint8_t MifareCardType;
-
 static bool IsTrailerAccessAllowed(uint8_t blockNo, uint8_t keytype, uint8_t action) {
     uint8_t sector_trailer[16];
     emlGetMem(sector_trailer, blockNo, 1);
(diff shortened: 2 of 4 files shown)


iceman1001:
excellent,  much nicer now.

Next time,  run the  _make style_  to ensure whitespaces etc is correct.

vratiskol:
Nice refactoring, my implementation look like a PoC compare to this :)
I used to add every type of card in precomputed command to (maybe one day) add a dual card simulation.
@McEloff  About Wrong ATQA, I tested it on my side and 4, 7 and 10 bits UID were simulate correctly.
Could you, just for my understanding, explain quickly what was incorrectly implement for you ?

McEloff:
Datasheet AN10833.pdf describes ATQA Coding according to the ISO/IEC 14443-3. Bits 7 and 8 defines UID size:  0-0 for single size, 0-1 for double size, 1-0 for triple size. In previos code of MifareSimInit() method, for example, wrong aqta was:
rATQA_Mini_4B[]  = {0x44, 0x00} - this atqa for 7-bytes uid, valid is 04 00
rATQA_Mini_10B[] = {0x44, 0x00} - this atqa for 7-bytes uid, valid is 84 00
and so on

iceman1001:
According to the pdf you link,
```
mini  4b  04 00
mini  7b  44 00
mini 10b  84 00

1k  4b   04 00
1k  7b   44 00
1k 10b   84 00

4k  4b   02 00
4k  7b   42 00
4k 10b   82 00
```

iceman1001:
The 2K plus cards is a different story.  I haven't seen much Classic 2k cards but  2k plus cards which can be adjusted to be classic via SL mode.

What do we mean when we say 2K cards?

McEloff:
2k Classic is not exists. But there are 2k Plus with SL0 emulation. Those cards has SAK=08, therefore many readers look at card as 1K classic. NXP Taginfo application tests this card via RATS+DESELECT support, then can operate whole of 32 sectors under SL0 mode as classic tag. Otherwise reading SAK 08 we can't determine difference between 1K and 2K card storage

iceman1001:
SAK 0x08,  ok,  but we were talking about ATQA.

McEloff:
ATQA not specified tag size. And datasheet strongly not recommends to use ATQA for chip identification. But datasheets describes how ATQA codes uid size for 4, 7, 10 bytes

iceman1001:
I think you missunderstand.  What ATQA does a 2K plus in SL0 mode have?

McEloff:
2k Plus: 04 00 or 44 00 for 4/7 bytes UID. ATQA not specified security level.

iceman1001:
Good, I updated the helptext to reflect this little fact.

McEloff:
Ok.
Today I have been known, that 'hf mf sim' does not works with reader ACR122U during Classic tag simulation with 7 bytes UID. Reader uses non-standard selection procedure. There is a trace log between ACR and 7-bytes Classic tag:

          0 |       1056 | Rdr |26                                 |     | REQA
       2244 |       4612 | Tag |44  00                             |     | 
      13312 |      15776 | Rdr |93  20                             |     | ANTICOLL
      16964 |      22852 | Tag |88  04  67  6e  85                 |     | 
      44144 |      54608 | Rdr |93  70  88  04  67  6e  85  ad  de |  ok | SELECT_UID
      55860 |      59380 | Tag |04  da  17                         |     | 
      68576 |      71040 | Rdr |95  20                             |     | ANTICOLL-2
      72228 |      78052 | Tag |8a  d6  49  80  95                 |     | 
      99408 |     109872 | Rdr |95  70  8a  d6  49  80  95  ce  21 |  ok | ANTICOLL-2
     111124 |     114644 | Tag |08  b6  dd                         |     | 

    3175728 |    3176720 | Rdr |52                                 |     | WUPA
    3177972 |    3180340 | Tag |44  00                             |     | 
    3193632 |    3204096 | Rdr |93  70  88  04  67  6e  85  ad  de |  ok | SELECT_UID
    3205348 |    3208868 | Tag |04  da  17                         |     | 
    3221904 |    3227824 | Rdr |95  50  8a  d6  49                 | !crc| ANTICOLL-2
    3229012 |    3231380 | Tag |80  95                             |     | 
    3248128 |    3258592 | Rdr |95  70  8a  d6  49  80  95  ce  21 |  ok | ANTICOLL-2
    3259860 |    3263380 | Tag |08  b6  dd                         |     | 
    6014688 |    6015680 | Rdr |52                                 |     | WUPA

After WUPA reader sends ANTICOL-2 '95  50' bytes with incomplete UID part. Tag responses of missing part of UID+BCC. Then reader complete selection tag.
Now Proxmark not supports given algorithm. This algorithm is often to use during multiple tag anticollision select procedure. I think this need to be corrected in further.
What do You think?

iceman1001:
Sounds like the reader is using a correct anticollision procedure and that the reader failed to read pm3 simulation where the transfer got collision bits.   Why?  Maybe they had 1card, 1pm3 simulating and present to reader.   The pm3 doesn't support multiple tags when reading,  nor do it follow a "genuine" anticollision when simulating.

If you are up to it,  fixing the simulation to support true ISO14443a anticollision would be great.

doegox:
True anticollision would be awesome, and opening the door for simulating N cards at once :) or at least N UIDs and mess up with readers.

merlokk:
almost all readers that i see just grabs the first card and thats all... (

iceman1001:
yup yup yup

McEloff:
I tests new release of 'hf mf sim' that supports anticollision procedure :) There is a refactoring of selection case, simplified code, and supports full-byte anticollision frame. Split-byte anticollision frame whill not be supported.
Commits ASAP :)

PR #170, opened 2019-04-18
'hf mf sim' full-byte anti-collision frame support

McEloff:
Refactoring of select tag case. Add precompiled responses for full-byte anti-collision on cascade levels 1/2/3


iceman1001:
A good stepping stone to full anticollision.  Dunno how often byte wise will help.

McEloff:
> Dunno how often byte wise will help

I found this only with acr122 reader.

PR #172, opened 2019-04-21
UL/NTAG new dump file format. Added counters support, simulation

McEloff:
1) New header format of dump files for Ultralight/NTAG. Added counters to header, removed PACK from header, because PACK and PWD for EV1 stores into dump.
2) Fix EV1/NTAG simulation.

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -1318,6 +1318,9 @@ void SimulateIso14443aTag(int tagType, int flags, uint8_t *data) {
                     p_response = NULL;
                 }
             }
+        } else if (receivedCmd[0] == MIFARE_ULEV1_VCSL) {
+            EmSend4bit(CARD_NACK_NA);
+            p_response = NULL;
         } else {
             // Check for ISO 14443A-4 compliant commands, look at left nibble
             switch (receivedCmd[0]) {
@@ -1371,8 +1374,10 @@ void SimulateIso14443aTag(int tagType, int flags, uint8_t *data) {
                 default: {
                     // Never seen this command before
                     LogTrace(receivedCmd, Uart.len, Uart.startTime * 16 - DELAY_AIR2ARM_AS_TAG, Uart.endTime * 16 - DELAY_AIR2ARM_AS_TAG, Uart.parity, true);
-                    Dbprintf("Received unknown command (len=%d):", len);
-                    Dbhexdump(len, receivedCmd, false);
+                    if (MF_DBGLEVEL >= 3) {
+                        Dbprintf("Received unknown command (len=%d):", len);
+                        Dbhexdump(len, receivedCmd, false);
+                    }
                     // Do not respond
                     dynamic_response_info.response_n = 0;
                 }
@@ -1388,7 +1393,7 @@ void SimulateIso14443aTag(int tagType, int flags, uint8_t *data) {
                 dynamic_response_info.response_n += 2;
 
                 if (prepare_tag_modulation(&dynamic_response_info, DYNAMIC_MODULATION_BUFFER_SIZE) == false) {
-                    DbpString("Error preparing tag response");
+                    if (MF_DBGLEVEL >= 3) DbpString("Error preparing tag response");
                     LogTrace(receivedCmd, Uart.len, Uart.startTime * 16 - DELAY_AIR2ARM_AS_TAG, Uart.endTime * 16 - DELAY_AIR2ARM_AS_TAG, Uart.parity, true);
                     break;
                 }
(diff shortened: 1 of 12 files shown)


client/cmdhfmfu.c
```diff
@@ -1619,11 +1619,15 @@ void printMFUdumpEx(mfu_dump_t *card, uint16_t pages, uint8_t startpage) {
...
     PrintAndLogEx(NORMAL, "Signature1| %s| %s", sprint_hex(card->signature, 16), sprint_ascii(card->signature, 16));
     PrintAndLogEx(NORMAL, "Signature2| %s| %s", sprint_hex(card->signature + 16, 16), sprint_ascii(card->signature + 16, 16));
+    PrintAndLogEx(NORMAL, "Counter0  | %-24s| %s", sprint_hex(card->counter_tearing[0],     3), sprint_ascii(card->counter_tearing[0],     3));
+    PrintAndLogEx(NORMAL, "Tearing0  | %-24s| %s", sprint_hex(card->counter_tearing[0] + 3, 1), sprint_ascii(card->counter_tearing[0] + 3, 1));
+    PrintAndLogEx(NORMAL, "Counter1  | %-24s| %s", sprint_hex(card->counter_tearing[1],     3), sprint_ascii(card->counter_tearing[1],     3));
+    PrintAndLogEx(NORMAL, "Tearing1  | %-24s| %s", sprint_hex(card->counter_tearing[1] + 3, 1), sprint_ascii(card->counter_tearing[1] + 3, 1));
+    PrintAndLogEx(NORMAL, "Counter2  | %-24s| %s", sprint_hex(card->counter_tearing[2],     3), sprint_ascii(card->counter_tearing[2],     3));
+    PrintAndLogEx(NORMAL, "Tearing3  | %-24s| %s", sprint_hex(card->counter_tearing[2] + 3, 1), sprint_ascii(card->counter_tearing[2] + 3, 1));
```
iceman1001: I think you can loop these three counter/tears instead.

client/loclass/fileutils.c
```diff
@@ -221,13 +221,16 @@ int saveFileJSON(const char *preferredName, const char *suffix, JSONFileType fty
...
-            JsonSaveStr(root, "$.Card.Counter", "N/A");
-
-            // size of header 48b
+            JsonSaveBufAsHexCompact(root, "$.Card.Counter0", tmp->counter_tearing[0], 3);
+            JsonSaveBufAsHexCompact(root, "$.Card.Tearing0", tmp->counter_tearing[0] + 3, 1);
+            JsonSaveBufAsHexCompact(root, "$.Card.Counter1", tmp->counter_tearing[1], 3);
+            JsonSaveBufAsHexCompact(root, "$.Card.Tearing1", tmp->counter_tearing[1] + 3, 1);
+            JsonSaveBufAsHexCompact(root, "$.Card.Counter2", tmp->counter_tearing[2], 3);
+            JsonSaveBufAsHexCompact(root, "$.Card.Tearing2", tmp->counter_tearing[2] + 3, 1);
+
```
iceman1001: same thing here,  loop it instead

McEloff:
> How backward compatible is this new format with all previous dumps?

I think write a conversation script. But I do not know, need to check format version during dump loading in 'hf mfu eload' and 'hf mfu restore'? If do it, then easier to convert old dump during loading without script using.

iceman1001:
It not too good to break all previous dumps without a plan how to solve it.
The json format is no problem, it was selected because we could just add stuff.
The binary header is another thing.  These bin dumps is used in chameleon mini,  that gui,  pm3 and MCT app... among others.   We don't just break that

McEloff:
We cannot add new capabilities of Ultralight EV1/NTAG to binary dump files without breaking  compatibility with other devices, because old header format was not be extendable. I add method to PM3 to convert old format to new on load dump and emul-file.

armsrc/iso14443a.c
```diff
@@ -1371,8 +1374,10 @@ void SimulateIso14443aTag(int tagType, int flags, uint8_t *data) {
                 default: {
                     // Never seen this command before
                     LogTrace(receivedCmd, Uart.len, Uart.startTime * 16 - DELAY_AIR2ARM_AS_TAG, Uart.endTime * 16 - DELAY_AIR2ARM_AS_TAG, Uart.parity, true);
-                    Dbprintf("Received unknown command (len=%d):", len);
-                    Dbhexdump(len, receivedCmd, false);
+                    if (MF_DBGLEVEL >= 3) {
```
iceman1001: please use the defines

armsrc/iso14443a.c
```diff
@@ -1388,7 +1393,7 @@ void SimulateIso14443aTag(int tagType, int flags, uint8_t *data) {
                 dynamic_response_info.response_n += 2;
 
                 if (prepare_tag_modulation(&dynamic_response_info, DYNAMIC_MODULATION_BUFFER_SIZE) == false) {
-                    DbpString("Error preparing tag response");
+                    if (MF_DBGLEVEL >= 3) DbpString("Error preparing tag response");
```
iceman1001: same as above

PR #179, opened 2019-05-02
Add libpcsclite-dev to package install list

MrButlerTron:
Build fails on new Lubuntu install, same error and solution as here: https://github.com/Proxmark/proxmark3/issues/774

### DIFF
--- a/Installation_Instructions/Ubuntu-Installation-Instructions.md
+++ b/Installation_Instructions/Ubuntu-Installation-Instructions.md
@@ -22,7 +22,7 @@ sudo apt-get update
 ### Requirements
 
 ```sh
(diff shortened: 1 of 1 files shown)


doegox:
Sorry, could you tell me which component has this dependency?
I couldn't find any "pcsc" or "winscard" occurrencies in the repo.

iceman1001:
This repo doesn't have PCSC support,  only in the offical repo. 

Can you show a output where this repo fails compilation on Lubuntu?

MrButlerTron:
Thanks, that actually makes a lot of sense.  Was moving all my instances of the software around, so likely I was compiling the official repo and not this one.  My bad.

PR #181, opened 2019-05-02
Update README.md

NinjaStyle82:
saw this issue on Ubuntu 18.10 and 19.04. Bricked my Proxmark3 several times until I removed modem manager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ The official PM3-GUI from Gaucho will not work.
 The new universial GUI will work. [Proxmark3 Universal GUI](https://github.com/burma69/PM3UniversalGUI) 
 
 ## Notices
-Kali and ArchLinux users usually must kill their modem manager in order for the proxmark3 to enumerate properly.   
+Kali, Ubuntu and ArchLinux users usually must kill their modem manager in order for the proxmark3 to enumerate properly.   
 `sudo apt remove modemmanager`
 		   
 ## Issues


iceman1001:
I have had Ubuntu 14.04, 1604. 1704, 1804 and never needed to remove the modemmanager when running the pm3.
The udev rules was enough for my case.   How come its different for you?

doegox:
@iceman1001 native or in a VM? I confirm this issue in different linux environments during my workshops and always told people to remove entirely modemmanager.
The risk of bricking a Pm3 is very high indeed.
Why some can live with udev rules while for others (including myself) it's not enough is still a mystery to me.

IMHO the best advice would be (and not only for Pm3 but also some other devices):
* For all linux-alike, remove modemmanager if you don't need it
* If you need modemmanager, install udev rules and hope for the best

iceman1001:
in VM,     I did however need to do it on parrotOS (seems to be a Debian offspring) also in VM.
I run vmware server (paid), you maybe run virtualbox from Oracle? 

Since you are the linux expert among us,  I go with that.  Can we make it in the Makefile like **make udev**  to remove the modemmanager?   So the process is as painless as could for endusers?

doegox:
Interesting post that shows the issue exists indeed on Ubuntu too and has maybe an explanation on why these different behaviors : https://askubuntu.com/questions/399263/udev-rules-seem-ignored-can-not-prevent-modem-manager-from-grabbing-device/1119498 
See the answer starting with _"Modem Manager can be configured to use different filter policies, and the udev tags such as ID_MM_DEVICE_IGNORE have no effect under the strict filter policy."_

iceman1001:
I can confirm this on Ubuntu 19.04,     I downloaded a vm image from osboxes.org and followed the instructions.   The modemmanager policy is strict.  

**sudo systemctl status ModemManager**

image

doegox:
Ì tried reinstalling ModemManager and experiment based on the link above.

```
apt-get-install modemmanager
```
Indeed on Debian, ModemManager runs with `strict` filter rule, ignoring our udev filters.
Trying the alternative way proposed in the link:
```
systemctl edit ModemManager.service
  [Service]
  Environment="MM_FILTER_RULE_TTY_ACM_INTERFACE=0"
```
This creates the following file: `/etc/systemd/system/ModemManager.service.d/override.conf` with these two lines.
```
service ModemManager restart
```
That's it.
To test, turn systemd debug on and watch logs, then set systemd log level back to ERR:
```
mmcli -G DEBUG
journalctl -f|grep "ModemManager.*\[filter\]"
mmcli -G ERR
```

When ModemManager interferes, you get:
```
ModemManager[xxxxx]: <debug> [filter] (tty/ttyACM0): port allowed: cdc-acm interface reported AT-capable
```
And when it's properly filtering /dev/ttyACM0:
```
ModemManager[xxxxx]: <debug> [filter] (tty/ttyACM0) port filtered: forbidden
```
Now flashing works fine and I see the "port filtered" line popping up during flashing.

iceman1001:
There is however a bullit in the ubuntu readme which says if you have issues,  remove the modemmanager.

doegox:
"if you have issues" : that advise might be already too late if bootrom flashing got corrupted :)

iceman1001:
Ok,  I had some issues on 19.04.    We need to address them. 
What should have be an easy remove modemmanager,  make udev,  wasn't. 

I am closing this PR since we need to update the instructions and maybe even the Makfile udev parts.

PR #187, opened 2019-05-10
client/Makefile: use brew libreadline on macOS

uhei:
Make sure we use libreadline from brew and not the default macOS shipped one.
Otherwise compilation fails for undefined rl_event_hook

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -72,6 +72,8 @@ else
     LUAPLATFORM = macosx
     OBJCSRCS = util_darwin.m
     LDFLAGS += -framework Foundation -framework AppKit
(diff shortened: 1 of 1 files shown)


doegox:
Thank you!
May I ask you to check the OS X installation instructions and if needed to submit a PR to update them?
https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/Mac-OS-X-Homebrew-Installation-Instructions.md
About this brew libreadline and more generally about OS X.
Thanks a lot!

uhei:
> May I ask you to check the OS X installation instructions and if needed to submit a PR to update them?

Yes, I will have a look.

Issue #189, opened 2019-05-10
proxmark3 client crash on macOS with pm3 RDV4 after starting


uhei:
After starting proxmark3 client it crashes with a buffer overflow:
```
tyrell:proxmark3 uhei$ PATH=/usr/bin /usr/bin/lldb client/proxmark3 /dev/tty.usbmodemiceman1 
(lldb) target create "client/proxmark3"
Current executable set to 'client/proxmark3' (x86_64).
(lldb) settings set -- target.run-args  "/dev/tty.usbmodemiceman1"
(lldb) run
Process 22554 launched: '/Users/uhei/Downloads/prox-rdv4/proxmark3/client/proxmark3' (x86_64)

          
██████╗ ███╗   ███╗ ████╗      ...iceman fork          
██╔══██╗████╗ ████║   ══█║       ...dedicated to RDV40           
██████╔╝██╔████╔██║ ████╔╝           
██╔═══╝ ██║╚██╔╝██║   ══█║     iceman@icesql.net          
██║     ██║ ╚═╝ ██║ ████╔╝    https://github.com/rfidresearchgroup/proxmark3/          
╚═╝     ╚═╝     ╚═╝ ╚═══╝  pre-release v4.0          

Support iceman on patreon,   https://www.patreon.com/iceman1001/          

          
[=] Using UART port /dev/tty.usbmodemiceman1           
2019-05-10 12:40:27.654094+0200 proxmark3[22554:158777] detected buffer overflow
Process 22554 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff68e4a2c6 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff68e4a2c6 <+10>: jae    0x7fff68e4a2d0            ; <+20>
    0x7fff68e4a2c8 <+12>: movq   %rax, %rdi
    0x7fff68e4a2cb <+15>: jmp    0x7fff68e44457            ; cerror_nocancel
    0x7fff68e4a2d0 <+20>: retq   
Target 0: (proxmark3) stopped.
(lldb) thread backtrace 
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff68e4a2c6 libsystem_kernel.dylib`__pthread_kill + 10
[...]

PR #190, opened 2019-05-10
client/comms.c: fix buffer overflow

uhei:
Don't copy more bytes into pm3_capabilities as its size.

Fix: RfidResearchGroup/proxmark3#189


doegox:
Thanks!
This hides a more serious problem that Proxmark3 FW is not compatible with client and PM3 needs to be flashed. I'll add a warning on top of this safety check.

doegox:
https://github.com/RfidResearchGroup/proxmark3/commit/3412e9d8c55eae57b0366a393b1aa6c84b642499

PR #195, opened 2019-05-14
Added paged moving to the plot GUI

mcd1992:
Adds bindings for HOME and END to jump to the start or end of the graph.
Adds bindings for PGUP and PGDOWN to move a page left or right on the graph.
Sets the PageWidth variable (wasn't set before) to contain the number of samples on the current page. (Fixes shift+arrow pagination)

### DIFF
--- a/client/proxguiqt.cpp
+++ b/client/proxguiqt.cpp
@@ -33,8 +33,8 @@ extern "C" {
 
 bool g_useOverlays = false;
 int g_absVMax = 0;
-uint32_t startMax;
-int PageWidth;
+uint32_t startMax; // Maximum offset in the graph (right side of graph)
+uint32_t PageWidth; // How many samples are currently visible on this 'page' / graph
 int unlockStart = 0;
 
 void ProxGuiQT::ShowGraphWindow(void) {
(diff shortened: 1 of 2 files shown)


client/proxguiqt.cpp
```diff
@@ -685,6 +689,26 @@ void Plot::keyPressEvent(QKeyEvent *event) {
...
+        case Qt::Key_PageUp:
+            if (GraphStart >= PageWidth) {
+                GraphStart -= PageWidth;
+            } else {
+                GraphStart = 0;
+            }
+            break;
+
+        case Qt::Key_PageDown:
+            GraphStart += PageWidth;
```
iceman1001: No check for larger than startMax?  ie overshoot number of samples available

mcd1992: It gets clamped in the paint function so I never bothered. I can easily check though, I'll throw it in tomorrow.

Thanks!

iceman1001: do another PR with that fix.  I merge this one meanwhile.

iceman1001:
And update Changelog.md  for documentation and eternal glory!

PR #201, opened 2019-05-22
fix: 'hf 14a sim' ultralight COMPATIBLE_WRITE command,

McEloff:
1. Bugfix. EV1 COMPATIBLE_WRITE command contains two parts: receive command and receive data
2. Reordered some conditions. We need to check any data before checking commands in received buffer, because data can be same to unexpected command
3. Additional length checks in  received command
4. Added EV1 virtual card support from config page of emulator memory
5. Defines used for constants
----
6. PACK-response to EV1 PASSWORD verification no more as precompiled sequence, because PACK can be changed dynamically to config page
7. Setting up minimal pages count for emulating Ultralight and EV1/NTAG tags, if loaded eml-file not contains whole dump

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -832,8 +832,6 @@ static bool SimulateIso14443aInit(int tagType, int flags, uint8_t *data, tag_res
     static uint8_t rSAKc2[3]  = { 0x00 };
     // dummy ATS (pseudo-ATR), answer to RATS
     static uint8_t rRATS[] = { 0x04, 0x58, 0x80, 0x02, 0x00, 0x00 };
-    // PACK response to PWD AUTH for EV1/NTAG
-    static uint8_t rPACK[4] = { 0x00 };
     // GET_VERSION response for EV1/NTAG
     static uint8_t rVERSION[10] = { 0x00 };
     // READ_SIG response for EV1/NTAG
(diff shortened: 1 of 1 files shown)


doegox:
Thanks !
While at hf 14a sim, can I ask you to have a look at standalone hf_colin ant hf_mattyrun ? 
They're broken since the API change of Mifare1ksim

iceman1001:
No need to, I fixed them.

McEloff:
Standalone 'hf_colin' and 'hf_mattyrun' uses current worked of  Mifare1ksim() method. I think, flag FLAG_UID_IN_EMUL must be replaced to FLAG_MF_1K simplify. In this case simulator would use uid placed into method parameter and conctructs sak/atqa by really tag size (1K) and given uid length.

I suspect, those standalones was not worked corrected earlier too, because uses hardcoded uid length = 4 bytes.  Previous calls of iso14443a_select_card() method can return uid of selected tag same to 4, 7, 10 bytes. But simulated tag is strongly 4 byte uid has. Try to apply attached patch.
[patch.txt](https://github.com/RfidResearchGroup/proxmark3/files/3210753/patch.txt)

iceman1001:
Thanks! That would fix the other issues with those standalones modes.  I pushed your fixes!

doegox:
Thanks a lot @McEloff !

PR #205, opened 2019-05-29
Specify that we need TCP and not UDP connection

phcoder:
Relevant man pages say that with our previous
code libc might return either. Glibc returns tcp but
bionic returns udp. So specify it explicitly


iceman1001:
Don't forget to add a line in the changelog :)

PR #210, opened 2019-06-09
<added keys for Transportes Insular La Palma (Public Transport)>

javastraat:
added keys for Transportes Insular La Palma (Public Transport)


iceman1001:
Thanks bro!

PR #221, opened 2019-07-04
Fix clock deadlock in hitag sniff

### LINKED ISSUE #220: Hitag sniff hangs if not first command
Hitag sniff works only if it's the first command you run. If it's preceded by any other command (e.g. `hf search`) it hangs here forever:

https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/hitag2.c#L742

ViRb3:
Fix #220


iceman1001:
Excellent!
Now a row in the changelog.md  please.

PR #222, opened 2019-07-04
Print help if unknown arg for hitag reader/writer

ViRb3:
Not the best solution, but better than the weird things happening now if no/invalid argument is passed.


doegox:
Thanks, that will do it for now

iceman1001:
Don't forget to add a line in changelog.md :)

PR #233, opened 2019-07-09
Add hitag2 write password auth

ViRb3:
**Tested and working.**

My only concern is the following write in crypto mode:
https://github.com/RfidResearchGroup/proxmark3/blob/cb453139813a7b1b5ff17e6bd66d3e80075ead35/armsrc/hitag2.c#L506-L511

Unless I'm missing something, this is useless since the write at the top of the method:
https://github.com/RfidResearchGroup/proxmark3/blob/cb453139813a7b1b5ff17e6bd66d3e80075ead35/armsrc/hitag2.c#L449-L453
will be called long before this one is reached.

For that reason I added only the top write in password mode.

### DIFF
--- a/armsrc/hitag2.c
+++ b/armsrc/hitag2.c
@@ -331,64 +331,6 @@ size_t blocknr;
 // Hitag2 operations
 //-----------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


doegox:
thank you!

PR #232, opened 2019-07-09
(OSX) Fix build failure and all build warnings

matrix:
Hi,
I'm my OSX environment the build process fail. This issue is related to the "binutils" package installed by brew:

```
$ ls -al $(which ar) $(which ranlib)
[...] /usr/local/bin/ar -> /usr/local/Cellar/binutils/2.32/bin/ar
[...] /usr/local/bin/ranlib -> ../Cellar/binutils/2.32/bin/ranlib
```

With the following patch I switch the "ar" and "ranlib" binaries to the working ones: "/usr/bin/ar" and "/usr/bin/ranlib".

Moreover I solved all the build warning on OSX and I found a little bug here:

```
[-] CC emv/cmdemv.c
emv/cmdemv.c:290:40: warning: if statement has empty body [-Wempty-body]
        if (pdol_data_tlv != &data_tlv);
                                       ^
emv/cmdemv.c:290:40: note: put the semicolon on a separate line to silence this warning
1 warning generated.
```

and I solved in this way:

```
--- a/client/emv/cmdemv.c
+++ b/client/emv/cmdemv.c
@@ -287,8 +287,8 @@ static int CmdEMVGPO(const char *Cmd) {
         PrintAndLogEx(ERR, "Can't create PDOL data.");
         tlvdb_free(tmp_ext);
         tlvdb_free(tlvRoot);
-        if (pdol_data_tlv != &data_tlv);
-        free(pdol_data_tlv);
+        if (pdol_data_tlv != &data_tlv)
+          free(pdol_data_tlv);
         return PM3_ESOFT;
     }
     PrintAndLogEx(INFO, "PDOL data[%d]: %s", pdol_data_tlv_data_len, sprint_hex(pdol_data_tlv_data, pdol_data_tlv_data_len));

```

using as reference the proxmark3 original source code [here](https://github.com/Proxmark/proxmark3/blob/master/client/emv/cmdemv.c#L330) 

Thanks,
matrix

PR #234, opened 2019-07-10
Fix compilation under MacOSX

matrix:
details here: #232


doegox:
Thank you.
Globally all compilation defines will have to be central and more easily tunable but for now I guess we can live with these ad-hoc changes.

PR #238, opened 2019-07-11
Update default_keys.dic with keys from an APK 

3ldidi94:
Update the default_keys.dic with keys from Android application "Scan Badge".
I did the diff between keys already present and keys from the application.


iceman1001:
Interesting the APK ScanBadge got all keys from this repo...

3ldidi94:
It seems that developers do not always contribute their new discoveries :disappointed:

iceman1001:
you got their default keys file?  How many keys do they have?

PR #242, opened 2019-07-12
Remove aes.c in favor of mbedtls implementation.

slurdge:
On my board, saves around 4% total size which is quite nice. I didn't investigate if we can reuse also the des implementations. Needs some testing as I do not know how to trigger this code path.


doegox:
Nice ! Did you get a chance to test the changes ?

slurdge:
Nope... Which card should I read ?

doegox:
A DESFire configured with some AES key and try sth like 
```
hf mfdes auth 3 4 0 404142434445464748494a4b4c4d4e4f
```

iceman1001:
Hm... Does that command actually work?   I never tested it on a card.   :)

iceman1001:
Great PR btw.    Using only mbtls imp.   Next step is remove the custom des aswell.

slurdge:
I removed custom des. However we lose some space (around 1%). Is it worth the trade with easiness of maintenance ?

doegox:
I'd say yes, better to get all our crypto from mbedtls.
There is still the opportunity to provide alternative implementation, still within the mbedtls api, see e.g.  https://github.com/ARMmbed/mbed-os/blob/master/features/mbedtls/targets/TARGET_NUVOTON/TARGET_NUC472/des/des_alt.c

iceman1001:
Yup,  only use mbedtls.

iceman1001:
@slurdge  did you test the mfdes command suggested by philippe?

doegox:
slurdge doesn't have a DESFire EV1 to test by himself, can someone else test it?

iceman1001:
that would be you and I ;)

iceman1001:
All my desfire cards wasn't programmed for AES.   
This PR compiles and I am fine.  I will merge it!

PR #244, opened 2019-07-14
HF_COLIN/STANDALONE: FIX Simulation/Writing Tag to flash/Reading back…

cjbrigato:
… from flash

* Simulation was broke by the 4/7/10 UID lenght pseudo-support (which, in regard of the whole, will never be of any use unless complete reimplementaiton). Kept this part as respect, but enforced UID_IN_EMUL, since it will _ALWAYS_ be the case, be it direct simulation after tag breaking or from reading back tag.
* Writing tag was broken when flashmem driver had SPI settings changed, enforcing 48MHz speed as necessity for some work (like in hf_bog), so now it is
* Reading back, however, should absolutely be done at 24Mhz to avoid those unecessary and unstable [On this chip] FAST_READS.

Should worsk flawlessly now.
If one has spare time, handling getting new keys or defining whole Schemes in Flashmem would be perfect. Just has to be a conditional and a 8/16key ordered answer.

### DIFF
--- a/armsrc/Standalone/hf_colin.c
+++ b/armsrc/Standalone/hf_colin.c
@@ -101,6 +101,8 @@ void ReadLastTagFromFlash() {
     size_t size = len;
     uint8_t *mem = BigBuf_malloc(size);
 
+    FlashmemSetSpiBaudrate(24000000);
+
     if (!FlashInit()) {
         return;
     }
@@ -147,6 +149,9 @@ void WriteTagToFlash(uint8_t index, size_t size) {
     uint8_t buff[PAGESIZE];
 
     emlGetMem(data, 0, (size * 64) / 1024);
+
+
+    FlashmemSetSpiBaudrate(48000000);
 
     if (!FlashInit()) {
         return;
@@ -424,7 +429,7 @@ failtag:
     //-----------------------------------------------------------------------------
     // also we could avoid first UID check for every block
 
-    // then let’s expose this “optimal case” of “well known vigik schemes” :
+    // then let's expose this “optimal case” of “well known vigik schemes” :
     for (uint8_t type = 0; type < 2 && !err && !trapped; type++) {
         for (int sec = 0; sec < sectorsCnt && !err && !trapped; ++sec) {
             key = cjat91_saMifareChkKeys(sec * 4, type, NULL, size, &keyBlock[0], &key64);
@@ -783,10 +788,18 @@ readysim:
         case 7:
             flags = FLAG_7B_UID_IN_DATA;
             break;
-        default:
+        case 4:
             flags = FLAG_4B_UID_IN_DATA;
             break;
+        default:
+	    flags = FLAG_UID_IN_EMUL;
+	    break;
     }
+
+    // Use UID, SAK, ATQA from EMUL, if uid not defined
+    //if ((flags & (FLAG_4B_UID_IN_DATA | FLAG_7B_UID_IN_DATA | FLAG_10B_UID_IN_DATA)) == 0) {
+        flags |= FLAG_UID_IN_EMUL;
+    //}
     Mifare1ksim(flags | FLAG_MF_1K, 0, cjuid);
     LED_C_OFF();
     SpinOff(50);
@@ -902,7 +915,7 @@ int e_MifareECardLoad(uint32_t numofsectors, uint8_t keytype) {
     return (isOK) ? PM3_SUCCESS : PM3_EUNDEF;
 }
 
-/* the chk function is a piwi’ed(tm) check that will try all keys for
+/* the chk function is a piwi'ed(tm) check that will try all keys for
 a particular sector. also no tracing no dbg */
 int cjat91_saMifareChkKeys(uint8_t blockNo, uint8_t keyType, bool clearTrace, uint8_t keyCount, uint8_t *datain, uint64_t *key) {
     DBGLEVEL = DBG_NONE;


doegox:
Hi Colin.
Did you enforce 48MHz on writes because you actually observed errors with the current code version or "to be on the safe side" ?
I thought commit a1bb2ec10685e9c887e2e25c651f29660da9d19b fixed the 24MHz write issues.

cjbrigato:
Hi Philippe !
Since I actually have been the one who rewrote the original flashmem.c when it first came, I had more than enough time to play with the flashmem but more important the absolutely awfull SPI of the AT91sam7sX (this 512 version being a massive winner in of the whole series).

I think I had explained in comments (or did I forgot ?) why these settings where chosen, specifically against what would be natural settings.

Not only the settings of a1bb2ec did not solved the bitflip problem but in my case they absolutely and admirably outperformed it ! 
I never had a problem with writing at 24Mhz with old settings _and_ a safer timeout, and ran the most aggressive benchmark I could against these. Still, live-writing required to be at 48Mhz and was, indeed more stable, not only because of the chip/spi (the flashmem documentation has some reference about write vs read).

In the same vein, going further 24Mhz reading (thus havign to use fastreads and others poisoned gifts) where unstable but known. I thought it was ok and somehow a good and working compromise all in all.

Found back a dump of the result when writing at 24Mhz on my standalone and device (BTAddon being up but not used, being standalone or connected did not change anything).

What it's been when writtent at 24Mhz : 
```
[usb] pm3 --> mem read o 0 l 1024
#db# FlashMem read | 0 - 1024 |
#db# FlashMem reading  | 0 | 0 | 512 |
#db# [Chunk: : 00/512] ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
#db# [Chunk: : 16/512] ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
#db# [Chunk: : 32/512] ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
#db# [Chunk: : 48/512] ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
#db# [Chunk: : 64/512] ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
#db# [Chunk: : 80/512] ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
#db# [Chunk: : 96/512] ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
#db# [Chunk: : 112/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
#db# [Chunk: : 128/512] ff 30 31 34 31 00 00 00 ff ff ff ff ff ff ff ff 
#db# [Chunk: : 144/512] ff ff 01 00 01 00 ec 03 70 01 ec 03 70 01 00 00 
#db# [Chunk: : 160/512] 00 00 ff 00 00 00 00 00 00 00 00 00 00 00 00 00 
#db# [Chunk: : 176/512] 00 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 
#db# [Chunk: : 192/512] af 76 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
#db# [Chunk: : 208/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 224/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 240/512] 00 00 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d 
#db# [Chunk: : 256/512] af 76 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 272/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 288/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 304/512] 00 00 00 88 29 da 9d af 76 7f 07 88 00 88 29 da 
#db# [Chunk: : 320/512] 9d af 76 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 336/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 352/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 368/512] 00 00 00 00 88 29 da 9d af 76 7f 07 88 00 88 29 
#db# [Chunk: : 384/512] da 9d af 76 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 400/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 416/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 432/512] 00 00 00 00 00 88 29 da 9d af 76 7f 07 88 00 88 
#db# [Chunk: : 448/512] 29 da 9d af 76 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 464/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 480/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 496/512] 00 00 00 00 00 00 88 29 da 9d af 76 7f 07 88 00 
```
What it should have been (and is when writtend ar 48Mhz) : 
```
[usb] pm3 --> mem read o 0 l 1024
#db# FlashMem read | 0 - 1024 |
#db# FlashMem reading  | 0 | 0 | 512 |
#db# [Chunk: : 00/512] 8e 11 39 7a dc 88 04 00 47 51 35 56 59 10 28 08
#db# [Chunk: : 16/512] 01 07 7e 09 01 22 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 32/512] 55 55 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 48/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
#db# [Chunk: : 64/512] 00 00 00 00 58 d0 28 6a 01 00 00 00 00 b2 7d 3f
#db# [Chunk: : 80/512] 41 15 11 00 00 00 00 00 00 00 00 00 00 01 00 00
#db# [Chunk: : 96/512] 00 00 00 00 00 00 00 00 00 00 00 00 f0 00 00 00
#db# [Chunk: : 112/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
#db# [Chunk: : 128/512] 30 31 34 31 00 00 00 ff ff ff ff ff ff ff ff ff
#db# [Chunk: : 144/512] ff 01 00 01 00 ec 03 70 01 ec 03 70 01 00 00 00
#db# [Chunk: : 160/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 176/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
#db# [Chunk: : 192/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 208/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 224/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 240/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
#db# [Chunk: : 256/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 272/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 288/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 304/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
#db# [Chunk: : 320/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 336/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 352/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 368/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
#db# [Chunk: : 384/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 400/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 416/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 432/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
#db# [Chunk: : 448/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 464/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 480/512] 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
#db# [Chunk: : 496/512] 88 29 da 9d af 76 7f 07 88 00 88 29 da 9d af 76
```

cjbrigato:
Ooook I found strenght to go through the Archives, and definitely the a1bb2ec commit did what I was Absolutely trying to avoid.

The whole flash chip documentation is made against a SPI Mode 0, with some ~mode3 addition as comparison here and there, since these are the supported modes.

Quoting the doc : 
```
SPI bus operation Modes 0 (0,0) and 3 (1,1) are supported. The primary difference between Mode 0
and Mode 3 concerns the normal state of the CLK signal when the SPI bus master is in standby and
data is not being transferred to the Serial Flash. For Mode 0 the CLK signal is normally low on the falling
and rising edges of /CS. For Mode 3 the CLK signal is normally high on the falling and rising edges of
/CS.
```

Then among my tests and Historical data of collegues, I went againts the very thing I wrote in the file : 
SPI modes *SHOULD* be :  
```
            Mode CPOL CPHA NCPHA
            0    0    0    1       clock normally low    read on rising edge
            1    0    1    0       clock normally low    read on falling edge
            2    1    0    1       clock normally high   read on falling edge
            3    1    1    0       clock normally high   read on rising edge
```
But As Stated : 
```
However, page 512 of the AT91SAM7Sx datasheet say "Note that in SPI
            master mode the ATSAM7S512/256/128/64/321/32 does not sample the data
            (MISO) on the opposite edge where data clocks out (MOSI) but the same
            edge is used as shown in Figure 36-3 and Figure 36-4."  Figure 36-3
            shows that CPOL=NCPHA=0 or CPOL=NCPHA=1 samples on the rising edge and
            that the data changes sometime after the rising edge (about 2 ns).  To
            be consistent with normal SPI operation, it is probably safe to say
            that the data changes on the falling edge and should be sampled on the
            rising edge.  Therefore, it appears that NCPHA should be treated the
            same as CPHA.  Thus:
            Mode CPOL CPHA NCPHA
            0    0    0    0       clock normally low    read on rising edge
            1    0    1    1       clock normally low    read on falling edge
            2    1    0    0       clock normally high   read on falling edge
            3    1    1    1       clock normally high   read on rising edge
```
You fallback to CPOL =: 0 and NCPHA = 1 makes in our singular case makes a mode 1 behavior which was all but stable.

As I said in the past, in case of instability, a safe thing to do is to write @48Mhz up to 0.8/ms per page (4ms for a 1k write, which was were the driver got up to on benchmarks and standalone), and to read at 24Mhz and avoid FASTREADS which in our case were , when it was working (~15% of the time) , not a real gain for the pain (going down to 350ms for FASTREAD From a 406ms for 24Mhz READS).
Another Thing to make it more stable is just to play with the clock and accept up to two time slower performance to ensure stability.

iceman1001:
@cjbrigato  I missed you bro!
Never thought of reusing the defines in flashmem.h  for calling the setspi speed instead?
```
#define FLASH_MINFAST 24000000
#define FLASH_FASTBAUD MCK
```

cjbrigato:
@iceman1001 I'm alive and as you see, still wordfull as ever :)
I don't understand what you're talking about.
Here we discuss the inability to write at 24Mhz in a stable way and Read at more than 24Mhz in a neither usefull nor stable too. (We are close to non-event, unless I missed something)

The MINFAST and FASTBAUD defines where here to handle the fact that in READ and only in READ case, if we go past the 24Mhz, then we have to use FASTREADS. But since we have different mode of operation for masterclock, going up to 48Mhz made the whole SCBR calculation and spi speed setting needing change. So we made everything operating SPI at MCK instead of MCK/2.
This seems related but here this is not.

I understand that interestingly in our case the very same value would apply to some defines made for READ vs /WRITE, but this is along coincidences and it would be misleading to reuse them.
Still, havin self-telling defines would be more "leading" than droping magic 24/48e10(n) ...

Maybe 
``` 
#define FLASH_READBAUD FLASH_BAUD
#define FLASH_WRITEBAUD MIN(2*FLASH_BAUD,MCK)
```

iceman1001:
:)  I notice the summer heat is strong in Paris 
We talk about your PR,  even if @doegox  and you were discussing the existence of a problem (or not)
or whether you or @doegox  solved it in previous PR,  I on the other hand just reflected on the code style with the values,  where we tend want to have a define in order to document / make clear the purpose of the value.  
So pick a define ;)   call it FLASH_BAUD_COLIN_IS_PERNICKETY_RD ,  I don't mind.  haha!

cjbrigato:
Problem is that to pick such define we should definitely agree on going back to former settings, and then agree that standard behavior is to read at 24mhz and write at 48Mhz.
I think my standalone still have the test read and test write flash commented available on to benchmark and test. 
If not, I they are on my old kigiv fork.

doegox:
Hi Colin!

> I thought it was ok and somehow a good and working compromise all in all.

Well, when the explanation on how to safely use the flash operations by changing the SPI speed for read<>write operations lies in a closed issue and is not handled automatically by the firmware, guess what happened: I missed your explanations, got bitflip issues and tries to solve them by myself.

> Not only the settings of a1bb2ec did not solved the bitflip problem but in my case they absolutely and admirably outperformed it !

I wrote this little script: https://gist.github.com/doegox/c35925b14c7b8dc4baf42d61d9c71f1b
which wipes the first three pages (196608b), fills them with fresh random data, read them back and compare contents.
Running 100x each on 3 different RDV4 tonight didn't reveal any issue.
What kind of test would exhibit problems with the current firmware settings ?

The only issue I had is that we should add some delay after erase/write operations in the firmware to let them finish before executing other flash operations.

cjbrigato:
For a reason I can't understand and that is the very reason why I did not include a rollback of settings in my pr, when CMD_FLASHMEM are launched by client I did not manage to reproduce the problem. Still, compile my standalone or bog one but deleting the line where SPI baud is set at 48Mhz and poof!  Note that he was more affected than me since he was writing if I recall in context where SPI was used in near context and so he had to get to 48imhz early whereas I did not need to (until yor fix).   Interesting thing that the same fix makes my standalone unable to write at 24mhz definitely while it permit fixing writing at 24mhz in some other cases.
I think I knew we were running toward this kind of surprise at the very moment I saw that the SPI on at91sam7X needed a double reset to hope being able to be considered configurable. I then came all the errata and SPI Mode act-as-pic32 wannabee etc.
So, to summary more kind of tests needs to be done, from and to device, the plugued and upluged, etc. I still in either case I'm quite sure your or my settings will lead to the same result : reading at 24mhz and writing at 48mhz is both without drawbacks and very stable, also being nearly the best performance we can get.

cjbrigato:
Your busy timeout is clearly low and edgy, but why did you lower it so much BTW?

doegox:
> Your busy timeout is clearly low and edgy, but why did you lower it so much BTW?

Yes you're right, I measured it against mem write (needed 20ms, took 50ms), but apparently mem wipe needs more time (70ms for a page?), and longer BUSY_TIMEOUT doesn't hurt, I changed it for 200ms now.

doegox:
I'm ok with playing with 24MHz & 48MHz but this should be handled transparently, memorizing current SPI speed and changing it automatically when needed depending on read/write/erase.
Still, I'd like to understand when writing at 24MHz would fail in specific cases, I'll try on your standalone mode.

doegox:
Hmm I ran your standalone mode from the master branch, acquired an empty MFC, then did `mem save` of first page and found the MFC dump correctly written.
I tested it while connected to the client and on another RDV4 with Blue Shark while disconnected and on battery.

iceman1001:
Lets see,   we can merge the PR since its limited to colins standalone only.
And open a issue where we can continue the discussions.

cjbrigato:
I'm on something as a result of your test but still can you ensure the test has been made in both connected and disconnected mode with btaddon as battery with bt in both case?
If the result are the very same, then the can you DO this :
- pm3 on master before my pr, and client and full image wise, and with this specific platform extras for in makefile.Platform : btaddon fpc_usart_dev but disable smart card and support (should be stated as absent in the client welcome message) 
- connect pm3 with btaddon off battery stated off via USB 
- launch the standalone, and the give it a fob until success, let the flash write then clic to get from simulation to a clone failure by not presenting a gen1a card.
- you should have the shell back available : try a read, not a save, from client in this state with : mem read o 0 l 1024 (and get a 512 chunk back :p / cc @iceman1001.
- the chunk has strong chance to be shown as in the same way I showed you.
- in such case, I without disconnecting try a standalone again, I then click the butto' while in waiting tag so the lasttag is replayed instead
- should see a 0000 uidbsim that you can stop.
- HD mf ekeyprn should show 1byte each sector offset Ed key with one or two sector set at 0

Ifbyou got time, and I would be please to have your result. 
In fact your commit DID solve the writing at 24mhz in a larger scale then with former settings.
But it seems that actually reading back with the same settings without a full reset after data have been written at 24mhz with your settings is actually where the problem lies. Considering how the client can operate its then normal that we had very low odd to face this case.

I think you can do a very quick standalone mode that just write a static 1kb from pages 5-8 back to 1-4 or even an hardcoded 1k after wiping the 4 first pages then read it back and compare to make the issue trigger. 

About the delays these are quite well documented in the chip manual so safe and still low timeouts constants can be defined. Consider hardware capping values in case of lacking the info (like the 0.8ms/page states hardware cap is for write operations, and it tells quite a lot about what is an efficient and still a safe timeout)

Sorry for awfull formatting and typos. Quite the war here.

PR #246, opened 2019-07-14
Fix. More flexible anti-collision frame check (hf_mf_sim)

McEloff:
Removed hard coded check for anti-collision during selection of emulated Mifare Classic tag


iceman1001:
One of these days we have a better anticollision :)
Don't forget a line in CHANGEME.md

PR #249, opened 2019-07-15
`hf 14a apdu` improvement

merlokk:
added:

1. print APDU (if it can be decoded)
2. construct extended and normal size apdu


client/emv/apduinfo.h
```diff
@@ -31,4 +32,28 @@ typedef struct {
 const APDUCode *GetAPDUCode(uint8_t sw1, uint8_t sw2);
 const char *GetAPDUCodeDescription(uint8_t sw1, uint8_t sw2);
 
+typedef struct {
+    uint8_t cla;
+    uint8_t ins;
+    uint8_t p1;
+    uint8_t p2;
+    uint8_t lc[3];
+} __attribute__((packed)) ExtAPDUHeader;
```
iceman1001: Don't we have a define PACKED for this?

client/emv/apduinfo.h
```diff
@@ -31,4 +32,28 @@ typedef struct {
...
+    uint8_t cla;
+    uint8_t ins;
+    uint8_t p1;
+    uint8_t p2;
+    uint16_t lc;
+    uint8_t *data;
+    uint32_t le;
+    bool extended_apdu;
+    uint8_t case_type;
+} __attribute__((packed)) APDUStruct;
```
iceman1001: See above comment

doegox:
and changelog... ;)

client/emv/apduinfo.h
merlokk: yes, but i see it in
`include\hitag.h` and `include\pm3_cmd.h`

have added it to `util.h`

doegox:
thanks!

iceman1001:
I think we already have the PACKED defined somewhere before????

merlokk:
here `include\hitag.h` and `include\pm3_cmd.h`

maybe ill add include\proxmark.h?
and add there?

iceman1001:
sounds like a good suggestion

merlokk:
just need to choose right name...
or maybe  `common_client.h`?
there is another proxmark3.h

iceman1001:
just need to choose right header file :)  it must be one in common folder since we use the PACKED define on both sides.

merlokk:
in include folder there is another common.h, but it for arm. almost all headers there for arm.

maybe lets client lay in folder client...

iceman1001:
as I said,  we use that define on both device and client.

merlokk:
ill create new one and put there min max and this

doegox:
common.h is already used both in client & arm

doegox:
while you're at it, please replace the ` __attribute__((__packed__))` still present in `client/elf.h`, `armsrc/Standalone/hf_young.c`, `include/mifare.h` and `include/common.h`

merlokk:
Ok.
I see that common.h is place there was added many arm related stuff. Try to get rid of that

doegox:
ok, yes include files are a bit messy, rationalization is welcomed :)

PR #251, opened 2019-07-16
added PACKED attribute for structures and reorganize include folder

merlokk:


client/deprecated-hid-flasher/flasher/usb_cmd.h
```diff
@@ -12,16 +12,8 @@
...
-typedef BYTE uint8_t;
-#define PACKED
-// stuff
-#else
-#include <stdint.h>
-#include <stdbool.h>
-#define PACKED __attribute__((packed))
-#endif
+
+#include "common.h"
```
doegox: client/deprecated-hid-flasher/flasher/usb_cmd.h should be left untouched because it's standalone without access to pm3 include files.
Or you can add the definition of PACKED in its client/deprecated-hid-flasher/flasher/proxmark3.h

merlokk: btw. it cant compile

merlokk: unknown type name 'PVOID'
unknown type name 'UCHAR'
etc...

merlokk: stuff from windows only.

merlokk: may you check?

doegox: client/deprecated-hid-flasher/flasher compiles fine under Linux.
Feel free to fix compilation under Windows but don't worry too much, it's deprecated ;)

merlokk: ok, let it be in this state...

iceman1001: yeah,  don't use it.  If the ppl want to upgrade from HID -> CDC they have plenty of JTAG options today.  I would say remove it completely

iceman1001:
Nice refactoring!   Code is getting cleaner!  Great!

PR #256, opened 2019-07-16
Add the needed qt package on ArchLinux

slurdge:
qt5-base is enough on my Arch

### DIFF
--- a/doc/md/Installation_Instructions/Linux-Installation-Instructions.md
+++ b/doc/md/Installation_Instructions/Linux-Installation-Instructions.md
@@ -41,8 +41,10 @@ Additional AUR packages:
 ```sh
 yaourt -S termcap
 ```
-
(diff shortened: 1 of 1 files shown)


doegox:
ok we trust you :)

iceman1001:
I was just about to say that the instructions used another package name.. 
We use both these ones in the document.
yaourt  vs pacman

I never tried Archlinux,  so excuse my question,  but do they do the same thing?

slurdge:
Almost. "pacman -S" is the equivalent of "apt install", gets official packages.
yaourt is a wrapper around pacman and makepkg. makepkg is used to build packages from community (AUR packages).

slurdge:
Oh. And I made a typo. It's "-Sy" (y == yes) not "-Su"

iceman1001:
thats your area of expertise...   *nix ppl unite!

PR #264, opened 2019-07-17
Fix build for linux on travis

slurdge:
And move to xenial

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
 ## Note: all dependencies on MacOS should be resolved by the brew install command
   if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then 
     sudo apt-get update -qq;
-    sudo apt-get install -y gcc-arm-none-eabi libnewlib-arm-none-eabi libnewlib-dev binutils-arm-none-eabi;
+    sudo apt-get install -y gcc-arm-none-eabi libnewlib-dev;
   elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
     brew update;
     brew tap RfidResearchGroup/proxmark3;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,16 @@
 # Travis-CI Build for RfidResearchGroup/Proxmark3
 language: c
 
-#default linux build env is: Ubuntu 14.04 trusty
+#default linux build env is: xenial
 compiler: gcc
 
 # Test on Linux and MacOS
 matrix:
   include:
-#    - os: osx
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -25,7 +19,7 @@ before_install:
 ## Note: all dependencies on MacOS should be resolved by the brew install command
   if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then 
     sudo apt-get update -qq;
-    sudo apt-get install -y gcc-arm-none-eabi;
+    sudo apt-get install -y gcc-arm-none-eabi libnewlib-arm-none-eabi libnewlib-dev binutils-arm-none-eabi;
```
doegox: Current ubuntu-alike instructions is to install
* gcc-arm-none-eabi libstdc++-arm-none-eabi-newlib

while here you install
* gcc-arm-none-eabi libnewlib-arm-none-eabi libnewlib-dev binutils-arm-none-eabi

I'd like to see both aligned. 
libnewlib-arm-none-eabi are libnewlib-dev dependencies of libstdc++-arm-none-eabi-newlib and binutils-arm-none-eabi a recommendation of libstdc++-arm-none-eabi-newlib but maybe this means libstdc++-arm-none-eabi-newlib itself wasn't needed. We should test it a bit further

slurdge: I think stdc++ is not needed. there is no c++ code AFAIK.
I just took the docker xenial and added deps until it compiled. Strangely the different ubuntu flavors do not package the same way arm-non-eabi.

doegox: I tested on a docker ubuntu:xenial
* you can skip binutils-arm-none-eabi, it's a dependency of gcc-arm-none-eabi
* it seems libstdc++-arm-none-eabi-newlib wasn't required.
* but it seems also that libnewlib-arm-none-eabi is not required, only libnewlib-dev

So ``apt-get install gcc-arm-none-eabi libnewlib-dev``
would suffice apparently
I tried removing libnewlib-arm-none-eabi and libstdc++-arm-none-eabi-newlib on my Debian and compile/flash/use runs fine

iceman1001:
Yeah,  I was curious to the new dependecies...
I guess xenial is ubuntu 16.04.  So go go go

slurdge:
Done. OSX still compiling but it's good for linux. We can enable travis and maybe even publish artifacts when we tag a release :)

doegox:
thanks ! @iceman1001 you can enable travis on this repo :)

iceman1001:
Since I am not a CI expert.  How do I enable travis CI?  Thought it needed a seperate branch etc?

PR #269, opened 2019-07-18
Update default_keys.dic with new keys from an APK

luminouw:
Following PR #238 I added some keys extracted from the Scan Badge APK that weren't present in the file and completing @3ldidi94 work with just 4 new keys.


client/default_keys.dic
```diff
@@ -944,5 +944,9 @@ A23456789123
 A00003000084
 675A32413770
 395244733978
+675A32413770
+395244733978
```
iceman1001: isn't these two duplicates from above?

luminouw: Yep, totally messed up the repos and commit here : D
Missing keys are:
4243414F5250
DFE73BE48AC6

PR #270, opened 2019-07-18
Update default_keys.dic with new keys from Scan Badge APK

luminouw:
Following PR #238 I added some keys extracted from the Scan Badge APK that weren't present in the file and completed @3ldidi94 previous work with 4 new keys.


RfidResearchGroup:
Don't forget to add a line on CHANGEME.md

PR #276, opened 2019-07-18
Flash 512k

slurdge:
Feel free to commend and/or rework.
Main benefit is in bootrom.c
The flasher had to be reworked a bit (connecting first) because it need to query the actual hardware. Other solutions (commandline switch, embedding flash size info inside elf, etc.) should be possible in the future.
We can also prevent _from_ bootrom the flashing outside the original region (since we have the chipid).
Works on my PM3RDV4. If you want to test, move the THUMB_SRC to ARM_SRC libs inside armsrc/Makefile.

### DIFF
--- a/bootrom/bootrom.c
+++ b/bootrom/bootrom.c
@@ -153,28 +153,35 @@ void UsbPacketReceived(uint8_t *packet, int len) {
         case CMD_FINISH_WRITE: {
             uint32_t *flash_mem = (uint32_t *)(&_flash_start);
             for (int j = 0; j < 2; j++) {
-                for (i = 0 + (64 * j); i < 64 + (64 * j); i++) {
-                    flash_mem[i] = c->d.asDwords[i];
-                }
-
                 uint32_t flash_address = arg0 + (0x100 * j);
-
+                AT91PS_EFC efc_bank = AT91C_BASE_EFC0;
+                int offset = 0;
+                uint32_t page_n = (flash_address - ((uint32_t)flash_mem)) / AT91C_IFLASH_PAGE_SIZE;
+                if (page_n >= 1024) {
+                    page_n -= 1024;
+                    efc_bank = AT91C_BASE_EFC1;
+                    // We need to offset the writes or it will not fill the correct bank write buffer.
+                    offset = 1024 * AT91C_IFLASH_PAGE_SIZE / 4;
+                }
+                for (i = 0 + (64 * j); i < 64 + (64 * j); i++) {
+                    flash_mem[offset+i] = c->d.asDwords[i];
+                }
+                
                 /* Check that the address that we are supposed to write to is within our allowed region */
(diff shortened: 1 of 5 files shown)


doegox:
don't forget the changelog for well deserved eternal gratitude :+1:

bootrom/bootrom.c
```diff
@@ -153,28 +153,35 @@ void UsbPacketReceived(uint8_t *packet, int len) {
...
             for (int j = 0; j < 2; j++) {
+                uint32_t flash_address = arg0 + (0x100 * j);
+                AT91PS_EFC efc_bank = AT91C_BASE_EFC0;
+                int offset = 0;
+                uint32_t page_n = (flash_address - ((uint32_t)flash_mem)) / AT91C_IFLASH_PAGE_SIZE;
+                if (page_n >= 1024) {
+                    page_n -= 1024;
+                    efc_bank = AT91C_BASE_EFC1;
+                    // We need to offset the writes or it will not fill the correct bank write buffer.
+                    offset = 1024 * AT91C_IFLASH_PAGE_SIZE / 4;
```
doegox: where did you find that magic offset value? Is there another more explicit way to express it?

client/flash.c
```diff
@@ -362,15 +356,20 @@ int flash_start_flashing(int enable_bl_writes, char *serial_port_name, uint32_t
         *chipinfo = resp.oldarg[0];
     }
 
+    uint32_t flash_end = FLASH_START + 256 * 1024;
```
doegox: what about FLASH_START + AT91C_IFLASH_PAGE_SIZE * AT91C_IFLASH_NB_OF_PAGES / 2 ?

client/flash.c
```diff
@@ -362,15 +356,20 @@ int flash_start_flashing(int enable_bl_writes, char *serial_port_name, uint32_t
...
+    if (((*chipinfo & 0xF00) >> 8) > 9) {
+        flash_end = FLASH_START + 512 * 1024;
```
doegox: FLASH_START + AT91C_IFLASH_PAGE_SIZE * AT91C_IFLASH_NB_OF_PAGES ?

bootrom/bootrom.c
doegox: ha ok I get it, it's to reconstruct the real address, I was puzzled by the /4 but that's because flash_mem is in dwords.
maybe write it as:
offset = (AT91C_IFLASH_NB_OF_PAGES / 2) * AT91C_IFLASH_PAGE_SIZE / sizeof(uint32_t);

slurdge:
Done. I checked on 512K and 256K (both <256K and >256K firmwares).

doegox:
Thank you!
I merged it by squashing both commits in one, to avoid people landing on an untested bootloader in case of git bisect or alike.

PR #281, opened 2019-07-19
BL versionning

slurdge:
This pull request adds basic bootloader versioning to help with migration. For now, the bootloader does not reports it understand the versionning and thus the flasher restricts the flashable area to 256k.

### DIFF
--- a/client/flash.c
+++ b/client/flash.c
@@ -356,8 +356,30 @@ int flash_start_flashing(int enable_bl_writes, char *serial_port_name, uint32_t 
         *chipinfo = resp.oldarg[0];
     }
 
+    int version = BL_VERSION_INVALID;
+    if (state & DEVICE_INFO_FLAG_UNDERSTANDS_VERSION) {
+        SendCommandBL(CMD_BL_VERSION, 0, 0, 0, NULL, 0);
+        PacketResponseNG resp;
+        WaitForResponse(CMD_BL_VERSION, &resp);
+        version = resp.oldarg[0];
+        if ((version < BL_VERSION_FIRST || version > BL_VERSION_LAST)) {
+            version = BL_VERSION_INVALID;
+        }
+    } else {
+        PrintAndLogEx(ERR, _RED_("Note: Your bootloader does not understand the new CMD_BL_VERSION command"));
+        PrintAndLogEx(ERR, _RED_("It is recommended that you update your bootloader") "\n");
+    }
+
+    bool allow_512k_writes = false;
+    if (version == BL_VERSION_INVALID) {
+        PrintAndLogEx(ERR, _RED_("Note: Your bootloader reported an invalid version number"));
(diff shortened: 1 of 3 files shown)


client/flash.c
```diff
@@ -356,8 +356,30 @@ int flash_start_flashing(int enable_bl_writes, char *serial_port_name, uint32_t
...
+            version = BL_VERSION_INVALID;
+        }
+    } else {
+        PrintAndLogEx(ERR, _RED_("Note: Your bootloader does not understand the new CMD_BL_VERSION command"));
+        PrintAndLogEx(ERR, _RED_("It is recommended that you update your bootloader") "\n");
+    }
+
+    bool allow_512k_writes = false;
+    if (version == BL_VERSION_INVALID) {
+        PrintAndLogEx(ERR, _RED_("Note: Your bootloader reported an invalid version number"));
```
doegox: Rather "a version not supported by the current flasher" ?
And maybe don't display this error message in cases when we already displayed "does not understand the new CMD_BL_VERSION"

client/flash.c
```diff
@@ -356,8 +356,30 @@ int flash_start_flashing(int enable_bl_writes, char *serial_port_name, uint32_t
...
+        PrintAndLogEx(ERR, _RED_("It is recommended that you update your bootloader") "\n");
```
doegox: I propose to change the msg for all occurrences of "you update your bootloader"
=> "you first update your bootloader alone, reboot the Proxmark3 then only update the main firmware"
Just to be on the safe side...

include/pm3_cmd.h
```diff
@@ -536,6 +537,15 @@ typedef struct {
 /* Set if this device understands the chip info command */
 #define DEVICE_INFO_FLAG_UNDERSTANDS_CHIP_INFO       (1<<5)
 
+/* Set if this device understands the version command */
+#define DEVICE_INFO_FLAG_UNDERSTANDS_VERSION         (1<<6)
+
+// Different versions here. Each version should increse the number
+#define BL_VERSION_INVALID  0
+#define BL_VERSION_1_0_0    1
```
doegox: I'm a bit puzzled between:
* the macro, using a major_minor_patch representation
* the actual value, not meant to be parsed as maj/min/patch (I'd expect sth like  1<<16+0<<8+0), maybe even just a BL_MAKE_VERSION(maj,min,patch) macro, see e.g. https://vulkan.lunarg.com/doc/view/1.0.30.0/linux/vkspec.chunked/apds03.html
* the BL_VERSION_FIRST/LAST where I would have expected a simple check on the major, such as BL_VERSION_MAJOR(version)==X

doegox:
FTR we're testing further on https://github.com/RfidResearchGroup/proxmark3/tree/bl_versionning so this won't come through this PR, closing it.

Issue #257, opened 2019-07-16
[Pull Requested] Implement SPIFFS


cjbrigato:
**Is your feature request related to a problem? Please describe.**
No. Unless we consider the low level way we manage flash a problem
**Describe the solution you'd like**
Implementation of Spiffs on customizable part of flash (64kb 128kb or 256kb)
**Describe alternatives you've considered**
Carrot cake. Did not yield to any comparable results'

**Additional context**
This is in fact my own request since I've managed to implement it with 256bytes logic pages and 4k blocks, write read and temporal cachrme with all three sizes. I'm currently opening this for further ideas and discussion on this idea and how things can go since this open a lot of new possibilities.
For the record, the implementation takes an overall 1kb ram with caches since we don't need more than one fd at a time. I think a full feature 4fd implementation will not be more than 2kb ram.
Binary size is another beast though, since this singleton instance will not be less than a 20kb size in flash. However a read-only version if we decide to let the write function a client thing is then less than 9kb

Let's discuss to make a PR not subject to immediate 50% invalid :)

cjbrigato:
Why did you refactor everything just when I started this work ? x)
Anyway, some PR might be incoming

cjbrigato:
Blocked as of #258 and #255

iceman1001:
You shouldn't be blocked anymore!

You got busy with your work before you could PR your stuff last autumn :)
But this time the device works much better than then.

cjbrigato:
So much. Unfortunately I had forgotten why the HAS_512_FLASH did not make it and so it is now bricked.
No interface available until this night so yeah, It suspended right now :(

iceman1001:
Larger than 256kb using the flasher is bad..
until fixed,  use jtag to flash 256> sized fullimages

doegox:
yes and SPIFFS implies > 256k so maybe priority is to manage to flash properly 512k from bootrom. I tried briefly one day to look at it and failed miserably...

iceman1001:
yeah.. its the swapping to bank1 after 256kb, that I failed.  I dunno how large each write is...

slurdge:
Don't feel bad, I'll brick mine soon enough 😄 
Do you have any place where the faulty swap is ? So I don't try it too...

iceman1001:
flash.c?  :)

cjbrigato:
[...]

PR #283, opened 2019-07-22
Early but functionnal SPIFFS implementation 

cjbrigato:
see #257, UPDATE HF_COLIN accordingly

Was a complicated merge, please correct any mistake or consequences on standard files (Makefile etc).

Let's discuss !

### DIFF
--- a/armsrc/Makefile
+++ b/armsrc/Makefile
@@ -34,6 +34,7 @@ SRC_CRAPTO1 = crypto1.c des.c desfire_key.c desfire_crypto.c mifaredesfire.c aes
 SRC_CRC = crc.c crc16.c crc32.c
 SRC_ICLASS = iclass.c optimized_cipher.c
 SRC_LEGIC = legicrf.c legicrfsim.c legic_prng.c
+SRC_SPIFFS = spiffs.c spiffs_cache.c spiffs_check.c spiffs_gc.c spiffs_nucleus.c spiffs_hydrogen.c
 #   SRC_BEE = bee.c
 
 # RDV40 related hardware support
@@ -97,6 +98,7 @@ THUMBSRC = start.c \
     $(SRC_SMARTCARD) \
     $(SRC_FPC) \
     $(SRC_HITAG) \
+    $(SRC_SPIFFS) \
     appmain.c \
     printf.c \
     commonutil.c \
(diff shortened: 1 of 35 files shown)


iceman1001:
Awesome!
Did you run a  MAKE STYLE ?    @doegox  did a neat addition to the makefile so we can enforce better codestyle in a easy way :)

cjbrigato:
Oh no I did not. Definitely nice addition.
Not around for some hours, is it blocking a review anyhow?

doegox:
Good job @cjbrigato !!
ok, far too much for a static review. I reviewed the parts interacting with the existing elements and I tested on my device. Let's merge and fix later whatever issues we could find. It shouldn't break anything in the existing features.

iceman1001:
thats the spirit!

cjbrigato:
Thanks guys! Next pr will be a rework of the standalone which will take configuration file from the spiffs regarding the schemes so there will be no more static schemes and it will be extensible. May even be able to create its own config files when it find new schemes :)

iceman1001:
now that here is the start of a new wave of standalone / device functions.  I see all dump can now be as fast as in your standalone mode.  current mf dump is so slow.

cjbrigato:
Yeah good idea that we can use the spiffs as a buffer for such transfer. But regarding the dump function I never used it : provide the key to emulator, but I use ecfill, and save emulator memory to file. It was always very quick, well at least with your fork in old days before the necessary was merged on original repo. Anyway spiffs has no real implication on the dumping speed if you use this method.
I've already made a to my standalone for having support needed for reading any schemes through json configuration file. This will be sick addition I'm sure.
Next will be to automatically add the new found schemes to the configuration when we find all keys without already a known scheme. 
I will add default keys as dynamically read from the device, and use circular buffer to test them all and so as a result we should definitely get through a standalone mode very powerful :)
I'll maintain some stats in another file so we know which scheme is actually a singular and anecdotical one and which one is more than that.
I'll add some connected mode to automatically upload results and 'ew schemes once bt connection is made. And then a connected mode where you can switch on a mode which add a darkside/nested triggering when no known keys are found or if there are some missing ones and you are in situation when you can afford these 30sec for such attack.

iceman1001:
current hf mf dump uses x calls to read accessrights  from each sector.  
Then it uses 1 call to read 1 block at the time from client to device. 
Much better to have all that on deviceside.  Read rights,  dump mem with the keys in a dump keys file?  and when every you want a copy ,  dump to the client is an simple download :) much faster.


your new json styled config was neat.   still,  BT transfers is slow,  so use NG format and keep comms to the minimun

PR #284, opened 2019-07-23
T55x7 Downlink mode support added

mwalker33:
This is a major update to the t55xx commands to support the different downlink modes.
Most options will support r <mode> where mode is:
0 - default, 
1- long leading reference, 
2 leading 0 
3 1 of 4 .
Some functions like password checks and detect support mode 4, which will try all 4 modes.
e.g. lf t55 detect r 4
This will try each mode until one works or all 4 have been tried.

I have updated the timing storage and reporting to store the single byte, rather then the value * 8.
e.g. we refer to a gap as 30 * 8.  I have applied the * 8 when used rather then store.  This saves a few bytes in both the code and flash storage needs (as there are now 4 sets of timings).
To keep code size down (and for discussion).  I removed the decimal value of the timings from the status report.  so less math and short message.  Since we always refer to the x * 8, not the result, I did not see the value for the space needed.
I added a z option to erase the timings from the flash.  this allows you to restore the code defaults (on reboot).

The flash stuct was updated to support the array of 4 modes.  The new stuct will also allow an easy way to add other fields if/when needed.  
e.g.
T55xxReadBlock function
how long to wait before acquire.
```
	// TurnReadLFOn(210*8); // issues with block 1 reads so dropping down seemed to help
	TurnReadLFOn(137*8); 
```
Some minor code clean up works.  and some tweaks i found gave better results in my testing on the rdv4


iceman1001:
Nice!
I dont really like the idea of saving a few bytes and enforce ppl to know that they need to multiply with 8 all the times they need to use a value from the config.
I like the old way better. 

There is already a  LF T55XX DEVICECONFIG command to use for setting values.  I suggest you use that instead.

mwalker33:
Ok, no problem with the * 8, happy to put back.
Not sure on the LF T55XX DEVICECONFIG, I was using that, I just extended to support a 4 sets of timing options.  Did I do something different ?
Side note: as per you issue comment re do more on chip for stand alone, do we want to proceed with this as its very client focused ?

iceman1001:
I usually say,  merge it and we take next PR for solving it even better.   I am very focused on getting functions and not spending energy on style.   :)  
There is also a  "make style" command,  the @doegox  added to the makefile in order for use to easily apply coding style on source.  If you didn't know about it,  try it out.  But not if you are on MINGW (only use on linux) 

There is a massive PR gonna be merge with SPIFFS,  so that one will get merged before this one.
You can already test so your changes doesn't get messed up

armsrc/lfops.c
```diff
@@ -1477,15 +1482,15 @@ void T55xxWriteBit(uint8_t bit, uint8_t downlink_idx) {
...
-		case 0 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_0 * 8);             break; // Send bit  0/00
-		case 1 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_1 * 8);             break; // Send bit  1/01
-		case 2 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_2 * 8);             break; // Send bits   10 (1 of 4)
-		case 3 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_3 * 8);             break; // Send bits   11 (1 of 4)
-		case 4 : TurnReadLFOn((T55xx_Timing.m[downlink_idx].write_0 + 136) * 8); break; // Send Long Leading Reference
+		case 0 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_0 );             break; // Send bit  0/00
+		case 1 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_1 );             break; // Send bit  1/01
+		case 2 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_2 );             break; // Send bits   10 (1 of 4)
+		case 3 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_3 );             break; // Send bits   11 (1 of 4)
+		case 4 : TurnReadLFOn(T55xx_Timing.m[downlink_idx].write_0 + (136 * 8));  break; // Send Long Leading Reference
```
iceman1001: I suggest a Define for this 136 * 8 value.  More descriptive

iceman1001:
looks like your PR didnt get broken with the SPIFFS one.  Great!

doegox:
Iceman told me ok for merging, so...
:)

iceman1001:
and a bit of cleaning up afterwards..  defines etc.. no biggie

PR #287, opened 2019-07-23
HF_COLIN : Makes Known Schemes in Struct form

cjbrigato:
This greatly DRY the code and is 1/2 Pass for having Schemes dynamicly read from SPIFFS as json.

### DIFF
--- a/armsrc/Standalone/hf_colin.c
+++ b/armsrc/Standalone/hf_colin.c
@@ -27,39 +27,78 @@ int curlline;
 
 // Colin's VIGIKPWN sniff/simulate/clone repeat routine for HF Mifare
 
-/*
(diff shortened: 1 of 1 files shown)


cjbrigato:
Note : This also reduce the overall footprint
before : 
```
   text	   data	    bss	    dec	    hex	filename
  10366	      0	      0	  10366	   287e	../../../spiffs/proxmark3/armsrc/obj/hf_colin.o
```
after : 
```
   text	   data	    bss	    dec	    hex	filename
   8062	    888	      4	   8954	   22fa	obj/hf_colin.o
```
This will be even better with pass 2/2. Might be a 3/2 pass for overall optimization footprint wise

iceman1001:
excellent!
a nice refactoring!

PR #288, opened 2019-07-25
Bug Fix : T55x7 timings save to flash

mwalker33:
This should now support the old flash saved settings for the t55x7 timings.  then allow the new ones to be added as needed.


mwalker33:
Hold on this till I get back.  I will change the z option from "erase" to set defaults in the client (as suggested).  and Auto set back to the pm3 if "FF"

mwalker33:
Ok, should be good to go when your ready.
Thanks

doegox:
thanks !

iceman1001:
Excellent!

PR #292, opened 2019-07-26
Update default_keys.dic

netvader:
added some keys from clever fit gym, hotel room turkey, and simonsvoss key card


iceman1001:
Thanks!

PR #298, opened 2019-07-29
Update termux_notes.md

netvader:
tested Termux successfully on Nexux 5X with LineageOS 15.1 which has USB_ACM enabled by default


doegox:
thanks!

PR #303, opened 2019-08-03
Fix password write offset by 1

### LINKED ISSUE #239: Hitag2 password write is offset by 1
`lf hitag write 1 00000000`
Expected write to page 1, actually writes to page 2.

Issue is in the `blocknr` incremention behavior in password auth, can be patched but I will try to refactor the whole thing the coming days.

ViRb3:
Fixes an issue introduced in #233, closes #239.
Now consistent with how crypto auth works.


iceman1001:
Excellent!

PR #304, opened 2019-08-03
Format and docs

ViRb3:
Formats and documents some hitag2 code, hopefully makes the logic easier to understand.


ViRb3:
Rebased on last merge to make the file changes easier to see.

iceman1001:
hm...
there is a command that @doegox  added which enforces our code style onto the source code.
Only run it on a *nix based OS,    **make style**

PR #305, opened 2019-08-03
Better warn user of hardcoded hitag info

ViRb3:
Got bitten by this, let's make it clear for other users.


iceman1001:
Dunno if you had to remove the text,     maybe add clear "not implemented"  text but keep your hardcoded warning text. its good.

ViRb3:
`Hitag2 tag information` is displayed as a part of `printHitagConfiguration`, so it was repeated twice.

iceman1001:
Ok Ok!

PR #308, opened 2019-08-04
FIXes # 174 and more (OSX > 10.4 num_Cpus) the good way

cjbrigato:
sysconf() is  part of IEEE Std 1003.1-1988 (POSIX.1).
Fortunately, it *is* available to OSX `>=10.5`.
At least on `~10.14`, the old unposix `sysctl` osx way was lost, at least in portable environments.
This fixes it, but will not be compatible with OSX > 10.4. 
So i'm sorry but here we loose compatibility with Macs olders than 2005. (🤣)

### DIFF
--- a/client/util.c
+++ b/client/util.c
@@ -9,7 +9,7 @@
 //-----------------------------------------------------------------------------
 
 // ensure gmtime_r is available even with -std=c99; must be included before
-#if !defined(_WIN32)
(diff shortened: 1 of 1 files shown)


iceman1001:
Great!
All OSX related issues is hard for me to deal with since I don't have a virtual image nor physical device.

cjbrigato:
@iceman1001 no worry I can take them with pleasure.

Issue #240, opened 2019-07-11
Last bit omitted in rawdemod


ViRb3:
Following @doegox's commands for manually reading a hitag2 UID:
```
lf cmdread d 50 z 166 o 116 c 000111
data ltrim 200
data norm
data rawdemod am
data printdemodbuffer o 5 x
```
Using the hitag2 module:
```
lf hitag read 26
// or
lf hitag info
// or
if hitag read 21 ... (makes dump)
```

Will result in the same UID, however, the last bit is `0` using the manual read, while it is a `1` using the hitag2 module. I tested this on two different cards, however, I cannot verify which one is the true UID. I suspect the rawdemod is at fault since the hitag2 module data has worked for me so far.

Potentially related:
* f1a0e57f65fc4715af15c7757d8a26b5df2ebcd
* https://github.com/RfidResearchGroup/proxmark3/issues/124#issuecomment-472800131

doegox:
Good catch, you're right.
I tried on a car key to get another ID and by chance its last bit is a 1.
My manual steps return an ID=xxxx4D6A but another reader (ACG-LF) returns xxxx4D6B.
Here we see the Manchester decoder stopped before the last weaker "1"(HL) bit:
![Screenshot from 2019-07-11 17-50-46](https://user-images.githubusercontent.com/60773/61066348-aa45be80-a405-11e9-97d4-89a820882255.png)

iceman1001:
there are most certain same issues with NR demod ..

iceman1001:
@ViRb3   Some news?

doegox:
😙

iceman1001:
@doegox   what is your output from car key and hitag today?

doegox:
Manual decoding gets still last bit wrong, but `lf hitag reader` works fine
```
[usb] pm3 --> lf cmdread d 50 z 166 o 116 c 000111
[+] Sending command...
[#] Done, saved 42216 out of 0 seen samples at 8 bits/sample

[+] downloading response signal data
[usb] pm3 --> data ltrim 200
[usb] pm3 --> data norm
[usb] pm3 --> data rawdemod am
11111011100110111101001001101011
0101.0
[usb] pm3 --> data printdemodbuffer o 5 x
[+] DemodBuffer: 737A4D6A[3]

[usb] pm3 --> lf hitag read 26
[+]  UID: 737a4d6b
```

iceman1001:
hitag support is ok now.

Issue #124, opened 2019-03-10
LF HITAG - fails 


iceman1001:
**Describe the bug**
the lf hitag commands fails to work

**To Reproduce**
1. put hitag2 tag on pm3 antenna
2. lf hitag reader 26
3. fails...

**Expected behavior**
fully functional / verified working card operations with pm3 and a hitag card.

**Desktop (please complete the following information):**
- rdv4
- latest source
- ubuntu / mingw

**Additional context**
This problem has existed for a while,  @doegox  brought this to my attention some days ago.
found a hitag2 card today and could verify that the hitag2 commands doesn't work.

iceman1001:
Looking at hitag.c in armsrc,  gives you that it uses its own clock definition.  
compared with the newly remade legic stuff from @drandreas  which also uses a custom clock but the code is neat and nice,   the hitag code is not on par with current coding style for pm3 when it comes to device (arm) usage.

iceman1001:
More ppl has also notice this,  see  https://github.com/Proxmark/proxmark3/issues/798
I don't think it has to do with i2c direct,  I would say it has to do with init of clocks etc.

iceman1001:
A missing TC0 init and bobs your uncle.

doegox:
Not quite :)

* RRG loaded on Pm3 Easy : :+1:
* RRG loaded on Pm3 rdv4 : :-1: 
  * With original highQ antenna : nothing
  * With lowQ antenna : sometimes it gets some part of the frame, but not all

iceman1001:
For it works for me.

-  RRG loaded on Pm3 rdv4 👍    LowQ antenna

iceman1001:
The reading distance is crap

iceman1001:
As seen below,   I need to position the tag until it reads.
```
pm3 --> lf hita read 26
#db# Configured for hitag2 reader
#db# Uknown frame length: 5
pm3 --> lf hita read 26
#db# Configured for hitag2 reader
#db# Uknown frame length: 31
pm3 --> lf hita read 26
#db# Configured for hitag2 reader
#db# Uknown frame length: 31
pm3 --> lf hita read 26
#db# Configured for hitag2 reader
#db# Uknown frame length: 2
pm3 --> lf hita read 26
#db# Configured for hitag2 reader
#db# Uknown frame length: 31
pm3 --> lf hita read 26
#db# Configured for hitag2 reader
Valid Hitag2 tag found - UID: ac20a810
[...]

PR #310, opened 2019-08-05
Rework hitag2 read/write help

ViRb3:
And add default keys

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Rework hitag2 read/write help (@ViRb3)
  - Add 'lf nedap' - encoding / decoding  (anon)
  - Add client option `-i` to stay in interactive mode after a script or command (@DidierStevens/@doegox)
  - Add VSCode tasks (@ViRb3)
--- a/client/cmdlfhitag.c
+++ b/client/cmdlfhitag.c
@@ -87,6 +87,7 @@ static int usage_hitag_reader(void) {
     PrintAndLogEx(NORMAL, "      01 <nr> <ar>     Read all pages, challenge mode");
     PrintAndLogEx(NORMAL, "      02 <key>         Read all pages, crypto mode. Set key=0 for no auth");
     PrintAndLogEx(NORMAL, "   Hitag1 (1*)");
+    PrintAndLogEx(NORMAL, "      Not implemented");
     PrintAndLogEx(NORMAL, "   Hitag2 (2*)");
     PrintAndLogEx(NORMAL, "      21 <password>    Read all pages, password mode. Default: 4D494B52 (\"MIKR\")");
     PrintAndLogEx(NORMAL, "      22 <nr> <ar>     Read all pages, challenge mode");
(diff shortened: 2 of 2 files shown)


iceman1001:
Hm.. so there is no write / read on hitag S?

ViRb3:
Sure there is? I only reworded the help text.

iceman1001:
Because your rewording removes "read all pages" etc from the help text..
So it that not true anymore?

ViRb3:
Oh, no, it's still valid. I just assumed that if it's under the `read` command it's obvious that it reads, and under the `write` command it writes. Think I should add that back to the help?

iceman1001:
Not sure its much clearer with your suggested changes.  
hitag reader...   acting as a reader doesn't mean read ,  read all pages,  etc..

ViRb3:
Sure, I'll re-write it to be clearer.

ViRb3:
How does that look like? :)

doegox:
looks good, thanks

pwpiwi:
Are there other default passwords? It doesn't work for me on a brand new Hitag2 tag:
```
[usb] pm3 --> lf hitag reader
Hitag reader functions
Usage: lf hitag reader [h] <reader function #>
Options:
       h               This help
   HitagS (0*)
      01 <nr> <ar>     Read all pages, challenge mode
      02 <key>         Read all pages, crypto mode. Set key=0 for no auth
   Hitag1 (1*)
      Not implemented
   Hitag2 (2*)
      21 <password>    Read all pages, password mode. Default: 4D494B52 ("MIKR")
      22 <nr> <ar>     Read all pages, challenge mode
      23 <key>         Read all pages, crypto mode. Key format: ISK high + ISK low. Default: 4F4E4D494B52 ("ONMIKR")
      25               Test recorded authentications
      26               Just read UID
[usb] pm3 --> lf hitag reader 21 4d494b52
#db# List identifier in password mode
#db# Password failed!
[usb] pm3 --> lf hitag reader 26
[+] Valid Hitag2 tag found - UID: aa8e6b11
```

iceman1001:
Better to use forum for questions.

ViRb3:
@pwpiwi those are the only default keys I'm afraid. I can confirm they work on my new Hitag2 cards. The distance and placement of the tag to the antenna is incredibly important, try moving it around and retrying the read.

pwpiwi:
Indeed, I could make it work after some fiddling! While `lf hitag 26` works at different distances, `lf hitag 23` is more sensitive to distance (which makes me think that sending is the issue rather than receiving). In general, possible distances are quite low compared to commercial LF readers.

doegox:
Indeed @pwpiwi lf hitag suffers from two problems: reading distance (while manually I could achieve much large distances) and last bit wrong/missing, see https://github.com/RfidResearchGroup/proxmark3/issues/124 and https://github.com/RfidResearchGroup/proxmark3/issues/240

ViRb3:
@pwpiwi any chance your could send that upstream hitag love back to this repo? I could use your refactor if I'm doing more work on hitag.

pwpiwi:
No need to send the latest changes from official to RRG. The PR adds latest additions (password write) formatting and refactoring from RRG to official. 😃

ViRb3:
I meant your touches, e.g. the commands extracted to the protocols header. That's good stuff :)

pwpiwi:
I had added this more or less for documentation only. Feel free to copy & paste. As discussed in #18 there is no easy way to reuse PRs and/or commits.

PR #311, opened 2019-08-06
Update Android Termux guide

ViRb3:
No more ncurses hacking, no more Makefile modifications!


iceman1001:
Why?  What has changed?

ViRb3:
Termcap was included in the official ncurses package, so no need to build and install it manually anymore. Also compilation succeeds without adding -fPIC to the Makefile.

netvader:
I can confirm that, works perfectly, thanks for the update!

doegox:
yep, tested, thanks.

iceman1001:
Excellent!

PR #312, opened 2019-08-09
summer restructuring:

doegox:
Here is a huge pile of changes, trying to get some order, especially in the mess of header files...
Unfortunately the process could not be really split into nice step by step commits as everything happened at the same time in some broken intermediate stages...

We can still factorize stuffs, but at least that's a start.

If you feel ok with it, we really need to test as many platforms as possible as header files can be platform-dependent and I have very probably broken stuffs on some platforms.

Tested on:
* all standalone modes
* Linux
* Mingw
* WSL

@iceman1001 @cjbrigato please help me with windows and macos :)

* .h include only the strict minimum for their own parsing
  * this forces all files to include explicitment their needs and not count on far streched dependencies
  * this helps Makefile to rebuild only the minimum
  * according to this rule, most standalone .h are now gone
  * big app.h is gone
  * remove seldom __cplusplus, if c++ happens, everything will have to be done properly anyway
* all non required include were removed
* split common/ into common/ (client+arm) and common_arm/ (os+bootloader)
  * bring zlib to common/
  * bring stuff not really/not yet used in common back to armsrc/ or client/
  * bring liblua into client/
  * bring uart into client/
  * move some portions of code around (dbprint, protocols,...)
* rename unused files into *_disabled.[ch] to make it explicit
* rename soft Uarts between 14a, 14b and iclass, so a standalone could use several without clash
* remove PrintAndLogDevice
* move deprecated-hid-flasher from client to tools
* Makefiles
  * clean now removes .d in client & armsrc => now in master too
  * fix nonce2key typo => now in master too
  * treat deps in armsrc/ as in client/
  * client: stop on warning (-Werror), same as for armsrc/


doegox:
ok, thanks CI, apparently it's ok on osx, but not on mingw

mwalker33:
Hi, would you like me to test anything (win10 user)?  

While you are doing some clean up, an (old) typo in the client\cmdhfmfdes.c
Approx Line 361
Assume it should read "... to change any key ..." not "... to change an ley..."

  str = "Authentication with the specified key is necessary to change any ley. A change key and a PICC master key (CMK) can only be changed after authentication with the master key. For keys other then the master or change key, an authentication with the same key is needed.";

doegox:
> Hi, would you like me to test anything (win10 user)?

Yes you can try this branch with your usual configuration or anything you could think of and see if any problem pops up.
I already did some tests locally on mingw and wsl and Appveyor CI is running fine too so this should work

> While you are doing some clean up, an (old) typo in the client\cmdhfmfdes.c

Thanks, pushed to master.

doegox:
stupid git couldn't merge with "conflict" while it could rebase the branch on master without any troubles...
So I force-pushed the rebased version.

mwalker33:
Win10 Setup.  All seems ok (quick test)
Run make clean and make all
Flashed OS (not tested boot ) - OK
Run a quick T55xx test - OK
Run a quick Mifare test - OK

So, good work.

Only 2 very minor things popped up during make (which did not affect compile and run)
(I downloaded the zip and expanded, so not sure it that was the cause v using a git client)

[=] LD flasher
[*] MAKE bootrom/all
[=] GEN version.c
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
[-] CC version.c


[-] CC cmd.c
[-] GEN version.c
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
[-] CC version.c
[=] LD obj/fullimage.stage1.elf

doegox:
Thanks for your tests!

doegox:
Tested on Android termux too
I'll merge soon as discussed with iceman.
Pushing first a rebase to not mess up history too much, as this one is quite larger than usual.

iceman1001:
....  I will start testing again...  
Awesome that you did this much needed restructuring.  I know some of you will think its breaking,  but we can never reach our goal to create a **libpm3**  if we don't do this kind of changes.

More to come.

PR #334, opened 2019-08-18
add make clean for linux

merlokk:


iceman1001:
How many cores / threads can we play with for travis?
Since we don't need to do  **make all** anylonger.   

if its possible for running  **make -j8**  ?

merlokk:
it looks it have only 2 cores. https://docs.travis-ci.com/user/reference/overview/

PR #338, opened 2019-08-18
Doc: Add notes about trace to Wireshark export

uhei:
Adding some notes how to bring trace output into Wireshark.


iceman1001:
Nice!

PR #340, opened 2019-08-19
PCF7931: Print found single/consecutive block(s), fixes to block 1 ch…

sguerrini97:
…eck and some comments

Printing blocks makes it easier to manually reorder blocks when block 0 and 1 are not transmitted


iceman1001:
Dont forget to add a line in CHANGELOG

PR #343, opened 2019-08-19
Client: fixed build on osx

drandreas:
On my system the build of the latest version failed with implicit declaration errors. Including `ctype.h` fixed the error for me.

### DIFF
--- a/client/cmdhflegic.c
+++ b/client/cmdhflegic.c
@@ -9,6 +9,7 @@
 //-----------------------------------------------------------------------------
 #include "cmdhflegic.h"
 
+#include <ctype.h>
 #include <stdio.h> // for Mingw readline
 #include <readline/readline.h>
 
--- a/client/proxmark3.c
+++ b/client/proxmark3.c
@@ -13,8 +13,10 @@
 #include <limits.h>
 
 #include <stdio.h> // for Mingw readline
+#include <string.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <ctype.h>
 #include <readline/readline.h>
 #include <readline/history.h>
 
--- a/client/ui.c
+++ b/client/ui.c
@@ -20,6 +20,7 @@
 
 #include <stdio.h> // for Mingw readline
 #include <stdarg.h>
+#include <string.h>
 #include <stdlib.h>
 #include <readline/readline.h>
 #include <complex.h>
@@ -155,8 +156,10 @@ void PrintAndLogEx(logLevel_t level, const char *fmt, ...) {
 }
 
 static void fPrintAndLog(FILE *stream, const char *fmt, ...) {
+#ifdef RL_STATE_READCMD
     char *saved_line;
     int saved_point;
+#endif
     va_list argptr;
     static FILE *logfile = NULL;
     static int logging = 1;


doegox:
And your changes break all platforms here: linux, OSX, Windows...

doegox:
Can you explain your osx env? Because here on xcode11 and on my Sierra VM things work properly so we need to understand which platform we should test more for the future... Thanks

drandreas:
Jup, travis just told me :-)

I've now wrapped `saved_line` and `saved_point` in the  `RL_STATE_READCMD` ifdef. That should fix the issue also for users with the `RL_STATE_READCMD` not set.

drandreas:
I fixed the following errors:
```
===================================================================
Platform name:     Proxmark3 rdv4
PLATFORM:          PM3RDV4
Platform extras:   No extra selected
Included options:  SMARTCARD FLASH LF HITAG ISO15693 LEGICRF ISO14443b ISO14443a ICLASS FELICA NFCBARCODE HFSNIFF
Standalone mode:   LF_SAMYRUN
===================================================================
[*] MAKE client/all
[-] CC proxmark3.c
proxmark3.c:79:26: error: implicitly declaring library function 'strlen' with type 'unsigned long (const char *)' [-Werror,-Wimplicit-function-declaration]
        script_cmd_len = strlen(script_cmd);
                         ^
proxmark3.c:79:26: note: include the header <string.h> or explicitly provide a declaration for 'strlen'
proxmark3.c:115:13: error: implicitly declaring library function 'memset' with type 'void *(void *, int, unsigned long)' [-Werror,-Wimplicit-function-declaration]
            memset(script_cmd_buf, 0, sizeof(script_cmd_buf));
            ^
proxmark3.c:115:13: note: include the header <string.h> or explicitly provide a declaration for 'memset'
proxmark3.c:161:21: error: use of undeclared identifier 'rl_event_hook'
                    rl_event_hook = check_comm;
                    ^
proxmark3.c:180:29: error: implicitly declaring library function 'isspace' with type 'int (int)' [-Werror,-Wimplicit-function-declaration]
            while (l > 0 && isspace(cmd[l - 1])) {
                            ^
proxmark3.c:180:29: note: include the header <ctype.h> or explicitly provide a declaration for 'isspace'
proxmark3.c:198:34: error: implicitly declaring library function 'strcmp' with type 'int (const char *, const char *)' [-Werror,-Wimplicit-function-declaration]
                if ((!entry) || (strcmp(entry->line, cmd) != 0))
                                 ^
proxmark3.c:198:34: note: include the header <string.h> or explicitly provide a declaration for 'strcmp'
proxmark3.c:263:9: error: implicitly declaring library function 'strncpy' with type 'char *(char *, const char *, unsigned long)' [-Werror,-Wimplicit-function-declaration]
        strncpy(my_executable_directory, my_executable_path, dirname_length + 1);
        ^
proxmark3.c:263:9: note: include the header <string.h> or explicitly provide a declaration for 'strncpy'
proxmark3.c:494:21: error: implicitly declaring library function 'strcpy' with type 'char *(char *, const char *)' [-Werror,-Wimplicit-function-declaration]
                    strcpy(ctmp, "script run ");
                    ^
proxmark3.c:494:21: note: include the header <string.h> or explicitly provide a declaration for 'strcpy'
7 errors generated.
make[1]: *** [obj/proxmark3.o] Error 1
make: *** [client/all] Error 2
```

iceman1001:
on which OS X are you running?

doegox:
We need to understand why you got these errors and see what is different in your environment.
* OSX version?
* Xcode version?
* gcc --version ?
* did you use the brew tap RfidResearchGroup/proxmark3 to install the requirements or did you get them from elsewhere? Especially readline.

drandreas:
I pulled latest from github, so `git clone ... && make clean && make all`. The readline issue is strange. I will look into it further first and report back next month.

doegox:
@iceman1001 told me you're running on Catalina.
I tried today on Catalina, it works fine.
I bet you didn't `brew install readline` and you're compiling against the macOS BSD libedit instead.

drandreas:
Sorry, for wasting your time. Non released versions of Mac OS are not supposed to be discussed in public. I was talking to iceman on a private channel than he asked me to do a PR. This in turn catched your attention.

You are right - i didn't install readline using brew. I have installed it using mac ports. All other forks of proxmarkv3 compile fine on my machine (official, iceman).

I assume f4c810f83bbed7e9fbc332646210ac9d46a5bedd broke the build on my system. What I do not yet understand: How do your makefiles detect the path where readline is installed in (mac ports as well as brew are using /opt/local as far as I know)?

doegox:
So far, official way is using brew.
And makefile adds the path to brew readline on OSX.
If you have a different setup you should be able to provide your own cflags. 
If the current makefile is not flexible enough for you to inject needed flags, tell me and we'll fix it.
But we won't hardcore some /opt for everyone just because some have specific setups.

iceman1001:
Anyone with developer account can download the catalina beta.  Some ppl even made vmware images for download.  

The change to Makefiles,  in order to become better for package maintainers of pentoo, gentoo, nethunter,  and in the long run creating a  system install,  aka  "make install"   is quite large.  So we want to test as much on what we can.  When someone says it doesn't work,  we have to dig into it to understand what happend.  

One step with this is the disconnecting of libs from pm3, as an example.
the termcap,  which has its own issue here,
Some Linux (e..g. Ubuntu) doesn't need -ltermcap,   can use -lncurses
Some Linux  (e.g. CentOS 5.5) need -ltermcap and don't have -lncurses.

So we have quite a few OS's to support, like Android, Raspberry, OS X, WSL, MINGW and we need to take in consideration when doing this.  Travis / Appveryor can check some for us,  and the rest is just what the community tells us about.

Non of us has OSX,   and so far we was happy with just caring about homebrew. 
We would need someone with OSX to help out adding consistant support for both Homebrew and mac ports since we can't do it. 
Asking how we are going to support things on OS X given these limitations, you tell me.

PR #353, opened 2019-08-22
Autopwn

matthiaskonrath:
I implemented a **hardautopwn** feature, that breaks all sector keys automatically.
It can be used with dictionaries, or / and a user provided sector key.
Every found key is then used to check if it works on other sectors (this helps limiting the amount of hardnested attacks needed to break all the sectors).

Screenshot from 2019-08-22 15-05-28

Screenshot from 2019-08-22 15-05-55

Screenshot from 2019-08-22 15-07-34

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -2059,7 +2059,7 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
     // Transfere the found keys to the simulator and dump the keys and card data
     PrintAndLogEx(NORMAL, "");
     PrintAndLogEx(INFO, "Dumping the keys:");
-    createKeyDump(sectors_cnt, e_sector, GenerateFilename("hf-mf-", "-key.bin"));
+    createMfcKeyDump(sectors_cnt, e_sector, GenerateFilename("hf-mf-", "-key.bin"));
     
     PrintAndLogEx(SUCCESS, "Transfering the found keys to the simulator memory (Cmd Error: 04 can occour, but this shouldn't be a problem)");
     for (current_sector_i=0; current_sector_i < sectors_cnt; current_sector_i++) {
@@ -2352,7 +2352,7 @@ out:
 
         if (createDumpFile) {
             fptr = GenerateFilename("hf-mf-", "-key.bin");
-            createKeyDump(sectorsCnt, e_sector, fptr);
+            createMfcKeyDump(sectorsCnt, e_sector, fptr);
         }
     }
 
@@ -2378,7 +2378,6 @@ static int CmdHF14AMfChk(const char *Cmd) {
     uint8_t keyType = 0;
     uint32_t keyitems = ARRAYLEN(g_mifare_default_keys);
     uint64_t key64 = 0;
-    uint8_t tempkey[6] = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
     char *fptr;
     int clen = 0;
     int transferToEml = 0;
@@ -2633,31 +2632,7 @@ out:
 
     if (createDumpFile) {
         fptr = GenerateFilename("hf-mf-", "-key.bin");
-        if (fptr == NULL) {
-            free(keyBlock);
-            free(e_sector);
-            return PM3_EFILE;
-        }
-
-        FILE *fkeys = fopen(fptr, "wb");
-        if (fkeys == NULL) {
-            PrintAndLogEx(WARNING, "Could not create file " _YELLOW_("%s"), fptr);
-            free(keyBlock);
-            free(e_sector);
-            return PM3_EFILE;
-        }
-        PrintAndLogEx(INFO, "Printing keys to binary file " _YELLOW_("%s")"...", fptr);
-
-        for (i = 0; i < SectorsCnt; i++) {
-            num_to_bytes(e_sector[i].Key[0], 6, tempkey);
-            fwrite(tempkey, 1, 6, fkeys);
-        }
-        for (i = 0; i < SectorsCnt; i++) {
-            num_to_bytes(e_sector[i].Key[1], 6, tempkey);
-            fwrite(tempkey, 1, 6, fkeys);
-        }
-        fclose(fkeys);
-        PrintAndLogEx(SUCCESS, "Found keys have been dumped to file " _YELLOW_("%s")". 0xffffffffffff has been inserted for unknown keys.", fptr);
+        createMfcKeyDump(SectorsCnt, e_sector, fptr);
     }
 
     free(keyBlock);
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -166,20 +166,24 @@ static int usage_hf14_hardnested(void) {
     PrintAndLogEx(NORMAL, "      hf mf hardnested 0 A A0A1A2A3A4A5 4 A FFFFFFFFFFFF");
     return 0;
 }
-static int usage_hf14_hardautopwn(void) {
+static int usage_hf14_autopwn(void) {
     PrintAndLogEx(NORMAL, "Usage:");
-    PrintAndLogEx(NORMAL, "      hf mf hardautopwn [k] <sector number> <key A|B> <key (12 hex symbols)>");
-    PrintAndLogEx(NORMAL, "                       * <card memory> [d] [f] [s] [i] [l]");
+    PrintAndLogEx(NORMAL, "      hf mf autopwn [k] <sector number> <key A|B> <key (12 hex symbols)>");
+    PrintAndLogEx(NORMAL, "                       [*] <card memory> [f] <dictionary>.dic [s] [i] [l]");
     PrintAndLogEx(NORMAL, "                       (card memory - 0 - MINI(320 bytes), 1 - 1K, 2 - 2K, 4 - 4K, <other> - 1K)");
+    PrintAndLogEx(NORMAL, "");
+    PrintAndLogEx(NORMAL, "Description:");
+    PrintAndLogEx(NORMAL, "      This command is used to automate the attack process on mifare classic nfc cards.");
+    PrintAndLogEx(NORMAL, "      The program tries to identify the prng type and then automatically attack it with the best algorithm.");
+    PrintAndLogEx(NORMAL, "      If all the sector keys are broken, they will be dumped to a file.");
     PrintAndLogEx(NORMAL, "");
     PrintAndLogEx(NORMAL, "Options:");
     PrintAndLogEx(NORMAL, "      h                          this help");
     PrintAndLogEx(NORMAL, "      k <sector> <keytype> <key> if a known key for a block is supplied");
-    PrintAndLogEx(NORMAL, "      d                          write keys to binary file");
     PrintAndLogEx(NORMAL, "      f <name>.dic               dictionary file for key discovery (the file has to end in .dic)");
-    PrintAndLogEx(NORMAL, "      s                          slower acquisition (required by some non standard cards)");
+    PrintAndLogEx(NORMAL, "      s                          slower acquisition (required by some non standard cards) for hardnested");
     PrintAndLogEx(NORMAL, "      l                          legacy mode (use the slow mfchk for the key enumeration)");
-    PrintAndLogEx(NORMAL, "      i <X>     set type of SIMD instructions. Without this flag programs autodetect it.");
+    PrintAndLogEx(NORMAL, "      i <X>     set type of SIMD instructions. Without this flag programs autodetect it. (for hardnested)");
     PrintAndLogEx(NORMAL, "        i 5   = AVX512");
     PrintAndLogEx(NORMAL, "        i 2   = AVX2");
     PrintAndLogEx(NORMAL, "        i a   = AVX");
@@ -188,13 +192,11 @@ static int usage_hf14_hardautopwn(void) {
     PrintAndLogEx(NORMAL, "        i n   = none (use CPU regular instruction set)");
     PrintAndLogEx(NORMAL, "");
     PrintAndLogEx(NORMAL, "Examples:");
-    PrintAndLogEx(NORMAL, "      hf mf hardautopwn d");
-    PrintAndLogEx(NORMAL, "      hf mf hardautopwn * 1 d f default_keys");
-    PrintAndLogEx(NORMAL, "      hf mf hardautopwn k 0 A FFFFFFFFFFFF d");
-    PrintAndLogEx(NORMAL, "      hf mf hardautopwn k 0 A FFFFFFFFFFFF * 1 d f default_keys");
-    PrintAndLogEx(NORMAL, "      hf mf hardautopwn k 0 A FFFFFFFFFFFF * 4 s i 5");
-
-    PrintAndLogEx(NORMAL, "");
+    PrintAndLogEx(NORMAL, "      hf mf autopwn");
+    PrintAndLogEx(NORMAL, "      hf mf autopwn * 1 f default_keys");
+    PrintAndLogEx(NORMAL, "      hf mf autopwn k 0 A FFFFFFFFFFFF");
+    PrintAndLogEx(NORMAL, "      hf mf autopwn k 0 A FFFFFFFFFFFF * 1 f default_keys");
+    PrintAndLogEx(NORMAL, "      hf mf autopwn k 0 A FFFFFFFFFFFF * 4 i 5");
     return 0;
 }
 static int usage_hf14_chk(void) {
@@ -1562,57 +1564,51 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
 }
 
 
-static int CmdHF14AMfHardAuto(const char *Cmd) {
+static int CmdHF14AMfAutoPWN(const char *Cmd) {
+    // Nested and Hardnested parameter
     uint8_t blockNo = 0;
     uint8_t keyType = 0;
-    uint8_t *keyBlock;
-    uint8_t sectorsCnt = MIFARE_1K_MAXSECTOR;
-    sector_t *e_sector;
     uint8_t key[6] = {0};
-    uint8_t tmpKey[6] = {0};
     uint64_t key64 = 0;
+    // Attack key storage variables
+    uint8_t *keyBlock;
     uint16_t keycnt = 0;
+    sector_t *e_sector;
+    uint8_t sectorsCnt = MIFARE_1K_MAXSECTOR;
+    uint8_t tmpKey[6] = {0};
     size_t datalen = 0;
-    uint32_t chunksize;
-
-    uint64_t t1; // For the timier
-    uint8_t foundKeysDictionary = 0;
-    uint8_t foundKeysReused = 0;
-    uint8_t foundKeysHardnested = 0;
-
+    // For the timier
+    uint64_t t1;
+    // Parameters and dictionary file
     char filename[FILE_PATH_SIZE] = {0};
     uint8_t cmdp = 0;
     char ctmp;
-
+    // Nested and Hardnested returned status
     uint64_t foundkey = 0;
     int16_t isOK = 0;
-    
-    int i, i2, i3; // Loop counter
+    // Loop counter
+    int i, i2, i3;
     int current_sector_i = 0, current_key_type_i = 0;
-
+    // Settings
     bool slow = false;
     bool nonce_file_read = false;
     bool nonce_file_write = false;
-    bool createDumpFile = false;
     bool know_target_key = false;
     bool legacy_mfchk = false;
-    bool firstChunk = 0, lastChunk = 0;
+    bool prng_type = false;
+    bool calibrate = true;
 
+    // Parse the options given by the user
     ctmp = tolower(param_getchar(Cmd, 0));
-    if (strlen(Cmd) < 1 || ctmp == 'h') return usage_hf14_hardautopwn();
-
     while ((ctmp = param_getchar(Cmd, cmdp))) {
         switch (tolower(ctmp)) {
             case 'h':
-                return usage_hf14_hardautopwn();
+                return usage_hf14_autopwn();
             case 'f':
                 if (param_getstr(Cmd, cmdp +1, filename, FILE_PATH_SIZE) >= FILE_PATH_SIZE) {
                     PrintAndLogEx(FAILED, "Filename too long");
                 }
                 cmdp ++;
-                break;
-            case 'd':
-                createDumpFile = true;
                 break;
             case 'l':
                 legacy_mfchk = true;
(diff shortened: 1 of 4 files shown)


iceman1001:
ooo,  interesting concept!

client/cmdhfmf.c
```diff
@@ -166,6 +166,33 @@ static int usage_hf14_hardnested(void) {
...
 }
+static int usage_hf14_hardautopwn(void) {
+    PrintAndLogEx(NORMAL, "Usage:");
+    PrintAndLogEx(NORMAL, "      hf mf hardautopwn [k] <sector number> <key A|B> <key (12 hex symbols)>");
+    PrintAndLogEx(NORMAL, "                       * <card memory> [d] [f] [s] [t] [i]");
+    PrintAndLogEx(NORMAL, "                       (card memory - 0 - MINI(320 bytes), 1 - 1K, 2 - 2K, 4 - 4K, <other> - 1K)");
+    PrintAndLogEx(NORMAL, "");
+    PrintAndLogEx(NORMAL, "Options:");
+    PrintAndLogEx(NORMAL, "      h                          this help");
+    PrintAndLogEx(NORMAL, "      k <sector> <keytype> <key> if a known key for a block is supplied");
```
iceman1001: use  s = sectore  t = keytype,   otherwise this will be a mess to keep track of.

client/cmdhfmf.c
```diff
@@ -166,6 +166,33 @@ static int usage_hf14_hardnested(void) {
     PrintAndLogEx(NORMAL, "      hf mf hardnested 0 A A0A1A2A3A4A5 4 A FFFFFFFFFFFF");
     return 0;
 }
+static int usage_hf14_hardautopwn(void) {
+    PrintAndLogEx(NORMAL, "Usage:");
```
iceman1001: no about the command phrase?

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
     return 0;
 }
 
+
+static int CmdHF14AMfHardAuto(const char *Cmd) {
+    /*
+        Author: Matthias Konrath
+        Company: Trustworks GmbH
+        Email: m.konrath@trustworks.at
+    */
```
iceman1001: you get your fame in the changelog and in the commit message..
or/and you can add a line on the top of this file.

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
...
+                PrintAndLogEx(WARNING, "Unknown parameter '%c'\n", ctmp);
+                usage_hf14_hardnested();
+                return 1;
+        }
+        cmdp++;
+    }
+
+    // Print parameters
+    PrintAndLogEx(NORMAL, "Used Parameters:");
+    PrintAndLogEx(NORMAL, "[+] Dumping the found keys: %s", createDumpFile ? "True" : "False");
```
iceman1001: use  SUCCESS to get [+]  :)

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
...
+                keyBlock = p;
+            }
+            int pos = 6 * keycnt;
+            memset(keyBlock + pos, 0, 6);
+            num_to_bytes(strtoll(buf, NULL, 16), 6, keyBlock + pos);
+            keycnt++;
+            memset(buf, 0, sizeof(buf));
+        }
+        fclose(f);
+        PrintAndLogEx(SUCCESS, "Loaded %2d keys from " _YELLOW_("%s"), keycnt, filename);
```
iceman1001: try looking inside fileutils.c  there is a function to load dictionary files

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
...
+    if (know_target_key == false) {
+        for (current_sector_i=0; current_sector_i < sectorsCnt; current_sector_i++) {
+            for (current_key_type_i=0; current_key_type_i < 2; current_key_type_i++) {
+                for (default_keys_i=0; default_keys_i < keycnt; default_keys_i++) {
+                    // Iterate over the keys
+                    for (i=0; i<6; i++) {
+                        tmpKey[i] = keyBlock[i + (6*default_keys_i)];
+                    }
+
+                    if (mfCheckKeys(current_sector_i*4, current_key_type_i, true, 1, tmpKey, &key64) == PM3_SUCCESS) {
```
iceman1001: hm... try   mfcheckkeys_fast,  its way better

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
...
+
+            foundkey = e_sector[current_sector_i].Key[current_key_type_i];
+
+            // Try the found keys
+            if (foundkey == 0) {
+                for (found_keys_i=0; found_keys_i < sectorsCnt; found_keys_i++) {
+                    // Iterate over the keys
+                    if (arr[found_keys_i + (current_key_type_i * sectorsCnt)] == 1) {
+                        num_to_bytes(e_sector[found_keys_i].Key[current_key_type_i], 6, tmpKey);
+                        if (mfCheckKeys(current_sector_i*4, current_key_type_i, true, 1, tmpKey, &key64) == PM3_SUCCESS) {
```
iceman1001: same as above

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
...
+                                current_key_type_i ? 'B' : 'A',
+                                tmpKey[0], tmpKey[1], tmpKey[2], tmpKey[3], tmpKey[4], tmpKey[5]);
+                            foundkey = bytes_to_num(tmpKey, 6);
+                            foundKeysReuse++;
+                            break;
+                        }
+                    }
+                    if (arr[found_keys_i + (((current_key_type_i+1)%2) * sectorsCnt)] == 1) {
+                        num_to_bytes(e_sector[found_keys_i].Key[(current_key_type_i+1)%2], 6, tmpKey);
+                        if (mfCheckKeys(current_sector_i*4, current_key_type_i, true, 1, tmpKey, &key64) == PM3_SUCCESS) {
```
iceman1001: same

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
...
+            }
+            // Try the default keys
+            if (foundkey == 0) {
+                for (default_keys_i=0; default_keys_i < keycnt; default_keys_i++) {
+                    // Iterate over the keys
+                    for (i=0; i<6; i++) {
+                        tmpKey[i] = keyBlock[i + (6*default_keys_i)];
+                    }
+
+                    if (mfCheckKeys(current_sector_i*4, current_key_type_i, true, 1, tmpKey, &key64) == PM3_SUCCESS) {
```
iceman1001: same

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
...
+                            break;
+                        default :
+                            break;
+                    }
+                    free(e_sector);
+                    return 2;
+                }
+
+                num_to_bytes(foundkey, 6, tmpKey);
+                PrintAndLogEx(SUCCESS, "[CRACKED  KEY]  Valid KEY FOUND: sector:%3d key type:%c key: " _YELLOW_("0x%02x%02x%02x%02x%02x%02x"),
```
iceman1001: we are trying to avoid words like "CRACKED".   I suggest "FOUND"  instead. 

[FOUND] Valid key:

client/cmdhfmf.c
```diff
@@ -1530,6 +1557,423 @@ static int CmdHF14AMfNestedHard(const char *Cmd) {
...
+    for (i=0; i<sectorsCnt; i++) {
+        for (i2=0; i2<2; i2++) {
+            e_sector[i].foundKey[i2] = arr[i + (i2 * sectorsCnt)];
+        }
+    }
+
+    // Dump all the data
+    printKeyTable(sectorsCnt, e_sector);
+
+    if (createDumpFile) {
```
iceman1001: this create dump keys bin file,..  we would need to move it to fileutils.c since its being used several times in this file

client/cmdhfmf.c
matthiaskonrath: DONE

client/cmdhfmf.c
matthiaskonrath: DONE

client/cmdhfmf.c
matthiaskonrath: This is more reliable and fast enougth. The fast key check looks a bit buggy when it's used with only one key (it returned mutiple times the wrong keys).

client/cmdhfmf.c
matthiaskonrath: This code has been removed.

client/cmdhfmf.c
matthiaskonrath: This code has been removed.

client/cmdhfmf.c
matthiaskonrath: DONE

client/cmdhfmf.c
matthiaskonrath: DONE

iceman1001:
Great!  ....now...  
this command is limited to hardnested however with some minor changes,  we can get a **hf mf autopwn** command.   Lets take it to the next level.
1. add simple check before dictionary checks,  if  card has weak/hard prng.
1a.  if weak -> attack process is: 
   - chkkey ;   
   -  if not found; 
           use darkside
    use nested
    save keys.
1b.  if hard  - continue your current process w hardnested:chkkey;
2.  make card dump -  a new last step.
2a. if all keys found, dump card

With these additions,   you have a C version of my improved hardnest_autopwn lua script.  Even if the name is a bit missrepresentation of whats in there. :)

matthiaskonrath:
The transformation is complete ;)
Now we have a very versatile command to attack all the mifare classic cards.
No more manual key extractions :)
I couldn't test the darkside attack (maybe someone with the right nfc card can check if it is working properly).

cjbrigato:
I can't but advice to not make a dump, but I as this is slow as f*k.

Throughout the process, always populate the emulator memory with found key. The when you got all the required keys, just do an ECFILL so the emulator memory is populated with actual badge dump (and ECFILL is f*king fast)  then just ESAVE locally. et voila

cjbrigato:
Btw, @iceman1001 @matthiaskonrath 
since we use dictionnarry for keys, why not reusing my "known scheme dictionnary" idea from my standalone ? and integrate it in the autopwn ?

All it takes si this struct 

```
typedef struct MFC1KSchema {
    uint8_t name[32];
    uint64_t trigger;
    uint64_t keysA[16];
    uint64_t keysB[16];
} MFC1KSchema;
```
You can take the two add/delete function I left for adding/deleting schemes
```
MFC1KSchema Schemas[MAX_SCHEMAS];

int total_schemas = 0;

void add_schema(MFC1KSchema *p, MFC1KSchema a, int *schemas_counter) {
    if (*schemas_counter < MAX_SCHEMAS) {
        p[*schemas_counter] = a;
        *schemas_counter += 1;
    }
}

void delete_schema(MFC1KSchema *p, int *schemas_counter, int index) {
    if (*schemas_counter > 0 && index < *schemas_counter && index > -1) {
        int last_index = *schemas_counter - 1;
        for (int i = index; i < last_index; i++) {
            p[i] = p[i + 1];
        }
        *schemas_counter -= 1;
    }
}
```
then you just throw in any json parser for these configs , e.g. for my classing Intratone/noralsy/cogeco etc : 

`❯ cat schemas.json`
```
[
   {
      "name":"UrmetCaptive",
      "trigger":"0x8829da9daf76",
      "keysA":[
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76"
      ],
      "keysB":[
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76",
         "0x8829da9daf76"
      ]
   },
   {
      "name":"Noralsy",
      "trigger":"0x414c41524f4e",
      "keysA":[
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e",
         "0x414c41524f4e"
      ],
      "keysB":[
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e",
         "0x424c41524f4e"
      ]
   },
   {
      "name":"Infineon/Hexact",
      "trigger":"0x484558414354",
      "keysA":[
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354",
         "0x484558414354"
      ],
      "keysB":[
         "0xa22ae129c013",
         "0x49fae4e3849f",
         "0x38fcf33072e0",
         "0x8ad5517b4b18",
         "0x509359f131b1",
         "0x6c78928e1317",
         "0xaa0720018738",
         "0xa6cac2886412",
         "0x62d0c424ed8e",
         "0xe64a986a5d94",
         "0x8fa1d601d0a2",
         "0x89347350bd36",
         "0x66d2b7dc39ef",
         "0x6bc1e1ae547d",
         "0x22729a9bd40f"
      ]
   }
]
```
and using them is just a matter of implementing a "trap" condition where if you find a key which is a "trigger" of any of the know scheme, you'll populate every keys from the scheme and try confirm it was ok by a simple ECFILL; or revert back to your previous state if anyhow it was a false positive.

This could be done inside the Chkey loops, like i did :  (this do not include the ECFILL -> test? -<?> Rollback I talked about but just to have an idea 
```
    for (uint8_t type = 0; type < 2 && !err && !trapped; type++) {
        for (int sec = 0; sec < sectorsCnt && !err && !trapped; ++sec) {
            key = cjat91_saMifareChkKeys(sec * 4, type, NULL, size, &keyBlock[0], &key64);

            if (key == -1) {
                err = 1;
                allKeysFound = false;
                break;
            } else if (key == -2) {
                err = 1; // Can't select card.
                allKeysFound = false;
                break;
            } else {
                num_to_bytes(key64, 6, foundKey[type][sec]);
                for (int i = 0; i < total_schemas; i++) {
                    if (key64 == Schemas[i].trigger) {
                         /* WE TRAPPED A KNOWN SCHEME TRIGGER */
                        uint16_t t = 0;
                        for (uint16_t s = 0; s < sectorsCnt; s++) {
                            num_to_bytes(Schemas[i].keysA[s], 6, foundKey[t][s]);
                            sprintf(tosendkey, "%02x%02x%02x%02x%02x%02x", foundKey[t][s][0], foundKey[t][s][1],
                                    foundKey[t][s][2], foundKey[t][s][3], foundKey[t][s][4], foundKey[t][s][5]);
                        }
                        t = 1;
                        for (uint16_t s = 0; s < sectorsCnt; s++) {
                            num_to_bytes(Schemas[i].keysB[s], 6, foundKey[t][s]);
                            sprintf(tosendkey, "%02x%02x%02x%02x%02x%02x", foundKey[t][s][0], foundKey[t][s][1],
                                    foundKey[t][s][2], foundKey[t][s][3], foundKey[t][s][4], foundKey[t][s][5]);
                        }
                        trapped = 1;
                        break;
                    }
                }
            }
        }
    }
```

and then next step would be to statistically create new "known schemes" dictionnary by keeping trace of random Pwned Tags for similarities and non-singularity of cases. Why not sending these anonimized results to somewhere secure online then parsing these data to API-provide these known schemes dictionnaries alongside with confidence and score, any ratings of sort we could compute, and providing also Verified known schemes.

This "autopwn" command makes too much for me to not profit from the occasion for such an integration (where anyway every programming aspect is 95% already done less than two `DIRENT` away :)

matthiaskonrath:
Now the keys get transferred to the simulator, and the card is dumped via the ECFILL and ECSAVE. This stuff is pretty fast ;)

iceman1001:
Great progress,   and most excellent enhancements suggestions from @cjbrigato  ..
Of course we should add those json triggered schemas.  
We should also use dictionaries on device side (fchk 1 m)  

I do notice that the dumps size here is 1K,   but if I remember correct the simulation nowdays can do 4K.  So the command should adapt to the different sizes  320b, 1K, 4K 
And some spelling misstakes.
And some obsession with LARGE_small letters mixed.  

I will merge the PR regardless,  but cleaning out the text output will be done.  :) 
And json schema triggers.. mmm.. 
and device dictionaries mmm

This will be a much appreciated command.

iceman1001:
About the output,   we really dont need to know much statistics,   we do like to know how long it took.  and which keys was found.   Ppl tend to like a "to the point" output.

matthiaskonrath:
> I do notice that the dumps size here is 1K, but if I remember correct the simulation nowdays can do 4K. So the command should adapt to the different sizes 320b, 1K, 4K

It already supports different card sizes.

iceman1001:
btw..  the underscore was the prefered way ... ;)    Give me some days to test this puppy,  and we be ready to merge it to the weekend.

matthiaskonrath:
> btw.. the underscore was the prefered way ... ;) Give me some days to test this puppy, and we be ready to merge it to the weekend.
Ooops, well I tried ...
Happy to see my code soon in this awesome repo ;)

iceman1001:
you did an awesome contribution,   we just gonna take it further to let it become ever more

matthiaskonrath:
I changed the variable names to include underscores.
We should implement a maxdatalen in loadFileDICTIONARY()! (to prevent heap overflows)
I also fixed a corner case when loading the keys to the simulator memory.

iceman1001:
funny,  I was chatting with @doegox  about the memory must be dynamically allocated for loaddictionaires yesterday.   Was working on it aswell,   lets see if I can get a fix pushed out.

@cjbrigato   you wanna play more with this PR? or you will adapt afterwards?  Maybe @matthiaskonrath  will do the json schemas in a next PR?

cjbrigato:
Hello, don't have to play more with this PR on my side. I'm getting on par with you dear uptstream as soon as you merge this, but I have a lot to do on a new standalone which is to be interactive with ESP32 client firmware i'm working on so I guess we cas just take out the json dictionnaries from this PR and Either @matthiaskonrath (since this command is his new baby) or me will do another PR with the Json dicts.
Since they are used by my standalone maybe it's better if I do  the PR then take it out of my standalone ? Let's add the dict natively insid the device firmware, take it out of my standalone, and make the autopwn use them as device-side capacity. This way it will be easier to keep dictionnaries device side since it's already the case in my standalone.
I'd be glad the json thing got out of my standalone so I can make it simpler to handle the ESP32 interactive mode better.

iceman1001:
yeah,  I want that json in the device aswell. 
@doegox  you fine with merge?

doegox:
Just one comment: `createKeyDump` is too generic as name for inclusion in fileutils.
=> `createMfcKeyDump` ?
I've no time to test, you can merge after this small change.
Thanks!

doegox:
ok for me!

iceman1001:
and finally one more thing,   add a line in CHANGELOG.md  :)

iceman1001:
The sensationalism,  you seem to have a flair for it.
Here is a good example.

Add:  'hf mf autopwn' -  executes darkside/nested/hardnested for easy extraction of keys.  (@matthiaskonrath)

iceman1001:
I merge instead.  Goes faster :)
Great addition!   Now @matthiaskonrath  can start looking into the other enhancements improvements from this PR.

cjbrigato:
hum why the hf mf autopwn sends the hardnested help ?

```
usb] pm3 --> hf mf autopwn ?
[!] Unknown parameter '?'

Usage:
      hf mf hardnested <block number> <key A|B> <key (12 hex symbols)>
                       <target block number> <target key A|B> [known target key (12 hex symbols)] [w] [s]
  or  hf mf hardnested r [known target key]

Options:
      h         this help
      w         acquire nonces and UID, and write them to binary file with default name hf-mf-<UID>-nonces.bin
      s         slower acquisition (required by some non standard cards)
      r         read hf-mf-<UID>-nonces.bin if tag present, otherwise read nonces.bin, then start attack
      u <UID>   read/write hf-mf-<UID>-nonces.bin instead of default name
      f <name>  read/write <name> instead of default name
      t         tests?
      i <X>     set type of SIMD instructions. Without this flag programs autodetect it.
        i 5   = AVX512
        i 2   = AVX2
        i a   = AVX
        i s   = SSE2
        i m   = MMX
        i n   = none (use CPU regular instruction set)

Examples:
      hf mf hardnested 0 A FFFFFFFFFFFF 4 A
      hf mf hardnested 0 A FFFFFFFFFFFF 4 A w
      hf mf hardnested 0 A FFFFFFFFFFFF 4 A f nonces.bin w s
      hf mf hardnested r
      hf mf hardnested r a0a1a2a3a4a5

Add the known target key to check if it is present in the remaining key space:
      hf mf hardnested 0 A A0A1A2A3A4A5 4 A FFFFFFFFFFFF
```

by the way being "auto" pown, why not using the defaut dictionnary as default parameter ?

matthiaskonrath:
> hum why the hf mf autopwn sends the hardnested help ?
> 
> ```
> usb] pm3 --> hf mf autopwn ?
> [!] Unknown parameter '?'
> 
> Usage:
>       hf mf hardnested <block number> <key A|B> <key (12 hex symbols)>
>                        <target block number> <target key A|B> [known target key (12 hex symbols)] [w] [s]
>   or  hf mf hardnested r [known target key]
> 
> Options:
>       h         this help
>       w         acquire nonces and UID, and write them to binary file with default name hf-mf-<UID>-nonces.bin
>       s         slower acquisition (required by some non standard cards)
>       r         read hf-mf-<UID>-nonces.bin if tag present, otherwise read nonces.bin, then start attack
>       u <UID>   read/write hf-mf-<UID>-nonces.bin instead of default name
>       f <name>  read/write <name> instead of default name
>       t         tests?
>       i <X>     set type of SIMD instructions. Without this flag programs autodetect it.
>         i 5   = AVX512
>         i 2   = AVX2
>         i a   = AVX
>         i s   = SSE2
>         i m   = MMX
>         i n   = none (use CPU regular instruction set)
> 
> Examples:
>       hf mf hardnested 0 A FFFFFFFFFFFF 4 A
>       hf mf hardnested 0 A FFFFFFFFFFFF 4 A w
>       hf mf hardnested 0 A FFFFFFFFFFFF 4 A f nonces.bin w s
>       hf mf hardnested r
>       hf mf hardnested r a0a1a2a3a4a5
> 
> Add the known target key to check if it is present in the remaining key space:
>       hf mf hardnested 0 A A0A1A2A3A4A5 4 A FFFFFFFFFFFF
> ```
> 
> by the way being "auto" pown, why not using the defaut dictionnary as default parameter ?

You have made a mistake there.
My code is working correct.
Try to make a fresh clone, or take a look inside the source code.
And it uses the default dictionary.
Sorry but the mistake has to be on your side ...

doegox:
@cjbrigato makes no mistake.
I fixed yours, @matthiaskonrath

matthiaskonrath:
> @cjbrigato makes no mistake.
> I fixed yours, @matthiaskonrath

OK yup with the wrong parameter this happens ... (thanks for the fix)
But the default dictionary is used, when no parameters are used.

iceman1001:
...just to keep it straight.  @cjbrigato  does do mistakes but very complicated ones.
We usually have an open mind and don't really play the blame game.  I do mistakes all the time.

The question is if its the default 20list array or default 800plus dictionary Colin references.  
I suggest we use default 20 list (just like nested normally works with) ...  And yes,  we have a bunch more dictionaries nowdays.  look under  client/dictionaries/ folders.   

Looking forward to the json scheme idea. 

and do we have to adapt @doegox  nested fixes for auto aswell,  since it uses a lot of copied code...  One of these days we need to break it out to methods instead.

matthiaskonrath:
> and do we have to adapt @doegox nested fixes for auto aswell, since it uses a lot of copied code... One of these days we need to break it out to methods instead.

I'll take care of this ;)

cjbrigato:
> just to keep it straight. @cjbrigato does do mistakes but very complicated ones.

Magad, you've made perfect definition it's frightening.

PR #359, opened 2019-08-26
Improved help message for the autopwn command + nested attack loop bugfix

matthiaskonrath:


iceman1001:
hm... interesting...   introducing  ?  as  help....    Not that it costs anything nor is hard to implement... but we don't introduce a client overall command in a single command PR.
If you want to introduce ?,  do it the proper way and add it everywhere...   make sure it doesn't fuck up other commands which might use ?..   :)

matthiaskonrath:
The nested fix from @doegox was already implemented, so it is fixed ...
I also found a bug in my nested attack loop (the retry was missing), that's now fixed.
I think we just forget the **?** as a extra help command ... this would be too much work and would mess with other existing parameters.

doegox:
f*ck I didn't see the PR before my changes, let me see...

matthiaskonrath:
Well use yours, it's pretty much the same ... ;)
But the help stuff is maybe useful?

doegox:
first part: I cherry-picked your help changes, removed dictionaries/ because it's now found automatically

doegox:
second: ok indeed it's about the same.
I'll close this PR.
BTW one improvement still to be done: 
As soon as one key is found, autopwn should try it on other A/B & sectors.
E.g. currently it doesn't try to read keyB with keyA freshly found with nested.
I won't touch the file for now to avoid any conflict ;)

matthiaskonrath:
Nice, I will look into that tomorrow ;)

matthiaskonrath:
> second: ok indeed it's about the same.
> I'll close this PR.
> BTW one improvement still to be done:
> As soon as one key is found, autopwn should try it on other A/B & sectors.
> E.g. currently it doesn't try to read keyB with keyA freshly found with nested.
> I won't touch the file for now to avoid any conflict ;)

Maybe I’m false, but I think this already exists in the loop in front of the nested loop?

doegox:
```
[=] [   NESTED   ] Sector no:  0, target key type:A          
[+] target block:  0 key type: A  -- found valid key [c0ffeec0ffee]          
[+] [TESTING  KEY] Valid KEY FOUND: sector:  0 key type:A key: 0xc0ffeec0ffee           
[=] [   NESTED   ] Sector no:  0, target key type:B          
[+] target block:  0 key type: B          
[-] Nested attack failed, trying again (1/10)          
[=] [   NESTED   ] Sector no:  0, target key type:B          
[+] target block:  0 key type: B  -- found valid key [deadbeefdead]          
[+] [TESTING  KEY] Valid KEY FOUND: sector:  0 key type:B key: 0xdeadbeefdead           
```

See, keyA/0 is found but keyB/0 is not simply read via keyA, it's found via nested

matthiaskonrath:
I don't understand what you mean.
This code loops over every sector and every key type (first with nested / hardnested and then with a new key, when one is discovered --> so if a found key is reused anywhere else, it will be found by the key reuse loop).

Screenshot from 2019-08-27 09-46-29

Screenshot from 2019-08-27 09-46-58

doegox:
sorry I expressed myself wrongly.
When you got a keyA, you can simply attempt to read the sector and get keyB (still you've to test the key you read). So when this works, no need to break keyB.

doegox:
so in your example, there was no need to try to break 7f..29, you could simply read it out of sector 5 with keyA 6a..21

matthiaskonrath:
> sorry I expressed myself wrongly.
> When you got a keyA, you can simply attempt to read the sector and get keyB (still you've to test the key you read). So when this works, no need to break keyB.

Ahh ok, well I have never done this before, but I'll give it a try ;)

doegox:
hf mf nested is doing it

iceman1001:
well.  It only works if the accessrights is set to allow to read keyB.  Which is not always the case.
However,  in nested,   it calls chk_fast,  which does do A/B tests.
in autopwn,  it calls chk,  which doesn't call A/B tests.
hence @doegox is saying,  loop A first,  for all found A,  try read keyB,   then loop B to nested the ones that wasnt able to be read with KeyA.

PR #360, opened 2019-08-27
Fixing filenames for "hf mf dump" command

mohemiv:
It's not convenient that "hf mf dump" creates files with only the first 4 bytes of a card UID, but "hf mf restore" requires files with the full UID of the card or manual argument transmission.

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -963,9 +963,11 @@ static int CmdHF14AMfDump(const char *Cmd) {
     PrintAndLogEx(SUCCESS, "\nSucceded in dumping all blocks");
 
     if (strlen(dataFilename) < 1) {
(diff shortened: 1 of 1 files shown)


doegox:
You're right.
I've the feeling there is more to do across the whole file.
There are still some other hardcoded `FillFileNameByUID(..., 4);` and having to call `GenerateFilename` sounds overkill (having to read again the card we just dumped, just to know its UID).
Imho all these functions should get/read the UID at the beginning and keep it somewhere during the entire function, not guessing UID from the dump itself, not interrogating the card again to know its UID.

iceman1001:
Yup,  but those fixes is for later.  I already took @mohemiv fixes and merged,  so I close this PR.
Thanks!

PR #361, opened 2019-08-27
New features and fixes for the autopwn command

matthiaskonrath:
Add feature to extract the B keys with known A keys (I couldn't verify if this is working correct, so someone should please try it first)
Add some of my found keys in the dictionary
Add a safer function to load dictionaries from the filesystem
Add fast key check for nested keys (I have no idea why this works there and not in the key reuse detection loop above...)

### DIFF
--- a/client/dictionaries/mfc_default_keys.dic
+++ b/client/dictionaries/mfc_default_keys.dic
@@ -982,3 +982,36 @@ ef4c5a7ac6fc
 b47058139187
 8268046cd154
 67cc03b7d577
+#
+# From the HTL Mödling, NÖ, AT
+#
+a5524645cd91
+d964406e67b4
+99858a49c119
+7b7e752b6a2d
+c27d999912ea
+66a163ba82b4
+4c60f4b15ba8
+#
+# CAFE + CO, AT
+#
+35d850d10a24
+4b511f4d28dd
+e45230e7a9e8
+535f47d35e39
+fb6c88b7e279
+#
+# Metro Card, AT
+#
+223C3427108A
+#
+# Unknown, AT
+#
+23d4cdff8da3
+e6849fcc324b
+12fd3a94df0e
+#
+# Unknown, AT
+#
+0b83797a9c64
+39ad2963d3d1
--- a/client/fileutils.c
+++ b/client/fileutils.c
@@ -649,6 +649,78 @@ out:
     return retval;
 }
 
+uint8_t* loadFileDICTIONARY_safe(const char *preferredName, uint8_t keylen, uint16_t *keycnt) {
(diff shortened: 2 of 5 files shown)


client/fileutils.c
```diff
@@ -649,6 +649,78 @@ int loadFileDICTIONARY(const char *preferredName, void *data, size_t *datalen, u
     return retval;
 }
 
+uint8_t* loadFileDICTIONARY_safe(const char *preferredName, uint8_t keylen, uint16_t *keycnt) {
```
doegox: better to return success codes as for loadFileDICTIONARY and to return the pointer to allocated space via pointer of pointer arg

client/dictionaries/mfc_default_keys.dic
```diff
@@ -982,3 +982,36 @@ ef4c5a7ac6fc
...
+# Unknown, AT
+#
+23d4cdff8da3
+e6849fcc324b
+12fd3a94df0e
+#
+# Unknown, AT
+#
+0b83797a9c64
+39ad2963d3d1
```
doegox: missing \n

doegox:
> Add feature to extract the B keys with known A keys (I couldn't verify if this is working correct, so someone should please try it first)

Yes it works!

matthiaskonrath:
> > Add feature to extract the B keys with known A keys (I couldn't verify if this is working correct, so someone should please try it first)
> 
> Yes it works!

Nice ;)

iceman1001:
hold on,  I have a rewrite of the autopwn...   You will need to merge against that one.. this pr will get conflict.

matthiaskonrath:
> hold on, I have a rewrite of the autopwn... You will need to merge against that one.. this pr will get conflict.

Could you include the feature to extract the B keys with known A keys in your rewrite.
I will then re implement the other stuff in a separate pull request, when you are done.

client/fileutils.c
matthiaskonrath: with the next pull request, when the rewrite from @iceman1001 is done.

iceman1001:
I just merged it.  So.. yeah.  How about you split things up.   Like the fileutils change,  (don't return pointer like @doegox  said,  use **)   do that in a separate PR.    And then we can use that as base for the rest of the changes in the files that uses loadDictionary..    However,  don't touch cmdhficlass / iclass stuff.  I am working on it.

doegox:
@matthiaskonrath be careful I pushed some style fix just after iceman, better to branch from there

doegox:
> Could you include the feature to extract the B keys with known A keys in your rewrite.

Still to be done again I'm afraid

PR #364, opened 2019-08-28
Implementation of a function to safely load a dictionary

matthiaskonrath:
I didn't test the code in cmdlft55xx.c, but I'm pretty sure it should just work fine.


client/cmdlft55xx.c
```diff
@@ -2127,19 +2127,11 @@ static int CmdT55xxChkPwds(const char *Cmd) {
...
-        keyBlock = calloc(4 * 200, sizeof(uint8_t));
-        if (keyBlock == NULL) {
-            PrintAndLogEx(ERR, "error, cannot allocate memory ");
-            return PM3_ESOFT;
-        }
-
-        int res = loadFileDICTIONARY(filename, keyBlock, &datalen, 4, &keycount);
-        if (res || keycount == 0) {
+        int res = loadFileDICTIONARY_safe(filename, &keyBlock, 4, &keycount);
+        if (res || keycount <= 0 || keyBlock == NULL) {
```
doegox: if res != PM3_SUCCESS

iceman1001:
Awesome!
Here I was fighting that :) and you swoop right in solving it for me.

client/fileutils.c
```diff
@@ -649,6 +649,79 @@ int loadFileDICTIONARY(const char *preferredName, void *data, size_t *datalen, u
...
+        keylen = 6;
+    }
+
+    // double up since its chars
+    keylen <<= 1;
+
+    char line[255];
+
+    // allocate some space for the dictionary
+    *data = (uint8_t*) malloc(keylen * allocation_size * sizeof(uint8_t));
```
iceman1001: use  `calloc`  instead.

matthiaskonrath: Is there something like realloc in calloc?
Otherwise this would be useless.

matthiaskonrath:
I'll leave the loadFileDICTIONARY calls in client/cmdflashmem.c to the pros to implement.

client/fileutils.c
matthiaskonrath: Or should I memset everthing new to 0 after the realloc?

iceman1001: malloc vs calloc,    calloc memsets to zero.
let realloc be for now,  its for later to make sure the memory reallocated is zeroed before using.

doegox: I'm not sure I understood your comment @iceman.
malloc + memset = calloc and realloc ~ malloc, so I'd say use:
* calloc then
* realloc + memset

matthiaskonrath: Sounds good

client/fileutils.c
```diff
@@ -649,6 +649,79 @@ int loadFileDICTIONARY(const char *preferredName, void *data, size_t *datalen, u
     return retval;
 }
 
+int loadFileDICTIONARY_safe(const char *preferredName, uint8_t **data, uint8_t keylen, uint16_t *keycnt) {
```
doegox: all loadFile* used to use void *data.
Any reason to move to uint8_t (and your numerous internal casts) ?
I'd also rename data into pdata, to show the indirection.
So void **pdata ?

matthiaskonrath: Yup that sounds like a good plan ;)

client/fileutils.c
```diff
@@ -649,6 +649,84 @@ int loadFileDICTIONARY(const char *preferredName, void *data, size_t *datalen, u
...
+    // read file
+    while (fgets(line, sizeof(line), f)) {
+        // check if we have enough space (if not allocate more)
+        if ((*keycnt) >= allocation_size) {
+            allocation_size += block_size;
+            *pdata = realloc(*pdata, keylen * allocation_size * sizeof(uint8_t));
+            if (*pdata == NULL) {
+                return PM3_EFILE;
+            } else {
+                // zero the new memeory (safety first)
```
doegox: typo

client/fileutils.c
```diff
@@ -649,6 +649,84 @@ int loadFileDICTIONARY(const char *preferredName, void *data, size_t *datalen, u
...
+                memset(*pdata + counter, 0, block_size);
```
doegox: is *pdata+counter really aligned on new mem?
I'd rather see *pdata+allocation_size-block_size

matthiaskonrath: If this isn't aligned, then this **num_to_bytes(key, keylen >> 1, *pdata + counter);**, would fail too, or wouldn't it?

doegox: see when keylen=6, counter progress by +3 so it can't be aligned on blocksize (512).
when zeroizing mem, you need to zeroize the new 512b block.
In the code you show, it's just the regular injection of keys that doesn't care about blocksize alignment, no problem there

matthiaskonrath: Ahhh ok, now I see...

matthiaskonrath:
This are far too much fixes and commits for such simple code...
I hope that works now ;)

iceman1001:
Yup,   but then we dont have to fix them later :)

Ready to merge @doegox ?

doegox:
ready :) done already I see :)

doegox:
BTW  @matthiaskonrath your master is not same as RRG master so every time we merge your PRs we get some useless Merge pull request #xx from RfidResearchGroup/master back in RRG repo.

Better if you realign masters without merges, for future PRs.
You can do it by e.g. resetting your master to d88b76bedab0362b26f0bc36a635b6fba05679d8 then merging RRG/master on it, this should be feasible by fast-forward, then force push your new master on your repo.

Then make new branches whenever you want to push code & PR, don't push code to your master.

PR #366, opened 2019-08-28
Restyled of the autopwn command output

matthiaskonrath:
I hope you like the new style :)
It helps to keep track of what does what (this is probably very helpful for beginners).

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -1733,8 +1733,7 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
         PrintAndLogEx(WARNING, "No known key was supplied, key recovery might fail");
     else {
         if (mfCheckKeys(FirstBlockOfSector(blockNo), keyType, true, 1, key, &key64) == PM3_SUCCESS) {
-            PrintAndLogEx(INFO, "Using key for the nested / hardnested | sector:"
-                          _RED_("%3d") " key type: "_RED_("%c") "  key: " _RED_("%s"),
+            PrintAndLogEx(INFO, "Using valid key: sector: %3d key type: %c  key: " _YELLOW_("%s") "(used for nested / hardnested attack)",
                           blockNo,
                           keyType ? 'B' : 'A',
                           sprint_hex(key, sizeof(key))
@@ -1745,7 +1744,7 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
             e_sector[blockNo].foundKey[keyType] = 3;
         } else {
             know_target_key = false;
-            PrintAndLogEx(FAILED, "Key is wrong. Can't authenticate to sector:"_RED_("%3d") " key type:"_RED_("%c") " key: " _RED_("%s"),
+            PrintAndLogEx(FAILED, "Key is wrong. Can't authenticate to sector:"_RED_("%3d") " key type: "_RED_("%c") " key: " _RED_("%s"),
                           blockNo,
                           keyType ? 'B' : 'A',
                           sprint_hex(key, sizeof(key))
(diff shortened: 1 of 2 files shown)


client/cmdhfmf.c
```diff
@@ -1980,7 +1988,10 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
 
                 if (current_key_type_i == 1) {
                     if (e_sector[current_sector_i].foundKey[0] && !e_sector[current_sector_i].foundKey[1]) {
-                        PrintAndLogEx(INFO, "Reading  B  key: sector: %3d", current_sector_i);
+                        PrintAndLogEx(INFO, "Reading  B  key: sector: %3d key type: %c  key: ?? ?? ?? ?? ?? ??",
```
doegox: I prefer to revert that one, no need to show unknown key before the read

doegox:
about style:
* you may report when reading B fails, so we don't jump from "reading" to "hardnested" without an explanation.
* in the keys table, why not using single char for results and showing one-line legend? e.g.
`D:Dictionary S:darkSide U:User R:Reused N:Nested H:Hardnested A:keyA`

client/cmdhfmf.c
```diff
@@ -2026,6 +2037,7 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
                     if (prng_type && (! nested_failed)) {
                         uint8_t retries = 0;
 tryNested:
+                        PrintAndLogEx(INFO, _YELLOW_("======================= START   NESTED   ATTACK ======================="));
```
doegox: put it before tryNested: else it'll repeat at each try

iceman1001:
The oneline suggestion is nice. 
The PR style changes is nothing I prefer.  The output from autopwn is verbose as is.  It should be minimized more.

matthiaskonrath:
Well minimizing requires that functions like mfCheckKeys_fast, mfnested and mfnestedhard are changed ...
Is this really what you want?
I try to make the used functions more visible, so the users understand what's happening.

doegox:
Personally I'm kind of ok with some separators, but right now they repeat far too much.
What about skipping "STOP" msgs and seeing it as a state machine?
Display START... when you change the strategy but don't display it repeatedly when you're still in the same strategy.

matthiaskonrath:
> Personally I'm kind of ok with some separators, but right now they repeat far too much.
> What about skipping "STOP" msgs and seeing it as a state machine?
> Display START... when you change the strategy but don't display it repeatedly when you're still in the same strategy.

That sounds good.
We have a nice over few and not too much information on the screen ...

matthiaskonrath:
Well It looks not too good in my opinion ...

Screenshot from 2019-08-28 16-15-30

matthiaskonrath:
I hope my compromise is acceptable.
Normally only the absolute minimum is shown to the user.
If the verbose flag is set, then the user can follow every step very closely.

Screenshot from 2019-08-28 18-43-21
Screenshot from 2019-08-28 18-42-50

matthiaskonrath:
An we now have this  gorgeous  thing :)

Screenshot from 2019-08-28 18-44-19

iceman1001:
the dictionary /darkside /user/reuser/nested ,,,  what information do we gain from it?

doegox:
> the dictionary /darkside /user/reuser/nested ,,, what information do we gain from it?

knowledge :) I like getting that info...

matthiaskonrath:
> the dictionary /darkside /user/reuser/nested ,,, what information do we gain from it?

How each key was broken

iceman1001:
Nevermind,  after I push my changes, this PR will break.  You will need to merge conflicts before this one can be merged,

matthiaskonrath:
Conflicts are now gone.
Now that most of the things are lower case, i changed a few messages to match that. Now only error messages should start with capital letter (at least most of the times).

Screenshot from 2019-08-29 06-29-24
Screenshot from 2019-08-29 06-28-51

iceman1001:
Excellent!
Now some testing with different cards.  I think I might have fucked up the parameter loop,  especially the "K"  option.

PR #371, opened 2019-08-29
remove termcap and use pkg-config for ncurses

ZeroChaos-:
termcap has been deprecated for over a decade, remove it.

ncurses *can* be split into two libs, ncurses and tinfo, most distros do this and the standard way of dealing with it is to just ask pkg-config for the right libs to link.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -90,7 +90,7 @@ else
     LIBS := -I/usr/local/opt/readline/include $(LIBS)
 else
     LUALIB +=  -ldl
-    LDLIBS +=  -ltermcap -lncurses
+    LDLIBS +=  `pkg-config --libs ncurses`
     LUAPLATFORM = linux
 endif
 endif


iceman1001:
```
PGKCFG = pkg-config --libs ncurses

LDLIBS +=  $(PKGCFG) 
```

ZeroChaos-:
don't lose the backticks, they are more than a little important.

Also, there really isn't a need to put it into another variable

doegox:
as said in the issue, better use  `$(shell pkg-config` than backticks, that's what we already use for QT5 libs

iceman1001:
```
LDLIBS = $(shell pkg-config --libs ncurses 2>/dev/null)
```

doegox:
I was about to tell the error message is important as ncurses is not optional, but after testing it seems ncurses is not needed at all... Maybe it was needed when the lua bins were uselessly built ?

iceman1001:
lets see what happens on all other os...  not like that every came back and bite us

iceman1001:
Pushed the fix.  Thanks!

Let see how much broke now :)

PR #382, opened 2019-08-31
debianize the package

PalinuroSec:


doegox:
Maybe you could rather add the /debian on a separate project on Debian Salsa, as we did years ago for libnfc on anonscm...

PR #383, opened 2019-08-31
appveyor make install

merlokk:


appveyor.yml
```diff
@@ -366,13 +352,13 @@ test_script:
 
     #file test
 
-    ExecTest "proxmark3 exists" "proxmark3.exe" {Test-Path C:\ProxSpace\Release\proxmark3.exe}
+    ExecTest "proxmark3 exists" "proxmark3.exe" {Test-Path C:\ProxSpace\pm3\Release\bin\proxmark3.exe}
 
-    ExecTest "arm image exists" "\arm\fullimage1.elf" {Test-Path C:\ProxSpace\Release\arm\fullimage.elf}
+    ExecTest "arm image exists" "pm3-flash-fullimage" {Test-Path C:\ProxSpace\pm3\Release\bin\pm3-flash-fullimage}
```
iceman1001: I think you are now checking if the flash-fullimage script exists,   when you want to check for the image..
same problem below

doegox: fixed

PR #388, opened 2019-09-07
autopwn: fix card detection

sh7d:
There is false assumption in autopwn code that function `detect_classic_prng()` returns only vaules 1 or 0 so its interpreted as boolean. So if there is no tag on the device - autopwn runs cracking procedure printing a LOTS of errors.

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -1660,7 +1660,7 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
     // Settings
     bool slow = false;
     bool legacy_mfchk = false;
-    int prng_type = 0;
+    int prng_type = PM3_EUNDEF;
     bool verbose = false;
     bool has_filename = false;
     bool errors = false;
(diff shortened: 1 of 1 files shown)


client/cmdhfmf.c
```diff
@@ -1660,7 +1660,7 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
     // Settings
     bool slow = false;
     bool legacy_mfchk = false;
-    bool prng_type = false;
+    int prng_type = 0;
```
doegox: better to init with PM3_ESOFT

client/cmdhfmf.c
```diff
@@ -1776,8 +1776,12 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
         }
     }
 
-    // card prng type (weak=true / hard=false)
+    // card prng type (weak=1 / hard=0 / select/card comm error fail = -vaule)
```
doegox: wording:  ... / error = negative value

client/cmdhfmf.c
```diff
@@ -1776,8 +1776,12 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
...
     prng_type = detect_classic_prng();
+    if (prng_type < 0){
+        PrintAndLogEx(FAILED, "\nNo tag detected or other tag communication error");
+        goto noValidKeyFound;
```
doegox: at this stage, don't goto but do
```
      free(e_sector);
      return prng_type;
```

doegox:
Thanks for spotting the issue ! please fix the few things I commented

client/cmdhfmf.c
doegox: better even PM3_EUNDEF  (even if the init value should never be used)

iceman1001:
we have more places that we mix PM3_SUCCESS/PM3_E* codes with result of functions where we call the device.   Pure client side functions can have bool etc,   but  @doegox  suggested keep return codes and pass a result variable instead..   So we can keep track of comms between device and result of the operations.  I am fixing 14a functions which suffers from this.  Lovely.

sh7d:
@doegox done

PR #398, opened 2019-09-14
T55xx Downlink Mode auto use

mwalker33:
Added the detected downlink mode to the config stuct.
lf t55 detect will auto try all downlink modes unless r <option> provided.
All commands that support read/write to card will by default use the detected mode unless r option supplied.
Password checks will default to r 0 (default), but support all downlink modes and option 4 to try all (slower)
Added the option to lf t55 config to set the default downlink mode (as per other config options)
Fixed detected config settings getting displayed when reading with password and password password not required.


iceman1001:
Excellent!

PR #409, opened 2019-09-21
subtle change for clarity in comment

amckee:
Both interface-raspberrypi.cfg and interface-raspberrypi2.cfg gave conflicting instructions for the dd return value. This should clear that up.


doegox:
thanks!

PR #410, opened 2019-09-22
lf em 4x05_dump - print output for each block read

mwalker33:
The lf em 4x05_dump was missing the code to print the data for the dump.
It was only showing the "special purpose" block 2 - password.
Now shows all 16 blocks of data.

### DIFF
--- a/client/cmdlfem4x.c
+++ b/client/cmdlfem4x.c
@@ -20,6 +20,7 @@
 #include "cmdparser.h"    // command_t
 #include "comms.h"
 #include "commonutil.h"
+#include "common.h"
 #include "util_posix.h"
 #include "protocols.h"
 #include "ui.h"
@@ -155,13 +156,15 @@ static int usage_lf_em4x50_write(void) {
 static int usage_lf_em4x05_dump(void) {
     PrintAndLogEx(NORMAL, "Dump EM4x05/EM4x69.  Tag must be on antenna. ");
     PrintAndLogEx(NORMAL, "");
-    PrintAndLogEx(NORMAL, "Usage:  lf em 4x05_dump [h] <pwd>");
+    PrintAndLogEx(NORMAL, "Usage:  lf em 4x05_dump [h] [s <filename prefix>] <pwd>");
     PrintAndLogEx(NORMAL, "Options:");
-    PrintAndLogEx(NORMAL, "       h         - this help");
-    PrintAndLogEx(NORMAL, "       pwd       - password (hex) (optional)");
+    PrintAndLogEx(NORMAL, "       h                     - this help");
+    PrintAndLogEx(NORMAL, "       s <filename prefix>   - overide filename prefix (optional).  Default is based on card serial number");
+    PrintAndLogEx(NORMAL, "       pwd                   - password (hex) (optional)");
     PrintAndLogEx(NORMAL, "Examples:");
     PrintAndLogEx(NORMAL, "      lf em 4x05_dump");
     PrintAndLogEx(NORMAL, "      lf em 4x05_dump 11223344");
+    PrintAndLogEx(NORMAL, "      lf em 4x50_dump s card1 11223344");
     return PM3_SUCCESS;
 }
 static int usage_lf_em4x05_read(void) {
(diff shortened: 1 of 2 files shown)


client/cmdlfem4x.c
```diff
@@ -1213,19 +1213,27 @@ static int CmdEM4x05Dump(const char *Cmd) {
...
-    PrintAndLogEx(NORMAL, "Addr | data   | ascii");
-    PrintAndLogEx(NORMAL, "-----+--------+------");
+    PrintAndLogEx(NORMAL, "Addr | data     | info");
+    PrintAndLogEx(NORMAL, "-----+----------+-------");
     for (; addr < 16; addr++) {
 
         if (addr == 2) {
             if (usePwd) {
-                PrintAndLogEx(NORMAL, " %02u | %08X", addr, pwd, word);
+                PrintAndLogEx(NORMAL, "  %02u | %08X |", addr, pwd, word);
```
iceman1001: You are missing to print all vars.

mwalker33: Yes, I copied the print from the read (single block function).  It was not printing the ascii, rather things like Lock.  What is the function to print the ascii, I can put back in.
I assume the write to file needs to be developed ?  or did that drop off somewhere as well ?

iceman1001: output  headers should follow other LF outputs.    Look inside client.util.c  for hex / ascii prints.

blk |  data | ascii  | lck
--

The dump save?   Make an array of all found blocks. 
send them into   savefile,  savefileEML,    for JSON we would need to add some more description.  We could add the lock data in it.

mwalker33: Yep, leave it with me, I will fix the headings etc.  and add the save.

iceman1001:
hm..  I don't think combining two different values inside one column is a good idea.
Even if the block is locked, you should be able to read it.  hence looking at data, ascii,  the lock information is seperate.

mwalker33:
Yep, agreed.  That was the draft for the return of the ascii.  Hold 5 till the next update :)

iceman1001:
Nice, one step closer!

- reuse swapendian macros in repo,  search for bswap.
- save; if no user filename was given,  then filename should be based on block1, UID. there is some support functions in repo.  
  

...which leads us to the small fact that ```lf t55xx dump``` doesn't save.  Yet, but it should. 
I might just as well add it. 
...which then later on leads to a  ```lf t55xx restore```  and ```lf em 4x05 restore```...

iceman1001:
Commands that should be implemented...
```
lf t55xx restore
lf em 4x05 restore/dump/read/write/wipe
lf em 4x50 restore/dump/read/write/wipe
```

mwalker33:
> Nice, one step closer!
> 
> * reuse swapendian macros in repo,  search for bswap.
> * save; if no user filename was given,  then filename should be based on block1, UID. there is some support functions in repo.
> 
> ...which leads us to the small fact that `lf t55xx dump` doesn't save. Yet, but it should.
> I might just as well add it.
> ...which then later on leads to a `lf t55xx restore` and `lf em 4x05 restore`...

Thanks for the reference for  swapendian macros (I thought it should be there somewhere).

Yep.  Lets get this little bit cleaned up and added,

I will open an issue of the other bits "LF save and restore" then we can work on it bit by bit.

mwalker33:
Should be close, let me know if anything else needs to be done (for this commit)
help updated.
used bswap_32 for endian change.
Uses serial number for default filename, unless set by user.

iceman1001:
Nice!
I am not sure the use of bswap32 twice.  Once when you store to array and once when you get uid for filename.  Is the data in LSB to start with or MSB?

iceman1001:
and lets use 'f' as parameter for filename

mwalker33:
the print %x is as read from the card and listed in the dump output (on the screen).  But if left as is will be backwards in the eml file (and bin file).  So I converted it as it was getting read.
As such, this meant that when creating the filename, it was "then" backwards, so needed to switch back.
So i see the options as
- leave as is
- have it not match somewhere 
- read, if block 1 then save somehwere for the filename build,  then convert for data save.

as is, everything reads the same and seems a little cleaner to me.

same output for reference
```
Addr | data     | ascii | info
-----+----------+-------+------
  00 | 00040072 | ...r  |
  01 | 6146D727 | aF.'  |
  02 | FEEDBEEF | ....  | password
  03 | 0000479F | ..G.  |
  04 | 0015805F | ..._  |
  05 | 18C631FF | ..1.  |
  06 | 06318C63 | .1.c  |
  07 | 11112222 | ..""  |
  08 | 22334455 | "3DU  |
  09 | 00000000 | ....  |
  10 | 00000000 | ....  |
  11 | 00000000 | ....  |
  12 | 00000000 | ....  |
  13 | 00000000 | ....  |
  14 | 00008002 | ....  | Lock
  15 | 00000000 | ....  | Lock
[+] saved 16 blocks to text file lf-4x05-6146D727-data-7.eml
[+] saved 64 bytes to binary file lf-4x05-6146D727-data-7.bin

eml file
00040072
6146D727
FEEDBEEF
0000479F
0015805F
18C631FF
06318C63
11112222
22334455
00000000
00000000
00000000
00000000
00000000
00008002
00000000
```

iceman1001:
oo.. nice!
now...  the lock bits info..   add a col for lck bits. to show wich block is lock or not...

mwalker33:
Something like this ?
lck marked with x if locked 
and f for filename

```
[usb] pm3 --> lf em 4x05_dump  f testname feedbeef
Addr | data     | ascii |lck| info
-----+----------+-------+---+-----
  00 | 00040072 | ...r  |   |
  01 | 6146D727 | aF.'  | x |
  02 | FEEDBEEF | ....  |   | password
  03 | 0000479F | ..G.  |   |
  04 | 0015805F | ..._  |   |
  05 | 18C631FF | ..1.  |   |
  06 | 06318C63 | .1.c  |   |
  07 | 11112222 | ..""  |   |
  08 | 22334455 | "3DU  |   |
  09 | 00000000 | ....  |   |
  10 | 00000000 | ....  |   |
  11 | 00000000 | ....  |   |
  12 | 00000000 | ....  |   |
  13 | 00000000 | ....  |   |
  14 | 00008002 | ....  |   | Lock
  15 | 00000000 | ....  |   | Lock
[+] saved 16 blocks to text file testname.eml
[+] saved 64 bytes to binary file testname.bin
```

iceman1001:
Yes! exactly like that. 
if you fine,  I will merge

PR #416, opened 2019-09-30
lf config s <samples to skip> option added

mwalker33:
This will allow the lf sniff to skip over x samples to allow capture of larger signals (e.g. cloners)

With the unint_32 for the samples to skip, the s <value> will be reset to 0 if not supplied on the lf config command.  If needed to retain the last value set, we could use 0xffffffff as the "default" to mean don't change from last set value set (comment lines left in code).


iceman1001:
...and since it is based on commit from @marshmellow42 , don't forget to mention that in the changelog.

mwalker33:
Good point.  Done.

iceman1001:
Excellent!

Issue #419, opened 2019-10-05
Color on windows 10


mwalker33:
**Describe the solution you'd like** 
Enable ansi color if a users windows can support it.

**Additional context**
I have the latest version of windows : Microsoft Windows [Version 10.0.18362.388]

In the proxmark3.c file (around line 684)
If I make this change     : session.supports_colors = true; 

Then colors seem to work (i.e. on my windows ansi colors will work if enabled)
To enable, a registry key needs to be set (for Windows 10.0.16257 and later)

[HKEY_CURRENT_USER\Console]
"VirtualTerminalLevel"=dword:00000001

Note: if the above key entry is not in the registry or is set to 0, then the ANSI colors don't print and session.supports_colors will still need to be set to false to ensure a clean output.

I have some test code working that will check the registry to see if the key is set to 1, and if so ,set session.supports_colors = true, else leave at false.

If we think this is a good idea, I will submit a PR for review and comments.

doegox:
Nice ! 
If registry test can be done in a reliable way without hurting the other platforms, why not.

iceman1001:
if we can easily access the registry under windows.  Go go go!

mwalker33:
If your happy, we can close this now.

PR #420, opened 2019-10-05
Color on windows 10

mwalker33:
For review and testing.  (Issue #419)

Might be worth testing on different setups prior to merge!

If your happy, I think we need to add the reg key change in the docs/help somewhere.  Where should that be done?


iceman1001:
...rebase?  anyway,  lets try it.

doegox:
doc: I'd say somewhere around https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/Windows-Installation-Instructions.md#launch-proxspace

mwalker33:
Another option would be to create a "Settings" file (say settings.json) stored/created in the same folder as proxmark3 executable.  Then save the flag to override the default.  i.e. if exists use this value, else what ever the code currently does.

While this is not automatic, it would be more generic and could serve as the base for other things we may want to client to store and use as a user preference.  Since, at the moment the user will need to patch the registry, to set a flag should be valid.

If you would prefer this, let me know (and supply the answer to these two questions)
- What would you prefer the settings name and format to be in ?
- Is there a variable that currently holds the path to the executable ?

doegox:
If there is a settings file, it should be e.g. `~/.proxmark3/resources/settings.json` and searched with `searchFile(&path, RESOURCES_SUBDIR, "settings", ".json", true);` so it can also be at other standard locations (/usr/local/share/proxmark3/resources/, (exec)/resources/ etc).

But back to the flag: I don't understand, you want to have the flag to set color in that file, regardless the status of the registry? It looks to me like twice the work, with a risk of flag & registry desync and ppl complaining about weird output...

mwalker33:
@doegox the settings file was a different way to allow it to work.  i.e. in response to "...rebase?" will the windows API for registry still work?  I am happy to leave the registry as is, as its simple and automatic.

iceman1001:
..rebase was a suggestion for your pr.  it has some commit messages that isn't involved in your PR.

mwalker33:
ok.   it should be clean for merge.  I checked all changes prior to the pull request (those changes had been undone prior to these works.

I added the quick reg. key to the windows install md file.

If you think there is an issue let me know.

Unless there are any other comments, I think we are good to go.
Happy to delay if you want to test on your platforms.

mwalker33:
T5577 guide
> that looks awesome, can we hope to see it back ?

Yep, I have updated it to match the RRG style.  
Like all these things plenty to do.
I want to add a few minor things, then the block 3 page 1 config (basics).
Once I get that roughed in, I will setup an issue to discuss and submit a PR to get it out there.  
Once published, we can then update and fix anything not clear or missing (and typos)

PR #423, opened 2019-10-07
Display high bit for Kastle HID to allow for lf hid clone

swg0101:
Currently, the high bit for Kastle HID is omitted from the output, which results in an invalid clone ID when lf hid clone [id] is used. Displaying the high bit along with the rest of the ID would allow these ID cards to be cloned properly.


iceman1001:
That was a sneaky one.  32bits format should have fit in u32 "lo" variable. 
Don't forget a line in Changelog.md

Issue #418, opened 2019-10-02
Password checking for T55xx detect and writes


swg0101:
Currently, it looks like the T55xx password is only checked on reads, but not detects or writes. However, manually specifying a password for these commands will still perma-brick a card or tag if no password is currently set.

Would it make sense for the client to disallow these commands with a password (unless forced) if a "t55xx detect" command succeeds without a password before conducting the actual commands?

This should stop more people from bricking their T55xx cards (which I learned the hard way).

Thanks!

mwalker33:
Not sure how a password protected write (when no password is set) can result in a "bricked" card.
At best it would set the lock bit (and stop you from changing that block), even then i cant see how it would do that as a password protected write can only be seen as a password protected write; which should result in a command sequence error (too many bits).
If you make a password protected read when the password is not set, this results in the command looking like a valid non-password write, which can set the lock bit (as the bit count is the same).

Do you have an example of a command that bricked the card and the exact card/chip type (e.g.. t5555, t5557, t5577) ?  I would like to re-read the specific data sheet.
 
Note: On many test runs of have sent password protected writes to a non password protected t5577  If the command was to be interpreted as a non password write it should have locked the block (as the Lock bit would have been set (1) based on the password i use in testing.  This has never happened.

swg0101:
The cards I had were T5577, and if I remember correctly I did:
lf hid clone 2
lf t5 wr b 0 d 0 p feedbeef

and the card bricked after that and won't respond to any commands.

iceman1001:
pwd checks for write is as @mwalker33 says,  not needed.
For t55xxDetect is makes sense to add the same checks as t55xxread has.

swg0101:
> That would write an invalid config to the card. Try the lf t55 wipe command and see if it comes back to life.

Nope (usually that's the first thing I try - along with writing something easily searchable like lf hid clone 2)... even a lf search u returns just noise. It could be that I have also mixed in a detect with password in my testing that broke the tags.

mwalker33:
Just did some tests on a t5577
the write with password when no password was set worked as a normal write command.  
i.e. the data was written and the password ignored (as expected)

The lf t55 detect  with password will have some undesired results depending on the password used.
e.g. if the most significant bit is set 0x80000000 then that will set the lock bit on block 0 and the card will not detect.  this can normally be recovered (for the t5577) with test mode.
    lf t55 write b 0 d 000880E0 t

[...]

PR #425, opened 2019-10-08
lf t55 detect without password first and help update

mwalker33:
some fixes for issue #418 

lf t55xx detect will try without password first and if not detected and a password supplied, will loop back and try with the supplied password.
The config structure now gets the pwd flag set and stores the password (if used)
The detect will show password status and the password (if password was used)

Updated the lf t55xx .... help to show the current downlink mode that will be used as the default if no r option supplied

Typo fix for the win10 help (proxmark.exe to proxmark3.exe)


doegox:
oops I see now you're filling the changelog from the bottom. New changes should be put at the top, not at the bottom. You may move your previous changelog entries as well

PR #438, opened 2019-10-15
WIP Attempt to document clocks and timers

doegox:
Feel free to comment / fix  / complete.
I wrote what I understood so far.
Still I don't understand why we have both Ticks and CountUS set of functions using the same timers for what looks like similar purposes

### DIFF
--- a/doc/clocks.md
+++ b/doc/clocks.md
@@ -1,0 +1,144 @@
+## Slow clock
+
+~32kHz internal RC clock
+
+Can be between 22 and 42 kHz
+
+## Main Oscillator / MAINCK
+
+cf `PMC_MOR` register
+
+16 MHz, based on external Xtal
+
+## PLL clock
+
+cf `PMC_PLLR` register
+
+96 MHz (MAINCK * 12 / 2)
+
+## Master Clock MCK, Processor Clock PCK, USB clock UDPCK
+
+cf `common_arm/clocks.c`
+
+cf `PMC_MCKR` and `PMC_SCER` registers
+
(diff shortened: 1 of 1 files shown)


iceman1001:
Great work compiling it to a document.  Long overdue.  In my newbie years I got bitten by these clock functions several times.  Never made sence but I tried to sort things out one by one,  with more comment in the code to document pitfalls like WDT...
Anyway,  having fewer clocks based on T0/T1/T2 would be nice.   With delta time.  And with clear comments saying which functions can't be used together.  And highest precision.

It is very popular among the code,  like you seen in legic,  or pcf5731, or hitag2-s,  etc to just take a timer (T) configure it and run their code.  Which is fine if you clean up afterwards and reset timer settings.  This is one reason for the device to become unresponsive.  We tried to clean up most of this but more cleaning is needed as you see.

The precision is very critical in most of our operations on the device side.
The 21.3 us tick was an eyeopener once upon a time.  Never understood why the pwm pulse was off until ...

Getting all clocks in order and with a precise purpose of the clocks we keep is a must.  This is one of our more important changes as I see it.

iceman1001:
the clock that counts to 0xC000 was used on one of those hitag/ti/pcf5731 styled commands.  
To be one timeperiod.    We don't use that style of code any longer but the clock lives on

iceman1001:
We might just merge this one and update the doc as we do things

doegox:
ok I just didn't  want to push wrong info but if you didn't detect probs so far yeah we can merge

PR #446, opened 2019-10-20
Fix pack in dump

grspy:
PACK wasn't included in the output dump due to incorrect check of ul_auth_select return value


iceman1001:
Good catch!

PR #447, opened 2019-10-22
Felica support: add command request service

7homasSutter:


iceman1001:
Great!

now comments...  Take it the right way.  
- The individual of parsing command line arguments isn't good.  We have support functions inside client/util.c    I know the FeliCa cmds haven't used much of them earlier but since we are remaking it lets start doing so. 
- Same thing with CRC,  I belive we use macros for them in the other places.  
- Helptexts,  we try using full commands in examples and text ,  not abbreviated.

iceman1001:
regardless,  it doable to merge and push fixes for it later.

7homasSutter:
> Great!
> 
> now comments... Take it the right way.
> 
> * The individual of parsing command line arguments isn't good.  We have support functions inside client/util.c    I know the FeliCa cmds haven't used much of them earlier but since we are remaking it lets start doing so.
> * Same thing with CRC,  I belive we use macros for them in the other places.
> * Helptexts,  we try using full commands in examples and text ,  not abbreviated.

Nice to know! I will fix that

client/cmdhffelica.c
```diff
@@ -339,12 +306,14 @@ static int CmdHFFelicaRequestService(const char *Cmd) {
     if (all_nodes) {
         for (uint16_t y = 1; y < 32; y++) {
             data[10] = int_to_hex(&y);
-            add_crc_bytes(&datalen, data, sizeof(data));
+            AddCrc(data, datalen);
```
iceman1001: you add crc,  but don't increase datalen,   and after sending you remove 2 bytes from datalen...

7homasSutter: Yeah true, I'm changing the parsing of the raw and request service with the methods from util.c and will test then everything again. Wait with merging please until I've done all the changes.

7homasSutter: Btw are there any unit tests, I could write?

iceman1001: not really... :(

7homasSutter: Users are the best unit tests anyway ;-D

The util.c helps a lot in parsing. I appreciate the feedback thanks for that! I will refactor the raw command in another pull request - then I will directly add a length parameter to the command.

iceman1001: yeah,  we get fast feedback when stuff breaks despite travis, appveyor and coverity  :)

7homasSutter:
If the tests run through, I think you can test it again and merge if everything is ok.

doegox:
ok let's merge, thanks for your contribution!

Issue #448, opened 2019-10-23
Broken "lf em 410x_sim"


temskiy:
The latest commit with worked function "lf em 410x_sim"  is 945fb82b48054d4096f82fda892abd21422a0025

**To Reproduce**
1. Build client & firmware at any commit after 945fb82b48054d4096f82fda892abd21422a0025
2. Flash firmwatre to PM
3. Start in client emulating em4100 tag with any ID

iceman1001:
...did you flash bootrom and fullimage?

temskiy:
Yes. After compiling i run "pm3-flash-all"

iceman1001:
And your output from 
```
hw status
hw tune
```

temskiy:
In attached file
[log.txt](https://github.com/RfidResearchGroup/proxmark3/files/3762176/log.txt)

iceman1001:
ok,  @mwalker33  confirmed the bug.  Its related to a coverity fix I tried.  He will have a look at it soon.

mwalker33:
This should now be fixed.  Can you update with the latest and confirm it's working for you.

temskiy:
Works fine
Thanks

iceman1001:
Thanks @temskiy  for reporting it in  and  thanks @mwalker33  for the fix!

temskiy:
It`s broken again.

PR #449, opened 2019-10-24
Broken "lf em 410x_sim" fix

mwalker33:
Issue  #448
Incorrect length was used in memcpy.
Added :
uint16_t len = MIN(BIGBUF_SIZE - offset,PM3_CMD_DATA_SIZE - 3);
this should ensure data copied does not go past the end of bigbuf and the correct number of bytes are copied (if room in bigbuf).


doegox:
thanks!

iceman1001:
...great catch and nice fix!
Maybe we should add support to BigBuff functions to    BigBuff_copy  which checks limits.
I think everytime we see  get_bigbuff_address calls,  and playing with its pointer we should keep a keen eye on it.  Luckily we don't have this many calls,  since we started with Malloc.   Which I think could also need checks...

PR #451, opened 2019-10-24
Felica support: add commands request response and read without encryption

7homasSutter:
Add two new commands:
- request response
- read without encryption


iceman1001:
excellent!

PR #453, opened 2019-10-26
Win10 color update

mwalker33:
I have found that in order to enable win10 color support (on the later versions of windows 10), a 2nd value may need to be set.  It is unclear if this is default or not.
The client also needs to un-check use legacy console (or update a 2nd key)

I have added a 2nd registry key check, to ensure windows 10 color will only be enabled if BOTH values are set (which is needed to work correctly).

MD install also update to reflect the 2nd settings.


doc/md/Installation_Instructions/Windows-Installation-Instructions.md
```diff
@@ -135,4 +135,12 @@ In later versions of windows 10 you may be able to get color to work by setting
...
 ```
-If after setting this key (and restarting proxmark3.exe) you get extra characters and no color text, either delete the key or set the value to 0
+You also need to disable "use legacy console" in the cmd.exe properties, or set the following registry key
+```
+[HKEY_CURRENT_USER\Console]
+    "ForceV2"=dword:00000001
+```
+After making these changes, you will need to start a new command prompt (cmd.exe) to ensure its using the new settings.
+
+If after making these changes (and restarting proxmark3.exe) you get extra characters and no color text, set either key to 0 or enable legacy mode again (and restart the command prompt).
```
doegox: last sentence: compared to the old sentence (if it doesn't work, delete VTL key or set key=0), this sentence seems to talk only about second key FV2.
I guess this should be rephrased such as "restore to the initial situation by either deleting the keys or setting their value to 0"

doegox:
BTW @mwalker33 what about Windows Terminal Preview 1910 Release ? Is pm3 color mgmt ready for it ?

doc/md/Installation_Instructions/Windows-Installation-Instructions.md
mwalker33: I will have a think and reword.

With the new code if 'either' of them are not set then the it will not enable color. (this will catch the users that have the legacy unchecked for some other reason.
i.e. I am treating them as prerequisites for color to be on.
So "set either key to 0" will turn off color, OR enable legacy mode (which is the user friendly way to turn it off.

mwalker33:
> BTW @mwalker33 what about Windows Terminal Preview 1910 Release ? Is pm3 color mgmt ready for it ?

A quick play.  I downloaded the terminal preview from windows app store.
It seems to run PowerShell (by default).
Color seems to just work. i.e. I can "disable" the registry checks and force color to be used and it works - So if/when that becomes the default it will be simpler..
I run a cmd.exe from inside the powershell/terminal and color still seems to work (with no tweaks).  

So as is, color will work in both if the reg settings are set (on cmd.exe and terminal preview),
and wont display color (in either) if either key is not set.

I think the settings file is still the better option (long term) so I see this as being more for the "as is" until I can get it completed.

i.e.
in proxmark3.exe … something like "settings WindowsColor <on|off>
Then some cheat sheets/scripts???? to help the user do what ever is needed.

I have started to test the json settings file and will raise an issue when ready for testing and comments (lots of different ways to do this, and want to make it expandable just incase other things need to be set latter).

doegox:
ok let's merge already what is available

iceman1001:
Yup,   the terminal used when compiling in proxspace doesn't allow for ansi art.   The cygwin terminal does,   and as seen in history of code we didn't allow for ansi art on Windows.
Then came WSL1 and 2.  Which can use ansi art.    Which made some interesting macro defines statements and now with the registry fix enables in certain scenarios.   Windows support... is a mess.

PR #456, opened 2019-10-29
Felica support: add write without encryption command

7homasSutter:
Add new command which allows to write data to the not encrypted memory.


iceman1001:
Excellent!

PR #457, opened 2019-10-29
make warning about ModemManager louder

rileytg:
#updatingforafriend


rileytg:
@mgdickinson

iceman1001:
the underlaying problem is ppl not reading instructions...   I doubt strongly that these kinds of enhancements help much against that given my emperical studies of the subject.

PR #459, opened 2019-11-01
Felica support sniffing

7homasSutter:
Hopefully this will fix the sniffing and allow to sniff all cards and readers.


iceman1001:
Excellent!

PR #462, opened 2019-11-02
Fixed hf_bog standalone and added new lua script to use SPIFFS

bogiton:


iceman1001:
Nice one!
Misses  "make style"  and the 0x40000 max length,  don't we have a macro defined for it?  I think we limit the spiffs to use 128 Kb at the moment. 

Will you also remove the old script?  Or shall we refurbish it?

bogiton:
Ah, make style. Damn, I have to get used to the new workflow.
Yeah, the 0x40000 is defined as FLASH_MEM_MAX_SIZE in pmflash.h but I wasn't sure if we should include this file in scripting.
The old script has some additions (by you I believe) which may be useful if mem is used without spiffs. Although, we could make another script that utilizes the spiffs for the dictionaries and configs of the last memory page instead and leave the new one for the pwds.

iceman1001:
yeah,  the new workflow is better.  Makes the code better and nicer.
The max isn't what we use for SPIFFS current. 

old script, ok, it is for dictionaries aswell.. Maybe remove your part inside of it,

bogiton:
Truth is that the FLASH_MEM_MAX_SIZE is never used, since the length is being acquired automatically from spiffs itself. And to be honest, I used more or less the logic of Colin's CmdFlashMemSpiFFSDump in cmdflashmemspiffs.c by adapting it for scripting.
Yeah, old script can be used just for printing the mfc keys, iclass keys and t55xx pwds, but maybe also change its name?

iceman1001:
changing name would be good.  "print_dic_spiffs.lua"  :)  Problem with lua file names is that we don't have tab completion nor "find the first matching of a short file name"  like pm3 cmds has.

PR #463, opened 2019-11-04
Felica support: Add new commands

7homasSutter:
Adds three new FeliCa commands:
Request Specification Version: Gives the Card OS information.
Reset Mode: Sets the Auth Mode back to 00h.
Request System Code: Gives a list of all system codes.

Bug-fix:
Fix a segmentation fault in the raw command when no card is on the reader.


iceman1001:
Excellent.

PR #464, opened 2019-11-04
Update aid list

merlokk:


iceman1001:
There is unicode problems with this commit.    @merlokk  you need to fix it.

merlokk:
yes(   fixed, but ill upload if tomorrow

PR #471, opened 2019-11-08
read_pwd_mem lua scripts update

bogiton:


iceman1001:
Excellent,   how about  **-d**  as "delete" instead of "r" as in remove  since we use "d" more elsewhere.

bogiton:
Sure, we could change it to "-d". I will include it in a future PR.

PR #478, opened 2019-11-19
switch from lua CRC16/ARC to CRC RevEng

0xDRRB:
CRC RevEng works if it is used correctly ;)
The Lua implementation of CRC16 / ARC is useless.
(+ update CHANGELOG.md)


iceman1001:
I was impressed you wrote a crc implementation in lua.    Now I can't even remember runmodel addition to lua.   Good thing it works :)

PR #480, opened 2019-11-22
Added support for some ISO15693 custom NXP commands

grspy:
plus some textual fixes


iceman1001:
Great!
But the dual implementation of defines is something that we should fix at one stage.

grspy:
As far as I can tell the definitions in the protocols.h file are used for the trace listing. At least for the iso15 operations.

iceman1001:
to be fare,  its the same function of them :)

PR #481, opened 2019-11-25
Add ISO15693 write AFI/DSFID commands and NXP Sysinfo for Slix2

grspy:


iceman1001:
Cool!
Now,  return codes..  
and turn off the rf-field...    This is the major downside with current ISo15 commands.

client/cmdhf15.c
```diff
@@ -945,7 +956,8 @@ static int CmdHF15FindAfi(const char *Cmd) {
 
     clearCommandBuffer();
     SendCommandMIX(CMD_HF_ISO15693_FINDAFI, strtol(Cmd, NULL, 0), 0, 0, NULL, 0);
-    return 0;
+    DropField();
```
iceman1001: This one is harder,   the FINDAFI command doesn't report back when it finished running.
So doing this, would interrupt the device side procedure aka command will not bruteforce AFi..

client/cmdhf15.c
```diff
@@ -1270,6 +1291,8 @@ static int CmdHF15Raw(const char *Cmd) {
             PrintAndLogEx(WARNING, "timeout while waiting for reply.");
         }
     }
+
+    DropField();
```
iceman1001: this one is also problematic.  RAW commands allows for the field to stay up...
this would interrupt that process..

grspy: Maybe we should add a "leave field on" option for it?

client/cmdhf15.c
grspy: Yes, I thought so too, but I tried it after that change and it did find a couple of AFIs. But maybe it was just pure luck it found those.

iceman1001: I think the solution is for the command to send a message back to the client once its done.
and have the client loop / wait for that response.

client/cmdhf15.c
iceman1001: Yes, I agree, there should be like 14a / 14b raw commands,   the option to restart/leave on/ etc..

iceman1001:
And add a line the the changelog.md file :)

client/cmdhf15.c
iceman1001: better to loop 2sec waittimeout,  look in "hf mf darkside" code for instance.   It will give a unified command execution.

armsrc/iso15693.c
```diff
@@ -968,12 +969,19 @@ void BruteforceIso15693Afi(uint32_t speed) {
 
         if (BUTTON_PRESS()) {
             DbpString("button pressed, aborting..");
+            aborted = true;
```
iceman1001: maybe assign BUTTON_PRESS to the bool?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Added `hf 15 writeafi`, `hf 15 writedsfid` and detailed info for SLIX2 tags in `hf 15 info`. Also did some refactoring in HF15 commands. (@grspy)
```
iceman1001: ok ok,  one line isn't enough.  make three entries.  one for each command change, please.

iceman1001:
Excellent!

PR #484, opened 2019-12-01
hf mfp check

merlokk:
check keys for mf plus


iceman1001:
Great!

iceman1001:
You could have divided this one with the loadkeys / new keys jsonformat in a seperate PR.
Not to mention we will most likely have a better key json format, see internal road map issue.

PR #504, opened 2019-12-31
Fix hitagS reader output

bosb:
`lf hitag reader 02` on a hitag S tag showed only the first byte of every page:
#db# Authenticating using key:
#db# 00 00 00 00 00 00
#db# Page[ 0]: 00 00 00 5F
#db# Page[ 1]: 00 00 00 C9
#db# Page[ 2]: 00 00 00 48
#db# Page[ 3]: 00 00 00 4D
#db# Page[ 4]: 00 00 00 00
#db# Page[ 5]: 00 00 00 00
#db# Page[ 6]: 00 00 00 00
#db# Page[ 7]: 00 00 00 57

fixed in `ReadHitagS`, so reader command now shows all bytes.
... but I also suspect function `SimulateHitagSTag` to be involved...
Only other reader to verify I have is a RFIDler which does not yet want to talk with proxmark rdv4 :-/
Will give it some more tries...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -398,6 +398,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Added trailer block decoding to `hf mf rdbl` and `hf mf cgetbl` (@merlokk)
  - Added `hf mf mad` and `hf mfp mad` MAD decode, check and print commands (@merlokk)
  - Add `script run luxeodump` (@0xdrrb)
+ - Fix `lf hitag reader 02` - print all bytes (@bosb)
 
 
 ### Fixed
--- a/armsrc/hitagS.c
+++ b/armsrc/hitagS.c
@@ -1273,7 +1273,7 @@ void ReadHitagS(hitag_function htf, hitag_data *htd) {
         } else if (tag.pstate == HT_SELECTED
                    && tag.tstate == HT_READING_PAGE
                    && rxlen > 0) {
-            //save received data
+            //save received data - 40 bits
             z = 0;
             for (i = 0; i < 5; i++) {
                 for (j = 0; j < 8; j++) {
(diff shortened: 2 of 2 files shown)


iceman1001:
Nice catch!
I get the feeling the sendnum is not increasing as it should given the old algo...

iceman1001:
Question is how often sendNum increases?  
And don't forget to add a row in CHANGELOG.md

bosb:
It's handled as a page structure instead of a byte by byte...so the counter variable name is misleading, will change that.

While reading through the whole file now:
How is philosophy about data represented and displayed in general in proxmark project?
General assumption: Order as the bytes go through air? Printed from left to right.

a) Example, first page from tag printed in hitag S datasheet:
MSByte -> LSByte
`page 0: UID_3 UID_2 UID_1 UID_0`
 first byte transmitted is `UID_0`

b) c data structure, I would assume:
each page has 4 bytes
```
page[0][0] = UID_0
page[0][1] = UID_1
...
```
(which seems to be the other way around in the file, means: page[0][0] = UID_3)

c) logtrace entry:
`... UID_0 UID_1 UID_2 UID_3 ...`

d) print out I would prefer in Dbprintf, because its easier to compare to datasheet
`"#db# Page[ 0]: UID_3 UID_2 UID_1 UID_0"`
(in rfidler the data output is the other way around, `"page 0: UID_0 UID_1 UID_2 UID_3"`)
(While looking at it, for consistency it would also be better to use the same as rfidler)

e) when talking about the UID of the tag in hex: UID_0 UID_1 UID_2 UID_3

f) tag in bin file:
`UID_0 UID_1 UID_2 UID_3 ...`

iceman1001:
mmm..  I know,  isn't it lovely how things are.   You have correct identified one of the many concurrency issues with rfid hacking and proxmark3 source code.   As you noticed its a mixed experience but usually there is a system behind all different design decisions.   

I tend to prefer to get the data as the datasheet states when it comes to   readblock/writeblock/dump/restore.   

But for commands like **info** / **reader**  I want to see UID like its used in the PACS system software.  Or how the company themselfs uses uid/csn.

UID are funny that way,  compare MFC / MFU / ISO15693 how they store it on board the card.

If you feel me,  that is how I want the hitag commands to behave.  Now,  to convert the rfidlers special attacks against hitag,  we will need to adapt to rfidler does.   Getting function in is more funnier that an idea of **correctness** or **codestyle**

iceman1001:
What is the purpose with printing the memory when simulating?

bosb:
I wanted to know which content I loaded and by printing if the config decoding at least for the size of the tag was right, indicated by the printed number of lines. Is it uncommon to print the data in that case?

iceman1001:
for your debugging stuff, no,
for the normal **sim** commands in the client,  yes.

Put a guard around it,    and you can access it whenever you want.
```
 if (DBGLEVEL >= DBG_EXTENDED)
```

iceman1001:
Excellent!

PR #508, opened 2020-01-01
add pac trace file

danshuk:


iceman1001:
Awesome!  Thanks bro!

iceman1001:
and... now that I merged it,  I remember,  you will need to add a line in /traces/README.txt aswell.
Trying to keep documentation on top of things.

iceman1001:
and maybe implement the  `lf pac sim`  ??

danshuk:
Sure, I'll update the readme now. I plan to implement the sim, but that needs some deeper thought.. I need to read more closely lfops.c etc. as maybe I need to implement nrz waveform generation, unless I missed something...

iceman1001:
Don't think you need to do much.  Search the other cmdlf* files for a lf system with nrz and look into sim ask.

PR #515, opened 2020-01-10
Correct usage for 'hf mf csave'

TheFreeman193:
Corrected csave usage to show `o` flag for output file instead of `i`.


iceman1001:
Good catch!

PR #518, opened 2020-01-13
fix: make libsndfile optional

micolous:
This makes `libsndfile` an optional dependency, in the same way as Qt (`SKIPQT`).

It is only used for one feature, and `libsndfile` depends on a number of other libraries.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -22,7 +22,7 @@ vpath %.dic dictionaries
 OBJDIR = obj
 
 LDLIBS ?= -L/usr/local/lib
-LDLIBS += -lreadline -lsndfile -lpthread -lm
+LDLIBS += -lreadline -lpthread -lm
 
 # RPi Zero gcc requires -latomic
 # but MacOSX /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld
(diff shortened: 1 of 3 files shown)


iceman1001:
So the reason for not supporting wav saving is because it depends on number of other libs?

Based on that, I don't see a reason for making optional.  The reason for skiping QT is because the troubles ppl having when compiling in QT4 vs QT5 vs non QT environments over multiple platforms.

You would have to present some real reasons for this change, where this would make an positive impact on the user experience.

For merging,  
The current methodolgy for the client is when not supporting a command is to hide that command.

We changed the client to have checks for which commands the client/device supports at run time.
See the command table.  Meaning when not compiled with snd,  the **data save w** option should not be visible.   

And the makefile changes needs to be verified on *nix, osx, mingw, wsl, ps322. 
The pkg-config checks for skipQT has a failsafe if pkg-config call fails.  You would need to add the same here. 

All of which adds some nice complexity to the client for _not_ supporting a command.

doegox:
The alternative is just to implement ourselves the static conversion to wav...

iceman1001:
I belive the suggestion from @doegox is the way forward.

doegox:
see https://github.com/RfidResearchGroup/proxmark3/pull/521

iceman1001:
That should sort out the issue @micolous  had with the extra dependencies.

micolous:
Yup, #521 works for me, thanks.

PR #521, opened 2020-01-13
Remove libsndfile dep and code our own savewav

doegox:
see if it's ok for you @iceman1001


iceman1001:
Excellent!
Besides the struct should be in the header file its looking good.

doegox:
Note that struct has no reason to be exposed through the header to other parts of the code, so I just moved it to top of fct

PR #522, opened 2020-01-14
Update mfc_default_keys.dic

h4waii:
Added various keys from hotel cards found through hardnested attacks


iceman1001:
Excellent,  do you have a indication over which hotels?

h4waii:
There should be a more anonymous and automated way to provide new keys. I'm assuming some users would like to share keys but don't want to give out information about where (geographically, organizationally, etc) they are to or were found.

So a way to check against duplicate and upload automatically would be a good way to get keys from users who aren't aware they can share their keys in an anonymous fashion.

Perhaps this could be added to `f/chk`, `hardnested` or `autopwn` script as an argument?

opt-in vs opt-out is the discussion.

PR #537, opened 2020-01-22
Support LTO-CM read block and add CM-Type info into "hf lto info" command.

Kevin-Nakamoto:
Tested Commands: 
hf lto info
hf lto rdbl 
hf lto rdbl h
hf lto rdbl s 0 e 254
hf lto rdbl s 0 e 0
hf lto list


iceman1001:
excellent +1

Don't forget to add a line in  CHANGELOG.md  for documentation

PR #544, opened 2020-01-29
added hf lto dump

Kevin-Nakamoto:
Tested commands:

- No file name specified:  
`[usb] pm3 --> hf lto dump`
`[+] saved 8160  bytes to binary file XXXXXXXXXX.bin`
`[+] saved 255  blocks to text file XXXXXXXXXX.eml`
where XXXXXXXXXX = UID of the tag in hex.

- File name specified by f option: 
`[usb] pm3 --> hf lto dump f test`
`[+] saved 8160  bytes to binary file test.bin`
`[+] saved 255  blocks to text file test.eml`

### DIFF
--- a/client/cmdhflto.c
+++ b/client/cmdhflto.c
@@ -20,6 +20,9 @@
 #include "ui.h"
 #include "cmdhf14a.h"
 #include "protocols.h"
+#include "fileutils.h"  //saveFile
+
+#define CM_MEM_MAX_SIZE     0x1FE0  // (32byte/block * 255block = 8160byte)
 
 static int CmdHelp(const char *Cmd);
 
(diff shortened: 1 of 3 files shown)


client/cmdhflto.c
```diff
@@ -428,19 +482,36 @@ static int CmdHfLTODump(const char *Cmd) {
...
+        return ret_val; 
+    }
+
+    // save to file 
+    if (filename[0] == '\0') {
+        memcpy(filename, sprint_hex_inrow(serial_number, sizeof(serial_number)), sizeof(serial_number) * 2);
+    }
+    saveFile(filename, ".bin", dump, dump_len);
+    saveFileEML(filename, dump, dump_len, 32);
+
```
iceman1001: I think you can use the fillUID function to get the uid from the first bytes.

The name convention for dump files is:   'hf_lto_xxxxxx.extension'  where xxxxx is UID (or serial number in this case.

iceman1001:
excellent!
Some minor thing with uid filename, next step json support?

PR #547, opened 2020-01-30
fix: logic inversion in loadFileDICTIONARY_safe use of CheckStringIsHEXValue

micolous:
Before this patch:

```
[usb] pm3 --> hf iclass chk f client/dictionaries/iclass_default_keys.dic
[+] loaded  0 keys from dictionary file client/dictionaries/iclass_default_keys.dic
[usb] pm3 -->
```

After this patch:

```
[usb] pm3 --> hf iclass chk f client/dictionaries/iclass_default_keys.dic
[+] loaded  7 keys from dictionary file client/dictionaries/iclass_default_keys.dic
[+] Reading tag CSN
(further communication continues)
```

It looks like the regression came from #484; it was fixed in 94eb741a4 for `loadFileDICTIONARYEx`, but not `loadFileDICTIONARY_safe`.

### DIFF
--- a/client/fileutils.c
+++ b/client/fileutils.c
@@ -1016,7 +1016,7 @@ int loadFileDICTIONARY_safe(const char *preferredName, void **pdata, uint8_t key
         if (line[0] == '#')
             continue;
 
-        if (CheckStringIsHEXValue(line))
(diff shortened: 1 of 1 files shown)


iceman1001:
Great catch!

PR #548, opened 2020-01-30
feature: standalone sniff iso14a to flash (HF_14ASNIFF)

micolous:
This standalone module sniffs ISO14a packets and writes it to flash as `hf_14asniff.trc`.

It is based on `HF_BOG`, but that only logs ULC/NTAG/ULEV1 auth; where as I'd like to get complete traces.

Because this module only logs the traces to flash and doesn't do anything special with it, it just calls out to `SniffIso14443a()`.

Documentation is included at the start of the file (similar to other modes).


iceman1001:
Great!
Is it fast enough to append to a file when sniffing like that?
and the **.trc** extension,  the one we use is  **.trace**

iceman1001:
And don't forget a line in CHANGELOG.md..
maybe add your name to the  **modinfo()**   "... by Micolous"

doegox:
Please add it also to the doc/md/Use_of_Proxmark/4_Advanced-compilation-parameters.md table (bottom of the page)

PR #549, opened 2020-02-03
appveyour fix

merlokk:


iceman1001:
cool, does it work?

merlokk:
ok, now it works

PR #553, opened 2020-02-05
Fixed the url for smartcard ATR lookup

TheDuchy:
A small change that just follows the redirect on the page as it was apparently moved.


iceman1001:
Thanks!

PR #559, opened 2020-02-14
hitagS print whole memory + print succeeded challenges

bosb:
1. hitagS read would stop after 2 pages when read with challenge or key.
2. additional to print failed challenges, print succeeded challenges in a format to easier use/grep with hitagcrack functions.


iceman1001:
Excellent

Issue #564, opened 2020-02-21
Convert all python scripts to python3


doegox:
- [x] ./fpga/tests/plot_edgedetect.py
- [x] ./tools/findbits.py
- [x] ./tools/findbits_test.py
- [x] ./tools/pm3_eml2mfd.py
- [x] ./tools/pm3_eml_mfd_test.py
- [x] ./tools/pm3_mfdread.py
- [x] ./tools/pm3_mfd2eml.py
- [x] ./tools/xorcheck.py

And this one can probably be fixed to python3 too:
- [ ] ./tools/pm3_mf7b_wipe.py #! /usr/bin/env python3.6

iceman1001:
@sigwinch28 
If you are up to it,  start converting to python3 would be nice.

sigwinch28:
I will see how far i get with `2to3`.

I'll catalog the dependencies that the scripts have, too: some have numpy dependencies, etc.

iceman1001:
that is a good way forward.

sigwinch28:
Dependency search:

* [x]  `./fpga/tests/plot_edgedetect.py`: `numpy`, `matplotlib`
* [x] `./tools/btaddon/hc06_factory.py`: `pySerial`?
* [x] `./tools/btaddon/hc06_console_AT.py`: `pySerial`?
* [x]  `./tools/xorcheck.py`: no external dependencies
* [x]  `./tools/pm3_mfdread.py`: `bitstring`
* [x]  `./tools/findbits_test.py`: no external dependencies, imports `findbits` locally.
* [x]  `./tools/pm3_mfd2eml.py`: no external dependencies
* [x]  `./tools/findbits.py`: no external dependencies
* [x]  `./tools/pm3_eml2mfd.py`: no external dependencies
* [x]  `./tools/pm3_eml_mfd_test.py`: no external dependencies, imports `pm3_eml2mfd` and `pm3_mfd2eml` locally
* [x]  `./tools/pm3_mf7b_wipe.py`: no external dependencies

Looks like the dependencies are:
- `fpga/tests`: `numpy` and `matplotlib`
- `tools`: `bitstring`
- `tools/btaddon`: `pySerial` (unconfirmed)

doegox:
Thanks for the list!
It is a bit overkill to require every pm3 user to install these dependencies.
These scripts are for very specific and rare use.
I prefer to catch properly the errors and guide the user.

sigwinch28:
[...]

PR #565, opened 2020-02-21
convert python2 mfd2eml and eml2mfd scripts to python3

sigwinch28:
Works towards #564 

Need someone to check my bytes<->string stuff is okay in python3, to make sure I'm not testing _nothing_ here.


doegox:
sounds good to me, tested locally
thanks!

PR #572, opened 2020-02-23
Partial Keri MS

mwalker33:
Added a function that can encode/decode the MS data.
Due to needing more samples for some bit offsets, I added a "bit map" to show which bits have been decoded and of those which bits may not be correct.
To complete the mapping, we need more samples for the FC and ID, then need to update the mapping arrays.


iceman1001:
excellent!

Issue #520, opened 2020-01-13
Android Termux crash running 'hf mf autopwn'


h4waii:
**Describe the bug**
Running `hf mf autopwn` on an Android device with Termux causes a crash due to:

> fortify pthread_mutex_lock called on a destroyed mutex

I collected a coredump through gdb:
```
Core was generated by `/client/proxmark3'.
Program terminated with signal SIGABRT, Aborted.
#0  0x0000007fb7ae1f38 in abort ()
   from /system/lib64/libc.so
[Current thread is 1 (LWP 23433)]
(gdb) bt
#0  0x0000007fb7ae1f38 in abort ()
   from /system/lib64/libc.so
#1  0x0000007fb7b442f4 in __fortify_fatal(char const*, ...) () from /system/lib64/libc.so
#2  0x0000007fb7b438cc in pthread_mutex_lock ()
   from /system/lib64/libc.so
#3  0x00000055555fb7c4 in free_statelist_cache ()
    at cmdhfmfhard.c:1624
#4  0x00000055555f8ef4 in mfnestedhard (
    blockNo=<optimized out>, keyType=<optimized out>,
    key=<optimized out>, trgBlockNo=<optimized out>,
    trgKeyType=<optimized out>, trgkey=<optimized out>,
    nonce_file_read=<optimized out>,
    nonce_file_write=false, slow=<optimized out>,
    tests=0, foundkey=0x7fffffe620, filename=0x0)
    at cmdhfmfhard.c:2382
#5  0x00000055555e4ee8 in CmdHF14AMfAutoPWN (
    Cmd=<optimized out>) at cmdhfmf.c:2201
#6  0x00000055556372f4 in CmdsParse (
    Commands=0x5555723720 <CommandTable>,
    Cmd=0x7fb722c996 "autopwn") at cmdparser.c:222
#7  0x00000055556372f4 in CmdsParse (
    Commands=0x55557229d8 <CommandTable>,
    Cmd=0x7fb722c993 "mf autopwn") at cmdparser.c:222
#8  0x00000055556372f4 in CmdsParse (
    Commands=0x5555726de0 <CommandTable>,
    Cmd=0x7fb722c990 "hf mf autopwn") at cmdparser.c:222
#9  0x000000555557e17c in main_loop (
    script_cmds_file=<optimized out>, script_cmd=0x0,
    stayInCommandLoop=<optimized out>) at proxmark3.c:265
[...]

PR #573, opened 2020-02-23
fix: #520 - free mutex after work is done

h4waii:
Why not just free after key found or not found?

### DIFF
--- a/client/cmdhfmfhard.c
+++ b/client/cmdhfmfhard.c
@@ -1845,14 +1845,15 @@ static bool TestIfKeyExists(uint64_t key) {
             num_keys_tested += count;
             hardnested_print_progress(num_acquired_nonces, "(Test: Key found)", 0.0, 0);
             crypto1_destroy(pcs);
+            pthread_mutex_destroy(&statelist_cache_mutex);
             return true;
         }
     }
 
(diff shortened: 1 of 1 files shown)


iceman1001:
dunno,  I guess it might be calling generate candidates several times?

h4waii:
Would it make sense to free them at https://github.com/RfidResearchGroup/proxmark3/blob/cfcb0494590d8dbe611e9e1eaf60f52495fd69de/client/cmdhfmfhard.c#L2258 after a key isn't found but the work continues?

Adding the destroy at these sections did not cause any appreciable slowdown on aarch64 or x86_64 for me, but I did **not** measure possible contention.

iceman1001:
destroying / creating a mutex isn't time consuming.

h4waii:
I added a free inside both if statements at line 2258 and 2385, as well as after work is complete with key or no key found.

I do not know if this is correct, but it makes sense in my head, compiles, and runs successful on both x86_64 and aarch64 /arm8 / Android now.

iceman1001:
sounds good to me.  Push your latest patches in this PR and I merge it.

iceman1001:
bugger,  this breaks the hardnested when running second time.

PR #575, opened 2020-02-25
hitagS fix sim + write

bosb:
I'm trying to get hitag S simulation to work.
Only partial success :-( and now stuck.

UID is D4 8E 09 8E ;   D4: 11010100 ; advanced mode 3 bits SOF
triggering for now with another proxmark by: lf cmdread d 50 z 116 o 166 c 011001
(but also via: lf hitag read 01)
I have a real tag with this id also to compare.
Any hints where I can influence 1+2 are appreciated...

Tag simulation sending UID:
hitags-sim-0

1. In the beginning the first half period is too long, and one period is missing
I can add the period in hitag_send_frame as a workaround by inserting at the beginning:

```
          AT91C_BASE_TC0->TC_CCR = AT91C_TC_SWTRG;
                // AC coding -_
                HIGH(GPIO_SSC_DOUT);
                while (AT91C_BASE_TC0->TC_CV < T0 * 16) {};
                
               LOW(GPIO_SSC_DOUT);
                while (AT91C_BASE_TC0->TC_CV < T0 * 32) {};
```
but it does not solve the length of the first period, for now it might not be that much of a problem, just looks not right.

2. I have no clue where that one comes from and I suspect it a problem, since the reader sends out a select, after it recieved the UID right, which the tag does not receive or recognize...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -420,6 +420,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Added `hf mf mad` and `hf mfp mad` MAD decode, check and print commands (@merlokk)
  - Added `script run luxeodump` (@0xdrrb)
  - Fix `lf hitag reader 02` - print all bytes (@bosb)
+ - Fix hitag S simulation (still not working), write, add example HITAG S 256  (@bosb)
 
 
 ### Fixed
--- a/armsrc/hitagS.c
+++ b/armsrc/hitagS.c
@@ -940,7 +940,7 @@ void SimulateHitagSTag(bool tag_mem_supplied, uint8_t *data) {
         memcpy((uint8_t *)tag.pages, data, 4 * 64);
     }
 
-    tag.uid = (uint32_t)tag.pages[0];
+    tag.uid = (tag.pages[0][3] << 24 | tag.pages[0][2] << 16 | tag.pages[0][1] << 8 | tag.pages[0][0]);
     tag.key = (intptr_t)tag.pages[3];
     tag.key <<= 16;
     tag.key += ((tag.pages[2][0]) << 8) + tag.pages[2][1];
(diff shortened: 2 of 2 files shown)


bosb:
A little better, but simulation does not want to enter select state (3)
(1) should be ok now, reader gets the uid right.

iceman1001:
You tell me when you feel its done.

bosb:
Sure, PR is in 'draft' mode, will then switch to 'ready for review' ;-)
I just expected some hints what causes (2)...
At least I'm not stuck anymore, found some workarounds to get it flowing...

iceman1001:
you would need to provide more details on how you are tackling and what the problem is when it comes to your nb 2

bosb:
Main problem: after sim tag sends, it does not get the answer/next command from reader on first try.

Setup:
**A**: proxmark rdv4 as reader with hitagS.c file from https://gist.github.com/bosb/ce82dc6569ee16fccdfe4dbd8d8dac93#file-hitags-c
modified to send repeated commands until tag answers
(if it matters, code base from proxmark for android app version, 08/2019, since I need the working sniffing hitag function)
```lf hit re 01```

**B**: proxmark rdv4 as tag simulation with this branch
load tag from gist https://gist.github.com/bosb/ce82dc6569ee16fccdfe4dbd8d8dac93#file-hitag_s_5fc21184-bin
```lf hit sim s b hitag_S_5FC21184.bin```

A: ```lf hit re 01```
A: give me uid 
B: my uid is 1234

A: I select you 1234
B: 1. no evidence that tag recognizes it
A: again select 1234
B: 2. ok here is my config

A: ok, give me page 0
B: ....
A: again
B: here ist page 0

.....
and so on....


It might be connected to this, but I'm not sure...

to reproduce (2)
A: ```data plot```
```lf cmdread d 50 z 116 o 166 c 011001```

B: sim file
```lf hit sim s b hitag_S_5FC21184.bin```

iceman1001:
The usual problem with your setup,  is that using two proxmark3 to do assumes that the proxmark3 code for the reader is perfect and according to specs.  
I have had that kind of idea many times and it all falls back on bad timings.  As your problem is.
The better setup is to use a valid reader to simulate and sniff against.   That is what I am aiming for in the hitag2 refactoring.  Haven't gotten the reader yet :(

So the problem in your setup is that you have most likely two problematic timings.  Trying to make it work in you setup doesn't guarantee us it will work against proper readers.

iceman1001:
You feel its ready to be merged?

bosb:
Let me clean it up a little and some last changes...maybe until end of week...

iceman1001:
don't forget to add a line in CHANGELOG.md
Then I will merge

iceman1001:
You think its ready?

bosb:
sure, go for it.

iceman1001:
Excellent!

PR #579, opened 2020-02-27
Add direction option to lf hid brute

capnkrunchy:
Adds the option in lf hid brute to only increment card number up, only increment down, or (by default) increment in both directions.


capnkrunchy:
Missed a couple formatting issues

PR #580, opened 2020-02-27
Add direction option to lf hid brute

capnkrunchy:
Added a command line parameter (x) to change the direction of card number incrementation for brute force attack. x 0 increments up and down, x 1 increments up only, and x 2 increments down only.

### DIFF
--- a/client/cmdlfhid.c
+++ b/client/cmdlfhid.c
@@ -448,12 +448,12 @@ static int CmdHIDBrute(const char *Cmd) {
 
         // Do one up
         if (datahi.CardNumber < 0xFFFF) {
-                if(direction == 0 | direction == 1){
+                if(direction == 0 || direction == 1){
                         datahi.CardNumber++;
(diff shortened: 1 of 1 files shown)


iceman1001:
...hm...  param 'x' is usually to denote "eXecute attack".
I suggest  "up" / "down"   instead.

capnkrunchy:
I've been having trouble getting the command parsing to take "up" and "down" without first giving a parameter.

I'm not sure I understand the command parsing and the switch case statement within the brute force section well enough to make that change.

I'm open to suggestions.

iceman1001:
See https://github.com/RfidResearchGroup/proxmark3/commit/c30bb354ac6d38a69a8e8d545e42c9020a1a0131

 lf hid brute f 21 c 5 i 0 o 0 d 2000 v down w H10301

You can use up/down and verbose to get more details.  The bruteforce now also exits when done.

I will close this PR.

capnkrunchy:
Awesome, thanks iceman!

PR #594, opened 2020-03-07
Add Mifare Ultralight commands to cheatsheet.md

dunderhay:


iceman1001:
Nice,   you could explain the dump with pwd, since it can also be a key etc..

iceman1001:
And don't forget a line in CHANGELOG.md

PR #595, opened 2020-03-07
EM4100 read/clone/simulate standalone mode

temskiy:
V1 - can only simulate predefined IDs
V2 - read (for RDV4 can store to flash) and simulate prdefined or readed IDs
V3 - see V2 + clone to T5555 tag


temskiy:
Miss some files. Sory

PR #597, opened 2020-03-10
Add Mifare Ultralight bruteforce support to hf_bruteforce lua script

dunderhay:


client/luascripts/hf_bruteforce.lua
```diff
@@ -60,46 +66,51 @@ local function help()
...
-    local end_id = 0xFFFFFFFF
+    local end_id = 0xFFFFFFFFFFFFFF
+    local mftype = ''
 
-    for o, a in getopt.getopt(args, 'e:s:t:h') do
+    for o, a in getopt.getopt(args, 'e:s:t:h:T:') do
         if o == 's' then start_id = a end
         if o == 'e' then end_id = a end
         if o == 't' then timeout = a end
+        if o == 'T' then mftype = a end
```
iceman1001: This idea of  two params that is the same but differ in capital or not,  is not recommend.
-t  vs -T  is confusing.
We need to come up with a better param name to make it clear and easy to understand.

I also suggest having **mfc** as default type,   making the new param optional from original expected behavior.

client/luascripts/hf_bruteforce.lua
```diff
@@ -1,27 +1,33 @@
--- Run me like this: proxmark3 /dev/rfcomm0 -l ./hf_bruteforce.lua
+-- Run me like this: ./client/proxmark3 /dev/ttyACM0 -l hf_bruteforce.lua
```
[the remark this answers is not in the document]
auradan: Ah okay I did try that but recall the exact error I got. Perhaps I had not run the make all command first tho and client was not there. Happy to update that if you can confirm will work tho :)

client/luascripts/hf_bruteforce.lua
auradan: Cool will update the script tonight and do a new pull request 👍

dunderhay: I changed the option to `x` for Mifare type and set the default to __mfc__.  Have also added a new line for running when attached via USB and kept original command to run via BlueShark addon.

dunderhay: Actually just noted that I didn't `git add` my final change to add the old BlueShark addon command back. Happy to redo that. Sorry for hassle

dunderhay:
I might stop making commits now and let you cherry pick any valid stuff out of my mess! haha

iceman1001:
Excellent!

Not to be that guy, but you seem to have added a GPL v3 license,  to origial code which wasn't yours.  Have you confirmed this with the orignal copyright owner?

dunderhay:
Sorry I noticed a bug as well.. `h` option should not require additional parameter. I have not checked that with the original owner because there was no license at all and I just want to make sure that anyone can use this for free. I can remove that too

iceman1001:
You didn't have to close the PR.  :)   I was about to merge it.

Short note on license used:
The lack of license doesn't mean it doesn't have a license.  All code published in this repo without direct license text inside the source code,  falls back onto the 
https://github.com/RfidResearchGroup/proxmark3/blob/master/LICENSE.txt

Since this is the prefered way for sharing the license used for this repo. 

ie: all code gets a GPL2.0 license by default here. If its not exclusive stated their own license compliance with GPL2.

PR #606, opened 2020-03-15
Add colour to command output

dunderhay:
Added colour to several commands:
`lf hid`
`hf 14a`
`hf mfu`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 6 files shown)


iceman1001:
It s gonna be fun looking at these new colors.  Awesome!

doegox:
yeah fun, but wait for phase 2 @iceman1001 :wink:

iceman1001:
What is Phase2 ??

iceman1001:
Hm,   I like seperatation of information. 
Like now UID etc became green  (was yellow before).  and the type is green.
I think  making  the type Yellow will make it easy to see the difference.  ie too much green.

Next thing.  The static nonce line should be:   Static nonce:  red(yes)  | white(no)

image

doegox:
> What is Phase2 ??

image

iceman1001:
Evil.

We most fight this evil.

PR #615, opened 2020-03-21
Update cmdlfkeri.c

mwalker33:
Updated ID/FC Mapping table.
Removed the bit location mapping as we now know the bit locations.


iceman1001:
Great!

PR #620, opened 2020-03-22
Keri ms clone support with fc and cardid

mwalker33:
As well as the original 
lf keri clone <internal cardid> 
it now supports
lf keri clone type ms fc 7 cardid 35423

Note: each command line option is based on the first letter (e.g. t for type, m for ms, f for fc etc)
so supports both readable and short version.

I have left the original command line options in tact, as such left the original help line as well.

let me know if I have missed something.


iceman1001:
nay its looking good.

You tell me when you feel its ready to be merged

mwalker33:
Think i have them all now :)

iceman1001:
Excellent bro!

PR #623, opened 2020-03-22
Update hf-writer.lua

Windslab:
After this fixes script works to Windows as well. I hope it doesn't will work at Linux at all... Could you please optimised code for both?


iceman1001:
You seem to be focused on one OS only,   when you make a PR you will have to test changes on many OS's to ensure it works on them all.
  
I suggest you test this one on  Ubunutu 19,  Kali 9,  OSX,  WSL2,  aswell and not just Proxspace / mingw.

PR #624, opened 2020-03-24
Updates to the Tesla AIDs

darconeous:
Now includes the AIDs that are used in the key fob and phone.
Removed some AIDs that can't be selected.

### DIFF
--- a/client/resources/aidlist.json
+++ b/client/resources/aidlist.json
@@ -2128,6 +2128,14 @@
         "Type": "EMV"
     },
     {
+        "AID": "7465736C6153746F7265",
+        "Vendor": "Tesla",
+        "Country": "",
+        "Name": "teslaStore",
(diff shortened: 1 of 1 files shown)


iceman1001:
Don't think you should remove AID's that isn't selectable because you don't have accessrights.

darconeous:
I had just assumed it was a non-selectable service, not that it was unselectable because of access rights. I wasn't even aware that you could configure GP in such a way that you could limit selection based on access rights.

But upon further inspection, I indeed do see a different response code when attempting to select `teslaStore`:

```
A>> T=1 (4+0010) 00A40400 0A 7465736C6153746F7265
A<< (0000+2) (27ms) 6999
```

Error `6999` (versus the normal `6A82`). So I think you are absolutely right, it is selectable but not without doing something else first. I'll put it back in.

PR #625, opened 2020-03-27
Standalone mode: Adding Legic Prime read/sim run

uhei:
This commit adds a standalone mode for Legic Prime.
It reads and simulates a Legic tag.
Tested with MIM1024 tags.
MIM256 and MIM512 should work to. We just read the data and for simulating
we pretend to be a MIM1024 card.

Co-authored-by: Stefanie Hofmann <>


iceman1001:
Cool!

You could add support to "save" all read cards to flash mem as a file.  Later on you can just download it from device.  Like save how it was when you first read the card,  then save it after simulation if reader updated something.  :)  two files.  before and after.

And don't forget to add a line in CHANGELOG.md  file.

uhei:
> You could add support to "save" all read cards to flash mem as a file. Later on you can just download it from device. Like save how it was when you first read the card, then save it after simulation if reader updated something. :) two files. before and after.

Great idea! Will it require platform specific code (like ```#if defined RDV4```)? Any good file recommendation for inspiration?

> And use **"make style"** in order to get the unified codeing style applied. 
> And don't forget to add a line in CHANGELOG.md file.
> And global variables is not recommend but atleast try using "g_variable" name style. So its easy to identify as a global var.

Thanks. I will push a new version of this commit.

iceman1001:
https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/Standalone/lf_em4100rwc.c#L105

iceman1001:
I hope you didn't get stuck.

uhei:
> I hope you didn't get stuck.

Partly. I'm unsure how to get the memory data as it seems there is no Legic specific function. Should/Can I use emlGetMemBt() to get the data?

I've fixed the other points

iceman1001:
I can merge this one,  and then we can have a look at adding flash file support

iceman1001:
Easier than you think.
legic reader and legic simualte works with  
```
    legic_mem = BigBuf_get_EM_addr();
```

You just need to same 1024 bytes from that address :)

iceman1001:
hold on,  I am rewrite it for you.

PR #629, opened 2020-03-30
Makefile: target checks for macOS

uhei:
recode is missing on macOS by default.
grep on macOS is BSD grep by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,13 +210,20 @@ style:
 
 # Detecting weird codepages and tabs.
 checks:
+	# Make sure recode is installed
+	@which recode >/dev/null || ( echo "Please install 'recode' package first" ; exit 1 )
 	@echo "Files with suspicious chars:"
 	@find . \( -name "*.[ch]" -or -name "*.cpp" -or -name "*.lua" -or -name "*.py" -or -name "*.pl" -or -name "Makefile" -or -name "*.v" \) \
 	      -exec sh -c "cat {} |recode utf8.. >/dev/null || echo {}" \;
 	@echo "Files with tabs:"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -210,13 +210,20 @@ style:
...
 	@find . \( -name "*.[ch]" -or -name "*.cpp" -or -name "*.lua" -or -name "*.py" -or -name "*.pl" -or -name "Makefile" -or -name "*.v" \) \
 	      -exec sh -c "cat {} |recode utf8.. >/dev/null || echo {}" \;
 	@echo "Files with tabs:"
 # to remove tabs within lines, one can try with: vi $file -c ':set tabstop=4' -c ':set et|retab' -c ':wq'
+ifeq ($(platform),Darwin)
+	@find . \( -name "*.[ch]" -or \( -name "*.cpp" -and -not -name "*.moc.cpp" \) -or -name "*.lua" -or -name "*.py" -or -name "*.pl" -or -name "*.md" -or -name "*.txt" -or -name "*.awk" -or -name "*.v" \) \
+	      -exec egrep -l '\t' {} \;
+else
 	@find . \( -name "*.[ch]" -or \( -name "*.cpp" -and -not -name "*.moc.cpp" \) -or -name "*.lua" -or -name "*.py" -or -name "*.pl" -or -name "*.md" -or -name "*.txt" -or -name "*.awk" -or -name "*.v" \) \
 	      -exec grep -lP '\t' {} \;
```
iceman1001: why not put the define around the ```-exec ...```line instead?

uhei: It's not possible as it is a one line command stretched over two lines by using the backslash for better readability.

PR #634, opened 2020-04-02
Add colour to hf mf chk and fchk output

dunderhay:


client/cmdhfmf.c
```diff
@@ -3576,14 +3576,14 @@ void printKeyTableEx(uint8_t sectorscnt, sector_t *e_sector, uint8_t start_secto
             if (start_sector == 0)
                 s = i;
 
-            PrintAndLogEx(NORMAL, "|%03d|  %s  | " _YELLOW_("%d")"|  %s  | " _YELLOW_("%d")"|"
+            PrintAndLogEx(SUCCESS, "| "_YELLOW_("%03d")"| " _GREEN_("%s")"  | " _YELLOW_("%d")"| " _GREEN_("%s")"  | " _YELLOW_("%d")"|"
```
iceman1001, line 3579: hm,...
so we get yellow and green regardless if we failed to get the key.
If we gonna do colors,  then  if "0",  it should be red,   all others may be green.

iceman1001:
Cool!

Issue #637, opened 2020-04-03
command execution time out


cosmo6110:
lf search 
EM410x tag, get "**command execution time out**", 

[usb] pm3 --> lf search
[=] NOTE: some demods output possible binary
[=] if it finds something that looks like a tag
[=] False Positives ARE possible
[=]
[=] Checking for known tags...
[=]
[+] EM410x pattern found
//*************
********//
[+] Valid EM410x ID found!

[!] command execution time out

[usb] pm3 --> hw status
#db# Memory
#db#   BIGBUF_SIZE.............40000
#db#   Available memory........39944
#db# Tracing
#db#   tracing ................1
#db#   traceLen ...............0
#db# Currently loaded FPGA image
#db#   mode.................... HF image built for 2s30vq100 on 2020-01-12 at 15:31:16
#db# Flash memory
#db#   Baudrate................24 MHz
#db#   Init....................OK
#db#   Memory size.............2 mbits / 256 kb
#db#   Unique ID...............0xD5690C23DF9F262A
#db# Smart card module (ISO 7816)
#db#   version.................v3.10
#db# LF Sampling config
#db#   [q] divisor.............95 ( 125.00 kHz )
#db#   [b] bits per sample.....8
#db#   [d] decimation..........1
#db#   [a] averaging...........Yes
#db#   [t] trigger threshold...0
#db#   [s] samples to skip.....0
#db# LF T55XX config
#db#            [r]               [a]   [b]   [c]   [d]   [e]   [f]   [g]
#db#            mode            |start|write|write|write| read|write|write
#db#                            | gap | gap |  0  |  1  | gap |  2  |  3
#db# ---------------------------+-----+-----+-----+-----+-----+-----+------
#db# fixed bit length (default) |  29 |  17 |  15 |  47 |  15 | N/A | N/A |
#db#     long leading reference |  29 |  17 |  15 |  47 |  15 | N/A | N/A |
#db#               leading zero |  29 |  17 |  15 |  40 |  15 | N/A | N/A |
#db#    1 of 4 coding reference |  29 |  17 |  15 |  31 |  15 |  47 |  63 |
#db#
#db# Transfer Speed
#db#   Sending packets to client...
#db#   Time elapsed............500ms
#db#   Bytes transferred.......166400
#db#   Transfer Speed PM3 -> Client = 332800 bytes/s
#db# Various
#db#   DBGLEVEL................1
#db#   ToSendMax...............-1
[...]

PR #645, opened 2020-04-05
Em4x05 bug

mwalker33:
command execution time out #637

Update the timing to ensure correct math order and not timeout.


iceman1001:
yeah,   that was me introducing it,  but it was your solution I tried to add  :) ha,  now you double fixed it.  

Thanks!

DonAnonymousio:
Finding this bug on the newest firmware. Would i have to edit this file from source and flash from source to fix it?

PR #647, opened 2020-04-05
Legic_clone.lua: Refactor to work with current command set

uhei:
Refactored to work with current command set and MIM256 and MIM1024 tags


uhei:
Have only been able to test it with MIM256 as my other tags are in the Covid-19 locked down office.

iceman1001:
Great!

PR #649, opened 2020-04-05
Improve 14443-a info

bkerler:
These changes attempt to improve the hf 14a info card detection to give better possible card recommendations and is based on the NXP AN10833 Rev 3.6 MIFARE Type Identification. It was tested with more than 20 cards (desfire ev1/mifare classic/mifare plus/ultralight c/jcop 4). It does now also contain a uid table for some 14443-a cards to improve detection.

### DIFF
--- a/client/cmdhf14a.c
+++ b/client/cmdhf14a.c
@@ -1245,6 +1245,122 @@ int CmdHF14A(const char *Cmd) {
     return CmdsParse(CommandTable, Cmd);
 }
 
+static void printTag(char *tag) {
+    PrintAndLogEx(SUCCESS, _YELLOW_("    %s"), tag);
+}
+
+
+typedef enum {
+    mtNone = 0,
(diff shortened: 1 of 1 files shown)


iceman1001:
Cool concept.

- The manufacturer lookup vs your two byte uid lookup?  Do ISO14443a have SRI cards?  I thought those were only ISO15693 based?
- And run ```make style``` on your code. 
- And enum we are trying to make them captalized,   an ongoing change, but for all new ones enforced :)
- And add "_t" to the type,  nxp_mifare_type ->  nxp_mifare_type_t
- TNP3 isn't "an other tag",  its a Mifare Classic with a different sak.
- Detect magic based on sak,  I wouldn't do that.  Since some magic cards can change sak/atqa.  If you identified it as classic,  or mfu,  you can try run magic capabilities detection.

PR #660, opened 2020-04-08
Improve hf mfdes client, complete restructure. Introduce client debugging.

bkerler:
The hf mfdes client was rewritten from scratch. On setting "data setdbgmode 2", in case of a need for debugging apdu, each apdu command and response is being printed.

### DIFF
--- a/include/protocols.h
+++ b/include/protocols.h
@@ -349,45 +349,58 @@ ISO 7816-4 Basic interindustry commands. For command APDU's.
 // 6x xx = ERROR
 
 // MIFARE DESFire command set:
-#define MFDES_CREATE_APPLICATION        0xca
-#define MFDES_DELETE_APPLICATION        0xda
-#define MFDES_GET_APPLICATION_IDS       0x6a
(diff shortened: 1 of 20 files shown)


include/protocols.h
```diff
@@ -381,6 +381,10 @@ ISO 7816-4 Basic interindustry commands. For command APDU's.
 #define MFDES_ABORT_TRANSACTION         0xA7
 #define MFDES_AUTHENTICATION_FRAME      0xAF
 #define MFDES_ADDITIONAL_FRAME          0xAF
+#define MFDES_ADDITIONAL_FRAME_RESP     0x91AF
+#define MFDES_SUCCESS_FRAME_RESP        0x9100
+#define MFDES_EAUTH_RESP                0x91AE
+#define MFDES_ENO_SUCH_KEY_RESP         0x9140
```
iceman1001, line 387: I dunno why you add the 0x91 part

bkerler: That's just temporary for better readability. It will be removed on the next commit.

PR #661, opened 2020-04-09
legic.lua: updated to work again

uhei:
Fixed/changed things:

* added info that virtual tag is always MIM1024
* changed reading files to read binary files (as written by 'hf legic dump')
* changed extension for writing files (*.bin and *.eml) to be on par with 'hf legic dump'
* CRC was calculated wrong when data was not padded with 0 for one char hex strings
* readTag (rt) can now be called multiple times without using the wrong filename
* tag length was calculated wrong as segment header length field includes the header itself
* bytes are XORed before writing them to a tag
* default name for file contains now the tag id


iceman1001:
Excellent!

Issue #659, opened 2020-04-08
Braces Error When Compiling cmdhfmfdes.c on Mac OSX Mojave/Catalina


capnkrunchy:
**Describe the bug**
When compiling any commit newer than 7594304 on Mac OSX, `make all` throws the following error(s):
`[-] CC cmdhfmfdes.c`
`cmdhfmfdes.c:713:30: error: suggest braces around initialization of subobject`
`      [-Werror,-Wmissing-braces]`
`    dfname_t dfnames[255] = {0};`
......
`cmdhfmfdes.c:886:30: error: suggest braces around initialization of subobject`
 `     [-Werror,-Wmissing-braces]`
 `   dfname_t dfnames[255] = {0};`

**To Reproduce**
1. Pull latest version or any version after 7594304
2. `make clean && make all`

**Expected behavior**
Compile without errors

**Screenshots**
![image](https://user-images.githubusercontent.com/47148444/78815003-38e76c00-7995-11ea-9ef3-310776dab285.png)

**Desktop (please complete the following information):**
 - OS: Mac OSX 10.14.6 and Mac OSX 10.15.4
 - Output of clang --version
        Mojave: `Apple LLVM version 10.0.1 (clang-1001.0.46.4)`
`Target: x86_64-apple-darwin18.7.0`
`Thread model: posix`
`InstalledDir: /Library/Developer/CommandLineTools/usr/bin`
        Catalina: `Apple clang version 11.0.0 (clang-1100.0.33.17)`
`Target: x86_64-apple-darwin19.3.0`
`Thread model: posix`
`InstalledDir: /Library/Developer/CommandLineTools/usr/bin`

**Additional context**

toucan12:
I have been experiencing the same symptoms since yesterday.
I don't know if it is the right way, but if I make the following modifications in the Makefile, it compiles normally.

In line 54 of client/Makefile,

**Before**
`CFLAGS ?= -Wall -Werror -g -O3`

**After**
`CFLAGS ?= -Wall -Werror -Wno-missing-braces -g -O3`

iceman1001:
.... or just  add a brace ...  ```{{0}}```  instead of ignoring it...

toucan12:
I've already tried it, but I get an error.

![2020-04-09_15-54-26](https://user-images.githubusercontent.com/63378223/78866594-a8835700-7a7a-11ea-9344-dd1b201a9e0d.jpg)

any ideas?

PR #662, opened 2020-04-09
cmdhfmfdes: fix clang warnings

uhei:
Make clang happy and fix several
error: expression which evaluates to zero treated as a null pointer constant of type 'uint8_t *' (aka 'unsigned char *')
      [-Werror,-Wnon-literal-null-conversion]
and
cmdhfmfdes.c:732:31: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]

Fixes RfidResearchGroup/proxmark3#659


iceman1001:
Nice,  you should have pulled before,  most of them was already fixed ;)

uhei:
> Nice, you should have pulled before, most of them was already fixed ;)

Haha. Look at the timestamps. We've committed more or less at the same time.

PR #664, opened 2020-04-09
HF MFDES updates. New aid create, aid delete and format card command. Auth fixed. Improved logging.

bkerler:
HF MFDES updates. New aid create, aid delete and format card command. Improved logging.
Auth DES, AES and ISO fixed and working.


iceman1001:
Superb!

PR #665, opened 2020-04-10
hfmfdes: make clang happy and fix warning

uhei:
fix warning error: expression which evaluates to zero treated as a null pointer constant of type 'uint8_t *' (aka 'unsigned char *')
      [-Werror,-Wnon-literal-null-conversion]


iceman1001:
Good catch!
I was already fiddling with stuff, so I think I got your fixes pushed already.  

Closing.

PR #668, opened 2020-04-10
Settings Update

mwalker33:
Added define to enable/disable settings file use (while we develop), disabled by default.
Added 
- plot window x,y,h,w options (overlay to follow)
- client debug level (off|simple|full)
- show hints
- show emoji
- draft os.supports.colors

If the settings files is not found then it will be auto created with defaults/discovered from main().
e.g. if main enables color or emoji then those will be true in the created settings.json file else defaults used.


iceman1001:
exciting times!

PR #680, opened 2020-04-13
Mifare Desfire functionality update (Read/Write/Clear files) and fixes

bkerler:
So now we can read, create, change, clear and write files in plaintext mode for
mifare desfire cards. CMAC + Enciphered Mode has to be implemented, as well as detection
for those modes (and error handling of course). Additional fixes for 'hf mfdes enum" as file sizes and infos were displayed incorrectly.


bkerler:
Added key display on hf mfdes enum as well

iceman1001:
Excellent!

PR #689, opened 2020-04-15
Update aid_desfire.json

netvader:
added aids from
https://www.telenot.com/ alarm system & https://microtronic.ch payment system


iceman1001:
Thanks!

PR #699, opened 2020-04-17
Add hf mf des chk and dump commands

bkerler:
Added generic keys for mf desfire, added chk command for checking keys and added dump command to print all file contents (if unauthenticated). Saving dumps to a file isn't support yet for dump command.


iceman1001:
Nice!

Looks like Travis reports its breaking on OSX,

image

uhei:
Compiles on macOS after:
```
diff --git a/client/src/cmdhfmfdes.c b/client/src/cmdhfmfdes.c
index 159e319f..814c9ebd 100644
--- a/client/src/cmdhfmfdes.c
+++ b/client/src/cmdhfmfdes.c
@@ -639,7 +639,7 @@ static nxp_cardtype_t getCardType(uint8_t major, uint8_t minor) {
     if (major == 0x11 &&  minor == 0x00)
         return PLUS_EV1;
 
-    return UNKNOWN;
+    return DESFIRE_UNKNOWN;
 }
 
 int handler_desfire_auth(mfdes_authinput_t *payload, mfdes_auth_res_t *rpayload, bool defaultkey) 
```

iceman1001:
now,  what is the purpose of the new key dump file in json?

bkerler:
Yeah, fixed that DESFIRE_UNKNOWN one. What new key dump file ?

bkerler:
If you meant the new json structure in fileutils, it's for saving the found keys using the chk cmd.

iceman1001:
yes,  but what is its purpose?  What is it you are going to use it for?  Tell me a user-scenario.
If you comparing with how  **mifare classic command (hf mf)** uses a keyfile in order to have keys for a key recovery attack,  and knowing if we have all to make a dump,    its not like that with a DESFire card.   What purpose will a keyfile have there?
If you look at the json dump file for MFC,  you see its reason to exist.  Its a dump file with keys.

A json dump file with file/data/keys,  would be very useful for DESFire.

PR #701, opened 2020-04-18
Otp tear update

fukmar:
OTP tear functionality with parameters.


doegox:
please redo your PR without binary files!

PR #721, opened 2020-04-28
Dfnames fix

bkerler:
This fixes issues with first gen Desfire cards that do not support iso df name support.


doegox:
It has been fixed by @iceman1001  differently in https://github.com/RfidResearchGroup/proxmark3/commit/d2e85b00405b4b0adde6a2f36ff03ba0734d5794#diff-0b3f8794029f47b6c1538963fd86d6c8L3316
Indeed it's better that the function still return an error, but the caller doesn't abort

PR #727, opened 2020-05-09
Rewrited LF_EM4100RWC standalone mode + card bruteforce added.

monster1025:
I have created separate mode - LF_EM4100RSWB(and not changed original one), because in this one is harder to switch between slots, but easier to work with one slot (more automatic switches between modes), also I reduced 16 slots to 4.

I think this 'menu/modes' layout is more suitable when you need fast read somebody's card, go to the door and sim it/write to empty card/try to brute nearest cards.

In this standalone mode I have added card bruteforce mode.


doegox:
Hi
Could you
* rebase your branch on current master
* change your authorship info in commits, right now it shows "Your Name"

thanks.

monster1025:
Sorry, my git skills are no so great (I don't figure out how to change username for many previous commits), so I can close this PR, and make a new one =))) Hope its ok.

Issue #730, opened 2020-05-12
Unable to compile client after 94133a8


swg0101:
It appears that as of the 94133a8 Merge branch 'strict' commit the compilation of the client has been failing when compiling for PM3OTHER.

See screenshot as attached.
![Screenshot_20200512-153758](https://user-images.githubusercontent.com/14957466/81752801-20bcae00-9467-11ea-8358-83f6cc5142de.jpg)

doegox:
which env is it ? termux ?

doegox:
does it compile if you do `make client NOERROR=1` ? (on latest commit)

swg0101:
Ah yes, NOERROR=1 works. This is on Termux Clang.

doegox:
Hmm when trying on the termux-packages cross-compilation docker image, it works fine.
But the compiler is older: Android (5220042 based on r346389c) clang version 8.0.7
I need to dig further

doegox:
ok I could reproduce, I pushed a fix to ignore these ones, still I find it weird, should be properly addressed

doegox:
btw could you share how you compile pm3 on termux ? I could compile the client but didn't find arm-none-eabi-gcc

swg0101:
Two-step process - compile pm3 client in Clang, compile boot image in Ubuntu:
~/start-ubuntu.sh "cd ~/proxmark3 && git checkout master && git pull && make bootrom fullimage PLATFORM=PM3OTHER -j4" && make -C ~/proxmark3 client PLATFORM=PM3OTHER -j4 NOERROR=1

doegox:
Thanks!
I think this strict warning revealed an issue in the headers which by chance defaulted into little endian instead of alerting about unknown endianess.
I pushed another fix in a branch https://github.com/RfidResearchGroup/proxmark3/commit/2c4eda6ab9b83e3d5853408c354d344d0255e0f0 
Let's see if it works properly on CI. If yes I'll merge.

swg0101:
Yes, it no longer errors out when compiling after make clean and without NOERROR=1, only the following appears as warnings:

`
[-] CC src/cmdsmartcard.c
src/cmdtrace.c:79:39: warning: cast from 'uint8_t *' (aka 'unsigned char *') to 'uint16_t *' (aka 'unsigned short *') increases required alignment from 1 to 2 [-Wcast-align]
[...]

PR #732, opened 2020-05-13
include machine/endian.h for os x

jamiefiedler:
referencing issue #730 / PR #731 
platform check is needed if bypassing <sys/types.h> on os x.
include <machine/endian.h> instead of <endian.h> to avoid the error below.

```
In file included from src/flash.c:20:
./src/proxendian.h:20:11: fatal error: 'endian.h' file not found
# include <endian.h>
          ^~~~~~~~~~
1 error generated.
```


doegox:
thanks !
I thought OSX CI succeeded before PR #731, apparently I did sth wrong.

PR #733, opened 2020-05-15
LF_EM4100RSWB - How I can add HF Mode?

monster1025:
I have idea to make standalone mode that will allow me to bypass simple locks in standalone mode.
I have already done with LF tags(EMs are most used in my place), but for now I also have some UID-based HF locks (mine is Xiaomi Aqara lock).

I am not pretty sure how to make dual-frequency standalone mode. Do I need to add HF to name or it is ok that "LF_xxx" mode can operate on HF or maybe we can add "DF_xxx" (dual frequency prefix)?
I need some help with naming - because (I think) it didn't fit current naming convention =)

### DIFF
--- a/armsrc/Standalone/lf_em4100rswb.c
+++ b/armsrc/Standalone/lf_em4100rswb.c
@@ -48,13 +48,19 @@
 #include "spiffs.h"
 #include "inttypes.h"
 #include "parity.h"
+//HF
+#include "iso14443a.h"
+#include "commonutil.h" // bytes_to_num etc
 
 #ifdef WITH_FLASH
 #include "flashmem.h"
 #endif
 
+#define TAG_TYPE_LF 0
+#define TAG_TYPE_HF 1
+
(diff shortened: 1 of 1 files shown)


doegox:
Hmm interesting concept to get both LF & HF in same standalone...
Yes maybe DF_ or 2F_ which speaks more by itself.
Obviously the name itself should be changed too as you're not limiting it to EM4100

doegox:
@iceman1001 reminded me that this is not necessarily a good idea to have a mode constantly moving between HF and LF because that implies flashing the FPGA every time and it could wear the eeprom sooner than expected

iceman1001:
not to mention standalone mode usually indicate the device will be used with a battery.  Thus when power is low or maybe dies,  that the flashing of the fpga will leave your device in a unknown state.

monster1025:
ok, now that it is bad idea. So, lets close this PR =)

PR #735, opened 2020-05-18
cmdtrace: fix clang warning cast alignment

uhei:


doegox:
thanks but we're aware of it and we need to address it properly, not just hiding it away. Soon hopefully

PR #738, opened 2020-05-19
Somes traces

FlUxIuS:
Uploading few traces for ISO/IEC 14443 Type A and B, and MIFARE DESFire.


iceman1001:
Excellent!   Thanks!

PR #740, opened 2020-05-20
Work directory supported.

xianglin1998:


xianglin1998:
I found that when I used some functions, the working directory could not be identified correctly, because I did not initialize the working directory, so I added the initialization of the working directory.

iceman1001:
Nice,
Never forget to initialize variables :)

In the process of following our naming convention,  I suggest you rename your global variables to be 
1. related to android name space
2. add the suffix  `g_`  to indicate variable is global.

ie:  
`my_***`
should be named,
`g_android_***`

xianglin1998:
OK, this is really a place to pay attention to. I will change this name next PR though pm3_main.c does not conflict with proxmark3.c compilation.

PR #749, opened 2020-05-29
Update mfc_default_keys.dic

hamperfait:
Added the keys of an old London Oyster Card for London Public Transportation.
[+] |-----|----------------|---|----------------|---|
[+] | Sec | key A          |res| key B          |res|
[+] |-----|----------------|---|----------------|---|
[+] | 000 | c298de483a13   | 1 | dd1867d87aac   | 1 |
[+] | 001 | dbf57f3e6a9e   | 1 | cc6aada1f15f   | 1 |
[+] | 002 | 6f4c3703348b   | 1 | cce8109d2b01   | 1 |
[+] | 003 | b8901fe2dd4b   | 1 | a8005cccf6c7   | 1 |
[+] | 004 | 05737afac0ee   | 1 | a658f4b77cbd   | 1 |
[+] | 005 | 09036dc72217   | 1 | 4a1f917f6d7a   | 1 |
[+] | 006 | 5f11583b4e0d   | 1 | e3fe86265d0c   | 1 |
[+] | 007 | 23c78fc0d9b1   | 1 | 995ce87e9330   | 1 |
[+] | 008 | 9de3922fe80e   | 1 | de01d9cf7a40   | 1 |
[+] | 009 | b2aa8f3c7782   | 1 | fb9f860b6aa9   | 1 |
[+] | 010 | c23730318074   | 1 | 4a3eb564e37d   | 1 |
[+] | 011 | 5b6853d7b26a   | 1 | 58132a59a16d   | 1 |
[+] | 012 | 76c4db09cf5c   | 1 | b499da6650f6   | 1 |
[+] | 013 | e60a3c9ccffd   | 1 | 111ce7d33be5   | 1 |
[+] | 014 | 69028bd32bbe   | 1 | 29464ba6d644   | 1 |
[+] | 015 | 28d7212ef1b5   | 1 | 56f53dfbda9f   | 1 |
[+] |-----|----------------|---|----------------|---|


doegox:
Oyster keys used to be diversified, not sure it's worth adding them...

hamperfait:
Oh? Didn't know that..
Sadly I don't have another card to check whether they're useful or not, and I think a few years ago they changed them and now use another protocol, not mifare anymore, so it might not be useful

doegox:
yeah that was the big diff with NL and its OV-chipkaart with one single key for the whole country, and why one country overcame more easily the Mifare crisis than the other one ;)

iceman1001:
We still need to collect this keys in order to maybe once figure out the old oyster card keygen algo.  :)

Here @doegox  and I differ,  I don't mind adding these keys to the dic.

iceman1001:
I would need your UID for your card

doegox:
Well yes we could keep track of a few Oyster cards for the algo (I've one too btw) but then better in another file, no ? Else putting a lot of diversified keys in the dict will just slow down the dict bruteforce.

iceman1001:
yes,  there is already diversifed keys inside the dic.   @ikarus23  has mentioned it and shares your approach.    I am more easy to it.   
Anyway,  yes, keeping track of the samples collected,  so I started a thread over at the forum where is should be.   If you see my previous post, I added a link there.

hamperfait:
The UID is E7 76 E8 0B

PR #750, opened 2020-05-31
emv/fix: Don't dump to executable directory

edgeyboy:
Currently EMV dumps are being written to the executable directory. If `make install` is used, this is a directory that is at best undesirable and at worst unwritable.

This also doesn't match the other dump functions, where proxmark3 uses a relative path based on current working directory.

This patch fixes that behavior and makes it behave like the rest of the dump functions.

### DIFF
--- a/client/src/emv/cmdemv.c
+++ b/client/src/emv/cmdemv.c
@@ -1420,7 +1420,7 @@ static int CmdEMVScan(const char *Cmd) {
     PrintChannel(channel);
     uint8_t psenum = (channel == ECC_CONTACT) ? 1 : 2;
     uint8_t relfname[250] = {0};
-    char *crelfname = (char *)relfname;
+    char *fname = (char *)relfname;
     int relfnamelen = 0;
     CLIGetStrWithReturn(12, relfname, &relfnamelen);
     CLIParserFree();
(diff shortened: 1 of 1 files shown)


iceman1001:
hm,  shouldn't save executing folder.   Should be the relative path,  and it does follow the fileutils.c save patterns.   It should use
https://github.com/RfidResearchGroup/proxmark3/blob/master/client/src/fileutils.c#L228
to always create a unique filename.  

So your patch doesn't really do the expected prefered solution.  
Do please update your PR accordingly and we should be able to merge pretty easy.

edgeyboy:
Considering I can't call that function directly (which seems to be the intended behaviour), should I instead create a new `JSONFileType` (`jftEmv`/`jftEmvScan`) and move the JSON dumping code for that jft from `cmdemv.c/CmdEMVScan` to `fileutils.c/saveFileJSONex`?

iceman1001:
That would be the best solution.

iceman1001:
@doegox  refactored filetutils json a bit,  it now can take a callback function.
ie,  you can call it with the fucntion you want to execute for a type.

https://github.com/RfidResearchGroup/proxmark3/blob/master/client/src/fileutils.c#L551

This should counter all specialized code in that function,  
Your solution should use it,   see cf preferences_load / preferences_save as examples.

edgeyboy:
I took a shot at this today, and to be completely honest, I'm not sure how to properly account for the arguments. (I now understand why this specific file save was done in the function itself)

I considered utilizing data field to pass any relevant information, but as it is right now it's not even passing along any fields, and while I could easily change that, I don't think that it'd lead to clean code.

I welcome any recommendations on how to tackle this. If there's none, I'll gladly close the PR and move to an issue instead.

iceman1001:
There was some other problems, like it always created one file,  overwriting the old.
it now creates unique files and when mergejson param is used it uses the user supplied name.
And ofcourse,  saves in current folder.

Fix:
https://github.com/RfidResearchGroup/proxmark3/commit/9ba03aec741c11ec9c56a99783a0c0e36310686f

PR #760, opened 2020-06-07
Move most of libs to thumb mode.

slurdge:


doegox:
I'm in favor of
* accepting the PR
* waiting for complains (some sim code timings got broken?)
* treating exceptions to thumb possibly in new files (e.g. iso14443a-sim.c) and document properly what has to be ARM & not Thumb

@iceman1001 ?

iceman1001:
....oyeah,  thats right,  this means time-critical stuff like darkside and simulations. ...  Thanks for reminding me. 

Well,  we are bleeding edge,  and its time Slurdges stuff gets merged :)   Go go go

PR #761, opened 2020-06-07
Fix RAMFUNC to be always in arm mode

slurdge:
Just to leave a trace of the fix. Feel free to include directly in main branch


iceman1001:
Did you pull latest before you made this PR?   I think that was already inside the master.

PR #762, opened 2020-06-09
remove duplicated keys (first occurrence kept)

lukaskuzmiak:
Randomly noticed a duplicated key, found some more when searching case-insensitive.
Kept first occurrence, removed subsequent one.


iceman1001:
Good catch!
I though I removed most of the duplicates.

PR #768, opened 2020-06-09
Decode MAD Info Byte and show info about sector it points to

### LINKED ISSUE #766: [idea] Mifare MAD - Card publisher sector enhancement
Now that we have the MAD database, we can perhaps enhance the Card publisher sector info:

```
[=] --- MAD Information -------------------------------
[=] ---------------------------------------------------
[+] Card publisher sector: 0x01
[=]  00 MAD 1
[=]  01 [7006] Hotel, access contr. & sec [Vingcard a.s.]
[=]  02 [7005] Energy Saving System For Hotels, Access Control [ENKOA System]
[=]  03 [7007] Hotel, access contr. & sec [Vingcard a.s.]
[=]  04 [7007] (continuation)
[=]  05 [7007] (continuation)
[=]  06 [7009] Access control data for electronic locks [Timelox AB]
[=]  07 [0000] (free)
[=]  08 [0000] (free)
[=]  09 [0000] (free)
[=]  10 [0000] (free)
[=]  11 [0000] (free)
[=]  12 [0000] (free)
[=]  13 [0000] (free)
[=]  14 [0000] (free)
[=]  15 [0000] (free)
```

Would be cool to actually read `Vingcard a.s.` from the sector1's info and list it with `Card publisher sector`.
Some refactoring might be needed as `Card publisher sector` is printed out before we know if MAD2 is supported and it *may* point into MAD2.

IMHO the best way to go about it is first parse (`MADDecode`), then print out output using the parsed data with `MAD1DecodeAndPrint` and `MAD2DecodeAndPrint`.

Feel free to assign to me if possible, I'll try to find some time for this soon.

lukaskuzmiak:
This closes #766

### DIFF
--- a/client/src/mifare/mad.c
+++ b/client/src/mifare/mad.c
@@ -241,22 +241,44 @@ static const char *aid_admin[] = {
     "not applicable"
 };
 
+static int MADInfoByteDecode(uint8_t *sector, bool swapmad, int MADver, bool verbose) {
+    uint8_t InfoByte;
+    if (MADver == 1) {
+        InfoByte = sector[16 + 1] & 0x3f;
+        if (InfoByte >= 0xF) {
+            PrintAndLogEx(WARNING, "Invalid Info byte (MAD1) value " _YELLOW_("0x%02x"), InfoByte);
+            if (verbose)
+                // I understand the spec in a way that MAD1 InfoByte should not point into MAD2 sectors, @lukaskuzmiak
+                PrintAndLogEx(WARNING, "MAD1 Info byte points outside of MAD1 sector space (0x%02x), report a bug?", InfoByte);
+            return PM3_ESOFT;
+        }
+    } else {
+        InfoByte = sector[1] & 0x3f;
+        if (InfoByte == 0x10 || InfoByte >= 0x28) {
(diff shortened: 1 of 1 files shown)


iceman1001:
Don't forget to add a line in CHANGELOG.md

iceman1001:
Question,  in what way do you think the printing of the line that we don't have before?
Is it the Sector 1 is Card publisher sector?  then we don't need the extra text,   we could just add a hint on the line  01 instead.   Makes for a more dense format.
```
01 [7006] Hotel, access contr. & sec [Vingcard a.s.]  (card publisher sector)
```
image

lukaskuzmiak:
So, the sector is:

> The information of the card publisher sector is particularly useful if somebody needs to find out the organization responsible for distribution of free card sectors for new
applications."

Ideally I would just print `company` out of that sector, but the `print_aid_description` kinda parses and prints all of it and it didn't seem worth changing the whole thing for just the company.

If you prefer a hint on the sector line instead, I can try that. Getting used to how things work around here, sorry if that's too verbose :-)

iceman1001:
Since we already print company,   adding a color on the sector line to emphase would be a nicer output.   Just go to yourself,  would you want to read the same info again?  Would that make it clearer?

So in the case above,   knowing that its Vingcard a.s.  that is reponsible would be the interesting bit.
Its a bit redundant now,  we can make it more crisp.

PR #787, opened 2020-06-13
Mifare Des authentification fixes

bkerler:
Fixes issues with accessing files after authentifcation on mifare desfire. Adds change key and getuid commands. Fixes format picc command.


client/src/mifare/desfire_crypto.c
```diff
@@ -854,3 +868,45 @@ void mifare_cypher_blocks_chained(desfiretag_t tag, desfirekey_t key, uint8_t *i
         offset += block_size;
     }
 }
+
+#define CRC32_PRESET 0xFFFFFFFF
+
+static void
+desfire_crc32_byte(uint32_t *crc, const uint8_t value)
```
iceman1001: These desfire crc32 fcts should be insid e/common/crc32....

bkerler: Yes, thanks for pointing me into the right direction. e7b2dbd removes that stuff.

drandreas:
😎 thx

PR #792, opened 2020-06-15
Fix em4x50

tharexde:
Fixed the following problem:
`lf search´ with an em4x50 tag placed on the device may result in `no known tags found´ and `lf em 4x50_read´ shows parity errors. The problem only occurs if the first bit of an em4x50 word is `0´. If it’s `1´ everything works as expected.

Reason: When performing the raw Manchester decoding in function `cleanAskRawDemod(…)´ the first transition of the signal is not evaluated due to the function call `getNextHigh(…)´. 

I tested the modification against the v4.9237 release by comparing the demodulation results with the available traces in the proxmark `traces´ directory. Doublecheck has been done with numerous available lf cards (em4x50, em41x, em42x, hid, …).
So far I couldn’t observe any differences…

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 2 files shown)


iceman1001:
Nice one!

Since you seem to be at LF,  have a better look at these two :)

- https://github.com/RfidResearchGroup/proxmark3/issues/240
- https://github.com/RfidResearchGroup/proxmark3/issues/149

PR #794, opened 2020-06-16
Resource clean up at cross-compilation added.

xianglin1998:


xianglin1998:
This is a bad commit. sorry.

PR #795, opened 2020-06-16
Resource clean up at cross-compilation added.

xianglin1998:


xianglin1998:
Oh my God, my modification is based on the old version, which is really a bad thing, I have to follow your changes to modify. i 'm sorry.

PR #796, opened 2020-06-16
Cross Compilation bug fixed

xianglin1998:


xianglin1998:
Some of the tested resources were not deleted.

PR #798, opened 2020-06-17
 EM4x50 missing functions 4x50_info, 4x50_ write, 4x50_write_password

tharexde:
Added lf functions for EM4x50:
	1. 4x50_info (displays all available info of EM4x50 tag)
	2. 4x50_write (writes a single word to a specified address of EM4x50 tag)
	3. 4x50_write_password (changes password of EM4x50 tag)
	
Decoding is done on the fly on device side (by directly interpreting pulse lengths, no GraphBuffer)
	- no memory limitations (a tag with 33 words is fully readable)
	- can be used in standalone mode
	
Functions currently only work for Opt64 data rate because I don't have any Opt32 EM4x50 tags. 
Implementation is heavily based on EM4450 datasheet and EM4150 application note (AppNote 406) with ideas from hitag2 code.

### DIFF
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -584,7 +584,8 @@ static bool login(uint8_t password[4]) {
             return true;
     
     } else {
-        Dbprintf("error in command request");
+         if (DBGLEVEL >= DBG_ERROR)
+             Dbprintf("error in command request");
     }
     
     return false;
@@ -605,7 +606,8 @@ static bool reset(void) {
             return true;
 
     } else {
-        Dbprintf("error in command request");
+         if (DBGLEVEL >= DBG_ERROR)
+             Dbprintf("error in command request");
     }
 
     return false;
@@ -634,7 +636,8 @@ static bool standard_read(int *now) {
         return true;
 
     } else {
-        Dbprintf("didn't find a listen window");
+         if (DBGLEVEL >= DBG_ERROR)
+             Dbprintf("didn't find a listen window");
     }
 
     return false;
@@ -667,7 +670,8 @@ static bool selective_read(uint8_t addresses[4]) {
                     return true;
         
     } else {
-        Dbprintf("error in command request");
+         if (DBGLEVEL >= DBG_ERROR)
+             Dbprintf("error in command request");
     }
 
     return false;
@@ -736,7 +740,8 @@ static bool write(uint8_t word[4], uint8_t address) {
         }
 
     } else {
-        Dbprintf("error in command request");
+         if (DBGLEVEL >= DBG_ERROR)
+             Dbprintf("error in command request");
     }
 
     return false;
@@ -774,7 +779,8 @@ static bool write_password(uint8_t password[4], uint8_t new_password[4]) {
         }
 
     } else {
-        Dbprintf("error in command request");
+         if (DBGLEVEL >= DBG_ERROR)
+             Dbprintf("error in command request");
     }
 
     return false;
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -12,6 +12,7 @@
 #include "ticks.h"
 #include "dbprint.h"
 #include "lfadc.h"
+#include "commonutil.h"
 #include "em4x50.h"
 
 // 4 data bytes
@@ -118,28 +119,16 @@ static uint8_t bits2byte(uint8_t bits[8], int length) {
     return byte;
 }
 
-static uint8_t msb2lsb(uint8_t byte) {
-
-    // returns byte with bit positions in reverse order
-    
-    uint8_t tmp = 0;
-    
-    for (int i = 0; i < 8; i++)
-        tmp |= ((byte >> (7-i)) & 1) << i;
-    
-    return tmp;
-}
-
 static void msb2lsb_word(uint8_t *word) {
     
     // reorders given <word> according to EM4x50 datasheet (msb -> lsb)
     
     uint8_t buff[4];
     
-    buff[0] = msb2lsb(word[3]);
-    buff[1] = msb2lsb(word[2]);
-    buff[2] = msb2lsb(word[1]);
-    buff[3] = msb2lsb(word[0]);
+    buff[0] = reflect8(word[3]);
+    buff[1] = reflect8(word[2]);
+    buff[2] = reflect8(word[1]);
+    buff[3] = reflect8(word[0]);
 
     word[0] = buff[0];
     word[1] = buff[1];
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -717,7 +717,7 @@ static bool write(uint8_t word[4], uint8_t address) {
     
     if (request_receive_mode()) {
 
-        // send selective read command
+        // send write command
         em4x50_send_byte_with_parity(EM4X50_COMMAND_WRITE);
 
         // send address data
@@ -753,7 +753,7 @@ static bool write_password(uint8_t password[4], uint8_t new_password[4]) {
     
     if (request_receive_mode()) {
 
-        // send selective read command
+        // send write password command
         em4x50_send_byte_with_parity(EM4X50_COMMAND_WRITE_PASSWORD);
 
         // send address data
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -791,7 +791,7 @@ void em4x50_write(em4x50_data_t *etd) {
     // write operation process for EM4x50 tag,
     // single word is written to given address, verified by selective read operation
     
-    bool bsuccess = true, blogin = false;
+    bool bsuccess = false, blogin = false;
     uint8_t word[4] = {0x00, 0x00, 0x00, 0x00};
     uint8_t addresses[4] = {0x00, 0x00, 0x00, 0x00};
     
@@ -826,6 +826,7 @@ void em4x50_write(em4x50_data_t *etd) {
                 word[3] = tag.sectors[etd->address][3];
                 msb2lsb_word(word);
                 
+                bsuccess = true;
                 for (int i = 0; i < 4; i++)
                     bsuccess &= (word[i] == etd->word[i]) ? true : false;
 
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -220,7 +220,7 @@ static void em4x50_setup_read(void) {
     //   TIMER_CLOCK0, used to measure exact timing before answering
     //   TIMER_CLOCK1, used to capture edges of the tag frames
     AT91C_BASE_PMC->PMC_PCER |= (1 << AT91C_ID_TC0) | (1 << AT91C_ID_TC1);
-    AT91C_BASE_PIOA->PIO_BSR = GPIO_SSC_FRAME; // wofür ist das?
+    AT91C_BASE_PIOA->PIO_BSR = GPIO_SSC_FRAME;
 
     // Disable timer during configuration
     AT91C_BASE_TC0->TC_CCR = AT91C_TC_CLKDIS;
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -102,7 +102,7 @@ static void init_tag(void) {
             tag.sectors[i][j] = 0x00;
 }
 
-static uint8_t bits2byte(uint8_t bits[8], int length) {
+static uint8_t bits2byte(uint8_t *bits, int length) {
 
     // converts <length> separate bits into a single "byte"
     
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -341,39 +341,23 @@ static void em4x50_send_byte(uint8_t byte) {
 
     // send byte (without parity)
     
-    int mask = 0x80;
+    for (int i = 0; i < 8; i++)
+        em4x50_send_bit((byte >> (7-i)) & 1);
 
-    for (int i = 0; i < 8; i++) {
-
-        if ((byte & mask) == 0)
-            em4x50_send_bit(0);
-        else
-            em4x50_send_bit(1);
-        
-        mask >>= 1;
-    }
 }
 
 static void em4x50_send_byte_with_parity(uint8_t byte) {
 
     // send byte followed by its (equal) parity bit
     
-    int parity = 0;
-    int mask = 0x80;
-
-    for (int i = 0; i < 8; i++) {
-        
-        if ((byte & mask) == 0) {
-            em4x50_send_bit(0);
-            parity ^= 0;
-        } else {
-            em4x50_send_bit(1);
-            parity ^= 1;
-        }
-        
-        mask >>= 1;
-    }
+    int parity = 0, bit = 0;
     
+    for (int i = 0; i < 8; i++) {
+        bit = (byte >> (7-i)) & 1;
+        em4x50_send_bit(bit);
+        parity ^= bit;
+    }
+
     em4x50_send_bit(parity);
 }
 
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -91,6 +91,9 @@ static em4x50_tag_t tag = {
 
 #define FPGA_TIMER_0                        0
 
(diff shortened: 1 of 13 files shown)


client/src/cmdlfem4x50.c
```diff
@@ -0,0 +1,720 @@
...
+    PrintAndLogEx(NORMAL, "Options:");
+    PrintAndLogEx(NORMAL, "       h         - this help");
+    PrintAndLogEx(NORMAL, "       p <pwd>   - password (hex)");
+    PrintAndLogEx(NORMAL, "       n <pwd>   - new password (hex)");
+    PrintAndLogEx(NORMAL, "Examples:");
+    PrintAndLogEx(NORMAL, "      lf em 4x50_write_password p 11223344 n 01020304");
+    return PM3_SUCCESS;
+}
+
+static uint8_t msb2lsb(uint8_t byte) {
```
iceman1001, line 61: Consider using:
https://github.com/RfidResearchGroup/proxmark3/blob/master/common/commonutil.c#L63

client/src/cmdlfem4x50.c
```diff
@@ -0,0 +1,720 @@
...
+    
+    uint8_t tmp = 0;
+    
+    for (int i = 0; i < 8; i++)
+        tmp |= ((byte >> (7-i)) & 1) << i;
+    
+    return tmp;
+}
+
+static bool check_bit_in_byte(uint8_t pos, uint8_t byte) {
```
iceman1001, line 73: Would need to unify these check bits functions...
https://github.com/RfidResearchGroup/proxmark3/blob/master/client/src/cmdhficlass.c#L379-L385

client/src/cmdlfem4x50.c
```diff
@@ -0,0 +1,720 @@
...
+    }
+
+    // binary stop bit
+    strcat(string, " | ");
+    sprintf(pstring, (word.stopparity) ? "%i" : _RED_("%i"), word.stopbit);
+    strcat(pstring, "  parities ");
+    strcat(string, pstring);
+    
+    // parities passed/failed
+    sprintf(pstring, (word.parity) ? _GREEN_("passed") : _RED_("failed"));
```
iceman1001, line 195: replace "passed" w "ok"     We are trying to unify these kinds of output in the client.

client/src/cmdlfem4x50.c
```diff
@@ -0,0 +1,720 @@
...
+            strcat(string, pstring);
+
+            for (int j = 0; j < 4; j++) {
+                sprintf(pstring, "%02x", words[i].byte[j]);
+                strcat(string, pstring);
+            }
+            
+            switch(i) {
+                case 0:
+                    sprintf(pstring, _BLUE_("  password, write only"));
```
iceman1001, line 290: BLUE is reserved for system.
Recommend using  YELLOW or GREEN,

client/src/cmdlfem4x50.c
```diff
@@ -0,0 +1,720 @@
...
+    }
+
+    // validation
+     if (errors) {
+         return usage_lf_em4x50_info();
+    }
+
+    // call info command
+    clearCommandBuffer();
+    SendCommandMIX(CMD_LF_EM4X50_INFO, 0, 0, 0, &etd, sizeof(etd));
```
iceman1001, line 457: OLD/MIX  shouldn't be use anymore,   use  NG format instead.

client/src/cmdlfem4x50.c
```diff
@@ -0,0 +1,720 @@
...
+
+    if (errors || !bpwd || !bnewpwd)
+         return usage_lf_em4x50_write_password();
+
+    clearCommandBuffer();
+    SendCommandMIX(CMD_LF_EM4X50_WRITE_PASSWORD, 0, 0, 0, &etd, sizeof(etd));
+
+    if (!WaitForResponse(CMD_ACK, &resp)) {
+        PrintAndLogEx(WARNING, "\n  timeout while waiting for reply.\n");
+        return PM3_ESOFT;
```
iceman1001, line 713: return code for timeouts,  PM3_ETIMEOUT   ;)

armsrc/em4x50.c
```diff
@@ -0,0 +1,877 @@
...
+        [25] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // user
+        [26] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // user
+        [27] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // user
+        [28] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // user
+        [29] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // user
+        [30] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // user
+        [31] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // user
+        [32] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // device serial number
+        [33] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // device identification
+    },
```
iceman1001, line 58: These kinds of solutions is not recommend,   we have scarse ram on device side,  so all these kinds of structures will be needed to be big_buff malloced and freed.   
For simulation later on,  try use the external flash memory for storage.

armsrc/em4x50.c
```diff
@@ -0,0 +1,877 @@
...
+static void init_tag(void) {
+
+    // initialize global tag structure
+    
+    for (int i = 0; i < 34; i++)
+        for (int j = 0; j < 7; j++)
+            tag.sectors[i][j] = 0x00;
+}
+
+static uint8_t bits2byte(uint8_t bits[8], int length) {
```
iceman1001, line 104: send ref?

armsrc/em4x50.c
```diff
@@ -0,0 +1,877 @@
...
+        byte |= bits[i];
+
+        if (i != length-1)
+            byte <<= 1;
+    }
+
+    return byte;
+}
+
+static uint8_t msb2lsb(uint8_t byte) {
```
iceman1001, line 121: see common utils.

armsrc/em4x50.c
```diff
@@ -0,0 +1,877 @@
...
+    AT91C_BASE_PIOA->PIO_OER = GPIO_SSC_DOUT;
+
+    // Disable modulation at default, which means enable the field
+    LOW(GPIO_SSC_DOUT);
+    
+    // Enable Peripheral Clock for
+    //   TIMER_CLOCK0, used to measure exact timing before answering
+    //   TIMER_CLOCK1, used to capture edges of the tag frames
+    AT91C_BASE_PMC->PMC_PCER |= (1 << AT91C_ID_TC0) | (1 << AT91C_ID_TC1);
+    AT91C_BASE_PIOA->PIO_BSR = GPIO_SSC_FRAME; // wofür ist das?
```
iceman1001, line 223: Comment?   Please, use english.

armsrc/em4x50.c
```diff
@@ -0,0 +1,877 @@
...
+
+static int get_next_bit(void) {
+    
+    // returns bit value (or EM4X50_BIT_OTHER -> no bit pattern) by evaluating
+    // a single sample within a bit period (given there is no LIW, ACK or NAK)
+    // This function is not used for decoding, it is only used for identifying
+    // a listen window (return value = EM4X50_BIT_OTHER) in functions
+    // "find_double_listen_window" and "check_ack"
+  
+    int dhigh = 230;    // 90% of maximum sample value (255)
```
iceman1001, line 256: these kinds of envelope values would need to be dynamic in order to get better distances.

armsrc/em4x50.c
tharexde: Oops - I forgot to delete it. What I wanted to say: "I have no idea what it means."

armsrc/em4x50.c
```diff
@@ -0,0 +1,877 @@
...
+
+static void em4x50_send_byte(uint8_t byte) {
+
+    // send byte (without parity)
+    
+    int mask = 0x80;
+
+    for (int i = 0; i < 8; i++) {
+
+        if ((byte & mask) == 0)
```
iceman1001, line 348: I think you can simplify this if statement to one line...

armsrc/em4x50.c
```diff
@@ -0,0 +1,877 @@
...
+        while (get_word_from_bitstream(bits) == EM4X50_TAG_WORD)
+            save_word((*now)++, bits);
+
+        // number of detected words
+        *now -= fwr;
+        
+        return true;
+
+    } else {
+        Dbprintf("didn't find a listen window");
```
iceman1001, line 664: use guard - if statements like   IF DBG_LEVEL >= ....  around  printing

iceman1001:
Great addition.  I have writting some comments which should be addressed. I didn't comment on all places but you understand the meaning.  

Almost so we need  read / sim   aswell on deviceside now ;)

tharexde:
Wow - that was a fast reply, thanks. And yes, I understand the meaning but it will take some time because I'm not a professional software architect and not used to work with git. 

I agree that there are still some functions missing, e.g.
- read: This is easy because it already exists ("standard read" and "selective read"). Only the corresponding part in cmdlfem4x50.c has to be added. What about the name of this function? "4x50_read" already exists. And besides: Any possible information is already presented by "lf em 4x50_info p <pwd>".
- sim: I started with the implementation but it seems to be slightly more complex than writing. At the moment it's possible to catch the readers command request but I'm still not able to catch the command itself. 

But before thinking about more functions - let's do the integration of this pull request first. :-)

client/src/cmdlfem4x50.c
tharexde: done

client/src/cmdlfem4x50.c
tharexde: done

client/src/cmdlfem4x50.c
tharexde: done

client/src/cmdlfem4x50.c
tharexde: done

armsrc/em4x50.c
tharexde: done

armsrc/em4x50.c
tharexde: deleted

armsrc/em4x50.c
tharexde: done

armsrc/em4x50.c
tharexde: done

armsrc/em4x50.c
tharexde: I will keep this in mind when realizing the "sim" function.

armsrc/em4x50.c
tharexde: What do you mean?

client/src/cmdlfem4x50.c
tharexde: done

armsrc/em4x50.c
iceman1001: bits[8]  ..  use  uint8_t *bits   instead

iceman1001:
For sure,  we prefer smaller PR's  which gets merged faster.

doegox:
Hi, See above the AppVeyor/Travis tests.

Apparently the client fails compiling with cmake:
```
CMakeFiles/proxmark3.dir/src/cmdlfem4x.c.o:(.data.rel+0x288): undefined reference to `CmdEM4x50Info'
CMakeFiles/proxmark3.dir/src/cmdlfem4x.c.o:(.data.rel+0x2a8): undefined reference to `CmdEM4x50Write'
CMakeFiles/proxmark3.dir/src/cmdlfem4x.c.o:(.data.rel+0x2c8): undefined reference to `CmdEM4x50WritePassword'
```
And on Windows we get more errors around strncpy usages, cf 
https://ci.appveyor.com/project/RfidResearchGroup/proxmark3/builds/33597928#L394

armsrc/em4x50.c
tharexde: What do you mean by "distances"? Would a #define statement regarding "dhigh" at the top of the file be sufficient?

tharexde:
Hi, 
thanks for the hint. So far I've only compiled it on macOS (10.15.5). 
I'll start with a windows installation to check the strncpy issue.

armsrc/em4x50.c
tharexde: done

client/src/cmdlfem4x50.c
```diff
@@ -0,0 +1,720 @@
...
+                    errors = true;
+                    break;
+                } else {
+                    bnewpwd = true;
+                }
+              
+                // prepare given password
+                etd.newpwd_given = true;
+                for (int i = 0; i < 4; i++)
+                    sscanf(strncpy(tmpbuff, new_password+2*i, 2), "%2hhx", &etd.new_password[i]);
```
iceman1001, line 684: try using  _param_gethex_  for getting a hex value from command line.

armsrc/em4x50.c
iceman1001: No,   with distances I mean read/write distance between tag and antenna.
Having a hardcoded value means you limit it. 

Usually if you sample the signal strength  32 or 64 times, you can get an average of signal strength.
that would be your "HI" value,  from that you calc your envelope trigger 90% value,    hence allowing for much better distances.

client/src/cmdlfem4x50.c
tharexde: Thanks - it looks much better now and there are no errors on WIndows. -> done

client/src/cmdlfem4x50.c
tharexde: function "check_bit_in_byte(...)" didn't really make sense - deleted

client/src/cmdlfem4x50.c
```diff
@@ -244,14 +237,14 @@ static void print_info_result(PacketResponseNG *resp, const em4x50_data_t *etd,
...
-    bpwc = check_bit_in_byte(7, words[2].byte[2]);   // password check
-    braw = check_bit_in_byte(6, words[2].byte[2]);   // read after write
-    fwr = reflect8(words[2].byte[0]);    // first word read
-    lwr = reflect8(words[2].byte[1]);    // last word read
-    fwrp = reflect8(words[1].byte[0]);   // first word read protected
-    lwrp = reflect8(words[1].byte[1]);   // last word read protected
-    fwwi = reflect8(words[1].byte[2]);   // first word write inhibited
-    lwwi = reflect8(words[1].byte[3]);   // last word write inhibited
+    bpwc = (bool)(words[2].byte[2] & 128);  // password check (bit 7)
+    braw = (bool)(words[2].byte[2] & 64);   // read after write (bit 6)
```
iceman1001, line 241: Usually you would use a enum or defines,  to specify what each bit does.
and compare it accordingly.  This way the code becomes much more readable and documented at the same time.
```
#define  EM4x50_CONF_BLK  2
#define  EM4x50_SETTING_BYTE  2

#define PWD_CHK  0x80
#define READ_AFTER_WRITE 0x40

if ( (words[EM4x50_CONF_BLK].byte[EM4x50_SETTING_BYTE] & PWD_CHK) == PWD_CHK) 

```

armsrc/em4x50.c
tharexde: That does make sense.
I added a EM4x50 specific function for this calculation, but I'm still using two hardcoded values ("noise", "sample_ref"). From my point of view these two remaining values don't have to be set dynamically. ;-)

client/src/cmdlfem4x50.c
tharexde: I agree and increased readability by adding numerous defines -> done

iceman1001:
OK, Looking great.
I am ready to merge.

PR #808, opened 2020-06-22
Little optimization for hitag2hell attack

aczid:
Thank you for including the hitag2hell attack code in the project. I noticed that the version of the code published during the WOOT proceedings was not generated using the very latest version of the memoization algorithm. The fixes in this PR make the CPU implementation about 10% faster in practice according to my benchmark. The published GPU version is not affected by this oversight.

Please let me know if you have plans to develop this implementation further, for example to add support for smaller or larger vector machines as was done for the crypto1 hardnested attack. I would like to point out in advance that such a modification would work better when the vector size is taken into account in the cracker code itself, because the work per layer is distributed differently as a consequence of the word size according to the memoization algorithm.

### DIFF
--- a/tools/hitag2crack/crack5/ht2crack5.c
+++ b/tools/hitag2crack/crack5/ht2crack5.c
@@ -330,7 +330,6 @@ static void *find_state(void *thread_d) {
                         const bitslice_value_t filter4 = f_c_bs(filter4_0, filter4_1, filter4_2, filter4_3, filter4_4);
                         bitslice_t results4;
                         results4.value = results3.value & (filter4 ^ keystream[4].value);
-
                         if (results4.bytes64[0] == 0
                                 && results4.bytes64[1] == 0
                                 && results4.bytes64[2] == 0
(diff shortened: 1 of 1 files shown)


iceman1001:
The idea is to incorporate them into the client aswell,  I guess cpu version.  
Or print hints what to run for gpu version,  like we do for _mfkey32_

If there is possibility to use AVR2, AVR, AVR512 adaptive versions it would be very interesting.

How is the vector size adapted?   I am in need of a cuda version of the gpu code. Its in OPENCL at the momement.

doegox:
Tested on fast examples from pm3_test.sh on my 8-core laptop:
* from 67.6 s to 60.5 s
* from 3.27 s to 2.77 s
Nice :)

doegox:
I would refrain from integrating the code entirely in the client @iceman1001 so we can use it independently as of today. But we can easily spawn a call to the executable (cpu/gpu) from the client.

iceman1001:
Agreed, just like we do with _mfkey32_ today.   Its independent tools and also incorporated in the client.  Exact how this will be done, is for later.

aczid:
> The idea is to incorporate them into the client aswell, I guess cpu version.
> Or print hints what to run for gpu version, like we do for _mfkey32_
> 
> If there is possibility to use AVR2, AVR, AVR512 adaptive versions it would be very interesting.
> 
> How is the vector size adapted? I am in need of a cuda version of the gpu code. Its in OPENCL at the momement.

The vector size determines how many bits are pre-set in the initial state using constants, and this in turn determines how many layers of code are needed to fill in the rest of the state in after that.

A different vector size also means that a different number of them that can reasonably fit inside the CPU cache.  Once we start trashing the cache we are likely in a worse position than before increasing the word size. Generating a new version for a different word size therefore takes a bit of trial and error to get the ideal balance between caching and execution. In practice this means not memoizing on every possible layer to save the valuable cache memory for the early / frequently used results.

To make a CUDA version you could just copypaste the OpenCL code (the underlying machine is the same and I've noticed that cache size is less of a concern for this algorithm on the GPU).

Issue #819, opened 2020-06-25
LF and sometimes HF antenna going hairwire in recent master revision


sh7d:
**Describe the bug**
In recent software revision (branch: master, commit: 1c275103948ca33580f621c7b3fc3266290de2d6) hw antenna tune going just hairwire and gives weird results:
```
[usb] pm3 --> hw tune
[=] Measuring antenna characteristics, please wait...
 🕕  12
[+] LF optimal: 542862,64 V -   0,00 kHz
[+] LF antenna is OK

[!] ⚠️  HF antenna is UNUSABLE


[-] ⛔ Not showing LF tuning graph since all values is zero.
```
Sometimes HF antena are working:
```
[usb] pm3 --> data tune
[=] Measuring antenna characteristics, please wait...
 🕕  12
[+] LF antenna:  1,33 V - 125,00 kHz
[+] LF antenna:  1,30 V - 134,83 kHz
[+] LF optimal:  1,46 V -  46,88 kHz
[!] ⚠️  LF antenna is UNUSABLE

[+] HF antenna: 46,80 V - 13.56 MHz
[+] HF antenna is OK


[+] Displaying LF tuning graph. Divisor 88 is 134,83 kHz, 95 is 125,00 kHz.
```
In any case lf tags are not found, so its not only the problem with the command:
lf search with tag on proxmark:
```
[usb] pm3 --> lf search
[=] NOTE: some demods output possible binary
[=] if it finds something that looks like a tag
[=] False Positives ARE possible
[=]
[=] Checking for known tags...
[=]
[-] ⛔ No data found!
[=] Signal looks like noise. Maybe not an LF tag?
[usb] pm3 -->
```

When i checkout v4.9237 its working like a charm:
```
[usb] pm3 --> hw tune
[=] Measuring antenna characteristics, please wait...
 🕚  10
[+] LF antenna: 70,81 V - 125,00 kHz
[+] LF antenna: 34,79 V - 134,83 kHz
[+] LF optimal: 71,36 V - 126,32 kHz
[+] LF antenna is OK

[+] HF antenna: 47,44 V - 13.56 MHz
[+] HF antenna is OK


[+] Displaying LF tuning graph. Divisor 88 is 134,83 kHz, 95 is 125,00 kHz.

[usb] pm3 -->
```
And yes, i reflashed firmware between the runs

**To Reproduce**
1. Check out master
2. Compile and reflash
3. hw tune
4. Check out v4.9237
[...]

PR #821, opened 2020-06-26
Stack size fix in ldscript

sh7d:
It's just fixes my problems in #819

### DIFF
--- a/common_arm/ldscript.common
+++ b/common_arm/ldscript.common
@@ -9,7 +9,7 @@ ms of the GNU GPL, version 2 or,
 -----------------------------------------------------------------------------
 */
 
-stacksize = DEFINED(stacksize) ? stacksize : 8488;
+stacksize = DEFINED(stacksize) ? stacksize : 8K;
 commonareasize = 0x20;
 
 /* AT91SAM7S256 has 256k Flash and 64k RAM */


iceman1001:
have you tested all functionality so nothing else breaks?

sh7d:
Well, hitag works, mf autopwn works, mf sim works, i dont have time for more

doegox:
@iceman1001 there last days we pushed the cursor back & forth a few times depending on the bugs, sometimes with bugfixes which relax constraints but we should track all use cases that claim a larger or a smaller stack, see which ones need to be fixed and know what are the real min/max where nothing breaks (if. there is such fine spot today :) )
Honestly I'm lost at this point.

doegox:
and poking @slurdge who opened the Pandora box ;)

slurdge:
Hey this is me, the pandora box opener!
What is happening is that we have two types of workloads: workloads that put everything on the stack (which is good), and workloads that use most of bigbuf (hopefully with a little bit of stack).

I guess that by chance 40K bigbuf was the sweet spot, what we can do is go back to 40K bigbuf, mark the end of it, and compute what was the max stack at that point.

However, moving forward we should:
 - Have less & less static buffers. Either alloc() them or stack them;
 - Be able to change the stack size depending on the workloads, either in the Makefile (with rules for Standalones etc.) or at the runtime (more difficult)

doegox:
> I guess that by chance 40K bigbuf was the sweet spot, what we can do is go back to 40K bigbuf, mark the end of it, and compute what was the max stack at that point.

Well that was the status with 40K BigBuff + 8488 stack wasn't it ? And for mysterious reasons it's buggy (#819)

slurdge:
Yeah it should... But that's why I suggested to go back. However I know remember some other fine point: we don't control that BigBuf is at the end.
So, what probably happens.
Before
```Code | BigBuf | Some other .bss content | Stack [Filling space]```
If "some other content" is not used with heavy stack usage, then everything is "fine".
Now
```Code | .bss | BigBuf Filling space | Stack [Fixed Size]```
So _by design_ BigBuf is at the end. We should check how much .bss there is.

iceman1001:
Right now its the bigbuff usage all cross the device side code which leads to weird stuff.
Sometimes it check available tracelen limit and some time there is a hard encoded 40000 size.  
We need to go through all of the code find all culprits.   Nothing else to do.

slurdge:
This man speaks the gospel.
BTW the .bss size is ~6.4K on latest trunk. So it is quite possible that the aforementioned scenario is taking place

iceman1001:
Closing since this one isn't a prefereed solution to the actual problem.

PR #822, opened 2020-06-26
EM4x50 function "sread" (standard/selective read)

tharexde:
Added new read function "sread" for EM4x50 (device side): 
 
- If no option is given „em4x50_sread“ carries out a „standard read“, i.e. the data transmitted by tag voluntarily is catched and decoded.
- If a dedicated address is given (option a) „em4x50_sread“ carries out a „selective read“, i.e. data stored at this address is requested and decoded.


iceman1001:
Nice!
I am curious how this is different from the current implemented 4x50_read commands?

tharexde:
There may be some points:
- "sread" decodes on device side, "read" on client side
- an EM4x50 tag contains up to 34 words, which are all decodeable by "sread", "read" can only decode up to max. 6 words (otherwise it stops with "No data found!..." or "Partial data...")
- "sread" realizes a two-way-communication, "read" decodes the GraphBuff filled by "lf read"
- "sread" can easily be used as the core of an EM4x50 standalone mode (would you like to have one?)
- "sread" can display e.g. read protected words if password is known, "read" displays only data that EM4x50 tag presents voluntarily (-> standard read mode)
- you asked for it ;-)

iceman1001:
Good,   because I don't see any use of the old read with this PR.
Might just remove the old one and replace it with this one

tharexde:
Ok, I'm going to do the following:
- delete the EM4x50 functions in cmdlfem4x.c
- rename "4x50_lread" to "4x50_read"
- integrate the new "4x50_read" into "lf search" process
Did I miss anything?

iceman1001:
I can merge this one and you can make a new PR for the next steps.

PR #825, opened 2020-06-30
replace em4x50 read

tharexde:
1. deleted EM4x50 functions in cmdlfem4x
2. renamed "4x50_sread" to "4x50_read"
3. replaced old "4x50_read" by new "4x50_read" in "lf search" process
4. fixed some timeout errors in em4x50.c/cmdlfem4x50.c

It should be noted, that the "lf search" command now takes about 2 seconds longer compared to the former version. This delay is due to the algorithm that decides "No, it's not an EM4x50 tag".


iceman1001:
Nice!

Lets see, the  em4x50read verbose if,   and param handling inside that function is  less prefered.
params should be inside the Cmdem4x50read,   to keep client cmd stuff seperated from functions.
If you need a pwd,   that should be parsed in cmd,   and used in function call to em4x50read(pwd).  You understand what I mean?

I can merge this and then you can make a new PR for those changes

PR #828, opened 2020-06-30
speed up "lf search" process (EM4x50)

tharexde:
Performance of "lf search" should now be comparable to the former version (if tag is not an EM4x50).

BTW: I've seen your improvements, thanks. So far I've two comments:
1. What should the output of "lf em 4x50_read" look like? I'm missing any information regarding the "standard read mode", which may be essential for copying a tag. In many cases the "standard read mode" equals transmitting the serial number, but that's not mandatory.
2. Sometimes there are wrong results (=serial) in "lf em 4x50_read". When using "lf search" the result is correct. I'll have a look at it.

### DIFF
--- a/armsrc/em4x50.c
+++ b/armsrc/em4x50.c
@@ -74,6 +74,7 @@ static em4x50_tag_t tag = {
 #define EM4X50_T_TAG_FULL_PERIOD            64
 #define EM4X50_T_TAG_TPP                    64
 #define EM4X50_T_TAG_TWA                    64
+#define EM4X50_T_WAITING_FOR_SNGLLIW        50
 #define EM4X50_T_WAITING_FOR_DBLLIW         1550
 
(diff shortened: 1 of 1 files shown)


iceman1001:
Yeah,   the whole parity output seems to me to be debug related more than what a user expect.

- lf em 4x50_info   -> should give detailed info about the card.  like config,settings, manufacturer etc.
- lf em 4x50_dump  ->  should print and write a bin/eml/json dump file of complete memory.  With settings and extra in the json format
- lf em 4x50_read  -> should give you the read data.
- lf search ->  if found,  it should give a tag serial and/or the emulated data.    

Since I am not sure if 4x50 can be programmed like 4x05 to emulate other credentials like HID, IO, Indala, 410x like we use t55x7 / Q5 for.    If it can,  the lf search should print that data ...

Feel free to reach out on email or signal to discuss more on this subject.

PR #829, opened 2020-07-03
Android library reused.

xianglin1998:
####
My project tree:

image

In bzip2.cmake. i complie source and generate static library.

####
Why i use git submodule for bzip2:

Because the AndroidStudio does not support "ExternalProject",  I only can put bzip2 into an external project and compiled with cmake.
tip:  AndroidStudio use Ninja to make all makefile, not UNIX makefile system.

####
Why android/CMakeLists.txt have the ExternalProject?

Because on unix system, the ExternalProject is work.

####
Some changed at pm3_main.c, i make it can work on android.

Please review, thanks. :)


client/android/pm3_main.c
```diff
@@ -28,26 +28,18 @@
...
-int push_cmdscriptfile(char *path, bool stayafter) {
-    return PM3_SUCCESS;
-}
-
 static char *g_android_executable_directory = NULL;
-static const char *g_android_user_directory = NULL;
+static char *g_android_user_directory = NULL;
 
-const char *get_executable_directory(void) {
+char version_information[] = {"ANDROID_LIBRARY 1.4.6 build by DXL"};
```
iceman1001, line 40: this needs to be removed.

PR #835, opened 2020-07-11
fix a bug in demodParadox()

jumpycalm:
Wrong offset leads to wrong hi2, hi and lo values.

### DIFF
--- a/client/src/cmdlfparadox.c
+++ b/client/src/cmdlfparadox.c
@@ -128,7 +128,7 @@ int demodParadox(void) {
     uint8_t error = 0;
 
     // Remove manchester encoding from FSK bits, skip pre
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice catch,  however I don't think your suggested fix will do it.

Offending line.
```
    for (uint8_t i = idx + PARADOX_PREAMBLE_LEN; i < (idx + 96 - PARADOX_PREAMBLE_LEN ); i += 2) {
```
```
Let's say 
 IDX is 5
 PREAMBLE LEN is 20

// starting point:  i = 5 + 20 = 25 

// endpoint,   
// i < 5 + 96 - 20 = 81
// i < 5 + 96 = 101
// i < 5 + 96 + 20 = 121

// endpoint minus startingpoint == desired length of 96.
//  81 - 25 = 56
// 101 - 25 = 76
// 121 - 25 = 96
```

So the correct fix should be a plus sign instead of a minus sign.

iceman1001:
No,  you were correct,   the line should skip the preamble (len =8) from those 96bits.

Issue #880, opened 2019-10-28
Windows Color Concept


mwalker33:
As we have seen the direct support of ansi colors in windows is tricky.  Windows 10 has added some support in cmd.exe as long as a registry key is set and legacy mode is turned off.
Windows terminal preview also supports ansi colors.
Proxspace (defterm) does not support ansi color either in terminal or cmd.exe.

I have had an initial play with converting the ansi colors to the cmd.exe 16 colors.
early tests show good results (code still needs work).  
With no changes to windows registry or any checkbox this seems to work.
I have tested the concept in proxspace, windows cmd.exe and windows terminal, all 3 show color as I expect.

To achieve this, i made some short term changes to force color on, then if win32 call my convert function from the logandprintex function.

For review and comment see the follow branch.

https://github.com/mwalker33/proxmark3-rrg/tree/Win10ColorUpdate

At this point I am looking for an initial comment on the concept.

If we think this could work i will spend some more time on the code and clean it, I am happy to call for some testers to test (off that branch) or what every way you think best.

Thanks.

mwalker33:
I think this on can be closed now.

Preference/settings is being driven via its issue.
Color, should be working as per the latest via  @Gator96100 pr.
If there are any VT/Esc screen position issues, lets open a new issueif/when needed.

PR #902, opened 2020-07-12
Refactor demodParadox()

jumpycalm:
1. Fixed the bug of Paradox ID, now the ID is 32 bits in length, first 8 bits is FC, next 16 bits is CN, last 8 bits is CRC;
2. Made code easier to read, code is much simpler now and fixed the color output;
3. Removed doegox's code on attempting to reverse engineering the 8 bit CRC value. I think new CRC reverse engineering needs to be based on the first 24 bits of the Paradox ID, so new codes to be written;
4. Added instruction in the header to allow other developers to easily understand Paradox.

### DIFF
--- a/armsrc/Standalone/lf_icehid.c
+++ b/armsrc/Standalone/lf_icehid.c
@@ -228,7 +228,8 @@ static uint32_t IceHIDDemod(void) {
 
     int dummyIdx = 0;
 
-    uint32_t hi2 = 0, hi = 0, lo = 0;
+    uint64_t hid_id_hi = 0;
+    uint64_t hid_id_lo = 0;
 
     // large enough to catch 2 sequences of largest format
 //    size_t size = 50 * 128 * 2;  // 12800 bytes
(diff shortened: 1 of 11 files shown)


iceman1001:
Interesting,

it isn't doegox crc comparisions,  its mine.  WIP,   it was there for another user to be able to compare his algo vs the builtin maxim-crc.  So yes,  its a bit messy in there as you seen.

About colors,  the exact style for to print and which to print,  is an ongoing process to unify all the LF demod outputs.  
There is a issue #880 about it.

Where do you figure that the preamble is that length? I am curious which sample set you have been using to come the that conclusion.

This PR will not be merged

jumpycalm:
Thank you iceman,

The purpose of removing your CRC code was not because it was messy, it was because we need to refactor the CRC check code based on demoded the FC and CN (24 bits in total) instead of the raw data. The CRC check is good thing to have in terms of reverse engineering the 8 bit CRC.

In terms of the color, the checksum value as well as the raw data's value was missed. If these results should not be highlighted by color, I can have it removed.

The preamble is 28 bits instead of 8 bits. The preamble is all the data before the 8 bit FC. The samples I tested: 0f55555555a995a5a9669a5a 0f5555555666a9aaa6a5aa6a 0f555555566a6959aa656a6a 0f55555559595aa559a5566a 0f55555559aaa69aa5a6565a 0f555555a5995566a699a6aa

One extra thing I want to add, based on some online research regarding Paradox, I found this link: https://clonemykey.com/copy-wizard/new-key/key-fob/paradox-key-fob/ We can tell two things from this link: 1. The FC is 8 bit; 2. Someone already managed to figure out the CRC calculation.

iceman1001:
Closing this one.  Too many changes across commands.

Try to keep PR to one change at the time.  Its a mess to test and verify otherwise.

Issue #832, opened 2020-07-05
NTAG215 old MFU format conversion not working?


shellster:
**Describe the bug**
If I use the very old (deprecated) iceman fork found here: https://github.com/iceman1001/proxmark3, I can use a tool like https://raw.githubusercontent.com/tomvanveen/samytools/master/mfubin2eml to clone an amiibo to a Proxmark3 Easy and successfully use it with a Nintendo switch.  However, if I use the latest version (proxmark3-4.9237) from this repo (and same amiibo file that works with the deprecated version) I get the following warnings:

```
proxmark3 /dev/ttyACM0 -c 'hf mf eload u amiibo'
[=] Session log <redacted>/log_20200705.txt
[=] Loading Preferences...
[+] loaded from JSON file <redacted>/preferences.json
[+] execute command from commandline: hf mf eload u amiibo

[=] Using UART port /dev/ttyACM0
[=] Communicating with PM3 over USB-CDC
[usb|script] pm3 --> hf mf eload u amiibo
[+] loaded 1020 bytes from text file amiibo.eml
[+] old mfu dump format, was converted on load to 243 pages
[=] Uploading to emulator memory
.................................................................................................................................................................................................................................................................

[-] ⛔ Warning, Ultralight/Ntag file content, Loaded 257 blocks into emulator memory

proxmark3 /dev/ttyACM0 -c 'hf 14a sim t 7 u 0418521a454b81'
[=] Session log <redacted>/log_20200705.txt
[=] Loading Preferences...
[+] loaded from JSON file <redacted>/preferences.json
[+] execute command from commandline: hf 14a sim t 7 u 0418521a454b81

[=] Using UART port /dev/ttyACM0
[=] Communicating with PM3 over USB-CDC
[usb|script] pm3 --> hf 14a sim t 7 u 0418521a454b81
[+] Emulating ISO/IEC 14443 type A tag with 7 byte UID (04 18 52 1A 45 4B 81 )
[=] Press pm3-button to abort simulation
```

[...]

PR #913, opened 2020-08-01
update bin2eml script from https://github.com/samyk/samytools/

mokus0:
As per discussion in #832, updates bin2eml script to generate new header format, calculate auth key, etc


tools/pm3_amii_bin2eml.pl
```diff
@@ -669,13 +691,15 @@
...
-00808000
-92580b4c
-45a9c42f
-a90145ce
-5e5f9c43
-09a43d47
-d232a3d1
-68cbade6
-7f8185c6
+010000
```
iceman1001, line 694: missing a byte here?

mokus0: It’s intentional; line 670 fills it in with the page count.

PR #922, opened 2020-08-13
Redefine the WSL docs, as WSL 2 is available now

linuxgemini:
WSL 2 currently does not support USB and Serial.

Although you can get around with usbip-win; it requires a reconfigured kernel to be compiled.


iceman1001:
Thanks!

PR #932, opened 2020-08-31
Registry entries are not needed when calling SetConsoleMode

Gator96100:
The registry key at `HKEY_CURRENT_USER\Console\VirtualTerminalLevel` sets the global default, but calling the SetConsoleMode API with the `ENABLE_VIRTUAL_TERMINAL_PROCESSING (0x0400)` flag will enable ansi color support, regardless of the global default.

Tested on Windows 10 2004 and Windows 7. On Windows versions prior to Windows 10 Anniversary Update SetConsoleMode will fail and ansi color support is turned off.

### DIFF
--- a/client/src/proxmark3.c
+++ b/client/src/proxmark3.c
@@ -678,55 +678,8 @@ finish2:
     return ret;
 }
 
-// Check if windows AnsiColor Support is enabled in the registery
-// [HKEY_CURRENT_USER\Console]
-//     "VirtualTerminalLevel"=dword:00000001
-// 2nd Key needs to be enabled...  This key takes the console out of legacy mode.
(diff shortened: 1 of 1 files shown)


iceman1001:
@mwalker33   something to look into

mwalker33:
Seems like a good idea.  I originally was not trying to turn on support, simply check for support. 
If we can safely enable for that session, then it should not affect any settings the end user needs/wanted for other things (by default).

Did you try it with the changes but using windows terminal (the multi tab console terminal in windows 10)?   Windows terminal by default supports Ansi Colors, so as long as that still works... It sounds good.

Gator96100:
It does work on the Windows cmd.exe as well as the new Windows terminal/Power shell terminal

PR #938, opened 2020-09-03
Add standalone mode hf_aveful: Mifare ultralight read/simulation

aveao:
This PR adds `hf_aveful`, which is a standalone mode that aims to read and simulate Mifare Ultralights (and Ultralight clones) on the go, with the process for reading a card taking less than a second. This doesn't include EV1, C etc.

C is not really a language I use a lot, so there may be potential improvements that can be done to a bunch of the code. Input welcome :)

Right now I am only accounting for Ultralights/clones with 16 blocks/pages, though I've supplied `define` options, so users that want to modify the code to work with other stuff (My-d move, my-d NFC, and Mifare Ultralight Nanos etc) can easily do so.

If anyone's interested in using this but aren't using master due to MERGEHELL: I've confirmed that the code works fine on both master and the latest stable release at the time of me writing this (v4.9237).

### DIFF
--- a/armsrc/Standalone/Makefile.hal
+++ b/armsrc/Standalone/Makefile.hal
@@ -38,6 +38,9 @@ define KNOWN_STANDALONE_DEFINITIONS
 | HF_14ASNIFF     | 14a sniff to flashmem                  |
 | (RDV4 only)     |                                        |
 +----------------------------------------------------------+
+| HF_AVEFUL       | Mifare ultralight read/simulation      |
+|                 | - Ave Ozkal                            |
++----------------------------------------------------------+
 | HF_BOG          | 14a sniff with ULC/ULEV1/NTAG auth     |
 | (RDV4 only)     | storing in flashmem - Bogito           |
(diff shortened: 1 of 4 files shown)


armsrc/Standalone/Makefile.hal
```diff
@@ -59,10 +59,13 @@ define KNOWN_STANDALONE_DEFINITIONS
 | HF_YOUNG        | Mifare sniff/simulation                |
 |                 | - Craig Young                          |
 +----------------------------------------------------------+
+| HF_AVEFUL       | Mifare ultralight read/simulation      |
```
doegox, line 62: cf readme: Add your mode to the `Makefile.hal` help and modes list (alphabetically) ;)
same for the STANDALONE_MODES list.
thanks

aveao: Sorry, I missed that one somehow. Should be good now.

iceman1001:
Well done!
Its not that hard to do a standalone mode,   now you can start improving it

PR #942, opened 2020-09-04
Return the Mfc gen3 writer to Mater

Windslab:
The all bugs has fixed


Windslab:
@doegox Could you please help @iceman1001, mb he is busy. You can do an merge.

iceman1001:
Lets see if it works,  I will merge.
@doegox   you can delete that branch later,

 I noticed that native pm3 client commands of gen3 is being developed so until then this script fills a void

PR #945, opened 2020-09-05
Magic gen 3 cards support

McEloff:
There are three operations for magic gen3 cards:
- changing UID of card without changes of manufacturer block
- full overwrite manufacturer block
- locking further UID changes


iceman1001:
Excellent !

PR #952, opened 2020-09-09
Fixed appveyor CI

Gator96100:
Updated appveyor CI to use the intended way of launching ProxSpace. 
Tested with ProxSpace v3.5 and v3.6.


iceman1001:
Excellent!

PR #958, opened 2020-09-16
lf t55xx sniff command added

mwalker33:
Added lf t55xx sniff to allow the extraction of commands as if sent to a T55xx card.

lf t55xx sniff - will call lf sniff (using existing lf config settings) to sniff the data, then auto attempt decoding.
lf t55xx sniff b - will use the data in the graph buffer.
lf t55xx w <0 width> 1 <width> - Will allow setting of custom widths to assist in getting a good extraction.

example of use.
lf t55 sniff
- present the cloner, and write.
review data.
if it looks a little out, 
lf t55 sniff b w <new 0 bit width> <new 1 bit width>


iceman1001:
Perfect addition!

PR #964, opened 2020-09-18
Appveyor run WSL installation in parallel

Gator96100:
Running ProxSpace and WSL installation in parallel saves about 6 minutes of appveyor runtime.

Shall I add a separate job for WSL compilation as well? This would save about 8 minutes of appveyor runtime. I would start the WSL compilation job in the background and would start printing its output once the ProxSpace compilation job is finished.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -60,7 +60,8 @@ clone_script:
                 break
             }
             
-            Start-Sleep -s 1
+            Start-Sleep -s 5
+            Receive-Job -Job $WSLjob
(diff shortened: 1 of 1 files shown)


iceman1001:
Go for that parallel compilation !

Gator96100:
Ready to merge. Parallel compilation only saves about 2.5 minutes. Probably because we hit a cpu bottleneck on Appveyor.

iceman1001:
excellent!

PR #965, opened 2020-09-19
add ST25TA ndef protection logic

tcprst:
This adds the ability to disable/enable read and write protection on the ST25TA series of tags with a known write password

### DIFF
--- a/client/src/cmdhfst.c
+++ b/client/src/cmdhfst.c
@@ -52,6 +52,21 @@ static int usage_hf_st_ndef(void) {
     return PM3_SUCCESS;
 }
 
+static int usage_hf_st_protect(void) {
+    PrintAndLogEx(NORMAL, "\n Change R/W protection for NFC Data Exchange Format (NDEF)\n");
+    PrintAndLogEx(NORMAL, "Usage: hf st protect [h] p <pwd> <r|w> <t>");
+    PrintAndLogEx(NORMAL, "Options:");
+    PrintAndLogEx(NORMAL, "    h          : This help");
+    PrintAndLogEx(NORMAL, "    p <pwd>    : 16 byte write password");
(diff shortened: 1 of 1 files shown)


tcprst:
If this looks good I also have the change password stuff ready to submit as well

client/src/cmdhfst.c
```diff
@@ -52,6 +52,21 @@ static int usage_hf_st_ndef(void) {
...
+    PrintAndLogEx(NORMAL, "Usage: hf st protect [h] p <pwd> <r|w> <t>");
+    PrintAndLogEx(NORMAL, "Options:");
+    PrintAndLogEx(NORMAL, "    h          : This help");
+    PrintAndLogEx(NORMAL, "    p <pwd>    : 16 byte write password");
+    PrintAndLogEx(NORMAL, "    <r|w>      : Change (r)ead or (w)rite protection");
+    PrintAndLogEx(NORMAL, "    t          : Enable / Disable protection");
+    PrintAndLogEx(NORMAL, "                 0 = Disable (default)");
+    PrintAndLogEx(NORMAL, "                 1 = Enable");
+    PrintAndLogEx(NORMAL, "Examples:");
+    PrintAndLogEx(NORMAL, _YELLOW_("          hf st protect p 82E80053D4CA5C0B656D852CC696C8A1 r 0"));
```
iceman1001, line 66: A bit unclear instructions vs your example.
`<r|w>  <t>`

tcprst: Would you suggest something other than the `<t>`, or should I omit the `<t>` and just put the `<0|1>`?

doegox: we use <> for named params and [] for verbatim and optional, so I'd suggest
* p <pwd>
* r|w
* [0|1]

tcprst: Perfect! Thanks for the clarification, I'll make these changes

PR #969, opened 2020-09-19
add ST25TA password change logic

tcprst:
This commit adds the ability to change both read and write passwords on the ST25TA series of tags with a known write password


doegox:
thanks!

iceman1001:
Great!

PR #971, opened 2020-09-20
'hf 14b dump' without params now calls help and corrected a typo

cberetta:
I found out that if I use the command "hf 14b dump" without any parameters instead of getting the help like other commands it tries to dump to file. I corrected this behavior so that calling the command without parameter the help will be shown.
I fixed also a couple of typo in the help itself.


iceman1001:
Thanks!

PR #974, opened 2020-09-25
Removed an status check that was blocking command 'hf 14b dump', …

cberetta:
…added SwapEndian64 to generated filename to be consistent with UID

First I added a missing "SwapEndian64" in the filename generation for the DUMP command _(hf 14b dump f)_ as without it the UID was "inverted" for what I call the "Endian-Effect".

Secondly I remove a status check from the DUMP command itself. The status check was always failing in my testing as the command was always returning "06" instead of the expected "00". I tried removing this check and the DUMP was successful. I'm not 100% sure this is correct, so I hope someone could do some testing. I own only one Srix4k _(one with Chip: 00, SRIX4K (Special))_ so I cannot test the code with other models.


iceman1001:
Nice,  I see that the check should have looked for expected size instead.
I merge this and adapt

PR #975, opened 2020-09-26
White cloner pwd

mwalker33:
Add lf t55xx chk e <EM4100> option.  

Checks calculated password based on the EM4100 id from some white cloners formula by forum user paleopterix


iceman1001:
Nice,  but  why don't you use the fct I added for it? 

https://github.com/RfidResearchGroup/proxmark3/blob/master/common/generator.c#L418

mwalker33:
Like that ?

PR #988, opened 2020-10-03
hf st commands - now use cliparser

tcprst:
This commit changes CLI parsing for all `hf st` commands to use cliparser

Thanks @mwalker33 for the awesome cliparser doc!

### DIFF
--- a/client/src/cmdhfst.c
+++ b/client/src/cmdhfst.c
@@ -14,6 +14,7 @@
 #include "cmdparser.h"     // command_t
 #include "comms.h"         // clearCommandBuffer
 #include "cmdtrace.h"
+#include "cliparser.h"
 #include "crc16.h"
 #include "cmdhf14a.h"
 #include "protocols.h"     // definitions of ISO14A/7816 protocol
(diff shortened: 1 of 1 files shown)


client/src/cmdhfst.c
```diff
@@ -473,57 +452,72 @@ static int cmd_hf_st_ndef(const char *Cmd) {
...
 
     //Validations
 
-    if (state[2] == 0x00) {
-        PrintAndLogEx(WARNING, "Missing action (r)ead or (w)rite");
-        errors = true;
+    if (enable_protection && disable_protection) {
+        PrintAndLogEx(ERR, "Must specify either enable or disable protection, not both");
+        return PM3_EINVARG;
+    } else {
```
iceman1001, line 496: No need for `else`  if you return.

client/src/cmdhfst.c
```diff
@@ -473,57 +452,72 @@ static int cmd_hf_st_ndef(const char *Cmd) {
...
+        }
     }
-    if (pwdlen != 16) {
-        PrintAndLogEx(WARNING, "Missing 16 byte password");
-        errors = true;
+
+    if (read_protection && write_protection) {
+        PrintAndLogEx(ERR, "Must specify either read or write protection, not both");
+        return PM3_EINVARG;
+    } else {
```
iceman1001, line 508: No need for `else` if you return

iceman1001:
Awesome! 
Just two minor comments

client/src/cmdhfst.c
tcprst: The else is if the error condition is not met.  I guess I could refactor it and do it differently if you would prefer.

iceman1001: obviously but since you are _returning_ ,   there is no use to have a _else_  
So you can make less indentation,  less "nested if's"   making the code a bit easier to read.   
Not much of a refactor to talk about.

PR #995, opened 2020-10-04
lf keri clone, sim - now use cliparser

tcprst:


iceman1001:
cool,
question is if we should have two lit instead of a t type option.
```
--ms
--
instead of
-t x
```

tcprst:
I proposed that to @mwalker33 last night, and he said that there are apparently many different types with Keri tags and `ms` is just the first one outside of plain internal ID that has been added. His thought was that if more types are discovered and added the --type argument made more sense.

iceman1001:
I see,   however  ...waiting for new types to be added might take some time.  Its not like it happens often.
Not a important thing, so I merge instead :)

PR #997, opened 2020-10-05
Appveyor use build cache for ProxSace packages

Gator96100:
Using the Appveyor cache to speed up the build process. Successful builds will take about the same time (WSL updating takes time), but builds that fail on ProxSpace will result in an earlier Appveyor abortion.

Fixed some small bugs as well.


iceman1001:
Is it irony or not when appveyor tests fails on your PR?  ;)

PR #1001, opened 2020-10-10
Replace awk with sed for WSL

socram8888:
The WSL distribution of Debian uses mawk instead of gawk by default, which does not support matching into an array as it is currently being done for extracting the COM port from the device name.

This commit replaces awk by sed, which also simplifies the building of /dev/ttySx paths.

### DIFF
--- a/pm3
+++ b/pm3
@@ -156,7 +156,7 @@ function get_pm3_list_WSL {
     # Need to look for this first,  the call to Win32_serialport "crashes" then native bt serial port.  Don't ask why.
     #BT direct SERIAL PORTS (COM)
     if $FINDBTRFCOMM; then
-        for DEV in $(powershell.exe -command "Get-CimInstance -ClassName Win32_PnPEntity | Where-Object Caption -like 'Standard Serial over Bluetooth link (COM*' | Select Name" 2> /dev/null | sed -nr 's#.*\bCOM([0-9]+)\b.*#/dev/ttyS\1#p'); do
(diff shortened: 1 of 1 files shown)


socram8888:
Second commit adds a fallback if WSLENV is not properly set.

I wanna point out that my installation (clean from less than one month ago) does not have WSLENV set, and thus Path is not populated with any of Windows' entries. Thus for making usage much easier for people who don't feel like tinkering with system-wide variables, I've decided to hard-code a fallback which should work for most of the users out there.

iceman1001:
Since the script still uses awk to get into the WSL part,  I am curious how that works for you
```HOSTOS=$(uname | awk '{print toupper($0)}')```

**Comment by socram8888** [comment]
No no, the problem is not that awk does not work at all. The problem is that having a third argument in the match call:
`awk 'match($0,/COM([0-9]+)/,m){print m[1]}')`
Is a GNU extension, and thus not properly supported in mawk, the version that by default comes with Debian:
```
marcos@DESKTOP-5666HIJ:~$ realpath $(which awk)
/usr/bin/mawk
marcos@DESKTOP-5666HIJ:~$ echo "Serial (COM9)" | gawk 'match($0,/COM([0-9]+)/,m){print m[1]}'
9
marcos@DESKTOP-5666HIJ:~$ echo "Serial (COM9)" | mawk 'match($0,/COM([0-9]+)/,m){print m[1]}'
mawk: line 1: syntax error at or near ,
```

doegox:
sounds good to me, thanks

PR #1005, opened 2020-10-12
Data save

mwalker33:
fix data save -f <filename> where filename > 7 characters.

Looks like when casting from char * to int8_t * in CLIGetStrWithReturn causes cliparser to limit length to 7 bytes (i.e. an issue determining the actual buffer length).
changed to use CLIParamStrToBuf and not CLIGetStrWithReturn
Since data save needs a filename, I also adjusted the -f to be required.


iceman1001:
Nice,  doesn't we need to do this for all commands that uses cliparser and save a filename?

mwalker33:
There will be a chance, depending on where the cast is done.  if the array data type is uint8_t and not cast, its ok.  (the other option would be to not cast in the cliparser call, rather cast when used)

I wanted to have  a good look at the function CLIGetStrWithReturn and see how it works out the size (as its not passed in).  It may be better to pass in the max size.
e.g. CLIGetStrWithReturn(ctx, 2, (uint8_t *)filename, &fnlen);
where fnlen starts of being "size" then on return holds the length transfered.

PR #1010, opened 2020-10-13
32 sector print incomplete for Mifare 4K

xianglin1998:
Hi guys,  when I executed the "hf mf rdsc" command to read the 4K tag, I found some problems: the information of sector 32 was lost. I guess the sector size conversion error, and read the source code verification

image

image

In the mifare 4K datasheet, subscript counts from sector 0 up to sector 31 are 4 blocks and from sector 32 to end are 16 blocks.
https://www.nxp.com/docs/en/data-sheet/MF1S70YYX_V1.pdf

image

Fixed:
image


iceman1001:
Good catch!
It should use the fcts instead but I'll merge

PR #1014, opened 2020-10-17
Removing duplicating passwords

will-caruana:
Removing duplicating passwords:
aabbccdd
bbccddee
ccddeeff


iceman1001:
Good catch!

Issue #1012, opened 2020-10-14
Refresh `hf 14a info` with latest info from AN10834 2020


doegox:
https://www.nxp.com/docs/en/application-note/AN10834.pdf p7

iceman1001:
So @VortixDev 's PR is merged.  
That document you linked to has a nasty decision tree

iceman1001:
Closing since we have gotten better magic detection and a `hf mf info`

PR #1015, opened 2020-10-17
Improve MIFARE type identification

VortixDev:
Partially fulfils https://github.com/RfidResearchGroup/proxmark3/issues/1012 (does not utilise GetVersion)

I modified the `detect_nxp_card` function to utilise the information in AN10834 and reduce false positives. Some changes:

1. Identification of MIFARE Classic 2K
2. Identification of SmartMX
3. Identification of MIFARE DESFire Light
4. Identification of 256B DESFire EV1s
5. Identification of MIFARE Hospitality
6. More specificity and accuracy with MIFARE Plus identification

More accuracy is also obtained by using `select_status` to determine whether or not ATS is supported. Further, checks are designed to be specific: there are instances with the current version where more possible matches are reported than necessary. For example, the `sak & 0x08` check is a subset of `sak & 0x09`, yet both checks are performed independently of each other. If a card passes the MIFARE Mini check of `sak & 0x09`, it will also pass the more generic check of `sak & 0x08`, and will thus will also report as potentially being a Classic or Plus, despite it being identified as a Mini.

I followed the decision of the current version that cards identified as NTAG 4xx should be of type `MTDESFIRE`, and cards identified as TNP3xxx should be of type `MTCLASSIC`, though am unsure if this was intended.

In the future, an adjustment could be made to use GetVersion to identify exactly what card is being detected.

Here's a before/after of some tests (pre-update on the left, post-update on the right):
_MIFARE Classic 1K:_
MIFARE Classic 1K

_MIFARE Classic 4K:_
MIFARE Classic 4K

_MIFARE DESFire EV1 4K:_
MIFARE DESFire EV1 4K

_MIFARE Ultralight EV1:_
MIFARE Ultralight EV1

### DIFF
--- a/client/src/cmdhf14a.c
+++ b/client/src/cmdhf14a.c
@@ -1455,96 +1455,133 @@ typedef enum {
     MTOTHER = 32
 } nxp_mifare_type_t;
 
-// According to NXP AN10833 Rev 3.6 MIFARE Type Identification, Table  6
-static int detect_nxp_card(uint8_t sak, uint16_t atqa) {
+// Based on NXP AN10833 Rev 3.6 and NXP AN10834 Rev 4.1
+static int detect_nxp_card(uint8_t sak, uint16_t atqa, uint64_t select_status) {
     int type = MTNONE;
-
+    
     PrintAndLogEx(SUCCESS, "Possible types:");
-
-    if (sak == 0x00) {
-        printTag("NTAG 20x / 21x / 21x TT / I2C plus");
-        printTag("MIFARE Ultralight / C / EV1 / Nano");
-        type = MTULTRALIGHT;
-    }
-
-    if (sak == 0x01) {
-        printTag("TNP3xxx (Activision Game Appliance)");
-        type = MTCLASSIC;
-    }
-
-    if ((sak & 0x04) == 0x04) {
-        printTag("Any MIFARE CL1 / NTAG424DNA");
-        type |= MTDESFIRE;
-    }
-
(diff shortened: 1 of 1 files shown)


iceman1001:
The identification is a interesting mess to begin with.
I agree that GetVersion() is now something that should be incorporated,  which it is done in its following relevant sub-commands , like 
- hf mfu info
- hf mfp info
- hf mfdes info

The 14a info wasn't support to identify all aspects of the tech,  it was more to identify the ISO14443a standard ,  which means a lot of hits.  The problem is that we are lacking a 
- hf mf info 
and there lays the beginning of the mess you see today.

I am not sure I agree with one -tech - one line,  Main reason , we have sparse space available on screen sizes with ppl using the client on android phones,  so a compact design is what we prefer. 

With that said,
hf 14a reader  ->  gives a fast select / anticollision for 14a.
hf 14a info ->  should give  ISO14443a -2, -3, -4  info and preferable a clear indication on what type it might be.  On other commands I have started calling it  "identification".    That identification should be enough for user to execute the more detailed info commands for each different tech.   You see for instance those "hint" lines there to help out newbies.

iceman1001:
Thanks!
I merge it.  
Since it is better than before,  but still not ideal.

doegox:
@iceman1001 The Gen3 detection is now wrongly done against MFUL/NTAG as well
image

iceman1001:
Not ideal to check Gen3 against ntag/ul...

iceman1001:
pull and test again,  I pushed some fixes

PR #1018, opened 2020-10-17
ST25TA drop field & cliparser buffer fixes

tcprst:
After doing the NFC nail sticker mod I noticed some of the ST25 commands would leave the field on when reaching an error condition.


iceman1001:
Do you still need that +1 fix for cliparser?  I thought I fixed it today

tcprst:
Oh, let me test!

PR #1038, opened 2020-10-30
RFID Brute forcing lists

will-caruana:
Lists of most common English words and most common passwords along with two lists of ASCII numbers to hex and an additional iClass list that is based off patterns seen in other default  password lists.


will-caruana:
Mistakes were made

PR #1043, opened 2020-11-01
Add Gallagher key checking/KDF on MIFARE Desfire

NZSmartie:
Both `hf mfdes auth` and `hf mfdes chk` now support Key Diversification for AN10922 and as special treat, Gallagher issued cards.

For `hf mfdes auth`:
```
    -d, --kdf <kdf>                Key Derivation Function (KDF) (0=None, 1=AN10922, 2=Gallagher)
    -i, --kdfi <kdfi>              KDF input (HEX 1-31 bytes)
```

And for `hf mfdes chk`:
```
    -f, --kdf <kdf>                Key Derivation Function (KDF) (0=None, 1=AN10922, Gallagher)
    -i, --kdfi <kdfi>              KDF input (HEX 1-31 bytes)
```

> Please note the `-f` option for `chk` since `-d` is used to specify the dictionary input

Examples:
- `hf mfdes auth -a 2081f4 -m 3 -t 4 -d 2 -n 2 -k 00112233445566778899aabbccddeeff`
  Will diversify the key for key `2` on AID `2081F4` for Gallagher issued cards

- `hf mfdes chk -f 1 -i 00112233 -d mfdes_default_keys`
  Will read in all the default keys from the dictionary, and diversify them
  using AN10922 with the input data `00112233`

- `hf mfdes chk -f 2 -d mfdes_default_keys`
  Will read in all the default keys from the dictionary, and diversify them
  using AN10922 but with input data generated from the card's UID, AID and
  key number.

### DIFF
--- a/client/src/cmdhfmfdes.c
+++ b/client/src/cmdhfmfdes.c
@@ -83,6 +83,16 @@ typedef struct mfdes_data {
     uint8_t *data;
 } PACKED mfdes_data_t;
 
+typedef struct {
+    uint8_t isOK;
+    uint8_t uid[7];
+    uint8_t uidlen;
+    uint8_t versionHW[7];
+    uint8_t versionSW[7];
+    uint8_t details[14];
+} PACKED mfdes_info_res_t;
+
+
 typedef struct mfdes_value {
     uint8_t fileno;  //01
     uint8_t value[16];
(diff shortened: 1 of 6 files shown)


common/generator.c
```diff
@@ -423,6 +423,47 @@ uint32_t lf_t55xx_white_pwdgen(uint32_t id) {
...
+    // Verify the AppID is a valid Gallagher AppID
+    if ((aid & 0xF0FFFF) != 0x2081F4) {
+        if (g_debugMode) {
+            PrintAndLogEx(WARNING, " invalid Gallagher aid %06X", aid);
+        }
+        return PM3_EINVARG;
+    }
+
+    int len = 0;
+    // If the keyNo == 2, then omit the UID.
```
NZSmartie, line 444: ah woops, meant to be keyNo == 1

client/src/cmdhfmfdes.c
```diff
@@ -4433,6 +4477,17 @@ static int CmdHF14aDesChk(const char *Cmd) {
...
+
+    mfdes_info_res_t info = {0};
+    res = mfdes_get_info(&info);
+    if (res != PM3_SUCCESS) {
+        return res;
+    }
+    // TODO: Store this UID someowhere not global
+    memcpy(tag->info.uid, info.uid, info.uidlen);
+    tag->info.uidlen = info.uidlen;
+
```
NZSmartie, line 4490: I had to retrieve the card's UID here as it's not stored with the client in any global/local variable. I'm sure there's a better way to do this. but it works for now. though I do occasionally get a _Failed to Select card_

iceman1001:
Awesome!

PR #1051, opened 2020-11-08
T55 detect with wake

mwalker33:
This will try the lf t55 detect as before, then if that does not find the config block, will repeat but with the lf t55 wake called then sleep for 90ms.
This is to address the delay needed after power on if init-delay is on.
It also allows a single detect command to detect if AOR is active.


iceman1001:
Awesome!

PR #1054, opened 2020-11-10
bsd/gnu-safe file byte calc

jamiefiedler:
either needs to use the correct bsd/gnu flag for bytes via stat:
**gnu** `stat -c "%s"`
**bsd** `stat -f "%z"`

or can use slightly slower `wc -c` which should be safe across both.


iceman1001:
Nice catch!

PR #1057, opened 2020-11-12
waveshare: fix dimensions of 2.7" screen, test 1.54" screen

nyanotech:
The 2.7-inch e-ink module's canvas is actually in portrait, trying to display a 276-wide by 176-tall bmp yields an image that's glitched in the way you'd expect for a canvas that's sideways.

Additionally I have tried one of the 1.54" 200x200 screens, and it seems to work perfectly (including tri-color).


iceman1001:
Nice one!

PR #1061, opened 2020-11-15
Fix old UID display on hf mf csetuid

aveao:
This PR fixes the issue where hf mf csetuid always displays an old UID of 00 00 00 00.

Before:

(Ignore the print debugging)

After:


doegox:
thanks!

PR #1066, opened 2020-11-23
hf iclass encrypt/decrypt - now use cliparser

tcprst:


client/src/cmdhficlass.c
```diff
@@ -1049,74 +1010,82 @@ static int CmdHFiClassEView(const char *Cmd) {
     free(dump);
     return PM3_SUCCESS;
 }
-
 static int CmdHFiClassDecrypt(const char *Cmd) {
+    CLIParserContext *clictx;
+    CLIParserInit(&clictx, "hf iclass decrypt",
+                  "3DES decrypt data\n"
+                  "This is naive implementation, it tries to decrypt every block after block 6.\n"
```
iceman1001, line 1017: The spelling mistake was before but shouldn't it be  "This is a naive" ?

client/src/cmdhficlass.c
```diff
@@ -1049,74 +1010,82 @@ static int CmdHFiClassEView(const char *Cmd) {
...
+                  "hf iclass decrypt -d 1122334455667788 --key 000102030405060708090a0b0c0d0e0f");
 
-    bool errors = false;
-    bool have_key = false;
-    bool have_data = false;
+    void *argtable[] = {
+        arg_param_begin,
+        arg_str0("f", "file", "<filename>", "filename of dumpfile"),
+        arg_str0("d", "data", "<encrypted blk>", "3DES encrypted data"),
+        arg_str0(NULL, "key", "<transport key>", "3DES transport key"),
```
iceman1001, line 1030: keep the   "k"   short param for keys.

tcprst: we got rid of it in others so I figured we should do the same here.

client/src/cmdhficlass.c
tcprst: Yup, I'll fix that!

client/src/cmdhficlass.c
iceman1001: did we?  in iclass commands?

iceman1001:
Great!

PR #1067, opened 2020-11-24
Initial T55xx guide

mwalker33:


iceman1001:
Awesome!
..and a line in changelog.md
...and link to it in the README.md.   I suggest in the column "use of proxmark"  as  "T55xx Guide"

PR #1070, opened 2020-11-25
hf iclass dump, restore, rdbl, wrbl cliparserization and argument normalization

tcprst:
hf iclass dump, restore, wrbl, & rdbl have been cliparserized.  Start block and end block commands now use `--first` and `--last` arguments and expect integers instead of 1 byte hex.  Blocks are now printed with decimal representation to match.

A new key index argument has been added when referencing keys in memory.

### DIFF
--- a/client/src/cmdhficlass.c
+++ b/client/src/cmdhficlass.c
@@ -71,26 +71,6 @@ static int usage_hf_iclass_sim(void) {
     PrintAndLogEx(NORMAL, "");
     return PM3_SUCCESS;
 }
-static int usage_hf_iclass_writeblock(void) {
-    PrintAndLogEx(NORMAL, "Write data to a iCLASS tag\n");
-    PrintAndLogEx(NORMAL, "Usage:  hf iclass wrbl b <block> d <data> k <key> [c|e|r|v]\n");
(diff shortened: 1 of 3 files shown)


client/src/cmdhficlass.c
```diff
@@ -1528,10 +1499,10 @@ static int CmdHFiClassDump(const char *Cmd) {
         .req.use_credit_key = false,
         .req.use_replay = use_replay,
         .req.send_reply = true,
-        .req.do_auth = auth,
+        .req.do_auth = have_debit_key,
```
iceman1001, line 1502: I think you missed the user-case  credit key.   The old auth was true for either debit or credit

iceman1001:
Excellent!

client/src/cmdhficlass.c
tcprst: ah, okay I can rename it back.  I was just trying to make things follow similar naming conventions as auth was generic

tcprst: okay, I see what I did there.  Should be fixed in the latest commit.

tcprst:
Just converted `hf iclass rdbl` and added to this PR as well.

PR #1072, opened 2020-11-28
hf iclass chk, managekeys, calcnewkey, lookup, sim - now use cliparser

tcprst:
This wraps up the `hf iclass` cliparserization.


iceman1001:
Awesome!

PR #1076, opened 2020-11-28
hf lto dump, restore, info, rdbl, wrbl - now use cliparser

tcprst:
The `hf lto` commands are now cliparserized and tested


tcprst:
Let me fix this, I must've done something stupid

tcprst:
Fixed!

PR #1079, opened 2020-11-30
MFU emulation now supports automatic exit after <num> blocks read.

cyberpunk-re:
Added `hf mfu sim t 7 n <numreads>` to allow the ultralight emulation to exit after `<numreads>` blocks. The approach was similar to the one that had been implemented for mifare classic tags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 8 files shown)


iceman1001:
May I ask what is the purpose of exiting a simulation after x number of reads on a MFU?

iceman1001:
You will need to solve the conflicts with yr PR.  I suggest you remove all "white space" changes,  and only have your actual changes in the PR.   We make a  `make style` PR kind of often.

iceman1001:
I get it,  I did something similar for MFC way back.
Nowdays we have  the usb cmd "break loop",   which should break out of the simulations.
but if you trigger an exit lets say after 4 reads, you know for sure the simulation has stopped.

iceman1001:
I will merge it after the CI has run,  Thanks!

iceman1001:
Thanks!

PR #1084, opened 2020-12-01
Iso15 sim

cyberpunk-re:
Fix `hf 15 sim` - Added basic responses to GET_SYSTEM_INFO and READBLOCK (returns empty block) requests to fix the currently broken iso15693 tag simulation.
A proper tag dump emulation should be implemented, but you can now successfully simulate the UID an ISO 15693 tag.


iceman1001:
Nice one!

That test file doesn't look like it belongs in the PR:

cyberpunk-re:
> That test file doesn't look like it belongs in the PR:

My bad. Test file removed!

iceman1001:
Awesome!

PR #1087, opened 2020-12-05
Initial commit for em4x70 support.

cmolson:
Initially, I only have an em4170 variant used for car transponders.
Also known as the ID48 and possibly Magic II.

This first feature is to read the tag id and user memory 1 and 2.

### DIFF
--- a/armsrc/em4x70.c
+++ b/armsrc/em4x70.c
@@ -488,12 +488,10 @@ static bool find_EM4X70_Tag(void) {
 
 static int em4x70_receive(uint8_t *bits) {
 
-    bool bbitchange = false;
     uint32_t pl;
     int bit_pos = 0;
+    uint8_t edge = 0;
 
-    // Set first bit to a 1 for starting off corectly
-    bits[0] = 1;
     
     bool foundheader = false;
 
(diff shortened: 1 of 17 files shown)


client/src/cmdlfem.c
```diff
@@ -25,6 +27,7 @@ static command_t CommandTable[] = {
     {"410x",  CmdLFEM410X,  AlwaysAvailable, "EM 4102 commands..."},
     {"4x05",  CmdLFEM4X05,  AlwaysAvailable, "EM 4205 / 4305 / 4369 / 4469 commands..."},
     {"4x50",  CmdLFEM4X50,  AlwaysAvailable, "EM 4350 / 4450 commands..."},
+    {"4x70",  CmdLFEM4X70,  AlwaysAvailable, "EM 4170 /4070 commands..."},
```
iceman1001, line 30: swap order,  and space

client/src/cmdlfem410x.c
```diff
@@ -10,6 +10,7 @@
 
 #include "cmdlfem410x.h"
 #include "cmdlfem4x50.h"
+#include "cmdlfem4x70.h"
```
iceman1001, line 13: I don't think it belongs here

client/src/cmdlfem4x70.c
```diff
@@ -0,0 +1,151 @@
...
+
+    // envoke reading of a EM4x70 tag which has to be on the antenna because
+    // decoding is done by the device (not on client side)
+
+    bool errors = false;
+    uint8_t cmdp = 0;
+
+    em4x70_data_t etd = {0};
+
+    while (param_getchar(Cmd, cmdp) != 0x00 && !errors) {
```
iceman1001, line 96: use the cliparser

include/em4x70.h
```diff
@@ -0,0 +1,18 @@
...
+//-----------------------------------------------------------------------------
+// Low frequency EM4x70 structs
+//-----------------------------------------------------------------------------
+
+#ifndef EM4X70_H__
+#define EM4X70_H__
+
+typedef struct {
+    bool parity;
+} PACKED em4x70_data_t;
```
iceman1001, line 16: packed struct with a bool??

client/src/cmdlfem410x.c
cmolson: Right!

include/em4x70.h
cmolson: I sort of copied the em4x50_data_t struct, I removed the items I wasn't using yet. I have added more items for the next pull request that includes writing support.

I'll remove the PACKED attribute in this push.

iceman1001: It just didn't make any sense to have an struct with a bool inside.  So its for next PRs  ok.

armsrc/em4x70.c
```diff
@@ -0,0 +1,594 @@
...
+    int num_bytes = len / 8; // We should have a multiple of 8 here
+
+    uint8_t bytes[8];
+
+    for(int i=0;i<num_bytes;i++) {
+        bytes[i] = bits2byte(bits, 8);
+        bits+=8;
+        Dbprintf("Read: %02X", bytes[i]);
+    }
+}*/
```
iceman1001, line 417: remove custom debug code.
there exists already a few  (see appmain.c)

iceman1001 (changes requested):
only convert to cliparser and remove your custom debug stuff and it looks fine

iceman1001:
And you will need to resolve the conflicting file

cmolson:
I will try and add the cliparser change today as well. Thanks!

I named this em4x70 as the transponder/tag I am using is very close to that, with the only difference from what I can see is in the parity of the commands. I will be continuing to update this and have a few more PR's to submit once this large one is in place.

The transponder/tag I am using is referred to ID48 in the car world.

iceman1001:
yeah,  we might need to rename it later on but for now it will do.
You should mention ID48 inside the help text of your `info` command in order to make things clear.

cmolson:
I've updated to use the cliparser. I also added some text about ID48 automotive transponders to the help text.

There was also a small bugfix to the regular EM4x70 devices that require parity bit in the command. We need to drop the most significant bit of the command before adding a parity so that the entire length remains 4 bits.

iceman1001:
As a initial PR,  its looks ok,   most likely some minor things but we sort it out.   There was a spelling mistake in the help text among others.

Thanks for the PR and the new supported type!

Issue #844, opened 2019-04-19
"t55xx config" => recompute block0


doegox:
Recompute block0 when config is changed rather that just wiping block0 info

Issue #845, opened 2020-07-11
"t55xx detect": check coherence with block0


doegox:
See if data read from block0 matches the auto config detection

PR #1090, opened 2020-12-07
Fix issue #844 - "t55xx config" => recompute block0 

cyberpunk-re:
`lf t55xx config` now recomputes block0 instead of resetting to 0.
Issue #844 may be closed after merge.


mwalker33:
Dont act on this, its more a comment and Iceman can give his view latter.

My understanding is the Block0: was what was "detected/read" from the card.  If custom settings are made to allow reading of the card, then we may or may not know the actual block 0 data, just the bits that allowed what we think is correct.

As such, do we need to "tag" the Block0: as being from the card read OR from adjustments to the config (which may or may not truly reflect the block0 on the card?

cyberpunk-re:
I see what you mean, but my take on it is that if we're setting the parameters manually, then the 'user configured' settings should be reflected as the hypothetically correct block 0. 

This is actually helpful since if you then go read block 0 and the settings are correct, the contents read from the card should match the computed block 0 (at least for the settings that were defined by the user). If there is a mismatch for these fields then something is wrong and the decoding will mismatch. 

In a way, you could even use this scheme as an auto-detection method. Here is an example:

**Correct settings:** (block 0 will match the one read from tag)
```
[usb] pm3 --> lf t55xx config b 50 d FSK2a o 33 i 1
(...)
[=]      Block0         : **0x00107060**
[=]      Downlink Mode  : default/fixed bit length
[=]      Password Set   : No
```

```
[usb] pm3 --> lf t55xx read b 0 (block 0 will be different)
(...)
[+]  00 | **00107060** | 00000000000100000111000001100000 | ..p`
```

**Incorrect bitrate:**
```
[usb] pm3 --> lf t55xx config b 40
(...)
[=]      Block0         : **0x000C7060**
[=]      Downlink Mode  : default/fixed bit length
[=]      Password Set   : No
```

```
[usb] pm3 --> lf t55xx read b 0
(...)
[+]  00 | **C0000303** | 11000000000000000000001100000011 | ....
```

Of course, another take would be to simply remove the block 0 print out on the `lf t55xx config` output instead of zeroing it out.

mwalker33:
The idea of setting the config by hand is to help get access to the card.  If you can detect you should be using that.

Also you have things like X-Mode where for example, the data rate will be represent completely different, yet to read the card, both may work.

Don't get me wrong.  When we need to set things by hand, it would be handy to know what the config "could/should" look like; as we can then store for future use.

Maybe, have "green" for as read from card. and "yellow" for modified by hand ?  (just thinking).

cyberpunk-re:
I agree you should stick to `detect` if that works. If it doesn't, knowing what should be in block0 is a quick way to check if the manual settings are correct. I believe it would be even better for config to attempt to read block 0 and warn the user if there is a mismatch (comparing the expected fields against both compatibility and x-mode config mappings).

The color coding is also an interesting idea. Let me know what you thinks its best. The idea is naturally to have something that can be useful and avoid misleading the user.

iceman1001:
....   just red/green will be hard for a user to know that its card read or user set...   
I think in that case,  we also need a "(tag read)"  "(user set)"  strings next to it.  In order to make it super clear.

cyberpunk-re:
Lets make it both ways then. `lf t55xx config` now prints source of block 0 data next to it (also color-coded):

(Unknown) - in red if yet unset
(Auto detect) - in yellow if auto-detected
(User set) - in yellow if user-defined with `lf t55xx config`
(Tag read) - in green if read from tag

Currently, the last case is non-existent but could be useful if we check block 0 during auto-detect in the future (issue #845 ?).

Let me know if you want any change.

iceman1001:
If you resolve the conflicting changelog ,  I can go a head and merge afterwards.

cyberpunk-re:
done

iceman1001:
Great!

PR #1092, opened 2020-12-10
Fix mistaken MIFARE increment and decrement constants

VortixDev:
The increment and decrement commands have their values mixed up - decrement is 0xC0 and increment is 0xC1. This is stated correctly in the comment of `include/protocols.h`, but there seems to have been a mistake when setting the constants. Although these constants are used in a few places, the only place it appears that this mix-up has an effect is in command annotations:

Trace


iceman1001:
Good catch!

PR #1093, opened 2020-12-11
Rawdemod parsing fixes

cyberpunk-re:
fix: `data rawdemod` parsing bug that prevented selection of psk2 modulation
Also rawdemod for psk was ignoring user-defined clock values (autodetection always took precedence).


iceman1001:
Nice catches

PR #1096, opened 2020-12-12
EM4x70 Unlock support

cmolson:
Support for unlocking a tag by providing the pin.
I had to rename the parity command-line switches, as I think 'p' makes more sense for the password/pin to unlock the tag.

### DIFF
--- a/client/src/cmdlfem4x70.c
+++ b/client/src/cmdlfem4x70.c
@@ -86,12 +86,12 @@ int CmdEM4x70Info(const char *Cmd) {
                   "  ID48 does not use command parity (default).\n"
                   "  V4070 and EM4170 do require parity bit.",
                   "lf em 4x70 info\n"
-                  "lf em 4x70 info -p -> adds parity bit to command\n"
+                  "lf em 4x70 info -x -> adds parity bit to command\n"
                 );
 
     void *argtable[] = {
         arg_param_begin,
-        arg_lit0("p", "parity", "Add parity bit when sending commands"),
+        arg_lit0("x", "parity", "Add parity bit when sending commands"),
         arg_param_end
     };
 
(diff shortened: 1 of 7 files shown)


client/src/cmdlfem4x70.c
```diff
@@ -86,12 +86,12 @@ int CmdEM4x70Info(const char *Cmd) {
...
                   "  V4070 and EM4170 do require parity bit.",
                   "lf em 4x70 info\n"
-                  "lf em 4x70 info -p -> adds parity bit to command\n"
+                  "lf em 4x70 info -x -> adds parity bit to command\n"
                 );
 
     void *argtable[] = {
         arg_param_begin,
-        arg_lit0("p", "parity", "Add parity bit when sending commands"),
+        arg_lit0("x", "parity", "Add parity bit when sending commands"),
```
iceman1001, line 94: use NULL instead of "x"  and I suggest shorten "parity" to "par"  
 x is used as  "eXecute attack" in other parts of the client.

client/src/cmdlfem4x70.c
```diff
@@ -181,10 +181,70 @@ int CmdEM4x70Write(const char *Cmd) {
...
 }
 
+int CmdEM4x70Unlock(const char *Cmd) {
+
+    // send pin code to device, unlocking it for writing
+    em4x70_data_t etd = {0};
+
+    CLIParserContext *ctx;
+
+    CLIParserInit(&ctx, "lf em 4x10 unlock",
```
iceman1001, line 191: you should replace to correct type throughout the file.

client/src/cmdlfem4x70.c
cmolson: Great suggestion, thanks!

iceman1001:
Great!

PR #1097, opened 2020-12-12
EM4x70 Authentication Support

cmolson:
Last of the primitive commands this tag supports.

### DIFF
--- a/client/src/cmdlfem4x70.c
+++ b/client/src/cmdlfem4x70.c
@@ -252,14 +252,14 @@ int CmdEM4x70Auth(const char *Cmd) {
     CLIParserInit(&ctx, "lf em 4x70 auth",
                   "Authenticate against an EM4x70 by sending random number (RN) and F(RN)\n"
                   "  If F(RN) is incorrect based on the tag crypt key, the tag will not respond",
-                  "lf em 4x70 auth -r 11223344556677 -f 11223344\n"
+                  "lf em 4x70 auth --rnd 11223344556677 --frn 11223344\n"
                  );
 
     void *argtable[] = {
         arg_param_begin,
-        arg_lit0("x", "par", "Add parity bit when sending commands"),
-        arg_str1("r", "rnd", "<hex>", "Random 56-bit"),
-        arg_str1("f", "frn", "<hex>", "F(RN) 28-bit as 4 hex bytes"),
+        arg_lit0(NULL, "par", "Add parity bit when sending commands"),
+        arg_str1(NULL, "rnd", "<hex>", "Random 56-bit"),
+        arg_str1(NULL, "frn", "<hex>", "F(RN) 28-bit as 4 hex bytes"),
         arg_param_end
     };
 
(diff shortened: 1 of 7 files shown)


client/src/cmdlfem4x70.c
```diff
@@ -240,11 +240,76 @@ int CmdEM4x70Unlock(const char *Cmd) {
...
+                  "Authenticate against an EM4x70 by sending random number (RN) and F(RN)\n"
+                  "  If F(RN) is incorrect based on the tag crypt key, the tag will not respond",
+                  "lf em 4x70 auth -r 11223344556677 -f 11223344\n"
+                 );
+
+    void *argtable[] = {
+        arg_param_begin,
+        arg_lit0("x", "par", "Add parity bit when sending commands"),
+        arg_str1("r", "rnd", "<hex>", "Random 56-bit"),
+        arg_str1("f", "frn", "<hex>", "F(RN) 28-bit as 4 hex bytes"),
```
iceman1001, line 262: "f " is used for  filename ,   so I suggest use NULL,     
"r"  usually indicate reverse..  
"x"  was the same comment as last PR,

I suggest you remove all one letter params with NULL.

client/src/cmdlfem4x70.c
```diff
@@ -240,11 +240,76 @@ int CmdEM4x70Unlock(const char *Cmd) {
...
+        PrintAndLogEx(FAILED, "F(RN) length must be 4 bytes instead of %d", frnd_len);
+        return PM3_EINVARG;
+    }
+
+    clearCommandBuffer();
+    SendCommandNG(CMD_LF_EM4X70_AUTH, (uint8_t *)&etd, sizeof(etd));
+
+    PacketResponseNG resp;
+    if (!WaitForResponseTimeout(CMD_LF_EM4X70_AUTH, &resp, TIMEOUT)) {
+        PrintAndLogEx(WARNING, "Timeout while waiting for reply.");
```
iceman1001, line 293: How is it set up on device on failure?  or timeouts,    does the device drop the field?  
because if not,  this will lead the device to become unresponsive if user starts issuing new commands when this one timeouts.

cmolson: I was trying to follow what the EM4x50 did where on-device there are timeouts during all the loops and limits setup to prevent getting stuck. When we fail to find a tag or perform the command it should call lf_finalize() before returning. I will review the on-device code for any potential areas it could get stuck (loops without a timeout).

So far no commands on the device should hit the client timeout, the device should be timing out first and return "Failed".

In a previous PR when adding write support, I fixed an issue where the receive function could get stuck if the tag is removed.

iceman1001: good,   or we need to loop in the client if it is enforced that the device must send an answer to this command.

iceman1001:
Let me know if you are ready so I can merge.

client/src/cmdlfem4x70.c
cmolson: All commands must return a success or failure response. So if there is no response it should be considered a bug in the em4x70 device code.

If I switch to "WaitForResponse" without the timeout, there will be a warning printed that pressing the button will cancel the operation, which isn't true in this case. I could perhaps update the error message to indicate the device may now be unresponsive?

I'm not sure the best option, in both cases I think the device would be unresponsive, but that could be true of a bug anywhere on the device I believe. I believe the watchdog would restart the device in this case?

iceman1001: making it fail safe on device side,  so it doesn't get stuff on device side in a loop.    
We already have a timeout on client side now,

cmolson: Sounds good, that should be how it works now. The device-side loops do have timeouts that should then cause us to return an error in the case of no tag or weak tag signal etc.

cmolson:
It is ready now.

iceman1001:
Great!

PR #1098, opened 2020-12-12
Reworked hf 14a apdufuzz (now apdufind).

ikarus23:
* Renamed to apdufind. Fuzzing is something different than this simple enumeration.
* Removed Le as parameter. This is just the maximum response bytes expected. Not much to find here with this simple tool.
* Sweep through all values, even if a start value was given (don't stop at 0).
* Changed sweep oder to INS->P1->P2->CLA. This way instructions are more quickly found.
* Show response data, if there are any (hex & ascii).
* Retry command with Le = 0x0100 (extended length APDU) if there was a length error return code.
* Improved the output with general information.
* Fixed bug: It's now possible to run the command without parameters.


iceman1001:
Awesome!

PR #1103, opened 2020-12-14
APDU finder: Reset the tag every x seconds (default is 5 minutes).

ikarus23:


iceman1001:
Isn't the activate_field always true?    If so the field is dropped and turned on each iteration.

ikarus23:
So far, yes. `activate_field` is always true. But `keep_field_on` is also true, so I figured if the field is still on, `activate_field` is ignored (no need to activate it.

iceman1001:
You need to follow the code paths...

ikarus23:
So what you are tying to say is: "no, the filed is deactivated before it is reactivated if `activate_filed` is true."

ikarus23:
Should be fixed.

PR #1104, opened 2020-12-14
em4x70: Use fixed threshold for high/low

cmolson:
I am basing this commit on tests using my PM3 easy and PM3RDV4.

I have found that reading/writing is much more reliable with a fixed level for the high/low values. With these updated settings I can read/write on the PM3 easy by resting the tag on the coil, or lifting it up about 1CM. With the previous auto-thresholds I had to carefully hold the tag maybe half a centimeter off the coil in a very specific location.

I plan to test using the RDV4 and the new updated antennas, as well as the larger LF antenna kit for it once I receive them. Until then, these are the best values I have found to use for making reading/writing the most reliable.

### DIFF
--- a/armsrc/em4x70.c
+++ b/armsrc/em4x70.c
@@ -55,14 +55,17 @@ static bool command_parity = true;
 #define EM4X70_COMMAND_WRITE                0x05
 #define EM4X70_COMMAND_UM2                  0x07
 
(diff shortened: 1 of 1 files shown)


iceman1001:
When you get the new antennas you can adapt to make the best signal / distance reading / writing

PR #1105, opened 2020-12-15
eMRTD: Bugfixes & basic cleanup

aveao:
Didn't expect to have to send this in so soon, but apparently there were some crashes on some versions of ubuntu, whoops. [I had only tested on my arch setup where it worked just fine](https://elixi.re/i/worksonmymachine.png).

This PR resolves those, and does some of the recommended cleanup of the code. More will happen over future PRs.

This one was confirmed working on WSL1, Proxspace, Ubuntu and Arch.


iceman1001:
Hard to tell the fix with all the renaming :)

PR #1116, opened 2020-12-17
4x50 eview

tharexde:
Added  lf em 4x50 eview - show uploaded EM4x50 data in emul memory


iceman1001:
Great!

doegox:
@tharexde Is it intentional that your standalone mode was deleted by this PR ?

tharexde:
Yes. The standalone mode would not have worked, because the file lf-tharexde.c was uploaded by mistake in a former PR. I do have a local standalone mode but it is not ready to be submitted. It may be an upcoming PR.

Issue #1117, opened 2020-12-17
eMRTD Checklist, issues and feature requests


aveao:
This issue is intended to be a meta progress tracker and discussion space for the `hf emrtd` commands.

# Checklist

- [x] ISO 14a/14b support
- [x] Non-BAC passport support
- [x] BAC passport support
- [x] `hf emrtd dump`
- [x] `hf emrtd info` that displays basic info (perhaps just based on EF_DG1)
- [x] `hf emrtd dump` dumping more detailed data from the files by parsing them, such as extracting the JPG and the cert file
- [x] Dumping EF_CardAccess when doing `hf emrtd dump`, as suggested by doegox
- [x] Automatically changing document number to uppercase and checking for length of DOB/Expiry, as suggested by doegox
- [x] `hf emrtd info` that displays extended info (EF_DG11, EF_DG12)
- [x] Ability to view info from a dump folder, as suggested by doegox
- [x] Ability to provide full MRZ line as option, as alternative to doc_nr/birth_date/exp_date (only on passports/TD3)
- [x] File hash checks based on EF_SOD
- [x] Ability to `hf emrtd dump` to a different folder, as suggested by doegox
- [x] `hf emrtd info` that displays a GUI with extended info (EF_DG1, EF_DG2, EF_DG5, EF_DG11, EF_DG12)
- [x] PACE support
- [ ] Bruteforce support (might be interesting to look at bruteforcing CAN when we add PACE support)
- [ ] Various vulnerability checks as suggested by doegox
- [x] Cert info display
- [ ] Cert verification as suggested by iceman

# Test results

## Working

- Turkey IDs (14a)
- Turkey ePassports (14a, 14b)
- Sweden ePassports (14b)
- Denmark ePassports (2020, 14b)
- Lithuania ePassports (14a)
- United States ePassports (14a)
- Pre-Brexit United Kingdom ePassports (14a)
- Canada ePassports (14a)
- Ukraine ePassports (14a)
- Serbia ePassports (14a)
- Belgium ePassports (14b)
- old Belgium ePassports without BAC (14a)
- Greek ePassports
- Russian ePassports (14b)
- Dutch ePassports (14a)
- Slovenia ePassports (14a)
- Taiwan ePassports (14a)

## Not working

- Phillipines ePassports (14b) -> Can't read, external auth results in `[!!] APDU: Small APDU response. Len=0`. ReadID android can read it, says BAC and AA are available, uses BAC.
[...]

PR #1119, opened 2020-12-18
eMRTD additions

aveao:
This PR:

- Dumps EF_CardAccess when doing hf emrtd dump
- Automatically changes document number to uppercase and checks for length of DOB/Expiry to ensure that stuff that has no chance of working will not be sent to the eMRTD

More information at #1117.

### DIFF
--- a/client/src/cmdhfemrtd.c
+++ b/client/src/cmdhfemrtd.c
@@ -1298,6 +1298,22 @@ static void text_to_upper(uint8_t *data, int datalen) {
     }
 }
 
+static bool validate_date(uint8_t *data, int datalen) {
+    // Date has to be 6 chars
+    if (datalen != 6) {
+        return false;
+    }
+
+    // Check for valid date and month numbers
+    char temp[4] = { 0x00 };
+    memcpy(temp, data + 2, 2);
+    int month = (int) strtol(temp, NULL, 10);
+    memcpy(temp, data + 4, 2);
+    int day = (int) strtol(temp, NULL, 10);
+
+    return !(day <= 0 || day > 31 || month <= 0 || month > 12);
+}
+
 static int cmd_hf_emrtd_dump(const char *Cmd) {
     CLIParserContext *ctx;
     CLIParserInit(&ctx, "hf emrtd dump",
@@ -1333,8 +1349,8 @@ static int cmd_hf_emrtd_dump(const char *Cmd) {
     if (CLIParamStrToBuf(arg_get_str(ctx, 2), dob, 6, &slen) != 0 || slen == 0) {
         BAC = false;
     } else {
-        if (slen != 6) {
-            PrintAndLogEx(ERR, "Date of Birth length is incorrect, cannot continue.");
+        if (!validate_date(dob, slen)) {
+            PrintAndLogEx(ERR, "Date of birth date format is incorrect, cannot continue.");
             PrintAndLogEx(HINT, "Use the format YYMMDD.");
             return PM3_ESOFT;
         }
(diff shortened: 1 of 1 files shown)


client/src/cmdhfemrtd.c
```diff
@@ -1296,6 +1291,15 @@ int infoHF_EMRTD(char *documentnumber, char *dob, char *expiry, bool BAC_availab
...
 
+static void text_to_upper(uint8_t *data, int datalen) {
+    // Loop over text to make lowercase text uppercase
+    for (int i = 0; i < datalen; i++) {
+        if ('a' <= data[i] && data[i] <= 'z') {
+            data[i] -= 32;
+        }
+    }
+}
+
```
iceman1001, line 1302: there is a  strup function to use...

aveao: I checked the code and couldn't find something like that, where is it?

client/src/cmdhfemrtd.c
```diff
@@ -1329,10 +1334,22 @@ static int cmd_hf_emrtd_dump(const char *Cmd) {
     
     if (CLIParamStrToBuf(arg_get_str(ctx, 2), dob, 6, &slen) != 0 || slen == 0) {
         BAC = false;
-    } 
+    } else {
+        if (slen != 6) {
```
iceman1001, line 1338: maybe not only check len,  but also if  MM is <=12 and DD <=31  (generic naive test)
or have  a array with MM, maxdd and make a lookup..

aveao: Will do.

client/src/cmdhfemrtd.c
doegox: toupper

aveao:
Resolved both issues. Input welcome on how I could handle that date thing better fwiw, strtol might not be ideal.

aveao:
The EF_CardAccess dump isn't working, will sort that out and then unmark as draft.

aveao:
Fixed, should be ready to re-review.

I've also added in some code improvements recommended by @ceres-c, thank you so much for all the help.

client/src/cmdhfemrtd.c
```diff
@@ -711,18 +710,17 @@ static bool emrtd_select_and_read(uint8_t *dataout, int *dataoutlen, const char
...
-    for (int i = 0; i < file_length - 6; i++) {
-        if ((file_contents[i] == 0xFF && file_contents[i + 1] == 0xD8 && file_contents[i + 2] == 0xFF && file_contents[i + 3] == 0xE0) ||
-            (file_contents[i] == 0x00 && file_contents[i + 1] == 0x00 && file_contents[i + 2] == 0x00 && file_contents[i + 3] == 0x0C && file_contents[i + 4] == 0x6A && file_contents[i + 5] == 0x50)) {
-            datalen = file_length - i;
-            memcpy(data, file_contents + i, datalen);
+    // Note: Doing file_length - 6 to account for the longest data we're checking.
+    for (offset = 0; offset < file_length - 6; offset++) {
+        if ((file_contents[offset] == 0xFF && file_contents[offset + 1] == 0xD8 && file_contents[offset + 2] == 0xFF && file_contents[offset + 3] == 0xE0) ||
+            (file_contents[offset] == 0x00 && file_contents[offset + 1] == 0x00 && file_contents[offset + 2] == 0x00 && file_contents[offset + 3] == 0x0C && file_contents[offset + 4] == 0x6A && file_contents[offset + 5] == 0x50)) {
+            datalen = file_length - offset;
```
iceman1001, line 723: Use memcmp  fct ?

aveao: Considered that, would likely have a slightly higher overhead, but I'll still do it for better visibility later today.

PR #1122, opened 2020-12-18
hf emrtd info: EF_DG11 parsing + Rework to make it easier to write support to read external files

aveao:
This PR adds EF_DG11 parsing to `hf emrtd info`, and fancy headers for files (with color, as the iceman tax):

The compat chart looks like this for EF_DG11:

I've also quickly gone over the code to make it possibly easier to read dumps from files. Wanted to push this out ASAP as doegox expressed interest in looking into that (and as I wanted to avoid merge conflicts), so this PR comes without any coverity fixes.

---

For more info on this PR series, see #1117


doegox:
nice!

PR #1139, opened 2020-12-27
apdufind: added option to skip instrcutions.

ikarus23:
(This helps if a tag might be locked permanently with a certain instruction. You can just skip it.)


ikarus23:
Did the `make style` break the commands.md documentation?

iceman1001 (changes requested):
Exclude the helptext changes for epa in this commit.

PR #1142, opened 2020-12-29
eMRTD support 3/?

aveao:
Big PR with lots of changes, sorry!

Changelog:

- New Feature: Partial EF_SOD parsing (rather unclean code-wise sadly)
- New Feature: `hf emrtd info` now parses and verifies hashes of files. The hash part is **NOT** verified against the signature in EF_SOD. The certificate in EF_SOD is **NOT** verified against authorities' signing certs from ICAO. Both of these are future goals.
- Improvement: emrtd_lds_get_data_by_tag improvements
- Improvement: Note that -m option is for passports (TD3) only
- Improvement: Introduce a new flag for files: EAC. While eMRTDs aren't required to enforce EAC for sensitive biometrics (fingerprints, iris scans etc), they are enforced on every eMRTD that I saw. The previous flag of PACE was incorrect for these in general. We may want to introduce a new flag to try to dump these files too. See ICAO 9303 p11 section 7 and TR-03110-1 for more info.
- Refactor: Change of a lot of ints ot size_ts. Unsure if I should also change the for loops that involve the returned size values, advice welcome.
- Refactor: Use memcmp for determining if the file is a JPEG or a JPEG2K. I did a bit of a hack to keep overhead low.

---

Opinions requested before merge:

- I'm not super happy with how I store the read hashes (`dg_hashes`).
- ~~or how I manage the sha256/sha512 thing. Perhaps smth similar to how `dg_table` is implemented might be a better idea? Input welcome.~~ I've implemented this change.
- ~~I can't find much info on if anything other than sha256 or sha512 are used. Info or sample EF_SODs welcome.~~ They may be used (no limits are specified), but it should be fairly easy to implement anything new due to the change right above this. ICAO 9309 p10 says it's AlgorithmIdentifier from RFC 3280, which in turn refers to RFC 3279 (which is incomplete as it lacks SHA256 and SHA512, so I based it on [RFC3447](https://tools.ietf.org/html/rfc3447#page-43)).
- I'm not happy with how I had to sneak in `dgnum` to `dg_table`. Opinions on better impls welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,7 @@ tools/andrew/*
 tools/jtag_openocd/openocd_configuration
 ppls patches/*
 *- Copy.*
+/EF_*
 
 client/lualibs/mfc_default_keys.lua
 client/lualibs/pm3_cmd.lua
--- a/client/src/cmdhfemrtd.c
+++ b/client/src/cmdhfemrtd.c
@@ -1485,7 +1485,6 @@ static int emrtd_print_ef_dg12_info(uint8_t *data, size_t datalen) {
 }
 
 static int emrtd_ef_sod_extract_signatures(uint8_t *data, size_t datalen, uint8_t *dataout, size_t *dataoutlen) {
-    // very very very very cursed code.
     uint8_t top[EMRTD_MAX_FILE_SIZE] = { 0x00 };
     uint8_t signeddata[EMRTD_MAX_FILE_SIZE] = { 0x00 };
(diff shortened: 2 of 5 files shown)


aveao:
Improved two bits I asked for opinions on, feeling more confident with a merge now, but opinions on improving what's in there is still welcome.

doegox:
Let's merge and improve later if needed

PR #1146, opened 2020-12-29
J-Link debugging

Gator96100:
Allow for J-Link debugging in VSCode of the proxmark3 firmware. OpenOCD debugging should also be possible, but would require a different launch configuration.

I did create installation instruction for VSCode for Ubuntu,WSL and ProxSpace. Testing and updating the installation instruction for other Linux distributions is still pending.

Tested to work with Ubuntu 20.10, WSL and ProxSpace
Happy debugging!


doegox:
Great!
Some remarks:
* `SKIP_COMPRESSION` => there is a DEBUG for client-side debugging (-g etc), so I rather see a `DEBUG_ARM` as the flag sets also -g on ARM (and skip compression to make it working properly)
* Instructions: ARM-side debugging won't be for regular users, better not to make it default to install all dev deps. (and currently only the Debian/Ubuntu section contains the info, not Arch etc). So I rather see one new MD file for ARM-side debugging, with a Windows and a Linux sections.

I would also simplify the armsrc/start.c ifdefs but I'll do it later, no need to block the PR on that

Gator96100:
Done.
BTW: All testing was done on a RDV2, I didn’t want to solder pins on my RDV4. However, I don’t think it should matter.

PR #1148, opened 2020-12-30
lf em 410x - textual

tcprst:
argument didn't match help text, user in discord pointed it out


iceman1001:
I renamed it to ID because EM410x is mentioned as ID / TAG ID in our demodulation,
and in its datasheet they reference its  64bit data.   The idea that its a unique ID doesn't fit.

tcprst:
Yeah I know it doesn't fit I just wanted to make it match the configured argument for the time being. I figured I'd go over all of them with a normalization pass for consistency once we get closer to the end

iceman1001:
I know,   you know,  it is for the next round after we added cliparser to all commands.   Then we will have a feeling for how we want to have it.    Its a "living" evolving going on.

PR #1149, opened 2020-12-30
eMRTD info: Split EF_SOD into a new section, add further case handling

aveao:
This PR splits EF_SOD into a new section and accounts for further cases such as file not being in EF_SOD, file being in EF_SOD but not actually being read.

This should conclude hash verifications for EF_SOD, hopefully... just need to do cert verifs one day.


iceman1001:
Nice!

PR #1150, opened 2020-12-30
eMRTD dump: Allow specifying a custom dump directory

aveao:
This PR adds support to supplying a path to dump stuff to. The directory needs to exist beforehand.

pm3dump

---

Poking #1117 as this was on checklist


iceman1001:
This one I wouldn't have merged.  I would have rather seen @mwalker33  's solution for the PM3 client instead.

Be aware that I will remove this functionality once the preferred solution is in place.

aveao:
@iceman1001 Sounds good to me. I'm not super happy with this impl anyways, so having it replaced will be nice.

PR #1154, opened 2020-12-31
hf 14a config - now use cliparser

tcprst:
Also fixed a textual bug device side with regards to BCC config 1.


iceman1001:
@doegox  mentioned that we could use  ` --atqa std|force|skip`   instead of 0,1,2.
I like that idea.

doegox:
> @doegox mentioned that we could use --atqa std|force|skip instead of 0,1,2.
> I like that idea.

I pushed a commit for that

PR #1158, opened 2021-01-04
Introduce Github Actions for macOS and Linux

aveao:
As discussed on Discord, Travis CI, which is used for macOS and Linux builds, now has quite limited build minutes.

---

This PR introduces 6 simple GitHub Actions (GitHub CI) workflows that do exactly what was done on Travis CI.

GitHub CI currently is free for public repositories and has unlimited minutes, but it shouldn't cause lock-in concerns as the CI format is quite simple, and would be fairly easy to port into anything else.

---

As a note, this runs the 6 actions **in parallel**, which is a little spammy:

but in return, improves build times and makes it easier to identify what exactly is failing, as the alternative of using steps isn't parallelized:

I can switch them over to steps if desired, but I don't think that's necessary.

---

Right now, this tests on Ubuntu 18.04 LTS and macOS Catalina (10.15). It should automatically switch over to newer versions (Ubuntu 20.04 LTS and macOS Big Sur (11.0)) when GitHub considers them stable (as I used their -latest variants).

Ubuntu 20.04 LTS move is more urgent, [it being rolled out these days](https://github.com/actions/virtual-environments/issues/1816), but I've also tested these by forcing them to 20.04 LTS and had no issues.

I'm also testing out macOS Big Sur (11.0) in the same way, ~~but it seems like the amount of runners is quite limited as my CI runners have been queued with no changes for 35 minutes now. In any case, state of those can be viewed [here](https://github.com/aveao/proxmark3/actions?query=branch%3Agithub-actions-testing).~~ and it also does have no issues, though GitHub Actions itself is a bit buggy with it. I assume it'll be fixed by the time it takes over macos-latest.

--- 

Some parts of the Ubuntu ones are based on @linuxgemini's CI configs for GitLab CI. Thank you linuxgemini!


iceman1001:
Awesome!

So,  @Gator96100 ,    just to be sure,   nothing interferes for you if we add this?

Gator96100:
Looks good to me.

PR #1159, opened 2021-01-04
android target client compability

ca1e:
fix for RFIDtools component


doegox:
thanks ! I'll merge and add client/src/cmdhfemrtd.c as well

PR #1160, opened 2021-01-04
Rework of GitHub Actions

linuxgemini:
This PR squashes down seperate workflows into one (per OS) workflow with seperate jobs inside. Each job in the workflow will run sequentially and won't depend on the status of the previous job.

_Note about macOS builds:_ Although I haven't experienced any failures during the rework, there still is a chance of failure due to still unknown reasons.


linuxgemini:
PS: I have added a condition where if `brew update` fails, the build will still continue.

PR #1169, opened 2021-01-20
dev 4x50 sim full

tharexde:
Added bidirectional communication to "4x50 sim" function so that proxmark can simulate an EM4x50 tag in a fully functional way.
A password can be specified as an optional parameter, since dump files usually do not contain the password.
Successful tests were performed against a second proxmark and a commercial reader/writer from Elatec (TWN4).


iceman1001:
Nice one!

PR #1170, opened 2021-01-22
Nonce acquisition improvements

socram8888:
This PR has two different, unrelated patches:
 - First it fixes the nonce acquisition for MFC tags. Previously, the counter "i" was being incremented by the "continue" in case of errors, which resulted in empty, zeroed nonces being sent to the host.
 - Second it replaces the error messages on the AcquireEncryptedNonces method so it's more clear which one is being called when an error pops up on the console.

### DIFF
--- a/armsrc/mifarecmd.c
+++ b/armsrc/mifarecmd.c
@@ -696,7 +696,7 @@ void MifareAcquireNonces(uint32_t arg0, uint32_t flags) {
 
     LED_C_ON();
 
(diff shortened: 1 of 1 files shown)


iceman1001:
Awesome!

Does this solve the mysterious autopwn/hardnested issue?

socram8888:
Nope. I just started working again on that, and found these minor issues along the way. I had been chasing red herrings for a while because the error on the console reported the Auth1 errors were detected on the AcquireNonces function until I discovered the errors on the actual AcquireEncryptedNonces said the same.

iceman1001:
:)

PR #1171, opened 2021-01-22
Fix out of bounds read

socram8888:
A very small bug that Address Sanitizer caught. This check inside the inner loop was using the outer's index variable.


iceman1001:
Good catch!

PR #1174, opened 2021-01-26
`trace list -t mf` - now can use external dictionary keys file

McEloff:


client/src/cmdtrace.c
```diff
@@ -626,6 +630,7 @@ int CmdTraceList(const char *Cmd) {
         arg_lit0("x", NULL, "show hexdump to convert to pcap(ng)\n"
                  "                                   or to import into Wireshark using encapsulation type \"ISO 14443\""),
         arg_strx0("t", "type", NULL, "protocol to annotate the trace"),
+        arg_strx0("d", "dict", NULL, "use dictionary keys file"),
```
iceman1001, line 633: we tend to use -d for data..
I suggest you replace it with  NULL,  "dict"   making a long param only.   Don't forget to change the command sample text above too.

iceman1001:
Awesome contribution!   Just minor thing and I can merge it

client/src/cmdtrace.c
McEloff: You are right. But "f" flag for file selection is used now for "show frame delay times".

iceman1001: yeah,  we are converting from old style to cliparser style and in that process we will have do go over stuff a second and third time to iron out those quirks.   So don't take mixes that still in the client as a justification.
We are trying to make the conversion work less.

iceman1001:
Awesome!

PR #1190, opened 2021-02-10
Fix Gallagher data decoding & CRC checks

DarkMatterMatt:
Hi, this PR fixes various bugs when parsing Gallagher card data.

Changes:
Descramble function was re-scrambling the data, rename to `scramble` and add `descramble` function
Skip constant magic prefix `7FEA` to get correct data to decode
Fix offsets & bitmasks to correctly decode card and region codes
Fix CRC refin & refout parameters (both now false)

Tested on several real cards to check that CRCs & embedded data was reading correctly.
Reference: https://github.com/megabug/gallagher-research/


iceman1001:
Interesting PR,
Your substitution table,  how did you generate it?   This different from Matt's  https://github.com/megabug/gallagher-research/blob/master/formats/cad.md  .  He also mentions that it can be used forward and backward,   and you say its two different ones.  What makes you divide from his ?    Either Matt has filled is repo with false data,...

Same goes for the CRC notes,   where your impl doesn't match his notes. 

Do you have a signal trace file from a 125 KHz Cardx/gallagher credential that you can share,  so I can test this PR?

DarkMatterMatt:
Hi, thanks for such a quick response!

The substitution table is the inverse of the scramble table provided [here](https://github.com/megabug/gallagher-research/blob/master/formats/cardholder/cardholder.md#encoding), such that `descramble(scramble(<any_byte>)) == <any_byte>`

The document you linked to refers to high frequency MIFARE cards, I believe this is completely different to the low frequency cards I've been working with?

The CRC implementation is referenced from [here](https://github.com/megabug/gallagher-research/blob/master/formats/card-specific/125khz/125khz.md#format), the two changes made were:
1) change both refin=false and refout=false, this was not specified in the linked document
2) change 0x107 to 0x7, these are two different representations of the same polynomial (and the underlying library implicitly converts them anyway, so either will work). I changed it to 0x7 for consistency with the other polynomials in the file.

I don't have a signal trace right now, but I can create one for you soon if need be. This PR can be tested by cloning and reading a test card from the two examples given in [encoding](https://github.com/megabug/gallagher-research/blob/master/formats/card-specific/125khz/125khz.md#encoding) and [decoding](https://github.com/megabug/gallagher-research/blob/master/formats/card-specific/125khz/125khz.md#decoding).
`lf gallagher clone --raw 7FEAA34562A96A351A8C59D2`
`lf gallagher clone --raw 7FEAA34BA4CB5A34068C922B`

These changes were made with Matt looking over my shoulder and helping me fix things :)

iceman1001:
Awesome!
Now the demod of the  /trace/lf_GALLAGHER.pm3  looks good too.  :)   

Thanks!  
...  and a encode?  ;)

DarkMatterMatt:
Thanks for the merge, hopefully you'll see an encode soon 🤞

PR #1191, opened 2021-02-10
Added the original VID and PID values

Gator96100:
Allow the pm3 script to find proxmarks with old firmware.

### DIFF
--- a/pm3
+++ b/pm3
@@ -148,16 +148,7 @@ function get_pm3_list_Windows {
     fi
 
     # Normal SERIAL PORTS (COM)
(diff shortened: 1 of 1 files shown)


iceman1001:
hm...  I thought the way was to add another where-object in the command...

Gator96100:
That was what I originally planned, but white BT dongle it is separated as well.

doegox:
white BT is separated because guarded by `if $FINDBTDONGLE` and much less mergeable in the Linux section.
Feel free to merge old and new VID/PID queries.

Gator96100:
I didn't notice the `if $FINDBTDONGLE`, now it makes more sense. Anyway, the queries are merged.

iceman1001:
Great!   Thanks @Gator96100 !

iceman1001:
...dang..   in WSL it just hangs now....

Gator96100:
Interesting, I did check the query in powershell. Didn’t test WSL though

PR #1192, opened 2021-02-10
Escape needed for WSL

Gator96100:
My bad for not testing WSL


doegox:
Do detections on Linux and osx need to be adapted as well?

Gator96100:
No, only Windows uses VID and PID all others use the vendor/device name. wmic doesn’t support vendor names only IDs.

PR #1193, opened 2021-02-11
Improve support for Gallagher cloning

DarkMatterMatt:
Add ability to clone Gallagher cards by specifying their region code, facility code, card number & issue level


iceman1001:
Awesome!

iceman1001:
You adding this to "lf gallagher sim" too?

PR #1195, opened 2021-02-11
Support specifying region/facility/card/issue numbers to Gallagher simulation

DarkMatterMatt:
Extract `region_code`/`facility_code`/`card_number`/`issue_level` validation to `isValidGallagherParams` function
Add `--rc/fc/cn/il` parameters to `lf gallagher sim`

Draft pull request as I don't have access to a reader (or another proxmark) for a couple of days, so hasn't been tested on real hardware. Correct data is printed on line 428 so I expect it to work.

### DIFF
--- a/client/src/cmdlfgallagher.c
+++ b/client/src/cmdlfgallagher.c
@@ -187,6 +187,34 @@ static int CmdGallagherReader(const char *Cmd) {
     return PM3_SUCCESS;
 }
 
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice!
Its still working with raw like before,  so its not an issue to merge.
Just let me know what you want to do.

DarkMatterMatt:
I'm happy to merge :)

iceman1001:
Don't forget to make a line in the CHANGELOG.md file...

PR #1200, opened 2021-02-14
4x50 standalone

tharexde:
Added standalone mode for EM4x50 (lf_tharexde).
It simulates EM4x50 dumps, reads words transmitted by EM4x50 tags in standard read mode and stores them in internal flash.
On entering stand-alone mode, this module will start simulating EM4x50 data. Data is read from eml dump file uploaded to flash. If reader sends password different from dump file password, it is saved in file lf_em4x50_passwords.log.
On switching to read mode by pressing pm3 button, module will start reading EM4x50 data. Each collected data set will be written/appended to logfile in flash (lf_em4x50_collect.log) as a text string.
RDV4 hardware (for flash and battery) is required.


iceman1001:
Nice standalone mode. 
I get the feeling this is actually two different PR's..  
One for the new standalone,   and the second is your refactoring of the em4x50 device side command fcts.

In order to make things easy to follow,  May I suggest you split this up in two PRs?

tharexde:
You're right. In the context of this standalone mode a refactoring of the device side was necessary. I'll do two separate PRs.

iceman1001:
Thank you!

tharexde:
closed, will be pushed again in two steps

PR #1204, opened 2021-02-16
Minor changes to hf jooki command

adadnc:
- Extended the parameter list of `hf jooki encode` command with `-y` for figurine type id and `-f` for figurine id.
- Made changes to command codes


iceman1001:
Nice,
some German comments that you should remove,  but you can do that in the next PR..
and I wouldn't use "-y / -f" ,  I would only use  "--tid / --fid"  instead.

PR #1206, opened 2021-02-18
Add another sig sample for MFC EV1

linuxgemini:
Signed-off-by: İlteriş Yağıztegin Eroğlu <ilteris@asenkron.com.tr>


doegox:
thanks, oops I was a bit quick to merge, I thought it was some missing samples but we had already completed for Classic EV1.
Anyway that doesn't hurt.

PR #1209, opened 2021-02-21
emrtd: Add EF_CardAccess parsing to offline info

aveao:
Whoops, forgot to add this to offline too.


iceman1001:
...and the changelog...

PR #1220, opened 2021-03-12
Add modemmanager check

r1ddl3rz:
Try to add modemmanager check. I could only test Linux and Arch. Need to
be reviewed :) These are my first attempts in bash.

### DIFF
--- a/pm3
+++ b/pm3
@@ -225,6 +225,29 @@ function get_pm3_list_WSL {
     fi
 }
 
+function check_interfering_packages() {
+    # Check for package modemmanager
+    compatible=1
+    package="modemmanager"
+    arch_file=/etc/arch-release
+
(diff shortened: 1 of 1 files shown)


doegox:
Thanks but I don't think this is the adequate approach.
* It's a rabbit hole to try to cover distributions. Here you check for Arch then call dpkg, which will fail on all distros except Debia/Ubntu based ones
* Having modemmanager installed or even running is not necessarily a problem, see https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/ModemManager-Must-Be-Discarded.md

r1ddl3rz:
Thank you @doegox . I was not sure myself. Therefore the hint that someone of you (old hands;) should check it.

PR #1228, opened 2021-03-22
Prevent Windows users from soft bricking their devices

Gator96100:
Many users use pm3-flash-all when coming from builds that are based on the official firmware, but that does result in the device not showing up. 

This addition checks if the PNPDeviceID has the Iceman suffix. It might not be the best check, but it should help inexperienced users.


iceman1001:
Got a training on thursday/friday,    so this PR will wait until afterwards to be merged.

PR #1230, opened 2021-03-31
Fix mfpauth error printout

VortixDev:
It seems that the wrong offset is used to get the error code in the `hf mfp auth` printout. I've also added a call to the `mfpGetErrorDescription` function to make it clear why the error is occuring, if known.

Before:
image

After:
image


iceman1001:
Nice!   Don't forget to add a line in changelog.md

PR #1231, opened 2021-04-01
Windows prioritise USB connections

Gator96100:
This is to help with COM ports that are created from Bluetooth devices.


iceman1001:
Thanks!

PR #1236, opened 2021-04-07
Fix swapped DESELECT and WTX annotations

VortixDev:
The MFDES annotation function currently checks for DESELECT and WTX the wrong way around.


iceman1001:
Thanks!

PR #1239, opened 2021-04-08
Fix RESTORE annotation

VortixDev:
The ISO 14443a annotation marks any data starting with `C2` as the RESTORE command, though this first byte is shared with the DESELECT S-block. The MFDES annotation function calls the ISO 14443a annotation function before applying its own annotations, and as it does not check the size of the command before labelling it as RESTORE, any DESELECT command is labelled as RESTORE(224) (treating the first CRC byte as the block parameter). I have modified the ISO 14443a annotation function to perform size checking.

Before:
image

After:
image


iceman1001:
Thanks!

PR #1248, opened 2021-04-15
show some general crypto description

merlokk:


iceman1001:
We are trying not to use `\t` in strings, and use 4 spaces instead.

merlokk:
@iceman1001  i'll change

PR #1250, opened 2021-04-19
hf 14a list: fix support for other options and provide specific help.

doegox:
@iceman as told in PM

* fix force the trace type also when other args are given. Otherwise e.g. `hf 14a list` will show 14a annotations but not `hf 14a list -c`
* short help text for the alias rather than dumping the whole trace list help when invoking one of the 16 aliases

If this looks ok to you we can do the same for the 15 other aliases of `trace list`

### DIFF
--- a/client/src/cmdhf14a.c
+++ b/client/src/cmdhf14a.c
@@ -171,11 +171,19 @@ static uint16_t frameLength = 0;
 uint16_t atsFSC[] = {16, 24, 32, 40, 48, 64, 96, 128, 256};
 
 static int CmdHF14AList(const char *Cmd) {
+    // leading space for when concatenated to other args
+    const char type[20] = " -t 14a";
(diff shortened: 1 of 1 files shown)


iceman:
Hey man, the work you’re doing here seems absolutely heroic and amazing.  Which I had time to be more invested.  But my GitHub name included on the group seems to be a mistake.  I am iceman on GitHub but I think that’s not who you want.  Maybe it’s a bug in GitHub or there is another...

iceman1001:
:)  Must been late for @doegox  ,   since he meant me and not you @iceman

iceman1001:
I think I rather not like the suggested solution.
I did another solution for `data rawdecode`   which I am thinking would suit the aliased list commands too.

doegox:
> :) Must been late for @doegox , since he meant me and not you @iceman

Damned, sorry for the HL :( it was late night indeed...

PR #1253, opened 2021-04-22
Desfire change key

mwalker33:
Desfire change key updates for DES/3DES
some minor tweaks and converting DES to 3DES to key change code.


iceman1001:
Awesome!

PR #1254, opened 2021-04-24
Update cmdhfmfdes.c

mwalker33:
Desfire changekey AES
Seems to work, but response validation fails.
Still not sure why the cmac check fails and we only got 8 bytes to check, rather then 9+ bytes
So added comment for end user to check with auth.
Also added a drop field as the field stayed active if no error from card.


iceman1001:
Great!

PR #1258, opened 2021-04-26
emrtd: Fix file reading and parsing

aveao:
This PR fixes two issues with the emrtd logic.

20ca241 fixes a small bug introduced in 2a9d064fd6fb53b1f628fc7d319b3fa9edda6382 on the `emrtd_get_asn1_data_length` function where file lengths on longer files were incorrectly determined and as such failed to read properly:

---

a20f3af fixes a small bug introduced in 501b31cf453596023cb0a79ebf133810f48eb473 on the `emrtd_mrz_determine_length` function where certain fields (such as Legal Name, Document Number, Optional elements) had the last character cropped off.

("old, working" is before 501b31cf453596023cb0a79ebf133810f48eb473, "old, broken" is after 501b31cf453596023cb0a79ebf133810f48eb473, "new" is this current fix)

I also ack that there was a bug in the logic in `emrtd_mrz_determine_length` before 501b31cf453596023cb0a79ebf133810f48eb473 where it could return -1 which made coverity unhappy. I have since fixed this issue.

---

thanks craftbyte for bringing these to my attention, btw

### DIFF
--- a/client/src/cmdhfemrtd.c
+++ b/client/src/cmdhfemrtd.c
@@ -1207,13 +1207,13 @@ static void emrtd_print_legal_sex(char *legal_sex) {
 
 static int emrtd_mrz_determine_length(char *mrz, int offset, int max_length) {
     int i;
-    for (i = max_length - 1; i >= 0; i--) {
-        if (mrz[offset + i] != '<') {
-            break;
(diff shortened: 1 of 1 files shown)


iceman1001:
Perfect!

PR #1262, opened 2021-05-02
hf_iceclass: Ensure dump filename is NULL-terminated

metalauricle:
This fixes the filename used in the save_to_flash function in hf_iceclass.

strlen will not count the null-terminator, thus this function previously 
trimmed off the null, resulting in garbage data in filenames.

### DIFF
--- a/armsrc/Standalone/hf_iceclass.c
+++ b/armsrc/Standalone/hf_iceclass.c
@@ -170,9 +170,10 @@ static void save_to_flash(uint8_t *data, uint16_t datalen, char *filename) {
                 data[4], data[5], data[6], data[7]
                );
     } else {
-        int fnlen = MIN(strlen(filename), SPIFFS_OBJ_NAME_LEN);
+        int fnlen = MIN(strlen(filename) + 1, SPIFFS_OBJ_NAME_LEN);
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice catch.
Maybe a zeroing of the fn array instead?
```
memset(fn, 0, sizeof(fn));
```

metalauricle:
> 
> 
> Nice catch.
> Maybe a zeroing of the fn array instead?
> 
> ```
> memset(fn, 0, sizeof(fn));
> ```

Wow, fast response. Thanks!
I considered this, but it would still leave the string un-terminated if it is longer than SPIFFS_OBJ_NAME_LEN.

iceman1001:
if its longer than SPIFFS_OBJ_NAME_LEN we'll have a overflow.

`SPIFFS_OBJ_NAME_LEN - 1`   is the longest filename possible to take in consideration.

iceman1001:
alright,  I pushed another fix that should take care of always have a null terminator in the different options.

metalauricle:
Ha, nice! You ended up with the same change as my updated branch. Didn't have a chance to change it till now. Should have checked the PR before doing so. Anyways, thanks.

PR #1263, opened 2021-05-05
Added standalone mode for ISO15693 cards.

tmudford:
Added a standalone mode that scans for an emulates ISO15693 cards.


iceman1001:
How about that,  from nowhere a new standalone mode.

May I suggest you follow these instructions,
https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/Standalone/readme.md#submitting-your-code

tmudford:
Will do. Thanks @iceman1001

iceman1001:
All good to go!  Thanks!

PR #1264, opened 2021-05-06
em4x50: fixed bug in eload and restore

tharexde:
fixed useless 0 == 0 comparison when checking for valid em4x50 data in dump file;
eload and restore functions should work again


iceman1001:
Good catch,
as a sidenote, the memcpy line  5 lines down,  could just have been moved up before the assignments.

PR #1266, opened 2021-05-08
Update 2_Configuration-and-Verification.md

boriskroeger:
Tested on macOS: According to the help from the pm3 command executable this should be "sc" instead of "smart". And "f" instead of "-f".


iceman1001:
I think you are looking from your old release,
the documents mirror the state of current master.

PR #1272, opened 2021-05-15
Update cmdlfem410x.c

mwalker33:
Draft EM4100 sim fix
For review: and feed back

The em4100 sim, by default adds 20 0x00's between tag ID repeats.
This did not allow my omnikey reader to see the sim ID.
Patch allows the end user to set the number of 0x00s (e.g. to none) with the default being the current 20.
If I set this to 0 the omnikey reader can see the Sim ID (tested on pm3 easy and rdv4).

Not sure what the "zero" argument should be called... so feedback welcome.

The EM 410x brute (for sim brute to a reader) also calls the some function.  I have added a place holder "zero" variable (set to 20) to compile and test.

Once we are happy with the initial tweak, i will repeat and add the some argument to the brute function.

Thanks


iceman1001:
Perfect!

PR #1280, opened 2021-05-21
C0decave requirements docu changes btaddon

c0decave:
hi,

i added some infos about the reader, the bluetooth addon and the firmware, also some lines for troubleshooting.


iceman1001:
Thanks!

PR #1284, opened 2021-05-23
Add hf 14a ndef command for type 4 tags

tcprst:
This PR adds support for detecting and retrieving NDEF data from type 4 tags.


iceman1001:
Awesome!

PR #1290, opened 2021-05-28
Fido fix

merlokk:


doegox:
Thanks @merlokk 
The PR is merged in branch *future* as *master* is currently under freeze.
Branch *future* will be merged when @iceman1001 will be back.

doegox:
https://github.com/RfidResearchGroup/proxmark3/tree/future

iceman1001:
the default filename needs to match the file name template, where we denotate hf vs lf as prefix.   
Please revert that change.

merlokk:
it will not work(
if i change the filename in the default folder and all the consts - will it ok?

PR #1301, opened 2021-06-04
Rewrote getIndalaBits()

chunkystew:
getIndalaBits() was not producing valid card data, examined a batch of genuine HID 4041X format cards and fixed up the transform from H10301 to 4041X. Now a read / clone using the same FC and SN will produce an identical card.


chunkystew:
Hi guys, had to do a lot of work on an Indala system and found some problems with the module for that. I believe this to be a worthy patch.

iceman1001:
Well,  I have not the time to go through your PR current,  but as I see it you swapped out the 26bit version to your own 4041x format.
You are aware of the fact there is thousands of bit formats for wiegand out there.   So on bases of this,  your PR can't be merged.


In Proxmark3 source code we tend to add functions, formats,  and usually never remove old ones,  since we don't have any good way of verifying old or new ones.   So we keep it until very much proven its wrong.   

Let me suggest you add a wiegand format instead and hook it up to the wiegand encode/decode instead.   
Indala commands should be refactored to use them instead. 

But regardless,   I am happy to see this PR from you.  Its not that often we see improvements to Indala formats.

chunkystew:
That's fair, although I cannot reuse the previous getIndalaBits function as it mangles the parity bits and some of the required structure for 4041X. Would you object to duplicating the function and adding a flag to 'lf indala clone' to specify the format?

iceman1001:
If you have a  trace file from one of your cards,  then that would be good to have for our test script.  ```lf read;  data save -f lf_indiala_4041x_<fc>_<cn>.pm3

chunkystew:
[lf_indala_4041x_234_21801.pm3.gz](https://github.com/RfidResearchGroup/proxmark3/files/6611568/lf_indala_4041x_234_21801.pm3.gz)

Here's a trace from a genuine HID 4041X card, I'll make the necessary code changes soon. Thanks!

chunkystew:
...and now I see that there is a whole folder full of traces. Added a trace there and rewrote as requested, with a flag to specify 4041X.

iceman1001:
getting there,
We dont do "credits" at function levels.  It messes up the source code.  You add a line in the beginning "modified by name or gh user"   and you add a line in CHANGELOG.md,   where you describe what you did in one-line.

iceman1001:
and there is a readme file in the traces folder where you need to add a line about your newly added trace file

chunkystew:
OK, done.

client/src/cmdlfindala.c
```diff
@@ -112,6 +112,17 @@ static void decodeHeden2L(uint8_t *bits) {
     PrintAndLogEx(SUCCESS, "    Heden-2L    | %u", cardnumber);
 }
 
+bool parityEven(uint16_t x) {
```
doegox, line 115: Please don't define and expose parity functions in indala src and header files.
Include parity.h
If you need a 16b version, include it there, sth like
```c
static inline uint8_t evenparity16(uint16_t x) {
    x ^= x >> 8;
    return evenparity8(x);
}
```

chunkystew: Friendly point of order, parity.c contains a single LUT for 8-bit odd parity only, and a commented out LUT for even parity... which I get, given that they are complimentary, but why a LUT at all? Using it requires loading the entire thing into memory, which is going to take more CPU time than computing parity using XORs, and then there's the memory usage. Granted that's not a lot, but on a microcontroller with 64K every byte counts, and I'm assuming that's why.

Would it not be more appropriate to calculate parity using XORs? This ought to be more efficient CPU and memory-wise. I'd be happy to add that to this PR but I don't want too much scope creep without checking.

doegox: Honestly I don't know how it came to take such choice, the only thing I'm pretty sure is that we need fast parity check in the proxmark. On the client of course that doesn't matter.
If you're up to make some tests and some profiling to see which approach would be faster on ARM, you're welcome :)

iceman1001:
How about not get creep spread and make this PR clean.
And for the potential parity related PR you do another one.

doegox:
Yes of course, when I said tests are welcome, not for this PR.
For now please just reuse the common parity fcts, even if less efficients, as it doesn't matter for the client side.
Thanks.

iceman1001:
What the status on this one?   Just the parity fct left or ?

doegox:
Yes

iceman1001:
@chunkystew   ping!

iceman1001:
Thanks!

PR #1304, opened 2021-06-07
Restored original getIndalaBits() and added getIndalaBits4041x() as an optional format

chunkystew:


chunkystew:
Guh. Duplicate.

PR #1309, opened 2021-06-11
Doc correction for windows, macos and troubleshooting

linuxgemini:
I'm doing a pass on docs and there will be more PRs coming soon.


doegox:
Thanks!
Is imgur permanent enough?
Maybe just bring the pic in the repo, or, personally I'd just see it replaced by one sentence and no screenshot...

linuxgemini:
imgur is tied to my account so it depends if my account gets suspended (or the service goes down)

we can link it inside the repo actually, I'll look into that in a bit.

PR #1310, opened 2021-06-11
ci: exclude changes to non-source folders

linuxgemini:
This PR will make Github Actions to not run on changes at non-sourcecode folders.


doegox:
interesting!
merged in future

PR #1312, opened 2021-06-13
Add support for NTAG 210μ

ah01:
Add support for NXP [NTAG 210μ](https://www.nxp.com/products/rfid-nfc/nfc-hf/ntag-for-tags-labels/ntag-210-micro-nfc-forum-type-2-tag-basic-ic:NT2L1001_NT2H1001)


doegox:
thank you!

PR #1321, opened 2021-06-19
Update cmdhfmfdes.c

mwalker33:
3K3Des Authentication and Change Key patch
Get Card UID (when authenticated for real UID
3Des session key patch when 1st half = 2nd Half


iceman1001:
Awesome!

PR #1327, opened 2021-06-21
Update Mac-OS-X-Homebrew-Installation-Instructions.md

Pondorasti:
Update misleading documentation. The current state of installation steps for macOS using Homebrew, recommends to export a global variable if using a generic device, instead of using `--with-generic` tag.


iceman1001:
Thanks!

PR #1338, opened 2021-06-25
Update for cmdlfnexwatch.c

Guilhem7:
Hi Iceman, this modification allows a user to find the magic number and to use the psk2 modulation to clone a t55 tag. Moreover the magic byte can also be added in parameter of the clone command if it isn't a known credential.

### DIFF
--- a/client/src/cmdlfnexwatch.c
+++ b/client/src/cmdlfnexwatch.c
@@ -24,6 +24,7 @@
 #include "cmdlft55xx.h"    // clone..
 #include "cmdlfem4x05.h"   //
 #include "cliparser.h"
+#include "util.h"
 
 
 typedef enum {
@@ -110,9 +111,8 @@ static int nexwatch_scamble(NexWatchScramble_t action, uint32_t *id, uint32_t *s
 }
 
 static int nexwatch_magic_bruteforce(uint32_t cn, uint8_t calc_parity, uint8_t chk) {
-    uint8_t magic = 0;
-    uint8_t temp_checksum;
-    for (; magic < 255; magic++) {
+    for (uint8_t magic = 0; magic < 255; magic++) {
+        uint8_t temp_checksum;
         temp_checksum = nexwatch_checksum(magic, cn, calc_parity);
         if (temp_checksum == chk) {
             PrintAndLogEx(SUCCESS, "    Magic number : " _GREEN_("0x%X"),  magic);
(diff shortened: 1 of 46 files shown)


common/lfdemod.c
```diff
@@ -295,6 +295,17 @@ int bits_to_array(const uint8_t *bits, size_t size, uint8_t *dest) {
     return PM3_SUCCESS;
 }
 
+// Ex: 0x56 and numbytes = 1 ==> return res = {0,1,0,1,0,1,1,0}
```
iceman1001, line 298: There is already a  `bytes_to_bytebits` fct in util.c    Use it instead

client/src/cmdlfnexwatch.c
```diff
@@ -108,6 +109,21 @@ static int nexwatch_scamble(NexWatchScramble_t action, uint32_t *id, uint32_t *s
     return PM3_SUCCESS;
 }
 
+static int nexwatch_magic_bruteforce(uint32_t cn, uint8_t calc_parity, uint8_t chk) {
+    uint8_t magic = 0;
+    uint8_t temp_checksum;
+    for (; magic < 255; magic++) {
```
iceman1001, line 115: cppchecker will say that you can lessen the scope of these two vars.   ie,  move them inside.

client/src/cmdlfnexwatch.c
```diff
@@ -285,6 +304,8 @@ static int CmdNexWatchClone(const char *Cmd) {
         arg_lit0(NULL, "hc", "Honeywell credential"),
         arg_lit0(NULL, "q5", "optional - specify writing to Q5/T5555 tag"),
         arg_lit0(NULL, "em", "optional - specify writing to EM4305/4469 tag"),
+        arg_str0(NULL, "magic", "<hex>", "optional - magic hex data. 1 byte"),
```
iceman1001, line 307: add `m"  - for a short param

client/src/cmdlfnexwatch.c
```diff
@@ -368,6 +407,17 @@ static int CmdNexWatchClone(const char *Cmd) {
         raw[10] |= nexwatch_checksum(magic, cn, parity);
     }
 
+    if (use_psk2) {
+        blocks[0] = 270464;
+        uint8_t * res_shifted = malloc(96 * sizeof(uint8_t));
```
iceman1001, line 412: 1. we use calloc in the repo instead of malloc.
2. I see no free of this allocated memory

iceman1001 (changes requested):

iceman1001:
Awesome!  
Some minor stuff to fix and its mergable

client/src/cmdlfnexwatch.c
Guilhem7: but m is already used for mode.

client/src/cmdlfnexwatch.c
Guilhem7: Ok, we fix it this afternoon then !

client/src/cmdlfnexwatch.c
iceman1001: I see.  bugger.    Ignore this one :)

client/src/cmdlfnexwatch.c
Guilhem7: do you mean we can remove the "static" ?

PR #1342, opened 2021-06-29
Change in hf 15 raw broke support hf_15_magic.lua

nemanjan00:
There was a change in `hf 15 raw` and this script was not updated to support that change


doegox:
thank you !

PR #1346, opened 2021-07-03
Desfire functionality update

merlokk:


iceman1001:
Interesting PR,  
but for the sake of easyness.   Can you split it up into two?  
1)  cliparser changes.
2) desfire changes.

PR #1347, opened 2021-07-03
add experimental support for macs with macports installed

linuxgemini:
currently requires symlinking python{version}(-embed).pc to python(embed).pc or python3(-embed).pc under /opt/local/lib/pkgconfig for python linking

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -14,10 +14,18 @@ vpath %.dic dictionaries
 OBJDIR = obj
 
 ifeq ($(USE_BREW),1)
-	INCLUDES += -I$(BREW_PREFIX)/include
-	LDLIBS += -L$(BREW_PREFIX)/lib
-	PKG_CONFIG_ENV := PKG_CONFIG_PATH=$(BREW_PREFIX)/opt/qt/lib/pkgconfig
-	PKG_CONFIG_ENV := PKG_CONFIG_PATH=$(BREW_PREFIX)/opt/qt5/lib/pkgconfig
+	ifdef MACPORTS_PREFIX
+		INCLUDES += -I$(MACPORTS_PREFIX)/include
+		LDLIBS += -L$(MACPORTS_PREFIX)/lib
+		PKG_CONFIG_ENV := PKG_CONFIG_PATH=$(MACPORTS_PREFIX)/lib/pkgconfig
+		PKG_CONFIG_ENV := PKG_CONFIG_PATH=$(MACPORTS_PREFIX)/libexec/qt/lib/pkgconfig
+		PKG_CONFIG_ENV := PKG_CONFIG_PATH=$(MACPORTS_PREFIX)/libexec/qt5/lib/pkgconfig
+	else
+		INCLUDES += -I$(BREW_PREFIX)/include
+		LDLIBS += -L$(BREW_PREFIX)/lib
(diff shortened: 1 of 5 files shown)


client/Makefile
```diff
@@ -117,8 +125,13 @@ INCLUDES += $(HARDNESTEDLIBINC)
...
 ifneq ($(SKIPLUASYSTEM),1)
-    LUAINCLUDES = $(shell $(PKG_CONFIG_ENV) pkg-config --cflags lua5.2 2>/dev/null)
-    LUALDLIBS = $(shell $(PKG_CONFIG_ENV) pkg-config --libs lua5.2 2>/dev/null)
+    ifdef MACPORTS_PREFIX
+        LUAINCLUDES = $(shell $(PKG_CONFIG_ENV) pkg-config --cflags lua-5.2 2>/dev/null)
+        LUALDLIBS = $(shell $(PKG_CONFIG_ENV) pkg-config --libs lua-5.2 2>/dev/null)
+    else
+        LUAINCLUDES = $(shell $(PKG_CONFIG_ENV) pkg-config --cflags lua5.2 2>/dev/null)
+        LUALDLIBS = $(shell $(PKG_CONFIG_ENV) pkg-config --libs lua5.2 2>/dev/null)
+    endif
```
iceman1001, line 134: What is the difference that we need to have this IFDEF ?   It looks the same on both sides.

doc/md/Installation_Instructions/Mac-OS-X-MacPorts-Installation-Instructions.md
```diff
@@ -0,0 +1,72 @@
...
+    ```bash
+    export C_INCLUDE_PATH="/opt/local/include"
+    export CPLUS_INCLUDE_PATH="/opt/local/include"
+    export LIBRARY_PATH="/opt/local/lib"
+    export LDFLAGS="-L/opt/local/lib"
+    export CFLAGS="-I/opt/local/include"
+    export CPPFLAGS="-isystem/opt/local/include -I/opt/local/include"
+    ```
+
+    other than just modifying PATH variable to `/opt/local/bin:/opt/local/sbin:/usr/local/bin:/usr/local/sbin:$PATH`.
```
iceman1001, line 22: unclear instruction here.  Do user have to modify PATH or not?

doc/md/Installation_Instructions/Mac-OS-X-MacPorts-Installation-Instructions.md
```diff
@@ -0,0 +1,72 @@
...
+    ```
+
+
+## Compile and use the project
+
+To use the compiled client, the only difference is that the Proxmark3 port is `/dev/tty.usbmodemiceman1`, so commands become:
+
+```sh
+proxmark3 /dev/ttyACM0  =>  proxmark3 /dev/tty.usbmodemiceman1
+```
```
iceman1001, line 66: I think we should add an instruction to use ./pm3  shell script.   I see many users thinking they need to call the client direct themself with a serial port.

iceman1001:
Nice one!
Some minor comments only

client/Makefile
doegox: THere is a diff, I see a lua-5.2 and a lua5.2

iceman1001: Sneaky dash.  Now I see it too.   Thanks!

iceman1001:
Thanks!

PR #1348, opened 2021-07-04
iClass Updates - AA1 Size

pcunning:
Fixes a variable pointer range issue and changes the client to handle cards with an impropper application limit.

I received some unpersonalized iClass cards that came with the config block set as `FF...` setting AA1 limit to 255. This caused a crash when attempting to dump the card. The crash seems to have been caused by the memcpy attempting to copy past the allocated variables size (starting at block 6 only 250 blocks would be left). I suspect this would cause issues if AA1 was ever > 250.

Also see https://discord.com/channels/738045391416852520/738046308929110076/856658801645977610

### DIFF
--- a/client/src/cmdhficlass.c
+++ b/client/src/cmdhficlass.c
@@ -1653,6 +1653,9 @@ static int CmdHFiClassDump(const char *Cmd) {
 
             app_limit1 = card_app2_limit[type];
             app_limit2 = 0;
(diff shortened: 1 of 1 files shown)


iceman1001:
Awesome!

iceman1001:
I pushed a changelog update to match this PR:

PR #1362, opened 2021-07-10
Update cmdlft55xx.c

scott4290:
Fixed two bugs:  >= preventing users from testing one password (start_password=end_password) and reporting that the start password is greater than the end password.  The second issue is when the end password _is_ the correct password, the found password reported was curr -1.

[usb] pm3 --> lf t55xx bruteforce -s 00000042 -e 00000042
[=] press <Enter> to exit
[=] Search password range [00000042 -> 00000042]
.[=] Trying password 00000042
[=]  Chip type......... T55x7
[=]  Modulation........ FSK2a
[=]  Bit rate.......... 4 - RF/50
[=]  Inverted.......... Yes
[=]  Offset............ 34
[=]  Seq. terminator... No
[=]  Block0............ 00107070 (auto detect)
[=]  Downlink mode..... default/fixed bit length
[=]  Password set...... Yes
[=]  Password.......... 00000042

[+] Found valid password: [ 00000042 ]
Downlink Mode used : default/fixed bit length

[+] time in bruteforce 0 seconds

### DIFF
--- a/client/src/cmdlft55xx.c
+++ b/client/src/cmdlft55xx.c
@@ -3301,7 +3301,7 @@ static int CmdT55xxBruteForce(const char *Cmd) {
     uint32_t curr = 0;
     uint8_t found = 0; // > 0 if found xx1 xx downlink needed, 1 found
 
-    if (start_password >= end_password) {
+    if (start_password > end_password) {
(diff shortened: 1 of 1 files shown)


iceman1001:
Thanks,

sidenote:
I think changing the `>=` introduced your second issue....

PR #1365, opened 2021-07-11
Update flash.c

scott4290:
Adding a suggested step to point the user in the right direction since the README.md has some dedicated notes on this topic.  It's likely that users seeing this message are compiling the firmware for the first time and may require the extra guidance.  Also, they have an unusable device until they re-compile and flash again.  I'm sure the extra note will be appreciated.

[+] Loading usable ELF segments:
[+]    0: V 0x00102000 P 0x00102000 (0x00040888->0x00040888) [R X] @0x94
[!!] 🚨 Error: PHDR is not contained in Flash
[!!] 🚨 Firmware probably too big for your device
[=] The flashing procedure failed, follow the suggested steps!
[+] All done


iceman1001:
thanks

PR #1366, opened 2021-07-12
Update bt_manual_v10.md

scott4290:
Added troubleshooting steps needed when bluetoothctl wasn't pairing and also didn't prompt for a PIN.


iceman1001:
Thanks!

PR #1371, opened 2021-07-12
Update cmdhf14a.c

TheArchitect0880:
changed "hf emv" to "emv" since its not under hf commands anymore


iceman1001:
Why close the PR?

TheArchitect0880:
Honestly, i have no clue how to test if this change actually does what i wanted.

iceman1001:
if you have a 14a emv card that will be picked up by 14a info and get a hint to the emv commands :)

PR #1372, opened 2021-07-13
Update bt_manual_v10.md

scott4290:
I realized I can replace an entire troubleshooting stanza with a preemptive remove command in the original instructions.
[bluetooth]# remove aa:bb:cc:dd:ee:ff

Pull if you think this is a cleaner approach.  Either is good with me.


iceman1001:
definitely cleaner

PR #1380, opened 2021-07-22
Add a standalone FacilityCode bruteforcer for HID

ss23:
This requires a known card number, but often works as 1 is a default.

My first stand-alone module, so let me know if there are any pointers etc :)

### DIFF
--- a/armsrc/Standalone/lf_hidfcbrute.c
+++ b/armsrc/Standalone/lf_hidfcbrute.c
@@ -1,0 +1,168 @@
+/**
+ * Bruteforce a HID system using a static card number but incrementing FC
+ *
+ * This is only going to work if the system has a card number registered that you know,
+ * or if you can determine whether a given FC is valid based on external information.
+ *
+ * Author: proxmark@ss23.geek.nz - ss23
+ * Based on lf_hidbrute
+ * 
+ * To retrieve log file from flash:
+ *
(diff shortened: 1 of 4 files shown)


iceman1001:
Nice,  
one question, what is it that you actually log?   There is no way for the proxmark3 to know when there is a successful attempt

ss23:
Since there's no automated way to tell, the only thing that is logged is when a user presses the button manually. The idea being that while doing the bruteforce, you wait for the system to accept a given FC, then you press the button so you log it (or the FC after it), that way you know what it was later.

It's not an ideal situation, but I figure this is more useful than trying to count manually.

armsrc/Standalone/lf_hidfcbrute.c
```diff
@@ -0,0 +1,168 @@
...
+#include "lfops.h"
+#include "BigBuf.h"
+#include "fpgaloader.h"
+#include "parity.h"
+
+// What card number should be used for the bruteforce?
+// In some systems, card number 1 is valid, so this may be a good starting point.
+#define CARD_NUMBER 1
+
+#define LF_HIDCOLLECT_LOGFILE "lf_hidlfcollect.log"
```
iceman1001, line 44: maybe rename the file to relate to _brute_ ?
like  `lf_hidbrute.log`

iceman1001:
ok,   so it will iterate 256 times regardless  and first button press starts the brute function. 
then every button press logs current number.   

ok,

PR #1381, opened 2021-07-23
Desf create/delete file

merlokk:


iceman1001:
I think we are keeping the other bytes to number conversions in  /common/commonutil.c

Issue #442, opened 2019-10-17
[feature] FeliCa Standard Support


7homasSutter:
Hello there!

I started to work on the FeliCa Standard card support and I will work on this for a couple of weeks. To keep you up to date with my progress I opened this issue and reference the commit id's. 

Protocol Commands:
- [x] Polling: #443  
  - [ ] Fix reader for RC-S888 / RC-S962 and RC-S886 / RC-S965
  - [ ] Add Parameter for specific System Code selection
- [x] Request Service #447
  - [ ] Add parameter for service code iteration
- [x] Raw: #443
  - [ ] Add parameter for calculation of length 
  - [ ] Refactor parameter parsing
- [x] Sniffing Communication
  - [ ] Fix #444 
- [x] Request Response
- [x] Read Without Encryption #451
- [x] Write Without Encryption #456 
- [ ] Search Service Code
- [x] Request System Code
- [ ] Authentication1
- [ ] Authentication2
- [ ] Read
- [ ] Write
- [ ] Request Service v2
- [ ] Get System Status
- [x] Request Specification Version
- [x] Reset Mode
- [ ] Authentication1 v2
- [ ] Authentication2 v2
- [ ] Read v2
- [ ] Write v2
- [ ] Update Random ID

Card Issue Commands:
- [ ] Register Issue ID
- [ ] Register Area
- [ ] Register Service
- [ ] Change System Block

Management Commands:
- [ ] Register Manufacturer ID
- [ ] Self Diagnosis

Proxmark Specific Commands:
- [ ] Info: General Card Info
- [ ] Dump
- [ ] Emulate

Another reason for opening this issue is, I own only one FeliCa Card and I don't know the keys of the card. So testing all the commands (which will need encryption) is probably impossible for me alone, but maybe someone in the community has some cards with known keys and can then test it. So when I'm ready for testing these commands I will announce it here in this issue as well. 

If you have anything I should know, do not hesitate to contact me directly.

Cheers
Tom

iceman1001:
Awesome!

I have some FeliCa card but no idea of about it.  This whole part is uncharted so getting support into proxmark3 is of essential value. Without it, no research :)

iceman1001:
I also suggest bringing in multiple minor PR, instead of a huge massive PR with all functions at once.
Much easier for us to analyse it that way.  :)  And it also brings curious ppl to the repo and might encourage them to be involved.

7homasSutter:
[...]

PR #1382, opened 2021-07-24
Add FeliCa model name to felica info command

ah01:
Sony provide [list of FeliCa products](https://www.sony.net/Products/felica/business/tech-support/list.html) with corresponding `IC Type` values. 

This PR adds ability to show _Model name of IC chip_ from given list into `hf felica info` command.

PS: Apparently there are more _IC Types_ out there then they are in this public list (e.g. in #442 somebody mentioned Suica (FeliCa standard (?) ) with type `0x0B` and some more are also in [nfcpy](https://github.com/nfcpy/nfcpy/blob/af5b13392f74a754fe4fa60f12463b7c392ffe8e/src/nfc/tag/tt3_sony.py#L60) code base). I did not include them because I am not sure about reliability of those informations. Or should I?

### DIFF
--- a/client/src/cmdhffelica.c
+++ b/client/src/cmdhffelica.c
@@ -167,6 +167,56 @@ static int print_authentication2(void) {
     return PM3_SUCCESS;
 }
 
+static const char *felica_model_name(uint8_t rom_type, uint8_t ic_type) {
+    // source: mainly https://www.sony.net/Products/felica/business/tech-support/list.html
+    switch (ic_type) {
+        // FeliCa Standard Products:
+        case 0x46:
+            return "FeliCa Standard RC-SA21/2";
+        case 0x45:
+            return "FeliCa Standard RC-SA20/2";
+        case 0x44:
+            return "FeliCa Standard RC-SA20/1";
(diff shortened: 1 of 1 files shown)


iceman1001:
Thanks!

iceman1001:
make another PR with nfcpy missing ic types.   So we can keep track of those changes individually

iceman1001:
these three seems to be missing
```
0x00: ("RC-S830",    8,  8),  # RC-S831/833
0x02: ("RC-S919",    1,  1),  # RC-S890
0x0B: ("RC-S???",    1,  1),  # new suica
```

iceman1001:
https://www.sony.net/Products/felica/business/tech-support/list.html

there are a range of Mobile FeliCa IC chips too which would be nice to have 
```
14-1F  Mobile 3.0  
10-13  Mobile 2.0
06-07  Mobile 1.0
```

iceman1001:
pushed in  https://github.com/RfidResearchGroup/proxmark3/commit/a7c02a4c52e398e8be8d599e90fa66d0fb9915ac

PR #1393, opened 2021-07-28
fc/cn support for iclass encode

bettse:
Pretty sure this isn't quite right (like, I think it works, but I think it isn't good code).  I wanted to open this to start getting feedback.  

 * More correct way to copy data to 'credential'
 * Should I support --format?
 * Is using `bin_len` an OK way to determine if the user is using --bin vs --fc/--cn ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Add `--wiegand/--fc/--cn` to `hf iclass encode` (@bettse)
  - Fix `hf 15 sample` - data collection works again (@iceman1001)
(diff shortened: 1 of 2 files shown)


tcprst:
Looks like you have H10301 hardcoded as the wiegand format.  Other commands take in the format as an argument, I think we need that argument for this to be useful.

iceman1001:
interesting,
maybe add a line or two in the help text about you should use the command with either bin param or fc/cn/fmt params to make it extra clear.

bettse:
> maybe add a line or two in the help text about you should use the command with either bin param or fc/cn/fmt params to make it extra clear.

Are there commands that have similar text that you'd like me to pattern off of?  Or if not, do you have a preference about if the comment precedes the examples or follows it?

iceman1001:
Nothing particular in mind,  you would just add a newline character and add some more clear instructions,  about here
image

iceman1001:
don't forget a line in CHANGELOG.md too

iceman1001:
Thanks!

PR #1398, opened 2021-07-29
Reword hex count for EM410x help

clach04:
The help text for the size/length of the EM410x id is ambiguous.

The length of 5 is the number of bytes but this is 10 hex digits. The example does clearly show a 10 hex digit number.

Rewording attempt to garner feedback. EM410x is not the only place where this wording is present for help/documentation.

### DIFF
--- a/client/src/cmdlfem410x.c
+++ b/client/src/cmdlfem410x.c
@@ -410,7 +410,7 @@ static int CmdEM410xSim(const char *Cmd) {
     void *argtable[] = {
         arg_param_begin,
         arg_u64_0(NULL, "clk", "<dec>", "<32|64> clock (default 64)"),
-        arg_str1(NULL, "id", "<hex>", "ID number (5 hex bytes)"),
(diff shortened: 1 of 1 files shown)


iceman1001:
Thanks for your suggestion, 

When we did the help text changes beginning of this year,  we unified from  hex bytes,  hex samples, hex digits into hex bytes.  It was a wild place with all kinds of denotations before this change.  Not very clear at all.

We settled for the hex byte concept,  where the normal denotation of a hex byte is with two digits.    

If a user would have problems with this then the whole proxmark client will be very hard to understand.
In the end we can't fool proof this software for users with limited IT knowledge.  It's not the intended user group.

Back to your suggested text change,  I understand that the error message when enter a id which isn't 5 bytes,  can be expanded to make it even clear but changing this one means changing all other instances of such feedback to the user in the client.

iceman1001:
fixed in #d0c1492a46a96b9393df86836ebda99c1112b744

iceman1001:
Thanks for the idea,  it should be even clearer now

Issue #1411, opened 2021-08-03
[Bug]14a anti-collision not working


AxisRay:
**Describe the bug**
Anti-collision not working, because the BBC in uid_resp is always 0x00
```
[#] BCC0 incorrect, got 0x00, expected 0xdc
[#] Aborting
[usb] pm3 --> trace list -t 14a
[=] downloading tracelog data from device
[+] Recorded activity (trace len = 71 bytes)
[=] start = start of start frame end = end of frame. src = source of transfer
[=] ISO14443A - all times are in carrier periods (1/13.56MHz)

      Start |        End | Src | Data (! denotes parity error)                                           | CRC | Annotation
------------+------------+-----+-------------------------------------------------------------------------+-----+--------------------
          0 |        992 | Rdr |52(7)                                                                    |     | WUPA
       2116 |       4484 | Tag |04  00                                                                   |     | 
       7040 |       9504 | Rdr |93  20                                                                   |     | ANTICOLL
      10564 |      16388 | Tag |99  bf! 3f! ff  de                                                       |     | 
     164480 |     168032 | Rdr |93  30  99                                                               | !crc| SELECT_XXX
     169140 |     173876 | Tag |ba  16  e9  dc                                                           |     | 
```
as the trace , tag have replied the remaining UID ba16e9  with BCC dc
But the pm3 print error log : BCC0 incorrect, got 0x00, expected 0xdc

**To Reproduce**
1. there are two mifare classic card with UID 99BA16E9 and 19B72F9F
2. put the two cards  on the pm3
3. execute `hf 14a info`
4. console output:
```
[#] BCC0 incorrect, got 0x00, expected 0xdc
[#] Aborting
```

**Expected behavior**
```
[usb] pm3 --> hf 14a info

[+]  UID: 59 B7 2F 9F
[+] ATQA: 00 04
[+]  SAK: 08 [2]
[+] Possible types:
[+]    MIFARE Classic 1K
[=] proprietary non iso14443-4 card found, RATS not supported
[+] Magic capabilities : Gen 1a
[+] Prng detection: weak
[#] Auth error
[?] Hint: try `hf mf` commands
```
**Additional context**

btw, the collision position detecting is unstable.

AxisRay:
the collision position detecting is unstable.
first try:
```
[usb] pm3 --> hf 14a list
[=] downloading tracelog data from device
[+] Recorded activity (trace len = 59 bytes)
[...]

PR #1412, opened 2021-08-03
Bug Fix : 14a anti-collision not working #1411

AxisRay:
#1411 

at line 2574 of armsrc/iso14443a.c
```c
for (uint16_t i = collision_answer_offset; i < (Demod.len - 1) * 8; i++, uid_resp_bits++) {
```
One less cycle 

should be

```c
for (uint16_t i = collision_answer_offset; i < Demod.len * 8; i++, uid_resp_bits++) {
```

But, it's still not perfect.
The collision detecting is very unstable.
I don't know how to solve it


iceman1001 (changes requested):
.gitigonore,  launch_linux, launsh_wsl and the proxmark3.c files doesn't relate to your bug fix

AxisRay:
Removed

AxisRay:
remove redundant commits  and do more test
then resubmit PR

PR #1418, opened 2021-08-04
Electra branch

doegox:
Not so easy to get a smooth integration of Electra support.
Please have a look before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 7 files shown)


doegox:
Example:
image

When parity also matches an EM XL ID:
image

doegox:
Line "Possible an Electra" doesn't appear if extra data is more than 3 niggles + zeroes.

iceman1001:
Cool,  and yes,   that was some nestled ifs to make this happen

NoSoup:
Hey, thank you for your work on adding support for electra. I've tried it just now with the changes from master and with a tag from Electra (from my intercom). Unfortunately `lf search` wasn't able to identify it :(.

doegox:
@NoSoup There are apparently different versions of Electra tag. 
Do you get some signal with `lf read` + `data plot` ? 
If yes, can you share the trace ? `data save -f myelectrafile`
You can reach me on the Discord server if that's easier

PR #1423, opened 2021-08-05
Improve user experience

AxisRay:
Improve user experience
Before:
image
After:
image

### DIFF
--- a/client/src/proxmark3.c
+++ b/client/src/proxmark3.c
@@ -127,6 +127,15 @@ static void prompt_compose(char *buf, size_t buflen, const char *promptctx, cons
 static int check_comm(void) {
     // If communications thread goes down. Device disconnected then this should hook up PM3 again.
     if (IsCommunicationThreadDead() && session.pm3_present) {
+#ifdef HAVE_READLINE
+        char* saved_line;
+        int saved_point;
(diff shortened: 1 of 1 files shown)


iceman1001:
Interesting find...

doegox:
Good catch but the solution sounds overcomplicated.
It seems this simple patch has the same effect.
```diff
diff --git a/client/src/proxmark3.c b/client/src/proxmark3.c
index d1507923a..d1f74bcba 100644
--- a/client/src/proxmark3.c
+++ b/client/src/proxmark3.c
@@ -135,7 +135,7 @@ static int check_comm(void) {
         char prompt_filtered[PROXPROMPT_MAX_SIZE] = {0};
         memcpy_filter_ansi(prompt_filtered, prompt, sizeof(prompt_filtered), !session.supports_colors);
         rl_set_prompt(prompt_filtered);
-        rl_forced_update_display();
+        rl_redisplay();
 #endif
         CloseProxmark(session.current_device);
     }
```
Do you confirm @AxisRay or did I miss sth ?

AxisRay:
Confirmed!
It seems that no different.
image

doegox:
thank you

PR #1428, opened 2021-08-05
Fix DESfire DF search and assignment to AID

craftbyte:
Previous implementation didn't work for cards with single DF app.


iceman1001:
Since @merlokk  is in the process of remaking stuff,  maybe leave PR's for desfire until he is done.
Start a issue to log found bugs,

merlokk:
thank you)

Issue #1426, opened 2021-08-05
[BUG] hf mfu dump return cmd error:00


AxisRay:
try to dump tag of the xiaomi air purifier filter with `hf mfu dump`
the key is `bc77b6ed` (sniff)
cmd returned `Cmd Error: 00` 
```
[fpc] pm3 --> hf mfu dump -k bc77b6ed
[+] TYPE: NTAG UNKNOWN  
[+] Reading tag memory...
[#] Cmd Error: 00
[#] Read block 4 error
[=] MFU dump file information
[=] -------------------------------------------------------------
[=]       Version | 00 53 04 02 01 00 0F 03 
[=]         TBD 0 | 00 00 
[=]         TBD 1 | 00 
[=]     Signature | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
[=]     Counter 0 | 00 00 00 
[=]     Tearing 0 | 00 
[=]     Counter 1 | 00 00 00 
[=]     Tearing 1 | 00 
[=]     Counter 2 | 00 00 00 
[=]     Tearing 2 | 00 
[=] Max data page | 2 (12 bytes)
[=]   Header size | 56
[=] -------------------------------------------------------------
[=] block#   | data        |lck| ascii
[=] ---------+-------------+---+------
[=]   0/0x00 | 53 5C 4D CA |   | S\M.
[=]   1/0x01 | EE 02 1F 00 |   | ....
[=]   2/0x02 | F3 48 00 00 |   | .H..
[=]   3/0x03 | E1 10 12 00 | 0 | ....
[=] ---------------------------------
[=] Using UID as filename
[+] saved 72 bytes to binary file hf-mfu-535C4DEE021F00-dump-8.bin
[+] saved to json file hf-mfu-535C4DEE021F00-dump-8.json
[!] ⚠️  Partial dump created. (4 of 42 blocks)
```
the trace shows that the pm3 didn't send READ BLOCK cmd
```
[fpc] pm3 --> trace list -t 14a
[=] downloading tracelog data from device
[+] Recorded activity (trace len = 193 bytes)
[=] start = start of start frame end = end of frame. src = source of transfer
[=] ISO14443A - all times are in carrier periods (1/13.56MHz)

      Start |        End | Src | Data (! denotes parity error)                                           | CRC | Annotation
[...]

PR #1430, opened 2021-08-06
Add support for Shanghai Feiju NTAG213

AxisRay:
#1426


iceman1001:
Nice!
Now it just need to be mapped to NTAG213  max blocks etc...

PR #1433, opened 2021-08-07
Add new standalone mode MFCSIM

AxisRay:
A really simple standalone mode
On entering stand-alone mode, this module will start simulating.
Data is read from bin dump file that uploaded to flash memory (hf_mfcsim_dump_1.bin).
Only support mifare classic 1k
hope someone will enjoy it


iceman1001:
Always welcome to  see code contributions.
You need to add a line in CHANGELOG.md  and  inside doc/md folder there is one more text file to modify

AxisRay:
Fixed!
I should have read the guide more carefully.

iceman1001:
Final textual thing , your standalone mode is targetted to rdv4 only.   So in the description,  you need to point that out  in your description  inside  armsrc/Standalone/Makefile.hal  

like add ` (RDV4 only) `

armsrc/Standalone/hf_mfcsim.c
```diff
@@ -0,0 +1,116 @@
...
+ * - LED C blinking: data transmiting
+ *
+ * To upload input file (eml format) to flash:
+ * - mem spiffs upload -s <filename> -d hf_mfcsim_dump_1.bin
+ * To delete the input file from flash:
+ * - mem spiffs remove -f hf_mfcsim_dump_1.bin
+ *
+ */
+
+#define HF_MFCSIM_INPUTFILE_SIM         "hf_mfcsim_dump_1.bin"
```
iceman1001, line 48: and maybe remove the `_1`  part in the file name.  a bit hard to remember

mistial-dev: I was thinking of doing a follow-up patch to allow a few different slots, though I admittedly haven't looked at how difficult that would be to do.

If there are multiple slots, having them be _1, _2, _3, and _4 would make a degree of sense.

iceman1001: that would make some sense.   But for now it doesn't support it.

AxisRay: The difficulty is how to indicate which card is currently being simulated.
the function "Mifare1ksim" has used all LEDs.
sadness....

iceman1001: you can flash like led the number of times == active slot?

iceman1001:
Thanks!

PR #1447, opened 2021-08-12
hf_mfcsim refactored

AxisRay:
Add multi cards support 
Use LED to indicate card slot


iceman1001:
Interesting, but are you trying to load 16 slots of 1kb memory into the 4kb reserved emulator memory?  Or was it 8kb nowdays. 
I would suggest you only read a file and load emul with that one,   every time you click the button to swap slot, you just load from spiffs the next one.  In a round loop.

AxisRay:
Done

PR #1457, opened 2021-08-16
Added keys from norwegian building site cards

micsen:
The sectors contain Name, Birthdate, Org no, Org name. etc


iceman1001:
Thanks!

PR #1487, opened 2021-08-28
Fix PAC/Stanley checksum calculation

rknoll:
The parity bit needs to be discarded when calculating the final
checksum byte value of PAC/Stanley card IDs. Verified by scanning
a tag which is now correctly identified.


doegox:
consider adding a line in CHANGELOG.md
thanks !

rknoll:
> consider adding a line in CHANGELOG.md
> thanks !

Done :-)

PR #1504, opened 2021-09-11
fix broken keycheck procedure in staticnested process

horrordash:


horrordash:
I think some clarification is needed here :)
in fact, the command parameters are inserted into the packet in the function mfCheсkKeys. when it is called a block of data should contain only keys. fix removes duplicate command parameters, otherwise the search for keys does not work

iceman1001:
Thanks for the explanation.
That code smells.  I rather would have a struct_t instead of those assignments.  Which would make things easier to understand whats going on.

iceman1001:
ok,  Looks like the `mfCheckKeys` call adds the header itself.
but the problem is that `mfCheckKeys_file` call doesn't add a header.

So this PR needs to adapt to both use-cases.

line 670-675  needs to be different according it `IfPm3Flash()` == true or not.

Does this make sense to you?

iceman1001:
@horrordash   Would you mind pulling latest and test,  I believe I got the correct process.

PR #1506, opened 2021-09-13
check coverity for va_start - va_arg - va_end

merlokk:


doegox:
Indeed cppcheck doesn't complain anymore on cmdhffido about variadic args

merlokk:
i don't see the other hints from emv part... it's strange...  have you see it?

doegox:
yes still complaining on emv: (but I was using cppcheck, not coverity)
e.g.:
```
data = emv_pki_decode_message(enc_pk, msgtype, &data_len,
                                  cert_tlv,
                                  5,
                                  rem_tlv,
                                  exp_tlv,
                                  add_tlv,
                                  sdatl_tlv,<--- Passing NULL after the last typed argument to a variadic function leads to undefined behaviour. [+]
                                  NULL);
```

doegox:
ok, all cppcheck warnings related to variadics are gone.
(it doesn't make any difference in coverity which doesn't care about these)

PR #1509, opened 2021-09-18
Windows: fixed usb prioritization

Gator96100:
There is a bug in the current pm3 script that will not prioritise USB connections. It has to do with skipping the usb part if there was a bt port detected. 

@Iceman1001 did add the comment `Need to look for this first, the call to Win32_serialport "crashes" then native bt serial port. Don't ask why.`. Since then, there where some changes to the queries and I can not replicate that issue. 

Thanks to @craftbyte for bringing this issue to my attention and help me fix it.

### DIFF
--- a/pm3
+++ b/pm3
@@ -135,7 +135,23 @@ function get_pm3_list_macOS {
 function get_pm3_list_Windows {
     N=$1
     PM3LIST=()
-    # Need to look for this first,  the call to Win32_serialport "crashes" then native bt serial port.  Don't ask why.
+	
+	# Normal SERIAL PORTS (COM)
+    for DEV in $(wmic /locale:ms_409 path Win32_SerialPort Where "PNPDeviceID LIKE '%VID_9AC4&PID_4B8F%' Or PNPDeviceID LIKE '%VID_2D2D&PID_504D%'" Get DeviceID 2>/dev/null | awk -b '/^COM/{print $1}'); do
+        DEV=${DEV/ */}
+        #prevent soft bricking when using pm3-flash-all on an outdated bootloader
+        if [ $(basename -- "$0") = "pm3-flash-all" ]; then
+            if [ ! $(wmic /locale:ms_409 path Win32_SerialPort Where "DeviceID='$DEV'" Get PNPDeviceID 2>/dev/null | awk -b '/^USB/{print $1}') = "USB\VID_9AC4&PID_4B8F\ICEMAN" ]; then
+                echo -e "\033[0;31m[!] Using pm3-flash-all on an oudated bootloader, use pm3-flash-bootrom first!"
+                exit 1
+            fi
+        fi
+        PM3LIST+=("$DEV")
+        if [ ${#PM3LIST[*]} -ge "$N" ]; then
+            return
+        fi
+    done
+	
     #BT direct SERIAL PORTS (COM)
     if $FINDBTRFCOMM; then
         for DEV in $(wmic /locale:ms_409 path Win32_PnPEntity Where "Caption LIKE '%Bluetooth%(COM%'" Get Name 2> /dev/null | awk -b 'match($0,/(COM[0-9]+)/,m){print m[1]}'); do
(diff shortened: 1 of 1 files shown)


iceman1001:
The original problem was that the wmi calls crashes the com port for the BT and it becomes unresponsive when we try to open it with the pm3 client.  Changing the order was a work around.  How did you solve the wmi crashing the com port ?

Gator96100:
I wasn’t able to replicate the original issue. I did try it with f80f12306 so the commit before the workaround.

The only issue I found was when switching between USB and BT connection is that there is some sort of timeout for switching between ports. When switching from USB to BT I had to wait a few sounds before I could open the BT port. Switching from BT to USB was usually fine, only once I couldn’t open the port, but running the pm3 script again worked.

Maybe it is a bug with your Bluetooth adapter?

iceman1001:
The port delay you are mentioning sounds like windows trying to recover the com port.   Also sounds like that timeout could be the culprit for the script.  Since the script doesn't wait for the com port to be available again.  

If my device is faulty?  That is a possibility if it wasn't for other people having the same issue. 

Did you try with WSL,  native BT or  WSL white dongle?

Gator96100:
I did try WSL and ProxSpace with a cheap USB dongle. 

It could also be a Windows bug that was fixed in an update.

Maybe @craftbyte can test if the bug exists on f80f123

iceman1001:
I still have it on my WSL1 and the white dongle,   before testing your PR.

Gator96100:
I don’t have a HC-06-USB dongle to test with, but maybe it has to do with needing to wait between wmi calls.  

Does `.pm3 –list` work currently? Is it just opening the com port that fails?

iceman1001:
WSL1  is a bit annoy with the white dongle

```
iceman@TAU:~/prox$ ./pm3 --list
1: /dev/ttyS3                     --< USB
2: /dev/ttyS12                   --< BT
``` 

```
iceman@TAU:~/prox$ ./pm3 -n 2                                                   
[=] Session log /home/iceman/.proxmark3/logs/log_20210919.txt                   
[+] loaded from JSON file /home/iceman/.proxmark3/preferences.json              
[=] Using UART port /dev/ttyS12                                                 
                                                                                
[!] ⚠️  ERROR: invalid serial port /dev/ttyS12                                  
                                                                                
[?] Try the shell script `./pm3 --list` to get a list of possible serial ports  
```


While Ubuntu21 over VM,  is quite ok

```
osboxes@osboxes:~/rdv44$ ./pm3 --list
1: /dev/ttyACM0
2: /dev/ttyUSB0
```

```
osboxes@osboxes:~/rdv44$ ./pm3 -n 2
[=] Session log /home/osboxes/.proxmark3/logs/log_20210919.txt
[+] loaded from JSON file /home/osboxes/.proxmark3/preferences.json
[=] Using UART port /dev/ttyUSB0
[=] Communicating with PM3 over FPC UART
[=] PM3 UART serial baudrate: 115200



  ██████╗ ███╗   ███╗█████╗ 
  ██╔══██╗████╗ ████║╚═══██╗
  ██████╔╝██╔████╔██║ ████╔╝
  ██╔═══╝ ██║╚██╔╝██║ ╚══██╗
  ██║     ██║ ╚═╝ ██║█████╔╝ 
  ╚═╝     ╚═╝     ╚═╝╚════╝     [ ❄️ Iceman ❄️ ]




 [ Proxmark3 RFID instrument ]

 [ CLIENT ]
  client: RRG/Iceman/master/v4.14434-7-g787098f3a 2021-09-19 17:58:35
  compiled with GCC 9.4.0 OS:Linux ARCH:x86_64

 [ PROXMARK3 ]
  device.................... RDV4
  firmware.................. RDV4
  external flash............ present
  smartcard reader.......... present
  FPC USART for BT add-on... present

```

iceman1001:
Most likely wait between WMI calls.    That wait seems very different in length on different systems.

iceman1001:
I realize now that we would need to test with native BT instead of dongle...

iceman1001:
Trying current code on native BT w WSl-1

First bug,  --list and -n doesn't match.  As seen,  -n 1 should have connected to ttyS3..  it didn't.
second bug,  can't connect over BT at all..

```
iceman@LAPTOP-02:~/rdv42$ ./pm3 --list
1: /dev/ttyS3
2: /dev/ttyS10
iceman@LAPTOP-02:~/rdv42$ ./pm3 -n 1
[=] Session log /home/iceman/.proxmark3/logs/log_20210919.txt
[+] loaded from JSON file /home/iceman/.proxmark3/preferences.json
[=] Using UART port /dev/ttyS10

[!] ⚠️  ERROR: invalid serial port /dev/ttyS10

iceman@LAPTOP-02:~/rdv42$ ./pm3 -n 2
[=] Session log /home/iceman/.proxmark3/logs/log_20210919.txt
[+] loaded from JSON file /home/iceman/.proxmark3/preferences.json
[=] Using UART port /dev/ttyS10

[!] ⚠️  ERROR: invalid serial port /dev/ttyS10

```

Gator96100:
The first bug was what we found yesterday, it has to do with the check if there was a pm3 found. I think that’s on me.
```
if [ ${#PM3LIST[*]} -ge "$N" ]; then
    return
fi
```
I’m sure I don’t have to tell you, but did you compile with btaddon? I did forget it and was wondering why it didn’t connect.

iceman1001:
Most people use termux and android w blueshark.
The windows wsl experience is not so nice.
Maybe its better with ProxSpace. 

I will merge anyway,   the messed up com port isn't our fault but Windows enumeration.

PR #1510, opened 2021-09-19
fix crack5opencl queue/threads.c

### LINKED ISSUE #1503: An improper locking bug(e.g., deadlock) on the lock ctx->queue_mutex
Hi, developers, thank you for your checking. It seems the lock **ctx->queue_mutex** is not released correctly when **ret != QUEUE_EMPTY** in the function **wu_queue_destroy**?

@matrix @doegox

https://github.com/RfidResearchGroup/proxmark3/blob/9da0073e0dd9758cece4bf215654c4550929c971/tools/hitag2crack/crack5opencl/queue.c#L374-L385

matrix:
fix #1503


iceman1001:
Thanks!

PR #1515, opened 2021-10-05
New 'lf hitag pwmdemod' cmd decoding the reader messages from graph b…

codeflakes0:
This new command is based on the 'data rawdemode' commands.
It is intended to be used on graph buffer data captured with a previous 'lf search'
Since the decode is done in variable bit range I've had to modify proxguiqt.cpp to take into account. Not sure
if this was the best way to deal with this problem.
There is a know bug when zooming and scrolling, the displayed demodulation bits are shifted and do not follow the signal.

### DIFF
--- a/common/lfdemod.c
+++ b/common/lfdemod.c
@@ -423,6 +423,17 @@ static size_t findModStart(uint8_t *src, size_t size, uint8_t expWaveSize) {
     return i;
 }
 
+static size_t findModStart2(uint8_t *src, size_t size, uint8_t expWaveSize) {
+    uint32_t i = 0;
+    uint8_t threshold = signalprop.amplitude / 10; // 10 %
+    for (; i < size - 20; i++) {
+        if (abs(src[i] - signalprop.mean) > threshold) {
+            break;
+        }
+    }
+    return i;
+}
+
 static int getClosestClock(int testclk) {
     uint16_t clocks[] = {8, 16, 32, 40, 50, 64, 100, 128, 256, 384};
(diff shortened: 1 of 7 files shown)


iceman1001:
hm.. interesting..
maybe it should have been under `data pwm`   and not under hitag.
Not sure about the graph buffer doing.
the normal way is to copy it to your local array,   put all demods into the demod array...
and thats it.

common/lfdemod.c
```diff
@@ -1836,6 +1847,83 @@ int nrzRawDemod(uint8_t *dest, size_t *size, int *clk, int *invert, int *startId
...
+        if (dest[idx] < wave_min) {
+            wave_min = dest[idx];
+            idx_min = idx;
+        }
+        if (dest[idx] > wave_max) {
+            wave_max = dest[idx];
+            idx_max = idx;
+        }
+
+        uint32_t period = idx - last_wave_start;
```
chrisfu, line 1884: Would this work as a `int32_t`? As it currently stands, builds are failing under the Darwin compiler due to `period` being unsigned.

mwalker33: What happens if "period" and "last_wave_start" are made size_t  (to match idx) rather then uint32_t ?
i.e. is size_t the same size as uint32_t ? (on some systems I believe size_t is 8 bytes, which would make it 64 bits, twice the size of a unit32_t.

iceman1001:
A complete rework of hitag is coming.     Closing.

PR #1516, opened 2021-10-08
Fix the exit code of build_all_firmwares.sh

sbrun:
Thanks for providing the build_all_firmwares.sh script.
We now use it for the package build in Kali but we had a small issue with the exit code of the script (= of the latest command in the script)

The command ``` $STATS && ( LANG=C arm-none-eabi-size armsrc/obj/[hl]f_*.o |grep -v "filename" >> standalones_stats.txt )``` is always false when $STATS is false (as it's defined by default).
So the exit code of the script is "1" and the build in Kali fails.


doegox:
Ha yes thank you!
Never hesitate to open a PR if you need small changes, rather than having to maintain a patch in the Debian build.

PR #1528, opened 2021-11-12
Added standalone mode that can load multiple different modes

dankar:
This mode adds a way of having multiple standalone modes at the same time, and choosing between them when entering standalone mode. It's at least useful for me to have that.

I have not tried this with all the different modes that exists, but a quick test with a few different ones seems to work fine.

The code is very messy since it's done just with some macros. It would be much cleaner to use some code generation, but this works.


iceman1001:
Interesting,   not quite how I saw multi standalone modes would look like.   A different solution indeed.

dankar:
> Interesting, not quite how I saw multi standalone modes would look like. A different solution indeed.

Feel free to ignore this PR if a more robust way is incoming. It was just a quick useful hack. :)

iceman1001:
I haven't seen anyone working on switching standalone modes functionality for a long time.

armsrc/Standalone/dankarmulti.c
```diff
@@ -0,0 +1,138 @@
...
+
+#define MODE_NAME sniff14a
+#define MODE_FILE "hf_14asniff.c"
+#include "dankarmulti.h"
+#define MODE_NAME em4100
+#define MODE_FILE "lf_em4100rswb.c"
+#include "dankarmulti.h"
+#define MODE_NAME icehid
+#define MODE_FILE "lf_icehid.c"
+#include "dankarmulti.h"
```
iceman1001, line 66: question,
how does re-defining the same macro work in the add_mode down below?

dankar: The same name is used again for ADD_MODE simply to add the different modes to a list. It simplifies enumerating the included modes.

iceman1001: yes,  but the redefine is done before the ADD_MODE macro..

dankar: The macros are undef:d in the "dankarmulti.h" header, at the end of it, meaning they can be redefined for the next mode.

iceman1001: .... sneaky way of dealing with macro defines...   Not sure about it,

iceman1001:
@doegox  what do you think?   Shall we merge or not?   I am skeptical to the "innovative" way of using macros,   then on the other hand...  bleeding edge.

@dankar    You think you can comment your code to explain how those macros works,  so it doesn't confuse ppl like me?

doegox:
Imho we can merge :)

dankar:
Give me a day and I can add to the comments about what it actually does :)

iceman1001:
You can make a new PR with comments :)

Thanks!

PR #1529, opened 2021-11-17
Fixed ledcontrol in SimulateTagLowFrequencyEx

zabszk:
Fixed the issue where SimulateTagLowFrequencyEx in some cases was controlling the LEDs even if ledcontrol argument was set to false.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Fixed `hf mfdes` - fix buffer overflow (@AxisRay)
  - Changed `hf mfdes` - detect LRP mode and info updates (@merlokk)
  - Fixed `hf mfdes` - increase response buffer length (@merlokk)
+ - Fixed `SimulateTagLowFrequencyEx` ignoring the `ledcontrol` argument (@zabszk)
 
 ## [crimson.4.14434][2021-09-18]
  - Fixed `hf mf staticnested` - flashmem / non loop now works (@horrordash)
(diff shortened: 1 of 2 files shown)


iceman1001:
I am curious,  where was the issue when calling `SimulateTagLowFrequencyEx` with ledcontrol false?

zabszk:
> 
> 
> I am curious, where was the issue when calling `SimulateTagLowFrequencyEx` with ledcontrol false?

`ledcontrol` is supposed to determine whether the function should control LEDs or not and that's how it works on lines 858 and 872. If it's set to true then the function turns on and off the LED, if it's set to false then the function doesn't control the LED.

But it is possible that the function executes `goto OUT;` on line 866 and then the LED is always turned off, ignoring the argument `ledcontrol` telling the script to not change the LEDs.

I had this issue in my standalone mode. I set `ledcontrol` to false, because I wanted to control the LEDs myself and in some cases (that goto) the function turns off the LED ignoring the `ledcontrol` being set to false.

iceman1001:
ok,  I haven't seen anyone using ledcontrol before.  Like in seven years. 
To be honest,  I would remove that option since its not used.   I just never got around to do it.

zabszk:
> 
> 
> ok, I haven't seen anyone using ledcontrol before. Like in seven years. To be honest, I would remove that option since its not used. I just never got around to do it.

To be honest I would love to see ledcontrol option in other lfops functions, such as functions used for reading and writing tags. If you work with stadalone mode like I do and these 4 LEDs are the only output you have, it can be sometimes problematic if r/w functions change the LEDs and you can't do anything about it.

iceman1001:
yeah,  either remove it complete or add it to all lf/hf device side commands.  The mix we have been having for ages isn't very good.  A code smell.

I will merge this one and feel free to add led-control to it all ;)

iceman1001:
Thanks!

zabszk:
> 
> 
> yeah, either remove it complete or add it to all lf/hf device side commands. The mix we have been having for ages isn't very good. A code smell.
> 
> I will merge this one and feel free to add led-control to it all ;)

Thanks! I agree that the mix isn't good. I was thinking about adding such argument when I noticed that one function for some reason already have it while the other functions don't.

I will add led control to all other functions.

PR #1530, opened 2021-11-18
armsrc/felica: fix crc offsets in felica_sim_lite

pieterg:
the sync bytes should not be included in the crc calculation
(and the crc bytes should be placed at the end of the response buffers)


iceman1001:
Thanks!

FeliCa could do with some more love!

PR #1531, opened 2021-11-18
Added ledcontrol to LF operations

zabszk:
As discussed in #1529 I have added ledcontrol argument to all LF operations.


iceman1001:
:)

PR #1532, opened 2021-11-18
armsrc/felica: fix handling POLL with Request Code 1

pieterg:
apparent typo, listenmode should be set to false in order to send the
prepared response


iceman1001:
Thanks!

PR #1534, opened 2021-11-19
armsrc/felica: add polling response time slot support to felica_sim_lite

pieterg:
Make sure we respond in a valid time slot.
We were responding too early, the first time slot (which is always allowed) starts after 512 * 64 / fc.

Also, respond in incremental time slots. We could have used a fixed time slot (#0 is always allowed), but by using incremental time slots we have a better chance the reader will get our response in the expected time frame.
(when our timing is wrong, we have a risk of responding too early when using the first slot, and too late when using the last slot. Slots in the middle of the allowed range have the best chance)

### DIFF
--- a/armsrc/felica.c
+++ b/armsrc/felica.c
@@ -735,6 +735,7 @@ void felica_sim_lite(uint8_t *uid) {
                     if (FelicaFrame.crc_ok) {
 
                         if (FelicaFrame.framebytes[2] == 6 && FelicaFrame.framebytes[3] == 0) {
+                            static uint8_t timeslot = 0;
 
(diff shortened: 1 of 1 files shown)


iceman1001:
Thanks!

I'm looking forward to more FeliCa love!

pieterg:
you're welcome!

Not sure whether I'll come across more FeliCa fixes, all I needed was a working simulator.
(which we have now)

iceman1001:
You can have a look at the normal FeliCa commands and make them perfect :)

Somehow, I get the felling you have the right knowledge.

Feel free to send me an email.

PR #1535, opened 2021-11-21
Fix possible race issues in crack5opencl

ryancaicse:
Fix possible race issues by recovering the lock

### DIFF
--- a/tools/hitag2crack/crack5opencl/queue.c
+++ b/tools/hitag2crack/crack5opencl/queue.c
@@ -382,6 +382,8 @@ int wu_queue_destroy(wu_queue_ctx_t *ctx) {
 #endif
         return ret;
     }
+    
+    pthread_mutex_lock(&ctx->queue_mutex); 
 
 #if TEST_UNIT == 1
     printf("ret from wu_queue_pop() (%d): %s\n", ret, wu_queue_strerror(ret));
@@ -395,7 +397,9 @@ int wu_queue_destroy(wu_queue_ctx_t *ctx) {
     ctx->queue_head = 0; //NULL;
     ctx->queue_tail = 0; //NULL;
     ctx->init = 0;
-
+    
+    pthread_mutex_unlock(&ctx->queue_mutex);
+    
     pthread_mutex_destroy(&ctx->queue_mutex);
     pthread_mutexattr_destroy(&ctx->queue_mutex_attr);
 


matrix:
hi @ryancaicse, honestly I don't understand why it is necessary to reintroduce the lock / unlock of the mutex in this position. 
The "wu_queue_destroy" function is called at the end of main, why should there be a deadlock here?

ryancaicse:
@matrix , there are no deadlocks but data races. Because there is an original lock `ctx->queue_mutex`, which is entirely removed by  #1510. The lock is necessary to guard these variables.

https://github.com/RfidResearchGroup/proxmark3/blob/9da0073e0dd9758cece4bf215654c4550929c971/tools/hitag2crack/crack5opencl/queue.c#L388-L399

matrix:
@doegox @iceman1001 for me it is not necessary, what do you think?

ryancaicse:
In the `wu_queue_pop` method, `ctx->queue_head, ctx->queue_tail` are protected by locks.
https://github.com/RfidResearchGroup/proxmark3/blob/9da0073e0dd9758cece4bf215654c4550929c971/tools/hitag2crack/crack5opencl/queue.c#L296-L324

But the lock is absent subsequently (originally have but be removed)
https://github.com/RfidResearchGroup/proxmark3/blob/9da0073e0dd9758cece4bf215654c4550929c971/tools/hitag2crack/crack5opencl/queue.c#L379-L399

iceman1001:
There was a dialogue about the why the changes was introduced to start with.

Now it seem to be about `data race conditions`  ...  

Can we get an example of when this happens?  a test case?

matrix:
> wu_queue_pop

the "wu_queue_pop" functions it's called many times, the "wu_queue_destroy" only one time at the end of program.

ryancaicse:
In this place, I guess the threads are destroyed before invoking wu_queue_destroy.
https://github.com/RfidResearchGroup/proxmark3/blob/88308ea7277329c5806a0eaddf7f6840c1b0d999/tools/hitag2crack/crack5opencl/ht2crack5opencl.c#L1198-L1215

In this place, would this test be invoked by multiple threads?
https://github.com/RfidResearchGroup/proxmark3/blob/88308ea7277329c5806a0eaddf7f6840c1b0d999/tools/hitag2crack/crack5opencl/queue.c#L440-L500

If both are safe, the lock is indeed unnecessary.

iceman1001:
@matrix    your call,   merge or not?

matrix:
Hi @iceman1001, not without a test case

PR #1538, opened 2021-11-27
Added explanation of how dankarmulti works

dankar:
As mentioned, a short description of how this works.


iceman1001:
:)
Thanks!

PR #1544, opened 2021-12-14
Create ntag_bruteforce.lua

keldnorman:
PRE: 

-- 1. Install lua rocks:             apt install luarocks
-- 2. Then install lua md5 module:   luarocks install md5

# ntag_bruteforce
LUA Script to Bruteforce NFC tags of the type NTAG21x using a Proxmark RDV4.0 - only for 32-bit codes.

Place this script in your local proxmark folder or in ${HOME}/.proxmark3/luascripts/

Then start proxmark3: 

root@pc:~/source/Proxmark# service ModemManager stop
root@pc:~/source/Proxmark# stty -F /dev/ttyACM0 115200 cs8 -cstopb -parenb raw -echo -echoe -echok
root@pc:~/source/Proxmark# proxmark3 /dev/ttyACM0 

[=] Session log /root/.proxmark3/logs/log_20211031.txt
[+] loaded from JSON file /root/.proxmark3/preferences.json
[=] Using UART port /dev/ttyACM0
[=] Communicating with PM3 over USB-CDC

  ██████╗ ███╗   ███╗█████╗ 
  ██╔══██╗████╗ ████║╚═══██╗
  ██████╔╝██╔████╔██║ ████╔╝
  ██╔═══╝ ██║╚██╔╝██║ ╚══██╗
  ██║     ██║ ╚═╝ ██║█████╔╝ 
  ╚═╝     ╚═╝     ╚═╝╚════╝     [ ❄ Iceman ❄ ]

 [ Proxmark3 RFID instrument ]

 [ CLIENT ]
  client: RRG/Iceman/master/v4.14434-17-g1cc1a3857 2021-10-06 14:24:04
  compiled with GCC 10.3.0 OS:Linux ARCH:x86_64

 [ PROXMARK3 ]
  device.................... RDV4
  firmware.................. RDV4
  external flash............ present
  smartcard reader.......... present
  FPC USART for BT add-on... absent

 [ ARM ]
  bootrom: RRG/Iceman/master/v4.14434-17-g1cc1a3857 2021-10-06 14:24:24
       os: RRG/Iceman/master/v4.14434-17-g1cc1a3857 2021-10-06 14:24:37
  compiled with GCC 8.3.1 20190703 (release) [gcc-8-branch revision 273027]

 [ FPGA ] 
  LF image built for 2s30vq100 on 2020-07-08 at 23:08:07
  HF image built for 2s30vq100 on 2020-07-08 at 23:08:19
  HF FeliCa image built for 2s30vq100 on 2020-07-08 at 23:08:30

 [ Hardware ]
  --= uC: AT91SAM7S512 Rev B
  --= Embedded flash memory 512K bytes ( 59% used )

HOW TO RUN THE SCRIPT:

[usb] pm3 --> script run ntag_bruteforce.lua -i  /root/.proxmark3/luascripts/keys/testkeys.list -h

Script      : ntag_bruteforce.lua
Author      : Keld Norman
Version     : 1.0.0
Description : Bruteforces 7 byte UID NTAG protected with a 32 bit password

      .-.
     /   \         .-.
    /     \       /   \       .-.     .-.     _   _
+--/-------\-----/-----\-----/---\---/---\---/-\-/-\/\/---
  /         \   /       \   /     '-'     '-'
 /           '-'         '-'
 
Usage:

script run ntag_bruteforce [-s <start_id>] [-e <end_id>] [-t <timeout>] [ -o <output_file> ] [ -p ] [ -h for help ]
script run ntag_bruteforce [-i <input_file>] [-t <timeout>] [ -o <output_file> ] [ -n | -x ] [ -p ] [ -h for help ]

DESCRIPTION 

This script will test either an 8 digit hexadecimal code or 4 char stings (will be converted to an 8 digit hex string )
against NFC cards of the type NTAG21x protected by a 32 bit password.

Read more about NTAGs here: https://www.nxp.com/docs/en/data-sheet/NTAG213_215_216.pdf

Arguments:

    -h                            This help
    -i       input_file           Read 4 char ASCII values to test from this file (will override -s and -e )
    -o       output_file          Write output to this file
    -t       0-99999, pause       Timeout (ms) between cards 1000 = 1 second (use the word 'pause' to wait for user input)
    -p                            Skip Ping

    # Either use the continuously test:

    -s       0-0xFFFFFFFF         Start HEX value
    -e       0-0xFFFFFFFF         End   HEX value

   # Or use a list of passwords from a file: 

    -x       Passwords in HEX     Password file (-i) contains HEX values (4 x 2hex -> 32 bit/line like: 00112233)
    -n       NTAG Tools format    Bruteforce with first 8 hex values of a md5 hash of the password 
                                  The password will be prefixed with hex value 20 (space) if the string/password is < 4 chars
EXAMPLES:
  
# Example of how to run the script with bruteforcing of continuously HEX numbers with 1 secound delay between tests:
 script run ntag_bruteforce -s 00000000 -e FFFFFFFF -t 1000  -o /var/log/ntag_bruteforce.log 

# Example of how to run the script and bruteforc the card using passwords from the input file with 1s delay between tests
 script run ntag_bruteforce -i /home/my_4_char_passwords_list.txt -o /var/log/ntag_bruteforce.log 

[+] finished ntag_bruteforce.lua

TEST A CARD WITH A CUSTOM PASSWORD LIST 
  
[usb] pm3 --> script run ntag_bruteforce.lua -i  /root/.proxmark3/luascripts/keys/testkeys.list -x

If you add -x and specify a password file to use for the bruteforce then the content will be treated as HEX so the file must contain 
one 32-bit HEX values / line
  
Example: 
  
root@pc:~/source/Proxmark# cat /here/is/my/password_list_file.txt
00000000
FFFFFFFF
12345678

Running the script with -i your_file but without the -x then the file will be treated as if the passwords were ascii
Because 32-bit = 8 bit then the ascii code must only be 4-chars long for this type of attack 
  If you add ascii strings instead then they must only be at max 4 chars/digits like this:
  
1234
abcs
NTAG
FFFF

### DIFF
--- a/client/luascripts/ntag_bruteforce.lua
+++ b/client/luascripts/ntag_bruteforce.lua
@@ -1,0 +1,424 @@
+#!/usr/bin/env lua
+os.execute("clear")
+core.console('clear')
+local DEBUG = true
+-------------------------------------------------------------------------------------------------------------
+-- PRE REQ:
+-------------------------------------------------------------------------------------------------------------
+-- 1. Install lua rocks:             apt install luarocks
+-- 2. Then install lua md5 module:   luarocks install md5
+-------------------------------------------------------------------------------------------------------------
+-- USAGE:
+-------------------------------------------------------------------------------------------------------------
+-- Run me like this (connected via USB): ./pm3 -l ntag_bruteforce.lua
+-- Run me like this (connected via Blueshark addon): ./client/proxmark3 /dev/rfcomm0 -l ./ntag_bruteforce.lua
+-------------------------------------------------------------------------------------------------------------
+-- VER  | AUTHOR         | DATE         | CHANGE
+-------------------------------------------------------------------------------------------------------------
+-- 1.0  | Keld Norman,   | 30 okt. 2021 | Initial version
+-- 1.1  | Keld Norman,   | 30 okt. 2021 | Added: Press enter to stop the script
+-- 1.2  | Keld Norman,   | 15 Nov. 2021 | Added: Check for correct hex values for bruteforcing
+-- 1.3  | Keld Norman,   | 15 Nov. 2021 | Added: Added a skip ping option
+-------------------------------------------------------------------------------------------------------------
+-- TODO:
+-------------------------------------------------------------------------------------------------------------
+-- Output file not implemented yet
+-------------------------------------------------------------------------------------------------------------
+-- SPEEDTEST
+-------------------------------------------------------------------------------------------------------------
+-- BRUTEFORCE ALL HEX COMBINATIONS:
+-- 
+-- With the -t 10 ( lowest possible delay ) and FFFFFFFF attempts or in decimal 4.294.967.295 combinations
+-- 
+-- My test showed that this script can do 255 password attempts in approxemately 170 seconds
+-- 
+-- That is : 255 / 170 = 1,5 attempt/second                                                                                                                                                                                      
+--                                                                                                                                                                                                                               
+-- So ..  4.294.967.295 combinations / 1,5 per second = 2.863.311.530 seconds and it is roughly 90 years                                                                                                                         
+--                                                                                                                                                                                                                               
+-------------------------------------------------------------------------------------------------------------                                                                                                                    
+-- PASSWORD LISTS:                                                                                                                                                                                                               
+-------------------------------------------------------------------------------------------------------------
+-- Crunch can generate all (14.776.336) combinations of 4 chars with a-z + A-Z + 0-9 like this: 
+--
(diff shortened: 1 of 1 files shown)


doegox:
Can't you use https://github.com/RfidResearchGroup/proxmark3/blob/master/client/lualibs/md5.lua rather than having to fetch an external lib ?

iceman1001:
... why does nft tool use md5 for NTAG pwds?  Am I missing something here?

iceman1001:
The existing MD5 lib,   if @keldnorman wouldn't mind testing this way
```
// how you include it
local md5 = require('md5')

// how you use it
local hash = md5.sumhexa(baseStr)
```

doegox:
> ... why does nft tool use md5 for NTAG pwds?  Am I missing something here?

No idea

iceman1001:
the md5 lib included in pm3 can do it,  I will push some minor changes.

Issue #1543, opened 2021-12-13
[idea]Upgrade HF_CRAFTBYTE.C


amerinoj:
Hi, 

When I went to test the standalone mode using hf_craftbyte I realized that isn't working to mifare 1k, reviewing the code I found the problem , isn't checking the tag type when the UID is emulated. 

If replace these lines:
--------------------------------------------------------------------------
Dbprintf("Starting simulation, press pm3-button to stop and go back to search state.");
                SimulateIso14443aTag(3, flags, card.uid, 0);
--------------------------------------------------------------------------

For these:
--------------------------------------------------------------------------
                Dbprintf("Starting simulation, press pm3-button to stop and go back to search state.");
				
					if (card.sak == 8 && card.atqa[0] == 4 && card.atqa[1] == 0) {
						Dbprintf("Mifare Classic");
						SimulateIso14443aTag(1, flags, card.uid, 0); // Mifare Classic
					} else if (card.sak == 0 && card.atqa[0] == 0x44 && card.atqa[1] == 0) {
						Dbprintf("Mifare Ultralight");
						SimulateIso14443aTag(2, flags, card.uid, 0);// Mifare Ultralight
					} else if (card.sak == 20 && card.atqa[0] == 0x44 && card.atqa[1] == 3) {
						Dbprintf("Mifare DESFire");
						SimulateIso14443aTag(3, flags, card.uid, 0);  // Mifare DESFire
					} else {
						Dbprintf("Unrecognized tag type -- defaulting to Mifare Classic emulation");
						SimulateIso14443aTag(1, flags, card.uid, 0); 
					}
--------------------------------------------------------------------------

The simulation work better.

Regards.

PR #1546, opened 2021-12-23
fix hf_craftbyte.c emulation card types

amerinoj:
Hello,

As commented:
https://github.com/RfidResearchGroup/proxmark3/issues/1543
I fixed emulation card type in hf_craftbyte.c . The original code  isn't checking the tag type when the UID is emulated.

Regards.


iceman1001:
Thanks,  looks like the same detection as in hf_young.c  now.

amerinoj:
yes, is the same piece of code.

PR #1550, opened 2021-12-30
Update ubuntu.yml

s00se:
Add upload of fullimage.elf

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -1,4 +1,4 @@
-name: Ubuntu Build and Test
+name: Ubuntu Build, Test, and Upload Firmware
 
 on:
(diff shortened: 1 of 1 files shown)


JessePinkman48:
> Add upload of fullimage.elf

iceman1001:
@doegox  and I have had a little chat about this one.   

We suggest instead that you can use the pre-compiled builds from @Gator96100  instead.

https://www.proxmarkbuilds.org/fileviewer.html

For instance,  latest available firmware with some windows binary (27mb download)  and you get a commit hash.  
Then you download that commit from repository and you can build your client from there.

image

iceman1001:
Thank you for the PR,  don't let this stop you @s00se   :)

PR #1551, opened 2022-01-04
Add `hf gallagher` commands

DarkMatterMatt:
Features:
- `reader`: Reads the Card Application Directory (CAD), then reads each credential file.
- `clone`: Supply a site key and cardholder credentials, this will create a new application on the card & add it to the CAD.
- `delete`: Delete the specified application and remove it from the CAD.
- `diversifykey`: Diversify the specified sitekey+card_uid+aid+key_num.

About me: I improved the `lf gallagher` commands about a year ago (https://github.com/RfidResearchGroup/proxmark3/pulls?q=author%3ADarkMatterMatt)


iceman1001:
Awesome stuff, 


one thing is your choice of license.   It would help if you would choose GPL3.0  and not the LGPL3.0..

doegox:
Well it's worse than that, proxmark3 codebase started long ago and it's license is GPL2.0 as at that time there were no GPL3... (https://github.com/RfidResearchGroup/proxmark3/blob/master/LICENSE.txt).
So we don't have much choice:
image

code in client/src is GPL2.0

iceman1001:
Not gonna have a discussion about the licenses here,  thanks for changing it.

The choice of adding default site key in the help text...  That will not do.  I rather see a check command (hf galll chk)  with the key inside a dictionary.   
or even just default to it w/o a help text entry.

iceman1001:
Ok.   It looks good. 
I say merge.

iceman1001:
Thanks!

PR #1553, opened 2022-01-05
Use pacman -Syu instead of pacman -Sy

zkldi:
`pacman -Sy` can result in partial upgrades and system damage if the user doesn't understand what it entails and never finishes the upgrade.

See: https://bbs.archlinux.org/viewtopic.php?id=241092


zkldi:
Not sure this is significant enough to get a changelog entry haha :sweat_smile:.

doegox:
thanks!

PR #1558, opened 2022-01-16
Update mfc_default_keys.dic

Phil242:
Adding keys for access cards of Eurothermes group in Switzerland.
Enjoy!


iceman1001:
nice,  looks like the keys doesn't change much.  Looks like a fixed number of bytes in each key.
Is this from one card or are these keys confirmed working on different cards?

Phil242:
Hi, 
I've only 2 cards issued the same day. Same keys. No way to obtain more cards.
Each key are used only one time, each sector have a different key A and B.
Except the last one, 100% different bytes, common for A & B on sector 15.
I agree, it's weird to have 4 bytes reused. 
But as it took 5200 seconds to extract them, I've published to avoid this pain to other dudez :).
Cheers

iceman1001:
Thanks!

PR #1559, opened 2022-01-17
Update wiegand_formats.c

bthedorff:
Adds support for PCSC's proprietary HID 37bit format P10004


iceman1001:
Cool,   do you have some raw hex or samples for it?   Where did you get the format specs from?

iceman1001:
Would you also mind adding a line in changelog.md?

bthedorff:
The bit structure was provided by PCSC and I referenced the DSX access control system programming to confirm same structure programming is used. See the below link for the specs as well but that site is off by 1. The only thing in programming that could be off is even/odd parity code (I based off pre-existing code) but all the cards I tested worked. 

https://helpdesk.rs2tech.com/kb/article/531-37-bit-wiegand-pcsc-awid-card-format/

Raw Data:
00000000000000061ee3085d (card # 202818 & FC 3133)

000000000000000ea63eaaa4 (card # 128341 & FC 7500)

000000000000000ea648bad4 (card # 148950 & FC 7500)

bthedorff:
@iceman1001 I'm sorry. I swear I tested it after my last change and it worked, however I tried it again and now it isn't. I must have made a mistake somewhere. Please use my latest version of code as I have confirmed it worked by downloading it straight from Github, compiled that version then tested and it worked on my cards. I also double checked the raw data and it is still the same as posted previously. So sorry for any confusion! Please let me know if you need anything else.

doegox:
I'm confused, which one is the working version? 
Is https://github.com/RfidResearchGroup/proxmark3/pull/1559/files ok ?
When I test it on your 3 samples, samples are decoded properly but parity fails.
(and btw that's a lot of hex for 37b, "0ea648bad4" should be enough)

bthedorff:
Yes, that is the correct version. I wanted to give the full output of raw data my proxmark3 gave in case it was needed/wanted.

doegox:
and about the parity check ?

bthedorff:
Going off of the site about (please ignore what I said previously, as the site is correct), I assumed it failed because the format does not use parity bits. Is that not the case? The even/odd parity bit was the one part of the code I wasn't 100% sure on but leaving this how a similar format was coded still gave me the correct card number so I left it.

doegox:
Ha ok, it might be worth to add a third state like parity: none
I'll merge your PR, thank you.

doegox:
I removed the parity scheme as we know it's not this one (as it fails on your samples) and I fixed the boundary checks on FC & CN

bthedorff:
Thank you!

PR #1560, opened 2022-01-17
Be more CONSTant with the parameters we will not modify

bettse:
If these are `const char` it makes it every so slightly easier when playing around with Swift integration.  I'm sure Swift integration shouldn't be one the strong motivations in the project, but if it doesn't hurt anything, I figure why not?

### DIFF
--- a/client/include/pm3.h
+++ b/client/include/pm3.h
@@ -18,8 +18,8 @@
 
 typedef struct pm3_device pm3;
 
-pm3 *pm3_open(char *port);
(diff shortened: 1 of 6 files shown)


bettse:

I already am in there

doegox:
Thank you!
Yes we do try to add const when possible, often guided by cppcheck but I must admit we didn't run cppcheck yet on the libpm3.
libpm3 API is very minimal at the moment, don't hesitate to propose API improvements if it can help your Swift integration.

PR #1561, opened 2022-01-25
Add from comment

bthedorff:
Adds from comment for PCSC 37Bit to match the other comments format. Adds uniformity to comments


iceman1001:
for the record,  those comment are there to track where the formats came from.  Not who did implement them.  There were a bunch of formats added at once.    Now we got AUTHORS.md  file for it and `git history`

bthedorff:
> for the record, those comment are there to track where the formats came from. Not who did implement them. There were a bunch of formats added at once. Now we got AUTHORS.md file for it and `git history`

Ahhh okay. I'm sorry, feel free to change it or I can submit a pull request to. I found a copy of it [here](https://helpdesk.rs2tech.com/kb/article/531-37-bit-wiegand-pcsc-awid-card-format/) so would you want the full link or just helpdesk.rs2tech.com ?

iceman1001:
nay,  just leave it as it is.  No worries.    It helps to know the source of the format when people want to debug things.

PR #1562, opened 2022-01-28
preprocessor macros for iOS

bettse:
More tweaks for Swift/iOS support of libpm3

The lualib one comes from http://lua-users.org/lists/lua-l/2017-09/msg00242.html


iceman1001:
Thanks!

PR #1564, opened 2022-01-28
Add a tip about poorly tuned antenna.

happyme531:


doegox:
Sorry, I feel uncomfortable with providing "official" tips which, if poorly followed, could lead to electrical shortages and permanent damages.

PR #1570, opened 2022-01-30
Add note about wiping password-protected T55xx tags

gabe-sky:
This commit adds an additional section showing how to wipe a password-protected T55xx tag, removing the password, if they don't care about preserving anything on it.


iceman1001:
hm,  hard one for me.  
I tend not to recommend people to wipe things as a general rule.   Destructive measures should be considered carefully in my view.
Maybe a `lf t55xx unprotect` command would be better?

gabe-sky:
Good thinking.  I'll open a new PR with `unprotect`.

PR #1573, opened 2022-01-31
Mifare debug

mwalker33:
Fix debug output as nr is an uint8_t of size 4 and was not printing the correct nr value.


iceman1001:
Nice catch with that debug statement!

Many commits in this PR that doesn't seem to releate....

mwalker33:
> 

mmm I thought that was cleaned up last time.
So you are correct.  The appmain ones should not be there.

PR #1574, opened 2022-01-31
Update mifareutil.c

mwalker33:
Mifare debug bug.  nr is a uint8_t size of 4 and did note output correctly.


iceman1001:
Thanks!

PR #1575, opened 2022-01-31
fix macos builds by not defining _POSIX_C_SOURCE in mfd_aes_brute.c if macos

linuxgemini:
unistd.h in macos has this snippet:

```c
#if __DARWIN_C_LEVEL >= __DARWIN_C_FULL
#define _SC_NPROCESSORS_CONF            57
#define _SC_NPROCESSORS_ONLN            58
#endif /* __DARWIN_C_LEVEL >= __DARWIN_C_FULL */
```

`__DARWIN_C_FULL` is `900000L`.

If `__POSIX_C_SOURCE` gets defined without `__DARWIN_C_SOURCE` being defined, `__DARWIN_C_LEVEL` will be assigned to `__DARWIN_C_ANSI`, which is `010000L`.

Thus this definition will remove `_SC_NPROCESSORS_CONF`.

If we don't define `__POSIX_C_SOURCE`, `__DARWIN_C_LEVEL` will be equal to `__DARWIN_C_FULL`, and `_SC_NPROCESSORS_CONF` will be available.

This commit also includes homebrew and macports prefix definitions for mfd_aes_brute.

Signed-off-by: İlteriş Yağıztegin Eroğlu <ilteris@asenkron.com.tr>

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -37,6 +37,13 @@ jobs:
       - name: Install dependencies
         run: brew install readline qt5 RfidResearchGroup/proxmark3/arm-none-eabi-gcc openssl
 
+      - name: Fix linkings for qt5 and openssl
+        working-directory: /usr/local/include
+        continue-on-error: true
+        run: |
+          ln -svf ../opt/openssl@3/include/openssl .
+          ln -svf ../opt/qt@5/include/qt .
+
       - name: Install Python dependencies
         run: |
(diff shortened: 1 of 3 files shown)


linuxgemini:
Also modified the macOS CI definition to handle brew-related symlink quirks

iceman1001:
Awesome!   Thank you,   don't forget a line in the CHANGELOG!

PR #1580, opened 2022-02-03
ARC support for AppNap toggle

bettse:
Another tweak on the road to swift support.  When compiling with xcodebuilt or swift, the `__has_feature(objc_arc)` being enabled cause it to throw errors with the `retain` and `release`.  ARC (automatic reference counting) makes them unnecessary.


iceman1001:
Thanks!

Issue #1583, opened 2022-02-03
hf mfd info - will verify signature only once after client starts


ah01:
**Describe the bug**

I have MFD EV2 (in factory default). When I start client and run `hf mfd info` everything is OK. When I run it **second time**, section _Tag Signature_ will fail with message `Card doesn't support GetSignature cmd`. Some other following information are missing as well, but signature is first.

It seems to be some problem with client. `hw reset` or physical reconnect of device will not solve the issue. I need to terminate client and start over.

I have tested several builds for windows from last ~3 months and build master in WSL with same result. One build that I have and that works for me is from August 2021. But this one is before @merlokk have rewritten most of mfd, right(?)

**To Reproduce**

1. Start client
2. Run `hf mfd info`
3. Result is OK, except error `error DESFIRESendApdu ...`, but that is visible only in full client debug mode.
2. Run `hf mfd info` again
4. Signature, free memory, ... fails.

**First run:**

```
[usb] pm3 --> hf mfd info

[=] ---------------------------------- Tag Information ----------------------------------
[+]               UID: 04 0F 28 CA 65 70 80
[+]      Batch number: CF 2D D4 51 90
[+]   Production date: week 38 / 2020

[=] --- Hardware Information
[=]    raw: 04010112001805
[=]      Vendor Id: NXP Semiconductors Germany
[=]           Type: 0x01
[=]        Subtype: 0x01
[=]        Version: 12.0 ( DESFire EV2 )
[=]   Storage size: 0x18 ( 4096 bytes )
[=]       Protocol: 0x05 ( ISO 14443-2, 14443-3 )

[=] --- Software Information
[=]    raw: 04010102011805
[=]      Vendor Id: NXP Semiconductors Germany
[=]           Type: 0x01
[=]        Subtype: 0x01
[=]        Version: 2.1
[=]   Storage size: 0x18 ( 4096 bytes )
[=]       Protocol: 0x05 ( ISO 14443-3, 14443-4 )

[=] --------------------------------- Card capabilities ---------------------------------

[=] --- Tag Signature
[=]  IC signature public key name: DESFire EV2
[=] IC signature public key value: 04B304DC4C615F5326FE9383DDEC9AA8
[=]                              : 92DF3A57FA7FFB3276192BC0EAA252ED
[=]                              : 45A865E3B093A3D0DCE5BE29E92F1392
[=]                              : CE7DE321E3E5C52B3A
[...]

PR #1584, opened 2022-02-03
Fix wrong log message value

ah01:
Found during troubleshooting #1583 (unfortunately this is not reason of the bug 😞 )


iceman1001:
Nice catch!

PR #1589, opened 2022-02-05
New luascript file to format Sector Trailers on the MiFare emulator

nisgola:
Added a new script that writes default or user defined Keys and Access bits to the emulator memory blocks.


iceman1001:
interesting one...
Try using parameter to the functions  
`
local myfct(sector, block, keya,keyb)
` 

call it with
`
myfct(1, 0, "ffffffffffff", "ffffffffffff")
`

nisgola:
Would it be better to remove the 4K option and just assign all available Sector trailers by default ? Trade emulator memory for code simplicity ?

iceman1001:
the globals isn't nice,
Have you considered skipping the sector loops,  and just just the blocks as base for the loops?

iceman1001:
thanks!

PR #1591, opened 2022-02-05
add silicon info logic to hf mfu info (only for nxp chips)

linuxgemini:
Signed-off-by: İlteriş Eroğlu <ilteris@asenkron.com.tr>


iceman1001:
Nice one!

PR #1592, opened 2022-02-07
Enhance hf mf uidbruteforces (UID length, HF MF 4K, help)

cactuschibre:
- Better interpretation of 4 or 7 bytes UID in UID bruteforcer
- Add Mifare Classic 4K in UID bruteforcer
- Fix help menu in UID bruteforcer


iceman1001:
Thanks

PR #1594, opened 2022-02-09
port cve-2021-34600 poc, fix device desfire aes crypto

y-x41:
Hey!

This PR adds support for performing the attack enabled by CVE-2021-34600. Details on how this attack works can be found on [our blog](https://x41-dsec.de/lab/blog/telenot-complex-insecure-keygen/).
This isn't 100% complete, yet, but since I won't be able to work on this until March I am already creating a pull request so that anyone who's interested can play around with it or potentially provide some feedback in the meantime.

A few issues remain:

  - Cancelling the simulator via the Proxmark's button does not work properly. While it seems that the simulation is stopped, it's not possible to enter any new commands on the client side. The Proxmark needs to be powercycled to get out of this state.
  - The current implementation only works for DESFire EV1/2 tags using AES keys. DES modes are not supported.
  - A `free()` in `desfire_crypto.c` had to be commented out due to linker errors.


iceman1001:
My guilty conscience.    I will take a look at the suggested changes and see what can be added to the latest code base to support this nicely.

PR #1596, opened 2022-02-11
Re-enabled the -k option in hf 15 raw (to keep the field on after sen…

Giraut:
…ding an ISO15693 command) and added a -n option to send the next command without resetting the field first.

So for example, to select a chip with UID 0E0401235708FE78, then read address 00h, then read address 01h in 3 commands, dropping the field at the 3rd command:

[usb] pm3 --> hf 15 raw -k -c -d 222578fe08572301040e
[+] received 3 octets
[+] 00 78 F0

[usb] pm3 --> hf 15 raw -k -n -c -d 122000
[+] received 7 octets
[+] 00 E1 40 80 09 3D 70

[usb] pm3 --> hf 15 raw -n -c -d 122001
[+] received 7 octets
[+] 00 03 10 D1 01 45 38

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -1236,7 +1236,11 @@ static void PacketReceived(PacketCommandNG *packet) {
             break;
         }
         case CMD_HF_ISO15693_COMMAND: {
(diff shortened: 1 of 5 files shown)


iceman1001:
... hm..  I think we have turn on antenna and keep antenna on as flows ...    Introducing a "next" would be a complete other flow.

PR #1602, opened 2022-02-15
Standalone mode hf_legicsim

uhei:
Add a standalone mode to simulate Legic Prime dumps stored on flash


iceman1001:
Interesting,
I am curious,  what is the difference from your first standalone mode?

And would you mind testing legic commands out,  since I refactored them a bit ?

PR #1603, opened 2022-02-15
Bugfix for NXP GET_SYSTEM_INFO command

c-barron:
NXP-specific get system info command is 0xAB instead of 0x2B, in section 9.5.3.18 of technical docs for NXP ICODE SLIX2:  https://www.nxp.com/docs/en/data-sheet/SL2S2602.pdf

SLIX2 supports this command, but earlier versions do not. To only run this command on SLIX2, it checks the version - which is stored in the 4th Byte of the UID (data[6] in the code).

If any other cards also support this command, let me know and I will revise. So far have confirmed that SLI, SLIX, SLIX-S, SLIX-L do not support it.


c-barron:
I should probably revise to allow SLIX2 and RFU now that I think about it - but I'm not aware of any cards that are using the RFU identifier at the moment

c-barron:
Ah yes - sorry about that, in my brain those sections were connected :upside_down_face:

PR #1606, opened 2022-02-18
fix slixdisable pass id

r1ddl3rz:
bug/typo


r1ddl3rz:
Got this Infos from dataheet: 

Table 15. Password Identifier
| Password identifier | Password |
| --------------------- | ---------- |
| 04h | Privacy |
| 08h | Destroy SLIX-L |
| 10h | EAS/AFI |

For ICODE SLIX-L it must be `04h`

iceman1001:
Yeah,   
0x04  would set e empty PRIVACY password
0x10  would set a empty  EAS/AFI password


I think we'll skip 0x08...

r1ddl3rz:
Better skip 0x08 ;-)

PR #1607, opened 2022-02-23
Fix support for more Mifare Desfire EV2 cards

gentilkiwi:
Was tempted to make a `major` mask with 0x0f == 1,2,3,... but...


iceman1001:
Which new Desfire card ev2 are those?

gentilkiwi:
> Which new Desfire card ev2 are those?

A 32k (version 0x22 0x00)

PR #1617, opened 2022-03-06
Update instructions for Termux

s00se:
Remove "tested setups". This seems extraneous, it should work on almost all recent Android hw+sw
Direct users to F-Droid for Termux
Use correct directory after cloning


doegox:
thanks

PR #1618, opened 2022-03-07
Add first and last block for ISO15693_READ_MULTI_BLOCK to command list

nemanjan00:
Small PR, but it was useful to me


nemanjan00:
Please, do not merge yet. I was sleepy when writing this and forgot there can be optional UID in which case data shown is wrong

iceman1001:
any work going on?

nemanjan00:
Just added check for optional param

iceman1001:
why not add the uid offset and get the blocks out?

nemanjan00:
I was unable to manually build raw command with UID that worked with my card.

nemanjan00:
Sorry, I was unaware of

```
hf 15 rdmulti -* -b 12 --cnt=3
```

iceman1001:
Thanks!

PR #1620, opened 2022-03-08
Make HF_15SNIFF standalone mode compatible with old PM3 (without flashmem)

lnv42:
Make HF_15SNIFF standalone mode compatible with old PM3 (without flashmem)


iceman1001:
Thanks!

PR #1621, opened 2022-03-08
Improve iso15 sniffing: add support for dual subcarriers sniffing

lnv42:
This PR add support for 2sc (FSK) sniffing for iso15.

I added a new FPGA image (to have enough space) : FPGA_BITSTREAM_HF_15.
I had to integrate some threshold in the FPGA part that may probably be optimized, but from my tests (not using actual reader), quality of sniffing is good, while working range is a bit smaller than with 1SC.
I had to modify slightly the 1SC part, because switching Fpgamode while sniffing (with FpgaWriteConfWord()) was sometimes too long so the tag answer start was lost... (the timing was varying a lot, i don't know why)
From my tests, my modifications do not affect quality of 1SC sniffing, but it comes without any warranty and may in theory affect very low strength/high range signals, but it doesn't seem to be observable.


armsrc/iso15693.c
```diff
@@ -1424,6 +1791,9 @@ void SniffIso15693(uint8_t jam_search_len, uint8_t *jam_search_string) {
     Dbprintf("  DecodeTag State........%d", dtag.state);
     Dbprintf("  DecodeTag byteCnt......%d", dtag.len);
     Dbprintf("  DecodeTag posCount.....%d", dtag.posCount);
+    Dbprintf("  DecodeTagFSK State........%d", dtagfsk.state);
+    Dbprintf("  DecodeTagFSK byteCnt......%d", dtagfsk.len);
```
iceman1001, line 1795: align?

iceman1001:
So,  now,  shall I understand this is a replacement for the normal 15 sniff?   does it work with singel carrier too?

armsrc/iso15693.c
lnv42: fixed

lnv42:
From my point of view, it's an improvement more than a replacement. 1SC sniffing is still working well and is still executing almost the same code (for both FPGA and ARM) as before.
I just had to reduce the sniffed amplitude (used for 1SC sniffing) returned by the FPGA code from 14bits to 12bits, this reduce slightly the amplitude precision, but not significantly, so, from my tests, the 1SC sniffing is still working as well as before.

PR #1625, opened 2022-03-14
Update T5577_Guide.md

joetomasone:
Added information on traceability data and Keysy

### DIFF
--- a/doc/T5577_Guide.md
+++ b/doc/T5577_Guide.md
@@ -66,6 +66,9 @@ the chip how to behave.
 
 ![](./t55xx_mem_map.png)
 
+ #### What is "Traceability Data"?
+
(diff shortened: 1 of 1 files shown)


iceman1001:
We definite need a note about lf cloning where it touches t55xx,  em4305, 4350, 4450, hitag,  and cloner tools like the gun models.
until then I will merge this PR.

Thanks!

joetomasone:
So I waited until Iceman made a decision before commenting, but as the
title of the note is "T5577 Guide", I figure it should be the intro that
people should read to understand the basics - what the chip is and can do,
and some of the things that the data sheet doesn't address or leaves
obtuse, plus some practical tips on what you may encounter in the wild.   I
agree that splitting things like the Keysy stuff and adding cloner gun tips
and the like could stand to be in a separate document, but I don't have the
exposure or experience with EMxxxx and the cloner guns to author such a
piece authoritatively, and the Keysy stuff wouldn't seem to merit its own
note.   Perhaps that is best left for someone who can write up the majority
of the meat and transfer the Keysys stuff?

iceman1001:
Until someone finds time,  we have it like we have it now.

PR #1628, opened 2022-03-19
update mfc_default_keys.dic

equipter:
added 2 keys. 4 different cards, 1 staff 3 student all had same keys in same places


iceman1001:
Thanks!

PR #1629, opened 2022-03-19
Update cmdhfmf.c

equipter:
changed "6 hex bytes" to "4 hex bytes" in hf mf restore, tiny typo :)


iceman1001:
Nice catch,
but it should actually accept 4,7 or 10 hex byte uid's.  I have pushed fix for it.  https://github.com/RfidResearchGroup/proxmark3/commit/f59f8dd5288ef40adc0747c7b53193a06a0ece71

and I will close this PR.

PR #1630, opened 2022-03-20
Create hf_i2c_recovery.lua

equipter:
helps unbrick soft locked i2cs (xsiids)


iceman1001:
:)  Nice one,   maybe a rename later since it seem to be a 14a related script

PR #1635, opened 2022-03-23
Fedora v4.14831.1

s00se:
For discussion. 

These were the changes made to cleanly build a Fedora RPM, hosting the SPEC file in the official repository means the existing change management process can be used.

I can remain a package maintainer if preferred, otherwise this can be changed to repository owner(s).


doegox:
we can avoid packaging the datasheets and old docs in the releases, but I don't think erasing them from the repo is the proper way to go.

what is the problem with the shebang in the cmd file ?

iceman1001:
and pm3generic ?   normally a distro offers both or more options like btaddon, rdv4,  generic,  etc

s00se:
> we can avoid packaging the datasheets and old docs in the releases, but I don't think erasing them from the repo is the proper way to go.

You are right, this isn't the correct way. I will remove them in the SPEC instead.
 
> what is the problem with the shebang in the cmd file ?

There are some scripts in the Fedora build system that I am not very familiar with, so I took the lazy way and modified the files. I will correct this in the SPEC instead.

s00se:
I will update the SPEC for the master branch and it can still target the release tarball.

Should the SPEC file live in a separate directory (hidden or otherwise) or in the root?

doegox:
> > and pm3generic ? normally a distro offers both or more options like btaddon, rdv4, generic, etc
> 
> Thanks for catching this, I will remove the override and build with the defaults.

Look at Kali package, there was quite some fine-tuning to get it there ; https://gitlab.com/kalilinux/packages/proxmark3/-/tree/kali/master/debian. 

you can use the same script tools/build_all_firmwares.sh to get a collection of firmware variants for distribution.

doegox:
yes, on Kali you get

* proxmark3 (client, host platform specific)
* proxmark3-common (scripts etc)
* proxmark3-doc
* proxmark3-firmwars

PR #1636, opened 2022-03-23
Re-enabled the -k option in hf 15 raw

Giraut:
(to keep the field on after sending an ISO15693 command), and made DirectTag15693Command not start by resetting the field if the previous commmand left the field on (i.e. only hf 15 raw -k).

This is the same idea as PR #1596 (https://github.com/RfidResearchGroup/proxmark3/pull/1596) but without adding the -n flag and extra command, which admittedly was a nasty solution.

As an added bonus, this implementation doesn't break backward compatibility - i.e. it will work exactly as before with an older client that can't pass the extra "field already on" flag to DirectTag15693Command()

### DIFF
--- a/armsrc/iso15693.c
+++ b/armsrc/iso15693.c
@@ -2357,9 +2357,12 @@ void BruteforceIso15693Afi(uint32_t speed) {
 
 // Allows to directly send commands to the tag via the client
 // OBS:  doesn't turn off rf field afterwards.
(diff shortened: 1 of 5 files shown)


iceman1001:
try https://github.com/RfidResearchGroup/proxmark3/commit/8d0b41a911ca308eac990cc6bdb21d3fd073ef1c  and see if this works better.

Issue #1616, opened 2022-03-02
Errors flashing the PM3 Easy devices with a generic build (MacOS)


maxieds:
**Describe the bug**
There are errors flashing the PM3 Easy line of devices using ``pm3-flash-all`` on MacOSX. The generic build of the firmware is installed by running
```bash
brew reinstall --with-generic proxmark3
```
where the Homebrew package is, as usual, the tap provided by this repo (Icemann/RFIDRes). Then when I plug in the Easy device over USB, and run the flash command, the following is the full output I get:
```bash
$ pm3-flash-all
[=] Session log /Users/localusernameredacted/.proxmark3/logs/log_20220302.txt
[+] loaded from JSON file /Users/localusernameredacted/.proxmark3/preferences.json
[+] About to use the following files:
[+]    /usr/local/Cellar/proxmark3/HEAD-ab52131/bin/../share/proxmark3/firmware/bootrom.elf
[+]    /usr/local/Cellar/proxmark3/HEAD-ab52131/bin/../share/proxmark3/firmware/fullimage.elf
[+] Loading ELF file /usr/local/Cellar/proxmark3/HEAD-ab52131/bin/../share/proxmark3/firmware/bootrom.elf
[+] ELF file version Iceman/master/v4.14831-404-gab5213126 2022-03-02 05:49:24 ea44b0c23

[+] Loading ELF file /usr/local/Cellar/proxmark3/HEAD-ab52131/bin/../share/proxmark3/firmware/fullimage.elf
[+] ELF file version Iceman/master/v4.14831-404-gab5213126 2022-03-02 05:49:38 ea44b0c23

[+] Waiting for Proxmark3 to appear on /dev/tty.usbmodem14601
 🕑  59 found
[+] Entering bootloader...
[+] (Press and release the button only to abort)
[+] Waiting for Proxmark3 to appear on /dev/tty.usbmodem14601
 🕓  59 found
[!!] 🚨 ====================== OBS ! ===========================================
[!!] 🚨 Note: Your bootloader does not understand the new CMD_BL_VERSION command
[!!] 🚨 It is recommended that you first update your bootloader alone,
[!!] 🚨 reboot the Proxmark3 then only update the main firmware


[!!] 🚨 ------------- Follow these steps -------------------

[!!] 🚨  1)   ./pm3-flash-bootrom
[!!] 🚨  2)   ./pm3-flash-all
[!!] 🚨  3)   ./pm3

[...]

PR #1637, opened 2022-03-27
Create PM3EasySpecificInstallationInstructions.md

maxieds:
As noted to @doegox in issue #1616 before @iceman1001 closed the issue. Hope this helps some others out now that my PM3 Easy device is running! 😺

### DIFF
--- a/doc/md/Installation_Instructions/PM3EasySpecificInstallationInstructions.md
+++ b/doc/md/Installation_Instructions/PM3EasySpecificInstallationInstructions.md
@@ -1,0 +1,94 @@
+# Installation tips and tricks for the Proxmark3 Easy devices
+
+## Installation instructions on MacOS (with Homebrew)
+
+See the documentation for the generic version of the Homebrew install of the 
+[Proxmark3 suite here](https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/Mac-OS-X-Homebrew-Installation-Instructions.md#install-proxmark3-tools).
+
+## Installation instructions from source (Linux and other OS)
+
+On MacOSX, see 
+[this document](https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/Mac-OS-X-Compile-From-Source-Instructions.md). 
+For Linux and other Unix platforms, see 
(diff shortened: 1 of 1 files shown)


iceman1001:
None of this information is new and not already documented in the current guides.
the correct makefile,   the button trick,  the instructions the flasher gives you.

maxieds:
@iceman1001 
I still contend that it is worthwhile to have a separate, centralized, easy to find resource for PM3 Easy users. The documentation provided by the manufacturer of the Easy devices and their online vendor really, really sucks! There is also some new information that is useful to Easy users that someone pointed out in #1616: the idea of holding down the button throughout the entire flashing procedure.

doegox:
> The documentation provided by the manufacturer of the Easy devices and their online vendor really, really sucks! 

We already did a favor to Easy owners to be able to use this fork made initially exclusively for RDV4 devices, but we're not there to help further those crappy pm3 clones vendors.

> There is also some new information that is useful to Easy users that someone pointed out in #1616: the idea of holding down the button throughout the entire flashing procedure.

New? Really?

image

maxieds:
@doegox 
Ok. You can do what you want. This seems to me like another attempt to get people to buy the RDV4 rather than the more affordable Easy devices by making it **harder**, but of course not impossible, to use the firmware and command line utilities that are actually under active development. It would have helped me very much to have all of that information in a centralized location starting a couple of months ago...

mwalker33:
I think this one can be a little tricky.  The documentation is already on (this) github.  If you follow the "trouble shooting" links it talks about the "button trick" as well as other things to check and review.
I believe the issue is more one of it not being front and center when an end user has an issue.  But everything cant be front and center; as with all things, the need to read to find the answer.

I have been following a lot of these discord and still trying to work out why people see to have issues with the firmware updates.  I suspect most of the time they are non-rdv4 proxmark's and not compiled with the correct options; leading to the main application failing.    At the some time, we have seen bad USB cables and broken pm3 easy's.   I believe this is what @doegox  was trying to refer to.  i.e. the repo cant have defaults that work for both out of the box, as the RDV4 is a different device with more things needed.  This repo was designed to build and support the RDV4 new features; as such are the defaults.  But every effort has been made to support other proxmarks (including the easy)

Given that the proxmark is also meant to be a device for advanced users and can have a steep learning curve for beginners; the documentation layout will be structured to support that user base.

Bottom line the button trick and how to get the pm3 easy running is already in the repo.

PR #1638, opened 2022-03-27
Sort filenames before hashing and force hashing in text mode

Doridian:
This avoids inconsistencies as can be seen between macOS and Windows platforms in how they hash the sources for the firmware mismatch checks.

Fixes done here:
- Use `sort` command to ensure filenames are sorted (not every OS sorts the files the same)
- Use `-t` argument to `sha256sum` to force hashing files in text mode, some OS (Windows) hash the .c and .h as by mistake by default

FWIW, this new hashing seems to produce the same hashes on macOS (and likely Linux), therefor I assume Windows is the odd one out here


iceman1001:
on wsl-1 / linux the text mode is default.   Lets hope this sorts out potential issues on proxspace then.

PR #1639, opened 2022-03-27
Fix compilation on M1 Macs

Doridian:
`mfd_aes_brute` was throwing a compilation error first because of using `-march=native` which does not exist on M1 Macs (yet), so I swapped that over to `-mcpu=native` conditionally.
Secondly, it did not have a method for checking AES-NI state on those machines either (the fallback being Linux only).
For now hardcoded to false, as M1 CPUs do not support AES-NI (and the check function was only used for display purposes in the first place)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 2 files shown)


iceman1001:
Interesting,  I pushed a change for not using AESNI detection on _APPLE_ / _MACH_  ,  which should solved that part in your PR.

Doridian:
@iceman1001 Rebased my changes on your push. And yes, disabling detection resolved that part of my PR.

iceman1001:
Thanks!

Talkwondo:
This not fix my issue. M1 mac can not compile. 
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

PR #1641, opened 2022-03-28
Minimal change to utilize ARM NEON capabilities

Doridian:
ARMv8 always ships with NEON capabilities supporting 128-bit operations.
This gives ~ +50% speedup on hardnested on Apple M1


iceman1001:
A simple but big bang for the buck PR  :)

PR #1643, opened 2022-03-29
Fix examples of mfdes chk

Doridian:


iceman1001:
conflict... 
How about you split this one up with,   the lowercase in one pr and the text fixes in another one

Doridian:
@iceman1001 fixed conflict and reduced it to only spelling fix

PR #1646, opened 2022-03-30
Fix USB communications when receiving empty frames

lnv42:
E.g. before this fix, iso15 (and others using the same way from ARM to client) raw commands expecting results but with no answer from the tag where showing the previously received commands instead of : "[!] /!\ command failed".

BEFORE:
[usb] pm3 --> hf 15 raw -c -d 260100
[+] received 12 octets
[+] 00 02 05 43 20 0B 66 24 16 E0 6C 56 
[usb] pm3 --> hf 15 raw -c -d 010203 # invalid command (tag is not answering)
[+] received 12 octets
[+] 00 02 05 43 20 0B 66 24 16 E0 6C 56

AFTER:
[usb] pm3 --> hf 15 raw -c -d 260100
[+] received 12 octets
[+] 00 02 05 43 20 0B 66 24 16 E0 6C 56 
[usb] pm3 --> hf 15 raw -c -d 010203 # invalid command (tag is not answering)
[!] ⚠️  command failed

### DIFF
--- a/client/src/comms.c
+++ b/client/src/comms.c
@@ -414,6 +414,15 @@ __attribute__((force_align_arg_pointer))
                         }
                     }
                 }
+                else if ((!error) && (length == 0)) { // we received an empty frame
+                    if (rx.ng)
+                        rx.length = 0; // set received length to 0
+                    else {  // old frames can't be empty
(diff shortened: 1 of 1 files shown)


lnv42:
Maybe a more specific error message saying 'no response received' or anything like that will be better than the generic "command failed" ?

PR #1647, opened 2022-03-30
Update hf_14a_i2crevive.lua

equipter:
added better correct APDUS, not sure how i missed that lol


iceman1001:
you might consider documenting what the apdu's does,  so one can understand how the revive works

PR #1650, opened 2022-04-03
Implement 14b sniff standalone mode

jacopo-j:
Very simple/straightforward. I did this several years ago because I needed it and then forgot about it. I thought I'd share it just in case anybody else finds it useful.


iceman1001:
Sometimes simple is all you need to get the job done

PR #1651, opened 2022-04-03
Update cmdhfmfu.c

R0BB13M:
corrected spelling of ultralight (from ultraligth) in the error message of hf mfu ndefread


iceman1001:
Nice catch!

PR #1657, opened 2022-04-08
Fix a bug in "lf config"

wh201906:
When executing "lf config --reset", the averaging will be 0 rather than 1(default).
This is because the config.averaging is set to 1 at first.(line 621)
Then this argument is overwritten unconditionally by
"config.averaging = (avg == 1);"(line 634)


iceman1001:
Good catch!

PR #1659, opened 2022-04-08
Show a tip in the plot window

wh201906:
This tip will be shown behind the "Apply" and "Restore" button


iceman1001:
Nice,   I make it italic to show more like a hint.

wh201906:
Yeah that would be much better.
Thx.

PR #1660, opened 2022-04-09
Fix the bug_report.md

wh201906:
Fix the link of troubleshooting guide, which should be a URL rather than a path.
Add 2 spaces after every title to get the same appearance on both Github and the editor.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,40 +7,40 @@ assignees: ''
 
 ---
 
-***Things to try before submitting bug report***
-read the [troubleshooting guide](/doc/md/Installation_Instructions/Troubleshooting.md)
+***Things to try before submitting bug report***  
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice catch with the broken link. 
I am not sure why you added two spaces?

wh201906:
In the normal markdown format, You need to add two spaces after a line to create a line break.
You can check [this](https://www.markdownguide.org/basic-syntax/#line-breaks) to see the format.
Two spaces is not necessary since this .md file is only used in Github, which will automatically create a line break as the raw input.

wh201906:
I add the two spaces just for the appearance in other general editors(like VSCode). How do you think of it? I can revert them if you like.

iceman1001:
our `make style`  will remove all double spaces if I am not mistaken.   I think you can remove the double spaces,  and we go with the fix for the broken link.

wh201906:
ok

iceman1001:
Thanks!

PR #1668, opened 2022-04-30
Add SUBE keys (Argentina public transportation system card)

CONIGUERO:


iceman1001:
IS this unique keys or are the system using diversified keys?   ie individual keys for each card

CONIGUERO:
> IS this unique keys or are the system using diversified keys? ie individual keys for each card

These are universal keys, however some card variants have different ones. Depends on the year of manufacture. All cards of the 2017 variant will have the same keys for example, while all the 2015 ones will have a different set but it will work on all cards made in 2015.

iceman1001:
Thats good.  

but... you removed all comments in your PR.   Just add the keys please.

iceman1001:
I added your keys,  closing this one.

https://github.com/RfidResearchGroup/proxmark3/commit/1b6b16d91a7546ce01339f02664005555a42807a

PR #1672, opened 2022-05-08
Update cmdhfmfdes.c

mwalker33:
Add the ability to use hex for the DFName to support non printable dfname values (e.g. as needed to NDEF AIDs)


iceman1001:
@mwalker33   the GH scanner wants  a size_t  as type for dfnamelen,   try type casting it.

PR #1675, opened 2022-05-09
Ndef notes

mwalker33:
This is a starting document to store ndef type4a notes.
It is hoped that others will add and build on this.
Initial contents is focused on setting up the NDEF type 4a records on a desfire ev1 (or later) card.


iceman1001:
Excellent!

PR #1676, opened 2022-05-13
Fixes a couple of new warnings [gcc 12.1.0]

gbin:
Stumbled on a couple warnings about array bounds that prevented proxmark3 to be compiled on a recent arch install.


gbin:
please ignore, that was fixed meanwhile on master. Thanks

PR #1678, opened 2022-06-05
Mifare Classic Value operations

mwalker33:
Added increment value, decrement value via the supplied commands and set value via the normal write command.  Also added "get" to read a value block and decode/display.

Initial implementation for review and feedback.


iceman1001:
Great PR with a great addition of value :)

Some minor spelling in armsrc comments,  and no use of  a struct for sending values to arm (NG Frames)  But that is all minor stuff

PR #1687, opened 2022-06-11
Switch usage of sprintf to snprintf

Doridian:
And some other code improvements along the way (in code related to the snprintf calls)


merlokk:
from:    sprintf(buf, "%02X%02X", sw1, sw2);
to:        snprintf(buf, sizeof(buf), "%02X%02X", sw1, sw2);
Maybe needs to add space for 0x00 at the end of line?
like: snprintf(buf, sizeof(buf) - 1, "%02X%02X", sw1, sw2);

Doridian:
@merlokk According to the C specifications, the length you specify in `snprintf` is the maximum number of bytes used (this is unlike some other strN functions, where it is the maximum resulting string length)

merlokk:
ok, thank you)

PR #1688, opened 2022-06-14
Massive update of keys

UberGuidoZ:
Greetings from the Flipper Zero world! I combined the Iceman dict with some various others I found, some keys I captured, and some that were provided. We've had a lot of luck with them and thought we should certainly share. [Source](https://github.com/UberGuidoZ/Flipper/blob/main/NFC/mf_classic_dict/mf_classic_dict.nfc)

- Comments have been all moved to directly above (instead of behind) keys
- Duplicates have been removed and verified by multiple contributors

This is In use by the leading 3rd party firmware devs ([RogueMaster](https://github.com/RogueMaster/flipperzero-firmware-wPlugins) and [Unleashed](https://github.com/Eng1n33r/flipperzero-firmware)) and thousands of individual users successfully.


iceman1001:
That one was created by merging the dictionaries on this repo...   There is a reason why those are not merged into the default dictionary. 

However, there seem to be a few new ones that are of interest if they can be confirmed to be default/static keys and not diversified.   Which I can't tell.  Seem to be some Russian keys that are new.

If you filter out the keys from this repo and then make a PR with the interesting ones,  it can be merged

UberGuidoZ:
Correct - there were merged dictionaries from this repo, as well as a number of other static keys collected from many sources. Sadly, the Flipper only works from a single dict file, so they were all merged. Comments were kept (though converted to capital letters for reasons). I don't expect you'll want the entire thing, but I didn't know a better way to get it ALL to you for browsing.

UberGuidoZ:
Closing this one to avoid having to recreate the files. New PR opened.
https://github.com/RfidResearchGroup/proxmark3/pull/1691

iceman1001:
yeah,   since I am certain that your run time doing a dictionary recovery takes forever one a flipper zero with your all-merged dictionary,   you intuitively understand why there are several dictionaries.  

I am glad that you found a way to clean it out, by the looks of your other PR

PR #1689, opened 2022-06-14
Change 'hf 14a list' to annotate more ECP frame formats

kormax:
Updated the logic behind 14a annotation to support more ECP (Apple Enhanced Contactless Protocol/Polling) frame variations with different parameters and versions, changed labels to indicate if it's version 1 or 2 (ECP1/ECP2).

As a result of this change, following frames (and possibly more still unknown ones) will be annotated:
- VASUP-A (Used by readers to wake up the phone to inform it about what passes to present for authentication after an unlock):
  `6a  01  00  00  00` 
- IPHONE (Used by iPhones so that other phones won't react to an NFC field and request an unlock, might be used for other purposes):  
 `6a  01  cf  00  00`
- More V2 frame variations for transit bypass.


iceman1001:
Nice one!

Thanks

PR #1690, opened 2022-06-14
Fix wrong length calcuation in *_to_buffer utilities

Doridian:
Tested with previously broken `data diff` command. Working now.


iceman1001:
Thanks!

PR #1704, opened 2022-06-27
Texkom tokens

merlokk:
a very strange "RFID" tokens at 13MHZ
sends into the field:
tk13 - 127 impulses (works)
tk15 - ???
tk17 - 65 impulses (works)


iceman1001:
Cool!

client/src/cmdhftexkom.c
```diff
@@ -241,6 +241,52 @@ static uint8_t TexcomTK17CRC(uint8_t* data) {
...
+    if (xlen < 10 || xlen > 90)
+        return -1;
+    if (xlen < 30)
+        return 0;
+    if (xlen < 50)
+        return 1;
+    if (xlen < 70)
+        return 2;
+    return 3;
+}
```
iceman1001, line 255: these return codes..  
use DEFINES in order to make them more understandable

iceman1001:
Excellent!

PR #1723, opened 2022-07-20
Add hf gallagher decode command and other minor Gallagher tweaks

nvx:
Adds a new `hf gallagher decode` command that decodes the Gallagher credential block.

Also tweaked the output format where the region code is shown to display it as an uppercase letter A-P in addition to the existing numeric display as this is how the region code is formatted in Gallagher software.

Finally there is a super simple one like fix to the Gallagher KDF algorithm to allow it to correctly generate diversified keys for the card master key -> The CSN was incorrectly being included in the KDF input when it shouldn't have been as per https://github.com/megabug/gallagher-research/blob/master/formats/card-specific/mifare-desfire.md#card-application as well as my own testing with Gallagher encoded cards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Added `hf gallagher decode` command and fix Gallagher diversification for card master key (@nvx)
  - Added mmbit-002 (kibi-002, kb5004xk1) russian tag to `hf texkom read` command (@merlokk)
  - Added `hf sniff --smode` skip/group adc data to consume less memory. Now it can sniff very long signals (@merlokk)
  - Added `hf fudan` skeleton commands (@iceman1001)
--- a/client/src/cmdhfgallagher.c
+++ b/client/src/cmdhfgallagher.c
@@ -840,9 +840,10 @@ static int hfgal_read_card(uint32_t aid, uint8_t *site_key, bool verbose, bool q
         uint32_t current_aid = cad_aid_byte_to_uint(&cad[i + 3]);
 
         if (verbose) {
-            if (region_code > 0 || facility_code > 0) {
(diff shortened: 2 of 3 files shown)


iceman1001:
Interesting PR,
`make style`  yeah, I tend to use on a proper linux system to avoid those extra linefeeds.

Not sure about the decode command.  I think I would rather see a `hf gallagher info` that read & decodes.

Once you make a decode,  you also want a encode command.

I will merge regardless.

nvx:
For reference there is already a hf gallagher reader command that does the read+decode from a desfire card, I've just found a few times where I've had the need to just perform the decode step.

One such case being on mifare classic which uses the same credential block format, although I've not played enough with Gallagher on mifare classic to be confident enough on identifying the block reliably to implement a classic reader and dump viewer (specifically does it always rely on the MAD to identify the CAD sector, or does it just search through the sectors for something that looks right, and does it always rely on the CAD on classic like it does on desfire or?)

iceman1001:
the idea is,  in order to make a more unified command set in the client. 

```
reader - simple identification of tag,  the necessary ones
info - deeper and verbose information about the tag
```

Which then gives,  if you have a deode command,  it needs an encode command to match it.  
the gallagher subsection of commands,  is to seperate it from the  classic,  desfire commands since they are getting to many as is.   And here we want the dedicated actions needed for Gallagher.  
I am sure you understand this already so I am just repeating for people who might read this later. 

Adding a  `hf gallagher mad` etc  would be fitting too.

nvx:
That makes sense, I actually wasn't too clear on the intended difference between reader and info so that clears things up a bit.

Considering this, would it make sense that reader doesn't take a key and instead just displays eg the CAD information like facility codes and application IDs (which are freely readable), while info goes a step further and requires the keys to display the actual card number?

As for handling mifare classic at the same time reader would be pretty easy if it was just doing that public info, but once keys and app IDs (or sectors) come into play that gets a little trickier since key lengths and the like are different for each card tech. Are there any examples of system specific commands that handle multiple card techs that could be used as a guide?

iceman1001:
reader,  should just more or less identify a tag as being gallagher.
info  should do all the rest. 

then you have dump/restore command to copy memory.
then you have clone command to "generate a new credential"  onto a card

then you have sim command w their e* commands,  but we don't have a proper desfire simulation yet.

Those are more or less the common commands,  that we have been trying to unify across hf commands..

PR #1724, opened 2022-07-21
iclass output improvements

nvx:
Fixed `hf iclass info` and `hf iclass view` key access info looking at the wrong card config bit

Changed `hf iclass list` to display matched keys on the CHECK command rather than the card response
This is useful as it shows what key a reader attempted to use with a card even if the card rejected the key. To make it more apparent when the CHECK command was successful though it now shows `CHECK SUCCESS` in green when the card responds to the CHECK command, and the key is not printed in green anymore since we don't know at that point if the key actually worked or not.

Also made `hf iclass list` key check lookup using the elite algorithm too
The display currently doesn't show any different output for elite vs standard kdf since there's fairly limited room, but I'd expect people would know what sort of keys they've loaded in, and worse case could double check with `hf iclass chk`

image

iClass dump views (ie, `hf iclass view`) now suppress repeated blocks, similar to how hexdump works. It will only suppress blocks in the application area, and will never suppress detected iclass credential blocks. For example a dump that has blocks 17-236 all containing just FF's:

```
<blocks 16 and prior output unchanged>
[=]  17/0x11 | FF FF FF FF FF FF FF FF | ........ |   | User
[=] *
[=] 236/0xEC | FF FF FF FF FF FF FF FF | ........ |   | User
```

Needless to say this is quite handy when dealing with complete card dumps when the AA2 area is blank.

I also fixed a small client memory leak in `check_known_default`, `prekey` wasn't being free'd, and fixed up the SIO detection logic (was using strstr on binary data so would have only been matching on `0x05` instead of the full `0x05, 0x00, 0x05, 0x00` pattern leading to false positives in some cases) and simplified a bit of the SIO printing code too (the duplicate logic was triggering me a little)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Changed `hf iclass view` to suppress consecutive blocks with repeated contents (@nvx)
+ - Fixed `hf iclass info` and `hf iclass view` key access info looking at the wrong card config bit (@nvx)
+ - Changed `hf iclass list` to display matched keys on the CHECK command rather than the card response, and made it check for elite keys too (@nvx)
  - Added `hf gallagher decode` command and fix Gallagher diversification for card master key (@nvx)
  - Added mmbit-002 (kibi-002, kb5004xk1) russian tag to `hf texkom read` command (@merlokk)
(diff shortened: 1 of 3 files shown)


iceman1001:
Interesting one,  
This isn't one PR.  It contains about three PR.  I suggest splitting it up.

Why? It is easier to look at commit history and see smaller PR / code changes.

nvx:
> Interesting one, This isn't one PR. It contains about three PR. I suggest splitting it up.
> 
> Why? It is easier to look at commit history and see smaller PR / code changes.

fair, some projects prefer larger PRs when they're semi-related for less effort reviewing - wasn't sure what was preferred here, will split it up.

nvx:
Closing this PR in favour of the split out #1726 #1727 #1728 #1729

PR #1726, opened 2022-07-22
Changed `hf iclass list` to display matched keys on the CHECK command

nvx:
Changed `hf iclass list` to display matched keys on the CHECK command rather than the card response
This is useful as it shows what key a reader attempted to use with a card even if the card rejected the key. To make it more apparent when the CHECK command was successful though it now shows `CHECK SUCCESS` in green when the card responds to the CHECK command, and the key is not printed in green anymore since we don't know at that point if the key actually worked or not.

Also made `hf iclass list` key check lookup using the elite algorithm too
The display currently doesn't show any different output for elite vs standard kdf since there's fairly limited room, but I'd expect people would know what sort of keys they've loaded in, and worse case could double check with `hf iclass chk`

image

Split out from #1724

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Changed `hf iclass list` to display matched keys on the CHECK command rather than the card response, and made it check for elite keys too (@nvx)
  - Added `hf gallagher decode` command and fix Gallagher diversification for card master key (@nvx)
  - Added mmbit-002 (kibi-002, kb5004xk1) russian tag to `hf texkom read` command (@merlokk)
  - Added `hf sniff --smode` skip/group adc data to consume less memory. Now it can sniff very long signals (@merlokk)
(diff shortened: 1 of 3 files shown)


iceman1001:
Lets see,
My baseline is:   I like output that is to the point.   No unnecessary text.   

Hence one line give me both key/color if success is useful to me as a user.

So for this idea to work in a way that fits that pattern I would suggest 
- keep the old one liner,   
- make key red or green if correct instead.

Does it make sense to you?

iceman1001:
printing the key is interesting piece of data.   

Lets go though some scenarios:

if the check fails,   you know be looking at the rest of the trace.  
Where is more or less it will do a new authentication w a anti collision.   

If check is success it will continue to read / write process the card.


Hence the "success" message is not bringing any interesting information.

nvx:
> printing the key is interesting piece of data.

indeed, happy to add some colour coding here (white perhaps being fairly neutral on if the key worked or not) to make it stand out more

> Lets go though some scenarios:
> 
> if the check fails, you know be looking at the rest of the trace. Where is more or less it will do a new authentication w a anti collision.

sometimes, an omnikey for example will just retry the CHECK command a few times (sometimes with multiple keys) without doing the full anti-collision until one works before eventually giving up and going back to the anti-collision

> If check is success it will continue to read / write process the card.
> 
> Hence the "success" message is not bringing any interesting information.

I guess, it's mostly just so at a glance you can see if the card accepted the CHECK command without having to infer it from the presence of other commands (which may be eg the reader continuing to read the card, or may be due to the reader retrying the anticollision or CHECK with a new key)

iceman1001:
yeah..  well...  you almost convince me about it. 

I don't like the text, but I can see where inexperienced users would benefit from it.  
If you fix the conflict,   I merge.

PR #1727, opened 2022-07-22
Fixed incorrect card config bit being checked for iClass key access flag

nvx:
Fixed incorrect card config bit being checked for iClass key access flag. Fixes `hf iclass view` and `hf iclass info` command output. The bit now matches what the PicoPass datasheet describes, as well as matching practical experience testing iClass cards in the wild.

Split from #1724

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Fixed `hf iclass info` and `hf iclass view` key access info looking at the wrong card config bit (@nvx)
  - Added `hf gallagher decode` command and fix Gallagher diversification for card master key (@nvx)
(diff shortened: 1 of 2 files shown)


nvx:
Just noticed the failed MacOS build, looks like the failure was unrelated to the PR though in setting up the environment.

iceman1001:
I believe this check is from Adam Lauries source,   
To be honest I find the picopass configurations quite confusing in terms how they relate to the physical card capacities. 
If the first bit decides keyaccess and not the inherent property of the a card configured with having books,  I shall merge.

nvx:
> I believe this check is from Adam Lauries source, To be honest I find the picopass configurations quite confusing in terms how they relate to the physical card capacities. If the first bit decides keyaccess and not the inherent property of the a card configured with having books, I shall merge.

Yeah I've tested a couple of cards and the current behaviour indicated that the Kc could be used to read AA1 with some of the cards, when in practice it wasn't permitted by the card (much to my disappointment).

The updated check in this PR matches the behaviour of all the PicoPass cards I have in my possession, as well as my interpretation of the datasheet.

iceman1001:
Then I shall merge,

Thank you!

PR #1728, opened 2022-07-22
Changed `hf iclass view` to suppress consecutive blocks with repeated contents

nvx:
iClass dump views (ie, `hf iclass view`) now suppress repeated blocks, similar to how hexdump works. It will only suppress blocks in the application area, and will never suppress detected iclass credential blocks. For example a dump that has blocks 17-236 all containing just FF's:

```
<blocks 16 and prior output unchanged>
[=]  17/0x11 | FF FF FF FF FF FF FF FF | ........ |   | User
[=] *
[=] 236/0xEC | FF FF FF FF FF FF FF FF | ........ |   | User
```

Needless to say this is quite handy when dealing with complete card dumps when the AA2 area is blank.

Split from #1724

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Changed `hf iclass view` to suppress consecutive blocks with repeated contents (@nvx)
  - Added `hf gallagher decode` command and fix Gallagher diversification for card master key (@nvx)
  - Added mmbit-002 (kibi-002, kb5004xk1) russian tag to `hf texkom read` command (@merlokk)
  - Added `hf sniff --smode` skip/group adc data to consume less memory. Now it can sniff very long signals (@merlokk)
--- a/client/src/cmdhficlass.c
+++ b/client/src/cmdhficlass.c
@@ -38,6 +38,7 @@
 #include "proxendian.h"
 #include "iclass_cmd.h"
 #include "crypto/asn1utils.h"      // ASN1 decoder
+#include "preferences.h"
 
 
 #define PICOPASS_BLOCK_SIZE    8
@@ -1100,6 +1101,7 @@ static int CmdHFiClassEView(const char *Cmd) {
         arg_param_begin,
         arg_int0("s", "size", "<256|2048>", "number of bytes to save (default 256)"),
         arg_lit0("v", "verbose", "verbose output"),
+        arg_lit0("z", "dense", "dense dump output style"),
         arg_param_end
     };
     CLIExecWithReturn(ctx, Cmd, argtable, true);
@@ -1107,6 +1109,7 @@ static int CmdHFiClassEView(const char *Cmd) {
     uint16_t blocks = 32;
     uint16_t bytes = arg_get_int_def(ctx, 1, 256);
     bool verbose = arg_get_lit(ctx, 2);
+    bool dense_output = g_session.dense_output || arg_get_lit(ctx, 3);
     blocks = bytes / 8;
 
     CLIParserFree(ctx);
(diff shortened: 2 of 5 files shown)


iceman1001:
interesting,  
So this will happen for  three or more lines?
and when the current line is the same as the next line?

nvx:
> interesting, So this will happen for three or more lines? and when the current line is the same as the next line?

correct, it will always display the first and last block of the sequence (so less than 3 lines wouldn't shorten anything), and each block must be identical or it ends - although if you had say a bunch of blocks with FF's, then a bunch of blocks with 00's it would condense both, but separately, ie

```
[=]  17/0x11 | FF FF FF FF FF FF FF FF | ........ |   | User
[=] *
[=] 100/0x64 | FF FF FF FF FF FF FF FF | ........ |   | User
[=] 101/0x65 | 00 00 00 00 00 00 00 00 | ........ |   | User
[=] *
[=] 236/0xEC | 00 00 00 00 00 00 00 00 | ........ |   | User
```

Of course this works with any non-repeating data in between blocks of repeating as well, like 20 blocks of repeating, then a few blocks of random info, then another 50 blocks repeating and it'll condense the first 20 and the last 50 separately, but show the data in between untouched

iceman1001:
The limit is three lines if too short and dump outputs will start look odd.   Groups of data can easily have 3 lines of repeating data and this would just confuse in such cases. 

Hm,   what about if we make it optional ?  

I like the concept of compress output,  but I do like normal output too.

nvx:
> Hm, what about if we make it optional ?

wouldn't be hard to add a flag to `view`, `eview`, `decrypt`, and `dump` commands (that's all the locations that end up displaying dumps), thoughts on what a suitable flag name would be, and if it should be enabled by default? On the other hand could do it as a client preference?

iceman1001:
in the long run,  a client preference for  "compressed output"   would be really nice.

we could reuse `-v` ... but no.   How about `-z  compressed output`   for those four commands?
It should not be enabled by default.   This could then be overridden by the  client pref setting if we implement it.

nvx:
`-z` sounds good, if we had a client preference to make the default compress then would specifying `-z` invert that behaviour (ie show the non-compressed output), or would it be worth having a "no-compress" flag as well to handle that case?

I can see this being useful for other formats in the long term too (mifare classic comes to mind, although I'd do it on a sector-by-sector basis rather than per-block since sectors have special blocks at the end for keys) so having the same flag options available there too would be useful.

iceman1001:
Ok, I see you thinking,   so for a client setting you would say
```
pref set output --dense
pref set output --normal
```

A) If you specify `hf iclass view -z` ,   what would happen.
B) If you specify `hf iclass view` ,   what would happen.

A), you will get compressed output regardless of client prefs.
B)  depends on client prefs.



Does that clear up how I am seeing it?

nvx:
> Ok, I see you thinking, so for a client setting you would say
> 
> ```
> pref set output --dense
> pref set output --normal
> ```
> 
> A) If you specify `hf iclass view -z` , what would happen. B) If you specify `hf iclass view` , what would happen.
> 
> A), you will get compressed output regardless of client prefs. B) depends on client prefs.
> 
> Does that clear up how I am seeing it?

Yup that's fine, the only gotcha is there's no way to always force the full output if the client preference is set to dense other than changing the preference back. I'm happy with that if you are though.

Will add the client pref option and `-z` flag and update the PR.

iceman1001:
well,  we can't have everything.   Meaning all commands shall have forceing options to override the client prefs.
It will just clutter down the client command params.  

for  normal,  I can with -z  at least get the compressed. 
for dense,  I get it by default.

for instance, I turn of hints,  I don't even want to have the option to see hints individually on commands.
You can see client prefs as turning off/on behaviors.

Another reason in this case,  is that prefs isn't well known to users.  We gotten users to start reading the help text with `-h` param.    So they will see a `-z`   and test things out.  

Much of how this repo tries to make the end user experience a bit simpler and more clear,  means we have overlaps with double commands.  We evolve slow but steady.  

In the long run,   `-z`  would be removed and only client pref would exist. 

a suggestion for you...   implement the save paths in the prefs ;)

nvx:
Added `-z` flag and client preference to control default behaviour, updated the changelog with info about the flag and preference, as well as resolved the conflict.

iceman1001:
This a good PR  :)  

Great contribution!

PR #1729, opened 2022-07-22
Fix some bugs in the iClass SIO detection and printing

nvx:
Fixes some bugs in the SIO detection and printing. Initially noticed when astyle reformatted some of the code I noticed the use of substr() on binary data containing 0x00's which substr will stop on, as well as incorrectly assuming the return of byte_substr will be 0 instead of -1 on a failed match, added some bounds checking, and simplified some of the code to reduce duplication.

The practical differences between the new and old code is the old code is more prone to false positives (matching on a single 0x05 character instead of the full 0x05, 0x00, 0x05, 0x00 sequence to identify SIO), as well as some potential undefined behaviour if sufficiently short dump files were loaded (I notice json files seem to allocate a fixed size buffer so these wouldn't have been too bad, but other formats seem to only allocate a buffer the size of the dump so it would have possibly overrun that buffer in some cases).

Split from PR #1724

### DIFF
--- a/client/src/cmdhficlass.c
+++ b/client/src/cmdhficlass.c
@@ -2443,8 +2443,9 @@ static int CmdHFiClass_loclass(const char *Cmd) {
 }
 
 static void detect_credential(uint8_t *data, bool *legacy, bool *se, bool *sr) {
-    char* r1 = strstr((char*)data + (5 * 8), "\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF");
-    char* r2 = strstr((char*)data + (11 * 8), "\x05\x00\x05\x00");
+    bool r1 = !memcmp(data + (5 * 8), "\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF", 8);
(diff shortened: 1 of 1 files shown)


iceman1001:
I implemented that byte_strstr,  and still I missed one usage.
And I had the memory overlap in mind,  but figured there are usually no dumps with small AA1 than the limits were.

Good fixing it regardless!

```
    int dlen = byte_strstr(sio_start, 8 * 8, pattern, sizeof(pattern));
    if (dlen == -1) {
        return;
    }
```

since you are cleaning out vars,   there is no need for `dlen` anymore.   Make one if-statement instead.

nvx:
> since you are cleaning out vars, there is no need for `dlen` anymore. Make one if-statement instead.

That's still used for the asn1 decode length, from memory I think it should be possible to infer the length of the asn1 object without needing the explicit length, but I haven't looked into the asn1_print function to see if it's  just being used as a maximum length safety check (at which point we could swap it to just be 7 block length that the SIO credential lives in) or if it actually needs to know the end of the message without the padding at the end.

iceman1001:
You are right.  it is used.

PR #1731, opened 2022-07-23
Added more details to the annotations of `hf mfdes list` output

nvx:
Parameters like appId, fileNo, etc are now shown in the `hf mfdes list` annotation column where available (and where I knew the details). Previously only keyNo on AUTH commands were shown.


iceman1001:
Nice one!

PR #1732, opened 2022-07-25
add 'hf mfu eload -a' to treat dump file as Amiibo

gtalusan:
This adds a parameter to make simulating decrypted Amiibo files a bit easier.

Prior to this, I was using `pm3_amii_bin2eml.pl` but it's a bit cumbersome, since the script only adds a fake 56 byte header to the dump file.  It looks like the client code already handles the lack of a header, so most of the script is unnecessary anyway.

`hf mfu -f myfile -a` will pull out the UID from the decrypted Amiibo binary and helpfully print out a hint to show which command to use.

```
[usb] pm3 --> hf mfu eload -a -f mm.bin
[=] 255 blocks ( 1020 bytes ) to upload
[+] loaded 540 bytes from binary file mm.bin
[=] detected plain mfu dump format
[+] plain mfu dump format was converted to 135 blocks

[=] MFU dump file information
[=] -------------------------------------------------------------
[=] Version..... 00 00 00 00 00 00 00 00 
[=] TBD 0....... 00 00 
[=] TBD 1....... 00 
[=] Signature... 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
[=]              00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
[=] Counter 0... 00 00 00 
[=] Tearing 0... 00 
[=] Counter 1... 00 00 00 
[=] Tearing 1... 00 
[=] Counter 2... 00 00 00 
[=] Tearing 2... 00 
[=] Max data page... 133 ( 536 bytes )
[=] Header size..... 56 bytes

[snip]

[=] MIFARE Ultralight override, will use 149 blocks ( 596 bytes )
[=] Uploading to emulator memory
[=] ......................................................................................................................................................
[?] You are ready to simulate an Amiibo.  Try `hf mfu sim -t 7 -u 04FE42302E5980`
```

### DIFF
--- a/client/src/cmdhfmf.c
+++ b/client/src/cmdhfmf.c
@@ -3866,6 +3866,7 @@ int CmdHF14AMfELoad(const char *Cmd) {
         arg_lit0(NULL, "2k", "MIFARE Classic/Plus 2k"),
         arg_lit0(NULL, "4k", "MIFARE Classic 4k / S70"),
         arg_lit0(NULL, "ul", "MIFARE Ultralight family"),
+        arg_lit0("a", "amiibo", "Treat MIFARE Ultralight as Amiibo"),
         arg_int0("q", "qty", "<dec>", "manually set number of blocks (overrides)"),
         arg_param_end
     };
@@ -3880,8 +3881,9 @@ int CmdHF14AMfELoad(const char *Cmd) {
     bool m2 = arg_get_lit(ctx, 4);
     bool m4 = arg_get_lit(ctx, 5);
     bool mu = arg_get_lit(ctx, 6);
+    bool ma = arg_get_lit(ctx, 7);
 
-    int numblks = arg_get_int_def(ctx, 7, -1);
+    int numblks = arg_get_int_def(ctx, 8, -1);
 
     CLIParserFree(ctx);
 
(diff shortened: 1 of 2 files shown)


iceman1001:
Thanks for your PR,

... its a lot of clutter for printing one line of optional HINT,   so I am not going to merge this one.

Normally,  the PM3 device side knows to use UID from data if not specifically given when simulating 14a based tags.
So the hint isn't even useful.

But the helptext can be improved regardless to make it even more clear,  from what eload/sim states.
```
hf mfu eload -h
hf mfu sim -h
```


I pushed a update to the help texts,  it should be bit more clear.

gtalusan:
Ah, understood.  Would a lua script be preferred?

iceman1001:
Now you updated the PR...  

with adding the pwd gen,   which already exist inside generator.c,  so there is no need to duplicate code. ref https://github.com/RfidResearchGroup/proxmark3/blob/master/common/generator.c#L110-L119

the `-t 7`  should already trigger 8080 pack if pack is zero. ref https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/iso14443a.c#L1648-L1650

you could easily add the pwd gen at the same spot where the pack is added.

gtalusan:
Sorry about that.  I did a forced-push a few minutes after opening the PR and it may have caused confusion.

In either case, you're right in your first remark - it is a bit cluttered when the end game is to only give a UID hint.

I've put it into lua script here https://github.com/RfidResearchGroup/proxmark3/pull/1733.

Thanks for the feedback!

PR #1735, opened 2022-07-29
add hf_mfu_amiibo_restore to restore an Amiibo dump to a blank NTAG215

gtalusan:
hf_mfu_amiibo_restore will restore an Amiibo backup to a blank NTAG215.

Two new Lua-callables are added to the scripting core: keygen_algo_b which is mostly a duplicate of _d, and ul_read_uid so the blank NTAG215 UID's can be used in this script.

This also sneaks in a functional bug fix and some spelling fixes to the Amiibo simulator -- dumps transformed by PyAmiibo's CLI tool have the lock bits unset, which would prevent the simulated tag from being recognized.

The Amiibo database is pulled out into lualibs so that it can be shared between hf_mfu_amiibo_restore and _sim.

A pyscript helper is added to call PyAmiibo with code that doesn't unset the lock bits.


iceman1001:
Looks like improvements indeed.  

So there is a dependency for a PyAmiibo?

gtalusan:
Yeah, it's not a hard dependency for proxmark, unless of course you want to use the Amiibo cloner.

I was tempted to modify the built in amiitool, but it was more work than `pip install`ing PyAmiibo and calling it via proxmark's pyscript.

I may yet do that in a future PR, but for now this works well enough for my purposes 🤣

iceman1001:
You will need to write the requirements inside the python script as a comment,   in order to make things easier for other users.

PR #1739, opened 2022-07-30
add comment indicating pyamiibo should be installed

gtalusan:
As per https://github.com/RfidResearchGroup/proxmark3/pull/1735#issuecomment-1200112432


iceman1001:
... a bit over the top.

I was thinking more in the terms some comment in the beginning of the file like this
```
# Requirements:
# python3 -m pip install pyamiibo
```

gtalusan:
That's easier.  Updated!

iceman1001:
Thanks!

PR #1740, opened 2022-07-31
add -i to hf emrtd info

gtalusan:
I was playing around with my Canadian passport and `hf emrtd info`.  I noticed there was Qt code to show pictures, but the code path wasn't being taken.

Here I add a `-i` / `--images` parameter to ignore the `fast` flag in the `dg_table`.

Qt's QImageReader is fairly smart so it'll know if any image is unsupported and constructs a null QImage if the data is an unknown format.  In my case, JP2 is supported, so the code to short-circuit the picture viewer is removed.

The temporary file creation/sleep code is also removed in favour of loading the image straight from the buffers passed to `emrtd_print_ef_dg2_info` and `emrtd_print_ef_dg5_info`.

### DIFF
--- a/client/src/proxgui.cpp
+++ b/client/src/proxgui.cpp
@@ -67,11 +67,13 @@ extern "C" void RepaintGraphWindow(void) {
 
 
 // hook up picture viewer
-extern "C" void ShowPictureWindow(char *fn) {
+extern "C" void ShowPictureWindow(uint8_t *data, int len) {
     // No support for jpeg2000 in Qt Image since a while...
     // https://doc.qt.io/qt-5/qtimageformats-index.html
-    if (strlen(fn) > 4 && !strcmp(fn + strlen(fn) - 4, ".jp2"))
+    QImage img = QImage::fromData(data, len);
+    if (img.isNull()) {
         return;
+    }
     if (!gui) {
         // Show a notice if X11/XQuartz isn't available
 #if defined(__MACH__) && defined(__APPLE__)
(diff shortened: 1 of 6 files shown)


client/src/proxgui.cpp
```diff
@@ -67,11 +67,11 @@ extern "C" void RepaintGraphWindow(void) {
 
 
 // hook up picture viewer
-extern "C" void ShowPictureWindow(char *fn) {
-    // No support for jpeg2000 in Qt Image since a while...
-    // https://doc.qt.io/qt-5/qtimageformats-index.html
```
iceman1001, line 72: Keep this comment for reference

iceman1001:
Interesting one,  
The image filename...  is information stored in the e-passport,    it would be nice to still print that information.
And when dumping,  we should store the image files.   Is this done in this PR?

gtalusan:
Yeah, nothing really changed with regard to the filename when dumping.  Running `dump` will write to the filenames in the usual spot.  I don't think there's much print/logging in that code path.

When `info` is run, this is what's printed out:

```
[=] -------------------- EF_SOD --------------------
[+] Hash algorithm: SHA-256
[+] EF_DG1: Valid
[+] EF_DG2: File was skipped, but is in EF_SOD.
[+] EF_DG15: Valid
```

When `info -i` is run, this is what's printed out:

```
[=] -------------------- EF_SOD --------------------
[+] Hash algorithm: SHA-256
[+] EF_DG1: Valid
[+] EF_DG2: Valid
[+] EF_DG15: Valid
```
... and the image is shown in the window.

client/src/proxgui.cpp
gtalusan: Updated!

PR #1742, opened 2022-08-02
Adding discovered Vigik (FR) keys

UberGuidoZ:
Source: https://pastebin.com/JRBfjvsi
Tracked to https://play.google.com/store/apps/details?id=fr.badgevigik.scanbadge
Website of Vigik included in commit above

Signed-off-by: UberGuidoZ <57457139+UberGuidoZ@users.noreply.github.com>


UberGuidoZ:
Also added to this slowly growing list:

https://github.com/UberGuidoZ/Flipper/blob/main/NFC/mf_classic_dict/Non-Prox_Keys_Only.nfc

UberGuidoZ:
https://github.com/RfidResearchGroup/proxmark3/pull/1743

iceman1001:
Odd,  very odd,   either Vigik changed their system,   with fixed keys...

Since these keys look bad,   they look like AID's     How reliant is the source of these  ?

UberGuidoZ:
Thanks for your thoughts. These keys were pulled directly from their official app linked above (Android/Google Play Store). Beyond that, I cannot speak to their reliability. (But they are hard-coded within the app and appear to be designed for cloning a badge by purchasing a badge direct from Vigik.)

iceman1001:
yeah,   ok,  so ripped from app,   they still could be apdus... 
Have they been verified against a legit vigik tag?

bettse:
I pulled down the APK and ran apkx on it.  I found the list and while the code is obfuscated, there are strong indicators that they are keys.  The list starts with the factory default, includes the MAD key, and then more than half those below the comment about the "Scan Badge" apk (https://github.com/RfidResearchGroup/proxmark3/blob/master/client/dictionaries/mfc_default_keys.dic#L1052).  I'm wondering if this a newer list of that same app being referenced there.

UberGuidoZ:
> yeah, ok, so ripped from app, they still could be apdus... Have they been verified against a legit vigik tag?

I have not personally verified them against a legit Vigik tag and do not have a way to do so. But I do know someone who may be able to. (@equipter ?)

UberGuidoZ:
> I pulled down the APK and ran apkx on it. I found the list and while the code is obfuscated, there are strong indicators that they are keys. The list starts with the factory default, includes the MAD key, and then more than half those below the comment about the "Scan Badge" apk (https://github.com/RfidResearchGroup/proxmark3/blob/master/client/dictionaries/mfc_default_keys.dic#L1052). I'm wondering if this a newer list of that same app being referenced there.

Thanks for your thoughts, @bettse - Equip mentioned something along the same lines when we were looking into the usefulness of the keys. I discovered the [previous keys](https://github.com/RfidResearchGroup/proxmark3/blob/master/client/dictionaries/mfc_default_keys.dic#L1052) and noticed they were not duplicates. If legit and valid, these would be quite useful.

iceman1001:
Not gonna lie,  they look like ISO7816 based commands.   Or AID's...
Too many zeros in them.     Besides the silly default keys,  we normally see  like a few zeros in the start of a key.

Not trying to hard about merging some keys, its just that trying to verify that they are actually static keys in use.

For this commit,  these ones are the ones I really doubt.  

```
A00000000007
A00000000008
A00000000064
A00000000075
A00000000113
A00000000137
A00000012684
A000000301FA
A00000030F1A
A00001000030
A00001000096
A00001000DAB
A000010012F8
A00002000057
A00002000084
A00002000503
A00003000028
A00003000036
A00005000021
A00005000468
A00006010A9B
A00008000AEA
A00009000064
A00009150B0C
A0000A000057
A0000C040BB8
A0000F000036
A00012000036
A00025000030
A00027000030
A0002D00014B
A0002F000099
A00030000028
A00031000057
A00038000036
A00039000036
A0003A070099
A0003D000030

B00000000048
```


As example of existing keys with lots of zeros.  I selected keys with `00000` zeros in them.

``` 
-- Silly test keys,  I strongly doubt any real system uses them. But users might for test vectors in their programs.
68:  000000000001 
69:  000000000002 
70:  00000000000a 
71:  00000000000b 
74:  100000000000 
80:  200000000000 
99:  a00000000000 
104:  b00000000000


-- first five zeros,  Possible 32b RND.  Also verified as default keys.
46:  00000ffe2488 # Västtrafiken KeyB 
624:  0000014B5C31 # Vingcard Mifare 4k Staff card

-- could be smaller RNDs
207:  1F1FFE000000
617:  0000000018de

-- user selected
24:  e00000000000 # iCopy-X

-- just odd,  most likely user selected too
996:  0000085f0000
```




Like,  staring with A0 ,  B0   is ISO7816 select file,   read binary commands..
This is why my spider senses tingles.

equipter:
those A0 ones are APDUs fwiw.  from the app they're mixed in with other assets but they are for the APDU commands not keys :p

UberGuidoZ:
> those A0 ones are APDUs fwiw. from the app they're mixed in with other assets but they are for the APDU commands not keys :p

Ahh, excellent info - then they should be removed to avoid bloat. Thank you for the insight! I knew it would be wise to invite you to the discussion. =)

@iceman1001, what are your thoughts on the rest? I'm going to pull those A0 ones out of my repo. (Your spidey sense has done you well.)

iceman1001:
Remove the bad ones,  and keep the rest.  Makes me doubt a bit more of the previous merged keys from APK application "Scan Badge"

We will need to make sure there is not bad ones left there too.

iceman1001:
@UberGuidoZ   I am going to close this PR

Make a new one with the corrections.   And add the changelog.md change in the same PR..

UberGuidoZ:
Will do, thank you.

quantum-x:
For what it's worth the "ScanBadge" application is *not* released via the VIGIK organisation.
It's run by a guy who was just sentenced by a french court for counterfeit, fraud and illegal use of the VIGIK brand. It's tied to the site 1badge.fr

quantum-x:
I can also verify that these keys are used by certain brands of badges in the French market.
They're also quite old (relatively speaking).

image

iceman1001:
How about that, @quantum-x lurking around. Better confirmation of keys can't be found :)

quantum-x:
Lurking until my beard is as impressive as yours ;)
I get the feeling these keys (based on dates released) correspond to firmware updates in French access control systems, where they added new keys to try to buy some time ..

iceman1001:
Almost don't have any beard left ;)
So even if those keys looks like APDU's,  they still where used.

@UberGuidoZ  you can add those confirmed to be keys cases to your new PR.

quantum-x:
@iceman1001 Correct :)

UberGuidoZ:
@equipter Check it out. Those keys are good to keep after all! We should add these as well as the "iGuard Simple and reverse keys" that you found today into the PR. Currently active one: https://github.com/RfidResearchGroup/proxmark3/pull/1749

PR #1744, opened 2022-08-02
Fix undefined symbol when build with clang

mywalkb:
I builded latest commit for termux, but I got an 

```
ld: error: undefined symbol: TexcomTK17Get2Bits
>>> referenced by cmdhftexkom.c:354 (src/cmdhftexkom.c:354)
>>>               obj/cmdhftexkom.o:(TexcomTK17Decode)
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
```

so I fixed


iceman1001:
Good catch!

PR #1745, opened 2022-08-03
Add installation guide for macOS Ventura Beta users

alexandrereol:
Add a note for macOS Ventura users on how to install with Xcode 14.0, notably that you need to rename Xcode-beta.app to Xcode.app in order for installation to succeed.

Installation succeeded on macOS Ventura Beta 4 using method described.

### DIFF
--- a/doc/md/Installation_Instructions/Mac-OS-X-Homebrew-Installation-Instructions.md
+++ b/doc/md/Installation_Instructions/Mac-OS-X-Homebrew-Installation-Instructions.md
@@ -29,6 +29,7 @@ Early users of macOS Ventura and Xcode 14.0 might run into an error saying that 
 
 If (and only if) you run into that error, here is the fix:
 - RE-download the *latest* Command Line Tools of Xcode Beta 14 and install them (again). (https://developer.apple.com/download/all/)
+- Run `sudo xcode-select -s /Applications/Xcode-beta.app` in Terminal.
 - Proceed with Brew installation
 That should normally fix the issue.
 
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice,
Maybe be even more clear in which steps and how to execute them?  Like where those apps are located etc.

gtalusan:
Hello, MacOS Ventura and Xcode14-beta user here.  I didn't have to rename the Xcode.app package at all.

Xcode has provisions to switch SDKs like so:

```
sudo xcode-select -s /Applications/Xcode-beta.app
```

and then you forget about it until every thing breaks in a few months.

Then you subsequently use the same command to switch back to the release version of Xcode.app!

iceman1001:
That sounds  like a smoother solution

alexandrereol:
macOS Ventura Beta 4 and Xcode 14.0 installation was not successful. When installing pm3 through brew, my setup was always getting an error saying that my Xcode (14.0) was out of date and that I needed to download the latest one from Apple (Even though I already was on the latest possible version). Executing the xcode-select command @gtalusan mentioned did not help. As soon as I renamed my Xcode.app to Xcode2.app and Xcode-beta.app to Xcode.app it did work. App renaming can be reversed after installation though.

alexandrereol:
I fully understand app renaming is not a pretty solution, but it was the only way of not getting this (false!) error saying my Xcode was out-of-date. It's only meant as notice for Beta users and can safely be removed when Xcode 14.0 RC releases in September/October.

gtalusan:
It's likely that your command line tools are out of date.  I remember seeing an Xcode 14 out-of-date message some months ago, and installing the newer CLT from the Apple Dev Center is what fixed it.  The issue is more to do with brew's diagnostic messages.  Indeed, I think believe you'll see similar messages for any other brew recipe and not just proxmark.

iceman1001:
The text,    it would need to  be rephrased.

Will @alexandrereol or @gtalusan   do it?

alexandrereol:
I've updated the instructions to reflect the changes discussed with @gtalusan 👍

iceman1001:
Excellent!

PR #1746, opened 2022-08-04
Fix gallagher KDF generation for Admin Cards aka Reader Config Cards

nvx:
Admin Cards (also called Reader Config Cards in some places) use slightly different KDF rules.

Verified functionality against a real system.


iceman1001:
Thanks!

PR #1747, opened 2022-08-05
Fix eload from json files failing

nvx:
Fix eload from json files failing (ie, hf mf eload -f file.json) the dump is the full size.

Also fixed an accidental client double free when loading dumps.


iceman1001:
I figure by the rearranging it was the last block that failed to load.

And good catch with the double free.

PR #1752, opened 2022-08-11
Improve ECP annotation, add mobile AIDs

kormax:
This PR includes 2 commits:

1) Improve annotations:
ECP2 payload can have varying length, if it didn't match the frame was not annotated, this is fixed now.
Reader type is now added to annotation if it is known.

2) Add AIDs used by mobile-first applets:
Note that some of them might not appear when testing via `hf 14a info --aidsearch`  because the device bails-out after a dozen of failed SELECTs. I've decided to add them for public knowledge nonetheless.

### DIFF
--- a/client/resources/aidlist.json
+++ b/client/resources/aidlist.json
@@ -2206,5 +2206,61 @@
         "Name": "SL Resekort",
         "Description": "transport card",
         "Type": "transport"
+    },
+    {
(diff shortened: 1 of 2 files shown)


kormax:
I've made a mistake by including a commit from the other PR here. If you accept this one, the other one can be ignored.

iceman1001:
Nice improvements!

PR #1755, opened 2022-08-17
hf_mf_hid_sim.lua

micsen:
Added a script that will allow for simulation of HID Mifare credentials
Currently only with H10301 support


iceman1001:
Cool script!

iceman1001:
Don't forget to add a line in the CHANGELOG.md

PR #1757, opened 2022-08-18
Update Makefile.defs

blshkv:
Do not overwrite global CC values (if specified in env)

Signed-off-by: Anton Bolshakov <blshkv@users.noreply.github.com>


iceman1001:
image

looks like `LD` didn't like this suggestion at all...

blshkv:
That basically means that there's a problem with your global LD setting. Sorry, I'm not an expert here. You can leave LD value unchanged then

iceman1001:
... that is our build actions on github...  

I will revert this and let your suggestions be hold until someone who can verify it is available.

blshkv:
The makefile is badly written,
```
 sable: $(OBJS)
-       $(LD) $(LDFLAGS) -o sable $(OBJS) $(LIBS)
+       $(CXX) $(LDFLAGS) -o sable $(OBJS) $(LIBS)
```

iceman1001:
Sadly enough it is out of my league.    Makefile is dark woodoo to me. 

If you have someone who can review it,  let me know

PR #1761, opened 2022-08-25
A few extra MF classic keys

borisgolovnev:


iceman1001:
Cool,
are these static keys or are they from a metro system that makes dynamic keys based on uid for example?
The dictionary wants static keys,   since dynamic keys is connected to one card only.

borisgolovnev:
Good question, I don't really know. Any way to find this out?

iceman1001:
Cool,  cool,
Usually if you have two or more cards from same system,  you can see if the keys are static.

borisgolovnev:
Ok, makes sense.

iceman1001:
did you check the keys on your hotel cards if they were static or?

PR #1763, opened 2022-08-30
do not use LD if you aren't using ld

ZeroChaos-:
All the uses of LD actually wanted g++, so just use CXX and remove an
unused LD define

Closes: https://github.com/pentoo/pentoo-overlay/issues/1259
Fixes: https://github.com/RfidResearchGroup/proxmark3/pull/1758


iceman1001:
Nice one!

PR #1765, opened 2022-08-31
Prep spiffs for bigger data files

mwalker33:
- added spiffs check after flase wipe to force it to update its status
- added spiffs write and append to write in 8192 byte chunks to allow spiffs space to be freed in time.
- fixed spiffs dump to correctly handle issues if it could not allocate bugbuff space.


iceman1001:
Excellent!

PR #1781, opened 2022-09-20
print_sr_blocks(): don't assume uid is at begining of dump

DidierA:
`print_sr_dump()` assumes uid is at begining of dump. This is not true for SRT512 tags (the uid is not stored anywhere in the data, it is obtained b sending `Get_UID()` to the chip), and this PR is an attempt to fix this.  
The UID is used to guess the chip type, and interpret the system block. A wong guess of the chip ID results in an inaccurate display of which blocks are write protected.

This fix adds and `uid` parameter to `print_sr_blocks()`, and provides it to the call if we are dumping a tag. If we are reading a dump file, it attempts to extract the uid from the file name, assuming uid has a len of 8 bytes, and the file has the standard format of `hf-14b-xxxxx.*`

This works for my limited test case (one SRT512 tag). If extracting the uid from the data works for some cases then this fix probably breaks them, and an alternative should be found. Is there a standard way to store the UID in the dump file?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Changed `hf 14b dump, view` to get correct chip type in case of SRT512 and friends (@DidierA)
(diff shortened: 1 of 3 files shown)


iceman1001:
hm.. haven't looked at it..
But I suggest we do a better json dump format for 14b,  and we use that as loading extra information.

DidierA:
You mean redefine json format in a way  such as what is described in mfu_binary_format_notes.md?
I’ll try to write something similar.

iceman1001:
merging anyway,   its better than nothing.

PR #1784, opened 2022-09-24
Update bignum.c

sun1ch:
Delete unused variable t = 0

Signed-off-by: Alex <98668583+sun1ch@users.noreply.github.com>

### DIFF
--- a/common/mbedtls/bignum.c
+++ b/common/mbedtls/bignum.c
@@ -1439,7 +1439,7 @@ static
 __attribute__((noinline))
 #endif
 void mpi_mul_hlp(size_t i, mbedtls_mpi_uint *s, mbedtls_mpi_uint *d, mbedtls_mpi_uint b) {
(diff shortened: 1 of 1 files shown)


iceman1001:
its a bit unclear this one.
Since the mbedtls uses ASM code in macros direct after the definition of variable t.  

`MULADDC_INIT`    etc..

And I am not sure at all if the ASM code actually fills the memory or uses the memory for this variable t.

sun1ch:
There's an error appears with variable t = 0  in bignum.c

sun1ch:
Without t = 0 in 1442 string and t++ in 1448 client compiled without error

iceman1001:
yes,   I understand that there is an warning in GCC 12,   but your fix doesn't verify if removal of the variable t, isnt used in the mbedtls macros.

iceman1001:
closing because of inactivity from poster

PR #1785, opened 2022-09-25
Add print flag to hf iclass encode

natesales:
Adds a `--print` flag to display encoded iCLASS blocks instead of writing them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Added print flag to `hf iclass encode` (@natesales)
  - Added `hf 15 eview` and `hf 15 esave` - Retrieve emulator image for ISO15693 simulation (@markus-oehme-pg40)
  - Changed `hf 15 sim` - now supports reader writes (@markus-oehme-pg40)
  - Added `hf 15 eload` - specify memory image for ISO15693 simulation (@markus-oehme-pg40)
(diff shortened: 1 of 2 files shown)


iceman1001:
Interesting...
the purpose of the command is to write to a card.  
there is no meaning to the command to look at encoded data.

From a user perspective,  and you want to look at how different encoded data is being produced,  maybe make its own command.   I don't commands being hijacked for different purposes,

natesales:
What would you suggest? Viewing encoded data uses much of the same code as writing it to a card, so would it make more sense to extract the encoding part of CmdHFiClassEncode to a new function and calling it in a new command in addition to "hf iclass encode"?

iceman1001:
In the end I see no use for this functionality.  If you need the encoded data,  for what?  Not going to help you with anything if you not going to use it with another software,   which means that software needs to be improved.

natesales:
This is useful so you can modify the encoded iCLASS blocks before writing them to a card, or save them to a file and emulate on the proxmark. Currently there isn't a way to emulate an arbitrary encoded iCLASS credential without writing it to a card, reading it, then emulating the dump from the card (which of course doesn't preserve immutable fields like the CSN).

iceman1001:
Let me decompose your suggetions.  First the specific command purpose is to write encrypted encoded data to picopass tags.

- These blocks are encrypted,  there  is no way you "modify" them as is.  
- Save them to file?  That sounds like a whole different command to me.
- Simulation of the encrypted credentials is a interesting concept but you would need more that those blocks to do so, and that would mean you would have a different command for it.

As I see it,  we use the "-v"  verbose flag to print out things of interest,   which would follow how the client is designed today.
The command would still encode the data and print what it wrote.  



To summarize,  
- I see  "Simulate Encoded credentials"  as a new command
- I see  "-v verbose mode"  to print out the encoded data and still write to tag.

iceman1001:
Closing, 

I look at making a verbose and the command already prints the output.   This PR will change the command to be something different.

PR #1787, opened 2022-09-29
Removed Duplicates from default dictonaries.

amec0e:
Was looking through the default dict as I felt I might need it soon and noticed some duplicates. So I thought I'd make a PR

t55xx_default_pwds - L73 same as L152 and L13 same as L42.
mfc_default_keys - L550 same as L1544, L548 same as L1550 and L632 same as L1749


iceman1001:
Good catch!
seems like there has been duplicates added..

amec0e:
Not a problem at all, it happens. I did not edit the change log as I was not sure if it was change log worthy, nonetheless happy to help 😄

PR #1788, opened 2022-09-29
Adds Uruguay transport system key

elcuervo:
Adds the shared key between all the cards in the Transport system of Montevideo, Uruguay


iceman1001:
Thanks!

PR #1789, opened 2022-09-30
Removed another duplicate

amec0e:
Good catch on the last duplicate I missed, due to the comments next to the key it went under the radar. After removing all the comments in a notepad I double checked and there was one more we missed.

L628 same as L1747


iceman1001:
good catch indeed,  
and here I was sure my `sort xxx.dic | uniq -c | sort -r`    would have given me the duplicates but no..

PR #1792, opened 2022-10-05
Clean up of t55xx_default_pwds

amec0e:
Removed 1 whitespace, grouped some keys, uppercased keys and moved line comments to above the keys. Duplicate checked after and all is good, just another little spring clean (wasn't much I'll admit that)


iceman1001:
Always good to have unified text files in style

PR #1802, opened 2022-10-27
Fixed "Residency Permit" Output

noahclements:
Bug discovered while scanning Swedish Residency Permit, resulted in document type of "German Residency Permit". Fixed for better accuracy.

Signed-off-by: Noah Clements <noah.clements@unb.ca>


iceman1001:
... more a textual fix than a bug,  but nevertheless it a good catch!

PR #1805, opened 2022-10-31
Modified logic to also try key B on MF cards.

jmichelp:
In configurations where keyA is unknown but ACLs are configured to allow all blocks to be read by keyB the command `hf mf dump` was failing. This commit attempts to fix this behavior by trying keyA first and swapping for keyB if half of the allowed tries failed.


iceman1001:
I pushed a fix for that

PR #1806, opened 2022-11-04
Fix prolematic return codes in mifare.

jmichelp:
Parts of the code returned positive values for error codes, which could result in the client exiting (return value 2).


iceman1001:
Great catches.   There are some more places where return codes doesnt follow our new prefered ways with PM3_SUCCESS / PM3_Exxxxx codes.

PR #1807, opened 2022-11-05
Add hf mf gload, ggetblk, gsetblk

DidierA:
Adds commands to load dump file to Magic Gen4 GTU cards.


iceman1001:
Great contribution,   nicely coded too.  +1

PR #1811, opened 2022-11-13
hf mf gview, gload : 2x speed

DidierA:
on Gen4 GTU cards, a select was made before each read or write command.
This commit adds a flag parameter to the read and write commands,
and gload and gview commands use this flag to tell when to select and end.

The trace buffer also contains the complete operation instead of the last
blokc read/write.

Speed gain:
`time ./proxmark3 -p /dev/ttyACM0 -c 'hf mf gview --4k'`
before: about 20s, now: about 7s

`time ./proxmark3 -p /dev/ttyACM0 -c 'hf mf gload --4k -f ../../../dumps/4k.bin'`
before: about 23s, now: about 10s


iceman1001:
Interesting, My guess is the 2 second timeout for each call which was the culprit for the slowness.

armsrc/mifarecmd.c
```diff
@@ -2684,63 +2684,125 @@ void MifareGen3Freez(void) {
...
     BigBuf_free();
 }
 
-void MifareG4WriteBlk(uint8_t blockno, uint8_t *pwd, uint8_t *data) {
-    iso14443a_setup(FPGA_HF_ISO14443A_READER_LISTEN);
-    clear_trace();
-    set_tracing(true);
+void MifareG4WriteBlk(uint8_t blockno, uint8_t *pwd, uint8_t *data, uint8_t workFlags) {
+    bool setup = workFlags & MAGIC_INIT & 0xFF ;
+    bool  done = workFlags & MAGIC_OFF  & 0xFF ;
```
iceman1001, line 2760: you extracting the flag here?

The following style is our preferred way 
```
bool setup = ((workFlags & MAGIC_INIT) == MAGIC_INIT):
```

DidierA: done.
While doing this, I noticed the led on/off pattern may not be consistent with other functions, is there a note defining what each led is assigned to, when to switch on/off each led?

iceman1001: yeah,   the LED is not unified across all different device side commands.   Feel free to contribute

PR #1813, opened 2022-11-16
hf_mfu_next.lua - new script for quickly modifying the DT NeXT/xNT implant (NTAG216)

flamebarke:
Signed-off-by: flamebarke <39644720+flamebarke@users.noreply.github.com>


iceman1001:
If you add a line in the CHANGELOG fle,     

Normally,  I am not fond of ASCII as input to passwords etc,  since we use bytes everywhere else but since it's a lua script. 

I notice there is no verification that you are operating on a NTAG216 before,  but blindly run the commands. 
Maybe something to improve in next version?

pcunning:
Makes more sense to name this hf_mfu_ntag since that is the chip line. It would also be nice to support the different sizes.

flamebarke:
> If you add a line in the CHANGELOG fle,
> 
> Normally, I am not fond of ASCII as input to passwords etc, since we use bytes everywhere else but since it's a lua script.
> 
> I notice there is no verification that you are operating on a NTAG216 before, but blindly run the commands. Maybe something to improve in next version?

> It would also be nice to support the different sizes.

No worries, when I have a bit of time I will update the script to have a validation check, include 213/215 and change the password input from ascii to hex.

iceman1001:
the renaming,  good catch @pcunning  ,   yeah   should definite be renamed, a simple thing,   can be done after a merge too.

flamebarke:
> the renaming, good catch @pcunning , yeah should definite be renamed, a simple thing, can be done after a merge too.

I've renamed the script to `hf_mfu_ntag`, and added in a line to the changelog. I'll do an update in the future on v1.0.1 and add in the recommended changes.

PR #1814, opened 2022-11-20
Hf 14a sim : fix parsing error and add hint for ultralight EV1

DidierA:
`hf mfu sim -t 7` or `hf 14a sim -t 7`  without any other option was stopping after 7 reads, the proposed fix coreects that.
Also added a hint in the help text for Mifare Ultralight EV1 to type 7.


iceman1001:
Nice catch with the value being set,  
Should look if we have more of those direct assignments in the client.

The text addition,    we have a limited amount of screen space.  The hint is now too long.
How about just   "MFU / NTAG 215  Amiibo"  for the hint
And in the description,   write a comment about type7 ...

DidierA:
I changed the help text, it now fits in 80 char width.

iceman1001:
Thanks!

PR #1815, opened 2022-11-20
Add hf mfu esave

DidierA:
Adds `hf mfu esave` command, to dump emulator memory into a mfu dump file.
Also changes `hf mfu eview` so that by default only the pages corresponding to the card size are shown.

### DIFF
--- a/client/src/cmdhfmfu.c
+++ b/client/src/cmdhfmfu.c
@@ -2718,6 +2718,7 @@ static int CmdHF14AMfUeLoad(const char *Cmd) {
     PrintAndLogEx(HINT, "Try " _YELLOW_("`hf mfu sim -t 7`") " to simulate an Amiibo.");
     return res;
 }
+
 //
 //  Simulate tag
 //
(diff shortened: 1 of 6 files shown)


client/src/cmdhfmfu.c
```diff
@@ -4098,41 +4099,135 @@ int CmdHF14MfuNDEFRead(const char *Cmd) {
     return status;
 }
 
+// utility function. Retrieves emulator memory
+static int GetMfuDumpFromEMul(mfu_dump_t **buf) {
+
+    uint8_t *dump = malloc(sizeof(mfu_dump_t));
```
iceman1001, line 4105: we use calloc,   to make sure the allocated memory is zeroed on all platforms.

iceman1001:
Nice!

Make sure for that you use calloc instead,

DidierA:
Yes ,it was originaly using calloc, but it caused a failure in the Windows build test.
Investigating the failure I did not really understood why calloc was used, since we're allocating a struct, and not an array, and thouht malloc was more appropriate.  If the concern is zeroing, would a `memset(dump, 0, sizeof(mfu_fump_t))` after the |malloc()| call be appropriate?
There's also the option of a stack variable, we're dealing with a little more than a kilobyte here :)

iceman1001:
that is what calloc does behind the scenes.   
look at other callocs we use in order to find how to correctly call it

DidierA:
I've changed it in my branch (commit 182d167aad4951da271952da3de39fbf069f8865) , should I open a new PR?

PR #1817, opened 2022-11-23
Added luascript for the NTAG I2C PLUS 2k

flamebarke:
New script for modifying user memory on sectors 0 and 1 on the NTAG I2C PLUS 2k. Can be used to read blocks, write blocks, write a file to a sector or dump a sector to disk.


iceman1001:
Nice!

PR #1818, opened 2022-11-23
use calloc(), not malloc()

DidierA:


client/src/cmdhfmfu.c
```diff
@@ -4102,20 +4102,20 @@ int CmdHF14MfuNDEFRead(const char *Cmd) {
 // utility function. Retrieves emulator memory
 static int GetMfuDumpFromEMul(mfu_dump_t **buf) {
 
-    uint8_t *dump = calloc(MFU_MAX_BYTES + MFU_DUMP_PREFIX_LENGTH, sizeof(uint8_t));
+    mfu_dump_t *dump = calloc(1, sizeof(mfu_dump_t));
```
DidierA, line 4105: hmmmm...
sizeof(mfu_dump_t) = 1080
MFU_DUMP_PREFIX_LENGTH + MFU_MAX_BYTES = 56 + (255 *4) = 1076.
This is probably what triggered the error in Widnows build test; but now we're reading 4 bytes more than the original code.

DidierA:
Now tested with dump of an MF0UL1101 containing a non-empty pack, after `hf mfu eload` and `hf mfu esave` the contents are the same.

client/src/cmdhfmfu.c
DidierA: GetFromDevice() call now retrieves only MFU_MAX_BYTES + MFU_DUMP_PREFIX_LENGTH.

iceman1001:
Nice!

PR #1819, opened 2022-11-23
Add initialization command response interpretation on the ksx6924 (Tmoney) card

toucan12:
Hi.
I added interpreting the initialization command response of the ksx6924 (Tmoney) card.
I also attach a screenshot.

image


iceman1001:
Nice one!
Not that I have such a card to test with .

PR #1820, opened 2022-11-24
Fix overflow in SPI memory when writing default key dictionnaries.

jmichelp:
Boundaries were defined in the memory layout but weren't enforced by the client, causing an overflow when trying to load a dictionnary that was too big.
It's too hard to enforce it on the ARM side as the command is a generic write.
Now that limits are defined, also outputs them as part of the `hw status` command.


iceman1001:
Nice enforcement of the different reserved space.  It very much possible to overwrite things.

iceman1001:
Since the default dictionary for mifare now is longer than 8kb,   its time to increase it with another block. 
```
// Reserved space for MIFARE Keys = 8 kb
````

iceman1001:
Great!

PR #1821, opened 2022-11-24
Add Magic Gen4 GTU detection

DidierA:
Adds identification of magic gen4 GTU cards, using the dump configuration command. Card will only be detected if the password is default (00000000).

Added symbolic names for magic gen4 gtu protocol in `protocols.h` and changed src to use them.

Also changed the mifare classic wipe function in `hf_mf_ultimatecard.lua`  to erase the whole 4K memory and add all sector trailers.

### DIFF
--- a/doc/magic_cards_notes.md
+++ b/doc/magic_cards_notes.md
@@ -998,9 +998,14 @@ Can emulate MIFARE Classic, Ultralight/NTAG families, 14b UID & App Data
 ### Identify
 ^[Top](#top) ^^[Gen4](#g4top)
 
-👉 **TODO** Tag doesn't get identified correctly by latest Proxmark3 client (it might get mislabeled as MFC Gen2/CUID, Gen3/APDU or NTAG21x Modifiable, depending on configured UID/ATQA/SAK/ATS)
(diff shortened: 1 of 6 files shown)


doc/magic_cards_notes.md
```diff
@@ -998,9 +998,12 @@ Can emulate MIFARE Classic, Ultralight/NTAG families, 14b UID & App Data
 ### Identify
 ^[Top](#top) ^^[Gen4](#g4top)
 
-👉 **TODO** Tag doesn't get identified correctly by latest Proxmark3 client (it might get mislabeled as MFC Gen2/CUID, Gen3/APDU or NTAG21x Modifiable, depending on configured UID/ATQA/SAK/ATS)
```
iceman1001, line 1001: This information is still relevant,   since the new grn4gtu identification has limitations.

iceman1001:
Interesting PR,   lot of mixed stuff,  we prefer one thing one pr since its easier to keep track of which change.

You will need to revert the text change in the magic note,  since that information is still relevant

doc/magic_cards_notes.md
DidierA: re-added the note, with precision on the default password requirement.

iceman1001:
Thanks!

PR #1823, opened 2022-11-29
Add keys

r1ddl3rz:


iceman1001:
Nice,
Are they verified to be non-diversified keys?

r1ddl3rz:
> Nice,
> Are they verified to be non-diversified keys?

I think so. I have different cards with the same key. Diversified keys shouldn't be the same? Or how could I verify that?

iceman1001:
different cards same key,  means its not diversified.  Which is good for this dictionary

PR #1824, opened 2022-11-29
Update Termux notes to recommend F-Droid

TrevorSayre:
- Updated requirements list to use links and ensure Bluetooth only requirements are clearly marked
- Update links to Termux to point to F-Droid
- Include link to F-Droid
- Include reason why F-Droid Termux is recommended


iceman1001:
Thank you!

PR #1825, opened 2022-11-30
Get the hitag2 dump command working as per docs

natmchugh:
I decided to try and understand some of the simpler hitag commands and this one looked like the simplest non working one. The command as written in the client didn't seem to be sending any commands. Hopefully I've understood it correctly and not broken anything.

**Example usage**
```
[usb] pm3 --> lf hitag dump -h

Read all card memory and save to fileIn password mode the default key is 4D494B52 (MIKR)                                                                  
In crypto mode the default key is 4F4E4D494B52 (ONMIKR)  format: ISK high + ISK low.                                                                      

usage:
    lf hitag dump [-h] [-f <fn>] [-k <hex>] [--nrar <hex>]

options:
    -h, --help                     This help
    -f, --file <fn>                specify file name
    -k, --key <hex>                key, 4 or 6 hex bytes
    --nrar <hex>                   nonce / answer reader, 8 hex bytes

examples/notes:
    lf hitag dump -k 4F4E4D494B52 
    lf hitag dump -k 4D494B52  
[usb] pm3 --> lf hitag dump -k AAAAAAAAAAAA -f testy
[=] Authenticating in crypto mode
[#] hitag2_crypto: key=0x555555555555 uid=0x48a41956
[#] Read successful!
[+] Dumping tag memory...
[=] 00 | 6A 98 25 12 | j.%.
[=] 01 | AA AA AA AA | ....
[=] 02 | 00 00 AA AA | ....
[=] 03 | 0E 91 34 08 | ..4.
[=] 04 | DE AD BE EF | ....
[=] 05 | 55 55 55 55 | UUUU
[=] 06 | AA AA AA AA | ....
[=] 07 | 55 55 55 55 | UUUU
[=] 08 | 00 00 00 00 | ....
[=] 09 | 00 00 00 00 | ....
[=] 10 | 00 00 00 00 | ....
[=] 11 | 00 00 00 00 | ....
[+] saved 48 bytes to binary file testy.bin
[+] saved 12 blocks to text file testy.eml
[+] saved to json file testy.json
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Fixed `trace list -t mf` - now also finds UID if anticollision is partial captured, to be used for mfkey (@iceman1001)
  - Added `hf mf gload, gsave, ggetblk, gsetblk` for Gen4 GTU in mifare classic mode (@DidierA)
  - Fixed SPI flash overflow when loading dictionnaries into flash. Breaking change: added 1 more sector for Mifare - dictionnaries should be loaded again (@jmichelp)
+ - Fixed `lf hitag dump` - Should now work as described in the command help (@natmchugh)
 
 ## [Radium.4.15864][2022-10-29]
  - Changed `lf indala sim` - now accepts fc / cn (@iceman1001)
--- a/client/src/cmdlfhitag.c
+++ b/client/src/cmdlfhitag.c
@@ -844,14 +844,39 @@ static int CmdLFHitag2Dump(const char *Cmd) {
         return PM3_EINVARG;
     }
 
-    PrintAndLogEx(SUCCESS, "Dumping tag memory...");
+    hitag_function htf;
+    hitag_data htd;
+    memset(&htd, 0, sizeof(htd));
 
+    if (keylen == 6) {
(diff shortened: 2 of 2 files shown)


natmchugh:
Good point let me see if I can fix that.

iceman1001:
Yeah,  here is the catch,  the device side refactoring isn't done.  So this command most likely is falling back to the defaults on device side when communicating with device.

natmchugh:
Right I have had a look and the reason is the message reverses the bits of the hex password before displaying it as debug info https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/hitag2.c#L779. So 1010 (A) becomes 0101 (5). I guess this is because that is the way it is sent to the function that sets up the crypto and that was more use when originally writting the function. Same with the uid.

I could remove the reverse macro for the message and in this example it would show `key=0xaaaaaaaaaaaa`. I guess that would make more sense for an end user. Or I could change the message to make it more clear these were bit reversed values. What do you think?

iceman1001:
Good figuring out stuff!

how about we print both values,   one normal and one reversed? 
That way we don't loose any information

iceman1001:
I merge this one,  so other adaptations can be made in a new PR.

PR #1826, opened 2022-12-02
Fix `trace list -r` (relative times)

nvx:
Fix `trace list -r` (relative times) not working unless the `-u` (micoseconds) flag was also provided.

Also makes the `--frame` option respect microseconds and relative times as well.


iceman1001:
Great catches!

```
trace load -f traces/hf_14a_mfu-sim.trace

trace list -1 -t 14a --frame -c -u -r
trace list -1 -t 14a --frame -c -u -r --frame

trace list -1 -t 14a --frame -c -u 
trace list -1 -t 14a --frame -c -u --frame
```

lets chat on discord about it

PR #1828, opened 2022-12-05
Fixing some ISO7816-4 logic.

jmichelp:
- Fixing overflow in BigBuffer if requesting to much data
- Fixing integer wrapping with the SIM module by capping APDU data length
- Fixing chaining logic which was loosing previously received data


iceman1001:
Thanks!

Maybe the apdu len macro could have a different name.  too generic,  but I merge it

PR #1830, opened 2022-12-05
Smartcard module firmware 4.12

jmichelp:
Major bump on the module firmware by Sentinel.
Length are encoded on 2 bytes now to avoid an integer overflow when reading data (firmware version is kept on 1 byte for backward compatibility).
This is a breaking change: ARM code is NOT backward compatible with 3.11 version when sending commands!


iceman1001:
Nice fixes,  
however the sim module asm and hash,  is something I PR.    You will need to remove that from your PR.

jmichelp:
Sure. I thought it would be easier to have both in 1 PR as you could check the hash and  the new firmware too. Removing it.

iceman1001:
I understand your thinking but its about origin,  which I only trust from the source.

PR #1834, opened 2022-12-07
Update iclass_default_keys.dic

Antiklesys:
Added an extra key from https://youtu.be/EH7dctOPxBA?t=805


jmichelp:
This is the same as the AA1 key. Just the DES parity hasn't been fixed.

iceman1001:
duplicate

Issue #1843, opened 2022-12-17
Compilation error on openSuse Leap 15.3


nitraiolo:
Compilation was working till a couple of weeks ago (sorry I didn't save the git pull log).

Now I get a this compilation error:

src/cmdscript.c: In function ‘CmdScriptRun’:
src/cmdscript.c:406:9: error: unknown type name ‘PyConfig’
         PyConfig py_conf;
         ^~~~~~~~
src/cmdscript.c:407:9: error: implicit declaration of function ‘PyConfig_InitIsolatedConfig’ [-Werror=implicit-function-declaration]
         PyConfig_InitIsolatedConfig(&py_conf);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
src/cmdscript.c:407:9: error: nested extern declaration of ‘PyConfig_InitIsolatedConfig’ [-Werror=nested-externs]
src/cmdscript.c:411:16: error: request for member ‘use_environment’ in something not a structure or union
         py_conf.use_environment = 1;
                ^
src/cmdscript.c:412:16: error: request for member ‘user_site_directory’ in something not a structure or union
         py_conf.user_site_directory = 1;
                ^
src/cmdscript.c:419:9: error: implicit declaration of function ‘PyConfig_SetBytesArgv’ [-Werror=implicit-function-declaration]
         PyConfig_SetBytesArgv(&py_conf, argc + 1, argv);
         ^~~~~~~~~~~~~~~~~~~~~
src/cmdscript.c:419:9: error: nested extern declaration of ‘PyConfig_SetBytesArgv’ [-Werror=nested-externs]
src/cmdscript.c:421:9: error: implicit declaration of function ‘PyConfig_SetBytesString’; did you mean ‘PyMapping_SetItemString’? [-Werror=implicit-function-declaration]
         PyConfig_SetBytesString(&py_conf, &py_conf.home, getenv("PYTHONHOME"));
         ^~~~~~~~~~~~~~~~~~~~~~~
         PyMapping_SetItemString
src/cmdscript.c:421:9: error: nested extern declaration of ‘PyConfig_SetBytesString’ [-Werror=nested-externs]
src/cmdscript.c:421:51: error: request for member ‘home’ in something not a structure or union
         PyConfig_SetBytesString(&py_conf, &py_conf.home, getenv("PYTHONHOME"));
                                                   ^
[...]

PR #1844, opened 2022-12-17
Conditional build of cmdscript to allow compilation with old python versions (<3.10.x)

nitraiolo:
To resolve issue #1843 and allow compilation of all binaries on systems with old python version.

### DIFF
--- a/client/src/cmdscript.c
+++ b/client/src/cmdscript.c
@@ -403,7 +403,7 @@ static int CmdScriptRun(const char *Cmd) {
         // hook Proxmark3 API
         PyImport_AppendInittab("_pm3", PyInit__pm3);
 #endif
-#if PY_MINOR_VERSION >= 10
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 10
         PyConfig py_conf;
         PyConfig_InitIsolatedConfig(&py_conf);
         // Despite being isolated we probably want to allow users to use
@@ -419,7 +419,7 @@ static int CmdScriptRun(const char *Cmd) {
         char *argv[128];
         argv[0] = filename;
         int argc = split(arguments, &argv[1]);
-#if PY_MINOR_VERSION >= 10
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 10
         // The following line will implicitly pre-initialize Python
         PyConfig_SetBytesArgv(&py_conf, argc + 1, argv);
         // This is required by Proxspace to work with an isolated Python configuration
--- a/client/src/cmdscript.c
+++ b/client/src/cmdscript.c
@@ -403,7 +403,9 @@ static int CmdScriptRun(const char *Cmd) {
         // hook Proxmark3 API
         PyImport_AppendInittab("_pm3", PyInit__pm3);
 #endif
-#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 10
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION < 10
+	Py_Initialize();
+#else
         PyConfig py_conf;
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice one but it breaks on WSL,  so some minor tweaks?

client/src/cmdscript.c
```diff
@@ -403,18 +403,23 @@ static int CmdScriptRun(const char *Cmd) {
         // hook Proxmark3 API
         PyImport_AppendInittab("_pm3", PyInit__pm3);
 #endif
+#if PY_MINOR_VERSION >= 10
```
jmichelp, line 406: Filtering only with the minor version might not be a good idea.

client/src/cmdscript.c
```diff
@@ -424,6 +429,17 @@ static int CmdScriptRun(const char *Cmd) {
 
         // clean up
         PyConfig_Clear(&py_conf);
+#else
+        wchar_t *py_args[argc + 1];
+        py_args[0] = Py_DecodeLocale(filename, NULL);
```
jmichelp, line 434: Using `Py_DecodeLocale()` will allocate memory which needs to be freed at the end.

nitraiolo: There is the free at line 426.
The code is the one from commit #38cb327991501325aab6b24310e5428dc0369ab0

nitraiolo:
> Nice one but it breaks on WSL, so some minor tweaks?

Some problem with preprocessor conditions? Can I see the error?

client/src/cmdscript.c
nitraiolo: You have right! Added MAJOR too, thanks!

client/src/cmdscript.c
```diff
@@ -424,6 +429,17 @@ static int CmdScriptRun(const char *Cmd) {
...
+        for (int i = 0; i < argc; i++) {
+            py_args[i + 1] = Py_DecodeLocale(argv[i], NULL);
```
jmichelp, line 436: Just noticed that the index might be off-by-one.
```c
for (int i = 0; i < argc + 1; i++) {
  py_args[i] = Py_DecodeLocale(argv[i], NULL);
```
This should be enough and will already handle filename (i.e. line 434 can ve removed)

nitraiolo: I've noticed it also that this code puts file name into args[0] and shifts all parameter by one. This was the way the code was working into commit #38cb327991501325aab6b24310e5428dc0369ab0. TBH I didn't dig more into this.

nitraiolo: @jmichelp tested with parity.py and you have right! That code was wrong!

[usb] pm3 --> script run parity.py 10 1234
[+] executing python /data/home/nitro/Downloads/NFC-RFID/iceman_builds/proxmark3_08/client/pyscripts/parity.py
[+] args '10 1234'
Traceback (most recent call last):
  File "parity.py", line 75, in <module>
    main()
  File "parity.py", line 61, in main
    numtype=  int(sys.argv[1], 10)
ValueError: invalid literal for int() with base 10: 'parity.py'

Fixing it as suggested works well

[usb] pm3 --> script run parity.py 10 1234
[+] executing python /data/home/nitro/Downloads/NFC-RFID/iceman_builds/proxmark3_08/client/pyscripts/parity.py
[+] args '10 1234'
numtype: 10
num: 1234 0x4D2
Even parity

[+] finished parity.py

nitraiolo:
> Nice one but it breaks on WSL, so some minor tweaks?

The last commit should have fixed this too.

nitraiolo:
This last commit is intended to help maintenance. This way the "default" when condition is not met (#else part) is always the new code.

iceman1001:
Builds everywhere!

PR #1846, opened 2022-12-20
Avoid memory leaks with Python <3.10

jmichelp:
Explained it during #1844 but it wasn't fixed before merging


iceman1001:
Perfect!

Issue #1854, opened 2023-01-03
Incorrect simulation of Mifare 1K card


szjiajin:
**Describe the bug**
According to this [data sheet](https://www.nxp.com/docs/en/data-sheet/MF1S50YYX_V1.pdf), in chapter 7.8.3 at page 13, at the end of table 8 the foot note says:

> If key B may be read in the corresponding Sector Trailer it cannot serve for authentication (see grey marked lines in Table
7). As a consequences, if the reader authenticates any block of a sector which uses such access conditions for the Sector
Trailer and using key B, the card will refuse any subsequent memory access after authentication.

It means that if the access bit (C1|C2|C3) for sector trailer is 0b000 or 0b001 or 0b010, the response should be an NACK.

But when I simulate a mifare card, I set the access bit of sector trailer to be 001 and perform authenticate with key B followed by a read operation, it seems to return 16 byte value with some of the bytes masked of if not accessible.
I checked the source code in [proxmark3/armsrc/mifaresim.c](https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/mifaresim.c) from line 887 to line 983. It does not seem to implement this feature.

**To Reproduce**
1. Obtain a normal mifare 1K card, change the access bit of one of the sector to 0xFF078069 (the transport configuration)
2. Authenticate with key B for one block in that sector and try to read the block, after the authentication is completed, the card will return an error code (0x01 or 0x05, I am not sure) for the ReadBlock command.
3. Reproduce it in the simulated card in pm3, it returns the value of the data block (or the masked value of the sector trailor as keyA and access bit is not accesssible), but not an error code.

**Expected behavior**
As mentioned above, it should return an error code instead of 16 byte value.

**Screenshots**
I think my explanation should be enough, but I will upload my screenshot of sniffing if necessary.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - inside proxmark3 client run the following commands and paste the output here.
 I am not with my workbench now, will submit the version later, but this should still be the issue in the latest firmware.
 - hw version
 - hw status
 - data tune

**Additional context**
I also checked repository of official proxmark but their repository stopped updating 2 years ago, so I decided to report the issue here.

iceman1001:
I am sure there are many more cases of the Proxmark3 simulation isn't 100% according to the datasheet.

Feel free to contribute!

iceman1001:
When fiddling with the proxmark3 simulation code its important to consider timings. 
[...]

PR #1856, opened 2023-01-06
Mifare compliance fix

szjiajin:
Fix the incorrect simulation of Mifare classic in [Issue #1854](https://github.com/RfidResearchGroup/proxmark3/issues/1854)  
Compiled with no errors and tested on my device. Don't think it is worth noting in the CHANGELOG though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Add `--shallow` option to `hf iclass` reader commands to do shallow (ASK) reader modulation instead of OOK (@nvx)
  - Improved NXP SLI/SLIX series tag identification (@nvx)
  - Fixed buffer overflow in "lf em 4x05 sniff" (@HeinrichsH)
  - Fixed potential NULL array printing (@jmichel)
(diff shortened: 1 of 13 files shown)


iceman1001:
Its an good PR,  
One concern I have is that when it comes to Mifare simulation on the Pm3 we are on severe time limitations.
While adding complete support of mifare classic protocols with all its quirks is nice  I can't help to think in which way it will be helpful.

Which readers have you run into that make use of this edge case?  
Have you measured against multiple different readers to make sure the timings still upholds?

szjiajin:
> Its an good PR, One concern I have is that when it comes to Mifare simulation on the Pm3 we are on severe time limitations. While adding complete support of mifare classic protocols with all its quirks is nice I can't help to think in which way it will be helpful.
> 
> Which readers have you run into that make use of this edge case? Have you measured against multiple different readers to make sure the timings still upholds?

Oh, sorry, I did not realize that there was time limitation.
The reader was a lock which its logic is implemeted by the manufacturer. It only use key B to authenticate and did not care about if anything is read out. (By the way, their card is Mifare classsic 1K from the ATQA and SAK but it has 18 sectors, have you seen this type of card?)  
Then I use pm3 as copied card to test it. While the official card correctly reply with an error code, pm3 return with zero data. It does not pass the check immediately, but after a few consecutive rounds of communication it finally accepted pm3.
After I made the change the reader accepted my pm3 in one authentication. I suspect that there is some sort of buffer overflow problem in that reader but I am not certain.  

Also that is the only reader I have come across. So I am not sure if this will have negative impact on other readers.

PR #1857, opened 2023-01-06
Initial rework on TLV to avoid mem leaks

jmichelp:
First stage here is to introduce a new API that doesn't leak memory for TLV parsing which will be used for PIV.
Once PIV is merged, I'll make a PR to move EMV code to the new API and deprecate `tlvdb_parse()` and `tlvdb_parse_multi()`


iceman1001:
Awesome!

PR #1858, opened 2023-01-06
First two PIV commands

jmichelp:
Unfortunately, `piv getdata` comes with the TLV parsing/printing of data which makes this PR quite big.
The other commands (`scan`, `authsign`, etc.) will be smaller and the `piv getdata` command really is the foundation.


iceman1001:
Awesome!

I can see in future improvements to move some descriptions to headers etc..

jmichelp:
Clearly. I want to put more thought about the TLV code. It's parsing untrusted input and is share between pretty much everything related to GlobalPlatform or smartcard so it's definitely worth treating it as a library so that it can be exercized/hardened through fuzzing.

PR #1863, opened 2023-01-09
fix off-by-1 when removing header in hf_mfu_amiibo_restore

gtalusan:


iceman1001:
Nice catch!

mzluckylady6969:
nice

iceman1001:
just remember that LUA is 1-based index,  vs  C which is 0-based index for arrays...

PR #1871, opened 2023-01-15
Fix mifare classic dump when block cannot be read

RouNNdeL:
Similarly to https://github.com/flipperdevices/flipperzero-firmware/pull/2296 there is an issue when dumping cards that have one block unreadable within the sector. While in case of the flipper, it was just authenticating once, here the for loop over the blocks within a sector will break, meaning the other blocks do not get read. Here's the log from the dump:

```
[=] Reading sector access bits...
[=] .................
[+] Finished reading sector access bits
[=] Dumping all blocks from card...
[...]
[+] successfully read block  0 of sector  3.
[+] successfully read block  1 of sector  3.
[+] successfully read block  2 of sector  3.
[+] successfully read block  3 of sector  3.
[+] successfully read block  0 of sector  4.
[+] successfully read block  1 of sector  4.
[!] ⚠️  access rights do not allow reading of sector  4 block   2
[!] ⚠️  command execute timeout when trying to read block  2 of sector  4.
[+] successfully read block  0 of sector  5.
[+] successfully read block  1 of sector  5.
[+] successfully read block  2 of sector  5.
[+] successfully read block  3 of sector  5.
```

No read for block 3 of sector 4 is even attempted, here's the resulting dump:

```
[=]    4 |  16 | 20 FE FF FF DF 01 00 00 20 FE FF FF 00 FF 00 FF |  ....... ....... 
[=]      |  17 | 21 FE FF FF DE 01 00 00 21 FE FF FF 00 FF 00 FF | !.......!....... 
[=]      |  18 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................ 
[=]      |  19 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................
```

### DIFF
--- a/client/src/cmdhfmf.c
+++ b/client/src/cmdhfmf.c
@@ -1044,11 +1044,9 @@ static int CmdHF14AMfDump(const char *Cmd) {
                     PrintAndLogEx(SUCCESS, "successfully read block %2d of sector %2d.", blockNo, sectorNo);
                 } else {
                     PrintAndLogEx(FAILED, "could not read block %2d of sector %2d", blockNo, sectorNo);
-                    break;
                 }
             } else {
                 PrintAndLogEx(WARNING, "command execute timeout when trying to read block %2d of sector %2d.", blockNo, sectorNo);
-                break;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


iceman1001:
I can imagine that the original flow is there to not try to read the rest of the blocks because it uses the same key.
However access rights might change that.  The PM3 should re-try three times reading a block in the code behind on device side.

It should be better if it check the access rights and try accordingly.

RouNNdeL:
Right now it will attempt a read with the A key even if the block is not readable at all, do you suggest checking the access bits outside of the retry loop and skipping that block entirely?

iceman1001:
well,  its better to have your changes,    maybe later we can find a better process

PR #1875, opened 2023-01-18
Add new Standalone mode LF_PROX2BRUT

lnv42:
Trivial card number brute forcer for when you know the facility code and want to find valid(s) card number(s).
It will try all card numbers from CARDNUM_START to CARDNUM_END one by one (max. ~65k tries).
This brute force will be a lot faster than Proxbrute that will try all possibles values for LF low, even those with bad checksum (~4g tries).
LEDs are used to know approximately which card number(s) worked.


iceman1001:
Nice,
should be some more places that would need your new standalone mode marker if I remember it correct

PR #1876, opened 2023-01-18
	modified:   client/dictionaries/mfc_default_keys.dic

colonelborkmundus:


iceman1001:
Interesting keys,  

are they static across cards from same hotel?   or are they diversified ?   
We are only interested in static default keys

iceman1001:
Cool,
there are two other files in your PR which doesnt belong there.  If you clean that out and its good to go

iceman1001:
All good,  thanks!

PR #1877, opened 2023-01-20
ADD OPEN SSL FOR CmdHF14AMfUCAuth

Alfa16bravo:
hello, can you add openssl and polarssl to use the CmdTestDES script in the CmdHF14AMfUCAuth file, to generate cryptographic keys, the code is already written, the imports are missing, if you explain to me or call the imports to compile I wants to try. thank you


iceman1001:
That code should be adapted to use MBEDTLS implementation instead. 

Feel free to contribute!

iceman1001:
And what is the purpose of this PR?   You write about openssl,   and have a PR for python compilation.

PR #1879, opened 2023-01-22
Fix the types of parameters in BitstreamOut_t

wh201906:
`BitstreamOut_t` in `client/src/loclass/cipherutils.h` is also used in `client/src/cmddata.c` for  down-sampling. 
However, the max number of `numbits` and `position` is 255(limited by uint8_t), which is too small. So I changed the type from uint8_t to uint32_t, keeping the struct the same with the one in `armsrc/lfsampling.h`

PR #1880, opened 2023-01-22
Fix the lf sampling when bits_per_sample is less than 8

wh201906:
This PR will make bps<8 work, which saves ROM for longer sample time.
This bug exists for a long time. 
***
This is the data graph before this fix
(bps=5, all data are messed up)
image

(bps=8, the expected waveform shape)
image
***
This is the data graph after this fix and PR #1879
(bps=5, the expected waveform shape)
image

The sample count is also bigger 
image


wh201906:
I'm not sure if the workaround in https://github.com/RfidResearchGroup/proxmark3/pull/1880/commits/f4879fdc172f43bf965a29f2420f7c42030e895c is good, or I should rewrite these functions.

iceman1001:
Good question,   The removal of the offset and compute signal properties should be applied to the signal regardless what it takes.

I can merge this PR,  and you can make the rewrite of them?

wh201906:
I will try to do this, but I don't know not how to do this now, so it might take some time. I'd appreciate it if someone can help me or just make a new PR for it.
By the way, I wonder the meaning of the `ON_DEVICE` macro there.
https://github.com/RfidResearchGroup/proxmark3/blob/master/common/lfdemod.c#L160

iceman1001:
that code is shared and is used in both ARMSRC and CLIENT.    
the if-case means that it is NOT on device,  ie client side.

wh201906:
I got it. Thanks!

PR #1881, opened 2023-01-23
Correct some parsing of the "get nxp system information" data from icode tags

mjacksn:
I believe that there were some minor issues with how the "nxp get info" command response was being interpreted. I tried to bring the code in line with the specification in the SLIX2 datasheet (page 33)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1363,6 +1363,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Mifare simulation, `hf mf sim` (was broken a long time) (@pwpiwi)
  - Major improvements in LF area and data operations. (@marshmellow42, @iceman1001)
  - Issues regarding LF simulation (@pwpiwi)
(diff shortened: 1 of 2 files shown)


iceman1001:
hm...  EAS support for SLIX ?   Dont remember,  current bit masks came from somewhere

mjacksn:
Sorry, I’m a brand new contributor, and I apologize if I’m misunderstanding something, but I do think the way the code currently interprets that info is a little off from the NXP documentation. I assume it probably went unnoticed because, just by chance, it results in an output that appears correct. I saw the bitmask 0x03 being used to check a single bit flag and that prompted me to look closer at the rest of that section of code.

Here are a couple of other examples:

- It looks like the current bitmasks don’t take the RFU bits 3 & 4 into account when checking the password-protect conditions flags 

conditions bits

- The current code doesn’t appear to take the existence of the RFU bits 15-31 into account when checking the EXT flag. 

Table 74  Feature flags bits

(those snips are from the SLIX2 datasheet just FYI)

PR #1882, opened 2023-01-23
Detect NTAG 5 devices and perform EAS status checks on other icode tags which support it

mjacksn:
Adds support to specifically detect NTAG 5 devices and updates hf 15 info command to performs EAS status checks on other icode tags which support EAS (SLI, SLIX, SLIX-L, and SLIX-S)


iceman1001:
Looks good!

Waiting for the actions to run

PR #1883, opened 2023-01-24
add extra keys from flipper unleashed dictionary

colonelborkmundus:
Added extra keys from the [Flipper Unleashed](https://github.com/DarkFlippers/unleashed-firmware) dictionary file while updating the Flipper dictionary (they have keys the pm3 doesn't have, and vice versa). Checked for dupes and commented out the lone entry I could find.


iceman1001:
Nice, 
are these keys static or are they diversified across cards from the same system?

colonelborkmundus:
They were part of the mfc_default_keys.dic Flipper file, which is mainly forked from this repo but hasn't been updated since last October, and there isn't much more information in the file comments. I see different keys from transit cards, and I don't know what the heck "MISC KEYS FROM MY OLD ACCESS CARDS" means. 

Is there any chance we can request that people be more specific when adding more keys? Having the extra information is, you know, nice.

iceman1001:
if the keys isn't verified to be static default across tags from the same system,   it is of little use in this dictionary.

You might ask in the flipper PRs about the background ...

colonelborkmundus:
Cool, I'll reopen this when I found out. Thank you!

PR #1884, opened 2023-01-24
add gen4 GTU magic card command 'Set Ultralight and M1 maximum sectors'

McEloff:
Gen4 GTU magic command to set maximum block/page for read/write standard (not magic) operations with card.
After executing command 'Dump configuration' (CF <passwd> C6) this parameter is backs to factory and need to be corrected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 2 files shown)


iceman1001:
Thanks!
I wonder if we would need to add that byte to the current gtu commands...

McEloff:
> Thanks! I wonder if we would need to add that byte to the current gtu commands...

I think we need to dump configuration before. If response config string contains 32 bytes, we need add new byte to new request, otherwise use 30-bytes config set request. If we sends old 30-bytes config string to 32-bytes current config, last bytes card fills from CRC-bytes at the end of request

iceman1001:
That makes sense.

PR #1885, opened 2023-01-24
Add a proper way to check if a FW is RDV4

Alejandro12120:
First of all I am neither an expert in C nor in iceman firmware so suggestions are welcome 👍 

Since checking if firmware is compiled with flash or smartcard, is a bad practice since you can flash them separately with a PM3GENERIC, I created this method to check if a firmware is RDV4, I don't have one so I couldn't test it, but it should work.

By the way, ignore commit da84eb62064db6387d7424f4974d2d8e46934ed5 since it was a quick fix.

img


Alejandro12120:
I tried flashing RDV4 firmware on my easy and it is working as expected.

img

iceman1001:
Nice,
the change to Makefile.platform.sample,  needs to be reverted before I can merge.

Alejandro12120:
You are right, fixed. @iceman1001

Issue #1861, opened 2023-01-08
hf_mfu_amiibo_restore succeed but invalid for the switch


saltbo:
**To Reproduce**
`script run hf_mfu_amiibo_restore -f myfile -k password`

**Expected behavior**
Restore a valid amiibo card

**Screenshots**
<img width="667" alt="image" src="https://user-images.githubusercontent.com/17308208/211202781-fe8f8a43-7b5b-425e-8065-30a537578336.png">

<img width="596" alt="image" src="https://user-images.githubusercontent.com/17308208/211202796-ce40713f-894a-48d6-ac04-dab61f3a9693.png">

<img width="821" alt="image" src="https://user-images.githubusercontent.com/17308208/211203131-a26272cc-2378-42a9-961f-5fa3d4465529.png">

dump eml content
```bash
00040402
01001103
00000086
4829121E
80649494
4B56C266
2CF4DD2E
7FE50E10
96FE7E08
1E048CAA
BCAEFC4A
00000000
00000000
00000000
0491E5F8
01EA4503
AD480FE0
F110FFEE
A5000600
3CE83A1F
907079D2
ED936922
235238B3
31D48C6C
26FA452F
80323FD8
E5C890C7
FA6E13BF
60A40B10
511C77AF
892B685D
0EC58BB7
D9F4F00C
86306CBC
EA099510
01030000
024F0902
0D123604
F59BCE70
BCAF328B
B4592BF0
E339B27D
F0AC8417
EE91F6D9
9EA425D1
725453D3
937BD2CE
7E8748E5
9DFEC870
12553F65
ECDD9360
C15AC77E
F18AA040
2C4835E5
EAE9F871
9D656E08
993AB0B6
72DC0F4F
C8BA1241
415A1193
AC7EB658
B424FA75
04DBDFFE
8B208ED2
2E2D130E
3963493A
40153AD4
8F31DE32
C686450A
9DC92512
73C6939C
3B911793
69A251D7
E4589A7F
6ACA863E
EFB403EF
1BE14E32
326319BD
7DD44C4B
D9981862
7DA8D43F
7765C8E5
3C97807A
37E94176
A527BC0C
D456EE31
AF4B3F81
081243E9
67DD4504
B3EA201B
E5AB5B56
1268CFB5
B0ECDBBE
69A26A96
4E331264
A6E2FF77
27F92B22
99277616
24B26C2F
BA0B4C88
F3900557
75C9808A
D01A348D
603303EC
[...]

PR #1886, opened 2023-01-25
use Amiibo PACK if simulator PWD matches generated Amiibo PWD

gtalusan:
I had pulled down and compiled latest after working through https://github.com/RfidResearchGroup/proxmark3/issues/1861 a week or so ago.

Last night, I dusted off my old 2DS XL and noticed that the Amiibo simulator wasn't working in the latest build.  I thought it must be my 2DS XL, but tested it on my kids' Switch and it wasn't working either.

It looks like in a63257799a0092460f86cce8bb3ded3e13cd2de2, the PACK was hardcoded to 00000000.

This patch will pull the PACK from the emulator memory and Amiibo simulation can work well again.

Do let me know if this not the preferred approach.

Thanks!

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -1229,6 +1229,17 @@ bool SimulateIso14443aInit(uint8_t tagType, uint16_t flags, uint8_t *data, tag_r
 
     AddCrc14A(rPPS, sizeof(rPPS) - 2);
 
+    if (tagType == 7) {
+        uint8_t pwd[4];
+        uint8_t gen_pwd[4];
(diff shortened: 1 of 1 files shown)


iceman1001:
Not sure if this kind of check is a good one.
Maybe verify that the pwd is amiibo and if so,  then set the pack?

gtalusan:
Okay, thanks for the feedback.

Here's a change -- the PWD in emulator memory is fetched.  If it matches the calculated Amiibo PWD, then the PACK is hardcoded to 8080 otherwise nothing happens.

iceman1001:
Thats more like it!

Thanks!

PR #1887, opened 2023-01-25
unconditionally recalculate pwd/pack for amiibo simulation

gtalusan:
https://github.com/RfidResearchGroup/proxmark3/pull/1886 will use whatever data has been uploaded into the emulator.  Depending on the source of the Amiibo binary data, the values could be 0.  If the user is running the amiibo simulator script, let's gently force the PWD/PACK to sane values.


iceman1001:
Thanks!

PR #1888, opened 2023-01-26
Fix uninitialised stack-based buffers

jmichelp:
Fixed more than necessary (e.g. the buffers immediately followed by memset) but that makes sure the pattern is consistent across the file and could prevent bugs from reappearing should the memset line go away :)


iceman1001:
Great catch!

PR #1889, opened 2023-01-27
Add 'lf em 4x70 brute' command

adite:
With help of this command, you can brute force the crypto key bits k95...k48 of an EM4x70.
To determine the remaining bits k47...k0 we have to know the Megamos crypto algo which is not the case at the moment,

According to paper "Dismantling Megamos Crypto", Roel Verdult, Flavio D. Garcia and Barıs¸ Ege.
§ 6.1 Partial Key-Update Attack (Optimizing the attack) steps 1...5.

Steps needed to brute force block 7, 8 and 9:
1. make sure, the EM4x70 is unlocked
2. we need an eavesdropped 56-bit Random and the 28-bit F(RN) 
3. write 0 to the desired crypto key block e.g. 9: 'lf em 4x7 write -b 9 -d 0000'
4. 'lf em 4x70 brute -b 9 --rnd 45F54ADA252AAC --frn 4866BB70' will increment the random until the authentication is valid and returns the crpyto key of block 9.
5. restore the found crypto key block: 'lf em 4x7 write -b 9 -d c0de'
6. process similary with block 8 and 7


iceman1001:
Interesting command.    The megamos recovery vectors is interesting.   Keep up the good work!

adite:
you're welcome!

henrygab:
@adite -- Just tried this, and was unable to get it to work.   Any ideas?  Here are the commands:

```sh
# Unlock the transponder ... this is successful
lf em 4x70 unlock --pin AAAAAAAA
# Write the example private key to the transponder ... this is successful
lf em 4x70 writekey -k F32AA98CF5BE4ADFA6D3480B
# Authenticate using the sample Nonce (`--rnd`) and Ac (`--frn`) values ... this is successful (9BD18)
lf em 4x70 auth --rnd 45F54ADA252AAC --frn 4866BB70
# Overwrite block 9 with all-zero data ... this appears to be successful
lf em 4x70 write --block 9 -d 0000
# Verify the private key changed ... this now properly fails
lf em 4x70 auth --rnd 45F54ADA252AAC --frn 4866BB70
# As expected: [-] ⛔ TAG Authentication Failed
# Use the new `brute` command to obtain partial key for block 9 ... fails after many, many attempts
lf em 4x70 brute --rnd 45F54ADA252AAC --frn 4866BB70 --block 9
```

Even if you could just indicate at what block you would expect to see a result output?  (e.g., occurs in block `F300`)
Having a closer `--start` parameter would allow quick validation, and simplify detecting if other factors are a cause.

Thanks!

adite:
Hello Henry,

Shit happens - i introduced a stupid bug with the very last modification i
made...
I am sorry for that and i will push the fix asap!

Best regards,
Adrian

Adrian Teuscher ***@***.***> schrieb am Mi., 1. Feb. 2023, 17:11:

PR #1890, opened 2023-01-27
Detect 50 pF version of NTAG 210u (micro)

mjacksn:
Small addition so that the 50 pF version of the NTAG 210u (micro) gets identified. Currently, it only looks for the 17 pF version.


iceman1001:
Thanks!

PR #1895, opened 2023-02-01
Fix broken 'lf em 4x70 brute' command

adite:
Unfortunately, a stupid mistake has crept in - the sizeof() a pointer instead an array was taken.
Sorry for that...
This pr fixes it and also enhance the timeout of the brute command to 1h.
This is because it can take about 45 minutes at worst case.

Regards,
Adrian


iceman1001:
Nasty little bug...

adite:
Not ideal for my first contribution to this project...

henrygab:
It's a common mistake.  The ARRAYSIZE() macros are **_inherently_** unsafe in C.   As bugs go, that's one that hits even the most experienced coders.  (I have a longer diatribe against that macro... but I'll spare you). 

Good news -- I confirmed with this fix that this cleanly found the partial for block 9.   Well done!

henrygab:
And now confirmed for block 8.  Seems like a solid implementation of that algorithm!

PR #1896, opened 2023-02-02
Added XOR key extraction and flag to Guardall G-Prox II

GuruSteve:
Guardall G-Prox fobs appear to store an arbitrary XOR key in the modulated data which is used to encode the FC and CN.  As a consequence of this, different data can decode to the same FC and CN.  Different installations appear to use different XOR keys as well, so without being able to specify the XOR key, it is highly unlikely that the resulting tag created from the clone command will be valid.  Previous code simply hardcoded the XOR key as 0x66;

### DIFF
--- a/client/src/cmdlfguard.c
+++ b/client/src/cmdlfguard.c
@@ -86,9 +86,9 @@ static int demod_guard_raw(uint8_t *raw, uint8_t rlen) {
     }
 
     if (unknown)
-        PrintAndLogEx(SUCCESS, "G-Prox-II - Unknown len: " _GREEN_("%u") ", Raw: %s", fmtlen, sprint_hex_inrow(raw, rlen));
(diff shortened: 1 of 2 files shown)


iceman1001:
Interesting,
maybe 0x66 should be the default value still ?

GuruSteve:
It seems that the value is as arbitrary as a FC value, so it should be mandatorily user-specified.  Doing so ought to also make it a slightly more user friendly experience

I was inspired to chase down the issue from the discord discussion here: https://discord.com/channels/738045391416852520/738046218558636052/1070439296638664744

There are still some unknowns I'm trying to figure out, but will need a larger sample size - would be helpful to get some different fob reads from the same building.  The code is far from perfect, but I'm confident being able to specify the XOR value is a big step in the right direction.

iceman1001:
Good!

After some community testing over at discord,  it seems to be better than before.

PR #1898, opened 2023-02-04
DRAFT - Skeleton template leaves PM3 in bad state?

henrygab:
This draft PR is not intended to be integrated.

I had hoped to have a nice skeleton of the minimum needed to add a new sub-command.

However, even this minimal skeleton, after calling the new command, leaves the proxmark3 device in a state where it refuses to accept additional commands.  Pressing the side button clears this error state.

I welcome any thoughts on this.

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -1232,8 +1232,8 @@ static void PacketReceived(PacketCommandNG *packet) {
             break;
         }
         case CMD_LF_EM4X70_NEW_COMMAND_XYZZY: {
-            Dbprintf("received CMD_LF_EM4x70_NEW_COMMAND_XYZZY command with %d bytes payload", packet->length);
             em4x70_NEW_COMMAND_XYZZY((em4x70_data_t *)packet->data.asBytes, true);
+            break;
         }
 #endif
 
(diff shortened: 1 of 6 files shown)


henrygab:
Steps for repro (current as of commit 45bff8e):

```
# Verify basic functionality and read capability
lf em 4x70 info
# exercise help ... works fine and so does subsequent command
lf em 4x70 xyzzy -h
lf em 4x70 info
# Then it falls apart....
# exercise something that uses the most basic FPGA functions ... works fine
lf em 4x70 xyzzy -k F32AA98CF5BE4ADFA6D3480B --rnd 45F54ADA252AAC --frn 4866BB70 -d 8
# the help capability also works fine...
lf em 4x70 xyzzy -h
# but if the next command uses the FPGA again ... times out with no response from PM3
lf em 4x70 info
```

So, it now appears the area to focus on is use / cleanup of the FPGA.  (best guess...)

armsrc/appmain.c
```diff
@@ -1231,6 +1231,10 @@ static void PacketReceived(PacketCommandNG *packet) {
             em4x70_brute((em4x70_data_t *)packet->data.asBytes, true);
             break;
         }
+        case CMD_LF_EM4X70_NEW_COMMAND_XYZZY: {
+            Dbprintf("received CMD_LF_EM4x70_NEW_COMMAND_XYZZY command with %d bytes payload", packet->length);
+            em4x70_NEW_COMMAND_XYZZY((em4x70_data_t *)packet->data.asBytes, true);
+        }
```
henrygab, line 1237: One significant problem was here...

Simple error ... I forgot the `break;` in this `switch`/`case` .   Oops!

PR #1899, opened 2023-02-05
Add a option to clear the scrollback buffer

wh201906:
When running `clear` in pm3 client, the scrollback buffer is not cleared. This behavior is different from a lot of the shells'.
This PR adds a `-b` option to clear the scrollback buffer.


iceman1001:
Thanks!

Never thought about the scrollback...

PR #1902, opened 2023-02-07
Rename 77-pm3-usb-device-blacklist in makefile

Foosec:
To align with the changes made in #1894

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ ifneq (,$(INSTALLSIMFW))
 endif
 ifeq ($(platform),Linux)
 	$(Q)$(INSTALLSUDO) $(MKDIR) $(DESTDIR)$(UDEV_PREFIX)
-	$(Q)$(INSTALLSUDO) $(CP) driver/77-pm3-usb-device-blacklist.rules $(DESTDIR)$(UDEV_PREFIX)/77-pm3-usb-device-blacklist.rules
+	$(Q)$(INSTALLSUDO) $(CP) driver/77-pm3-usb-device-blacklist-dialout.rules $(DESTDIR)$(UDEV_PREFIX)/77-pm3-usb-device-blacklist.rules
 endif
 
 uninstall: common/uninstall


Foosec:
Converted to draft, bad fix since it assumes dialout over uucp. Could add Linux check to udev target, and add udev to install. Waiting to hear preferred fix.

henrygab:
This appears to be a regression from my change.  I missed this reference to the udev rules, sorry!

The logic used elsewhere, and which I [duplicated](https://github.com/henrygab/proxmark3/blob/01d30e14fc80505f6bacf6efd71c8c3201aef939/Makefile#L258-L267) in my change:

```
ifneq ($(wildcard /etc/arch-release),)
	# If user is running ArchLinux, use group 'uucp'
	$(SUDO) cp -rf driver/77-pm3-usb-device-blacklist-uucp.rules    $(DESTDIR)$(UDEV_PREFIX)/77-pm3-usb-device-blacklist.rules
else
	# Else, use group 'dialout'
	$(SUDO) cp -rf driver/77-pm3-usb-device-blacklist-dialout.rules $(DESTDIR)$(UDEV_PREFIX)/77-pm3-usb-device-blacklist.rules
endif
```

Perhaps you can use the same conditions in that other makefile location?

henrygab:
@Foosec -- The regression was caused by my failing to test the `install` path.

I think PR #1903 may be what you were trying for.   Can you try it out, and let me know if it resolves your issue?

Foosec:
> 
It does, i was looking for a way to leave that logic in the udev target, perhaps make the udev target a dependency of install, and rather check in udev if the system is linux or not?

henrygab:
> _[[PR #1903]]_ does _[[resolve the issue]]_, i was looking for a way to leave that logic in the udev target, perhaps make the udev target a dependency of install, and rather check in udev if the system is linux or not?

Thanks for confirming this resolves the issue!  I'll leave #1903 open, as it resolves the immediate issue.

That said, I think making the `udev` target a dependency of `install` makes sense, and I'll look forward to seeing your PR with that change!

Foosec:
Fixed in #1903 Thank you @henrygab for your quick reaction!

PR #1903, opened 2023-02-08
Fix Makefile regression from #1894

henrygab:
Regression caused by #1894.
See also PR #1902, where this PR was confirmed to fix the issue.

<details><summary>Root Cause (click to expand)</summary><P/>

PR #1894 updated the udev rules to automatically assign a group owner, and to automatically give the group both read and write privileges.   However, ArchLinux uses a different group name for access rights (according to pre-existing Makefile rules) ... they need the group to be `uucp` instead of `dialout`.   Therefore, the udev rules were split into two ... one file for when using `uucp` and one file for when using `dialout`.   This worked, and the devices were getting assigned the right group names, and automatically were usable without using `sudo`.

Unfortunately, I failed to test the `make install` option, and also failed to notice that the file was referenced in that section.   Therefore, this caused a `make install` to fail, since the udev rules file it tries to copy no longer exists (it must select either the `uucp` or `dialout` file for the copy).

Keeping the final filename identical was an intentional choice, as it reduced visible changes after installation of the udev rules.

I reviewed the `uninstall` section also ... it does **_not_** appear to require any changes. 

</details>

This fix copies the existing logic to also select the proper udev rules to be copied in the `install` section of the Makefile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Fixed `Makefile` regression that broke `make install` (@henrygab)
  - Fixed `lf em 4x70 brute` - now works as expected (@adite)
  - Fixed the lf sampling when bits_per_sample is less than 8 (@wh201906)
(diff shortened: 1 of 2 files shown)


iceman1001:
Good to catch this!

henrygab:
> Good to catch this!

@Foosec -- all kudos go to you for discovering the regression, reporting it, and providing an initial PR.  Your quick action helped minimize the impact of my regression.  Thank you again!

PR #1905, opened 2023-02-09
Add new UID of Waveshare 1.54" NFC ePaper

ah01:
My new Waveshare 1.54" NFC ePaper has different UID then what was expected in PM code. Probably new HW revision or something.


iceman1001:
Thanks!

And line in the changelog file..

PR #1907, opened 2023-02-09
feat: add standalone ICECLASS read/sim mode

natesales:
This standalone mode starts by reading iCLASS credentials to flash, then after a button press, switches to emulating a previously saved credential.


iceman1001:
You don't have any instructions for your mode,  and it continues to loop,  is that the intention?

natesales:
> You don't have any instructions for your mode, and it continues to loop, is that the intention?

I've added a line to explain what this mode does.

And yes, this is the intended use. Like `ICE_STATE_DUMP_SIM`, the final behavior is to sim the credential until power off.

iceman1001:
Thanks!

PR #1908, opened 2023-02-10
fix infinity loop in SpinDelayUs() and SpinDelayUsPrecision()

lnv42:
I was debugging a test code of my own (including a lot of SpinDelayUs()), that had regular weird infinity looping and i figured out the issue was that AT91C_BASE_PWMC_CH0->PWMC_CCNTR value is never equal to 0, so if start+ticks was equal to 0, it was inifity looping...
This fix may produce bit longer wait than expected in some case, depending on if AT91C_BASE_PWMC_CH0->PWMC_CCNTR delay between 0xFFFF and 0x0001 is just 1 step or 2...

/!\ Figure out why AT91C_BASE_PWMC_CH0->PWMC_CCNTR is never 0 and fix it there is probably a better way to fix this infinity loop issue /!\


iceman1001:
That is one bizarre finding.   Cool and thanks!

PR #1913, opened 2023-02-16
Mark credentials as decrypted in decrypted dumps and show them when using the view command

fersingb:
Hello,

As suggested on the discord server, here is a small modification to the code to mark the iclass credentials as decrypted in the decrypted dumps. I also now show the credentials in the iclass view command if the source dump file is decrypted.

Cheers


iceman1001:
Interesting, 
Since the config bytes setting operates on bit level,   setting a whole byte or even comparing on byte level is wrong. 

Adapt to the checks and compares to the specific bit and its good to go.

fersingb:
Hi @iceman1001,

Thank you for your feedback. I've modified the code to only set/check the relevant bits. Also, I've added some checks to only attempt a decryption if the encryption is 3DES, as other encryption methods are not supported yet.

iceman1001:
Much better!

Thanks for the contribution!

PR #1919, opened 2023-02-18
Prevent buffer overflow in `AppendGraph()`

henrygab:
Comment marked this as a TODO.
So, I did it.


iceman1001:
always wondered why we don't just  memcpy the clocks...  

Keep up the good work!

iceman1001:
Thanks!

PR #1920, opened 2023-02-19
Prevent double-enumeration under WSL2

henrygab:
This makes enumeration under WSL2 work correctly.

I leaves in the workaround someone added (which does not affect all systems?).
At the same time, it explicitly avoids double-enumeration when the exact same device name is found.

### DIFF
--- a/pm3
+++ b/pm3
@@ -68,11 +68,12 @@ function get_pm3_list_Linux {
             fi
         fi
         # WSL2 with usbipd detection - doesn't report same things as WSL1
-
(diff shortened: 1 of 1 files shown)


iceman1001:
you should leave the `-qi` parameter in.    Since yours vs mine setup has different case.

henrygab:
> you should leave the `-qi` parameter in. Since yours vs mine setup has different case.

OK.  This now checks using case-insensitive matching.  This script works for WSL2.  What's currently checked in does not.

iceman1001:
Awesome!

PR #1922, opened 2023-02-21
bugfix: wrong text in hf mfdes list when no data exists in trace

DidierA:
I noticed when data is missing in the trace (probably due to buggy code on my side), and using `hf mfdes list`,  the hex part was correctly stating `<empty trace - possible error>`, but the Annotations  were displaying valid commands such as `READ RECCORDS` or `CREATE STD DATA FILE`. The `cmd` buffer was read despite `cmdsize` being 0.


iceman1001:
Good catch!

PR #1925, opened 2023-02-23
Add mfc signature support to gen4 lua script

augustozanellato:


iceman1001:
that disconnect &  connect seems odd...

DidierA:
> that disconnect & connect seems odd...

The 6B command changes the card's configuration, so I would expect it to  need a reset...
For the MFC EV1 signature, only blocks 0x45 to 0x47 are written, doesn't 0x44 also need to be changed? (changing the whole sector)

augustozanellato:
> > that disconnect & connect seems odd...
> 
> The 6B command changes the card's configuration, so I would expect it to need a reset...

That makes sense. 

> For the MFC EV1 signature, only blocks 0x45 to 0x47 are written, doesn't 0x44 also need to be changed? (changing the whole sector)

0x44 isn’t read by [read_mfc_ev1_signature](https://github.com/RfidResearchGroup/proxmark3/blob/02a801f6655ac82d7a7e168bc7120577e276e668/client/src/mifare/mifarehost.c#L1438) and I haven’t found nxp documents describing it

augustozanellato:
PR rebased so CI should pass on proxspace

PR #1928, opened 2023-02-27
Adding bruteforcing functionality to HID facility code

cha0tic87:


client/src/cmdlfhid.c
```diff
@@ -475,26 +475,27 @@ static int CmdHIDClone(const char *Cmd) {
...
         arg_lit0("v", "verbose",             "verbose output"),
         arg_str1("w", "wiegand", "<format>", "see " _YELLOW_("`wiegand list`") " for available formats"),
-        arg_u64_0(NULL, "fc",     "<dec>",    "facility code"),
-        arg_u64_0(NULL, "cn",     "<dec>",    "card number to start with"),
-        arg_u64_0("i",  "issue",  "<dec>",    "issue level"),
-        arg_u64_0("o", "oem",     "<dec>",    "OEM code"),
-        arg_u64_0("d", "delay",   "<dec>",    "delay betweens attempts in ms. Default 1000ms"),
-        arg_lit0(NULL, "up",                 "direction to increment card number. (default is both directions)"),
-        arg_lit0(NULL, "down",               "direction to decrement card number. (default is both directions)"),
+	arg_str1("f", "field", "<fc, cn>",   "field to bruteforce"),
```
iceman1001, line 491: param `f` is used for filename,   let me suggest not to use a short param name, replace with NULL and adapt help text accordingly

client/src/cmdlfhid.c
```diff
@@ -630,8 +656,8 @@ static command_t CommandTable[] = {
     {"reader",  CmdHIDReader,   IfPm3Lf,         "attempt to read and extract tag data"},
     {"clone",   CmdHIDClone,    IfPm3Lf,         "clone HID tag to T55x7"},
     {"sim",     CmdHIDSim,      IfPm3Lf,         "simulate HID tag"},
-    {"brute",   CmdHIDBrute,    IfPm3Lf,         "bruteforce card number against reader"},
-    {"watch",   CmdHIDWatch,    IfPm3Lf,         "continuously watch for cards.  Reader mode"},
+    {"brute",   CmdHIDBrute,    IfPm3Lf,         "bruteforce HID tag while maintaining specific parameter values"},
+    {"watch",   CmdHIDWatch,    IfPm3Lf,         "continuously watch for cards"},
```
iceman1001, line 660: No need to change text

client/src/cmdlfhid.c
cha0tic87: Updated per suggestions. Also updated help text to accommodate.

client/src/cmdlfhid.c
cha0tic87: I've reverted the removal of "Reader Mode", but I still think the change to line 633 is necessary as part of the PR (text updated).

iceman1001:
Perfect!

PR #1931, opened 2023-03-01
HF15: Write / Read  when is UID is passed as argument

chl337:
If UID is passed as argument, e.g. "-u E011223344556677", the sanity check
`if ((scan + unaddressed + uidlen) > 1`  will always fail, because **uidlen** is set as the length of the UID in `util.c::param_gethex_to_eol`.

### DIFF
--- a/client/src/cmdhf15.c
+++ b/client/src/cmdhf15.c
@@ -48,6 +48,7 @@
 #define Logic1                  Iso15693Logic1
 #define FrameEOF                Iso15693FrameEOF
 #define CARD_MEMORY_SIZE        4096
+#define HF15_UID_LENGTH		8
 
 #ifndef Crc15
 # define Crc15(data, len)       Crc16ex(CRC_15693, (data), (len))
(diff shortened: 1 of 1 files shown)


client/src/cmdhf15.c
```diff
@@ -222,6 +223,16 @@ static const productName_t uidmapping[] = {
...
+static inline void reverse(uint8_t *buff, uint8_t length) {
+    uint8_t upper_bound = (length % 2 == 0) ? length / 2 : (length / 2) + 1;
+    uint8_t tmp = 0;
+    for (int start = 0, end = length - 1; end >= upper_bound; ++start, --end) {
+        tmp = buff[end];
+        buff[end] = buff[start];
+        buff[start] = tmp;
+    }
+}
+
```
iceman1001, line 235: I think we have a array reverse function already...   look in utils.c  or in common/

iceman1001:
Good fix!  
I think we do have a reverse array function...  If not, I suggest you move it to utils.c...

PR #1932, opened 2023-03-03
hf mfdes: consistency fixes

DidierA:
In hf mfdes, in the example description of some commands, '-s' command was used (which does not exist) instead of '--schann'.

Also the error message for an invalid file number was giving the number in decimal, while it is given in hex in the arguments.


iceman1001:
Nice catch!

PR #1933, opened 2023-03-03
Desfire : check length of tag answers

DidierA:
While experimenting with an ultimate magic card with ATQ, SAK and RATS set to mimick a DESFire tag, I noticed `hf mfdes info` returns strange values in different locations.
Actualy, the tag replies `6D 00` (instruction code not supported) to some commands, and this was interpreted as valid data.

This PR adds information on the wrong length of the answers, but keeps going on and adds FF where no data has been received, to cover the most use cases.

### DIFF
--- a/armsrc/mifaredesfire.c
+++ b/armsrc/mifaredesfire.c
@@ -182,6 +182,12 @@ void MifareDesfireGetInformation(void) {
         return;
     }
 
+    if (len < sizeof(payload.versionHW)+1) {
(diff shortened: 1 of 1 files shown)


iceman1001:
Good catch,  
We never claimed to have perfect support for compliance and non-compliance tags  :)    But this is a step in the right direction

We should have it in the identification part of  `hf mfdes info` too,  that this is most likely a gen4 tag in desfire mode

PR #1935, opened 2023-03-07
Update Troubleshooting, libQt5Core.so.5 not found

wwin3286tw:
Update the version number of Ubuntu which has the same problem

I checked newer Ubuntu version 22.04 WSL 1 still have problem and solved by same solution sudo strip --remove-section=.note.ABI-tag /usr/lib/x86_64-linux-gnu/libQt5Core.so.5

### DIFF
--- a/doc/md/Installation_Instructions/Troubleshooting.md
+++ b/doc/md/Installation_Instructions/Troubleshooting.md
@@ -224,7 +224,7 @@ Try running it with
 ## libQt5Core.so.5 not found
 ^[Top](#top)
 
-On WSL1 / updated to Ubuntu 20.04,  there is a slight chance you experience problems when compiling the repo with QT5.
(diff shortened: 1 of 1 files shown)


wwin3286tw:
### Additional materials
* Image
image

* Error message
```
/usr/lib/x86_64-linux-gnu/qt5/bin/uic: error while loading shared libraries: libQt5Core.so.5: cannot open shared object file: No such file or directory
make[1]: *** [Makefile:775: src/ui/ui_overlays.h] Error 127
make: *** [Makefile:171: client/all] Error 2
```

```
# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
```
* current repository version
```
# git rev-parse --verify HEAD
7abfff4095c3d2eae9eb85d243e3ea7f1a030a33
```

iceman1001:
Thanks!
 
The text might even be improved by saying  ` or higher`...

PR #1936, opened 2023-03-08
New keys from The Horde

colonelborkmundus:
Cleaned up first patch for my own sanity (should be easier to distinguish that keys are diversified and come from different cards even though they are from the same hotel chain) and added new static default keys acquired from The Horde of key cards. Also removed the unconfirmed F0 Unleashed keys.


colonelborkmundus:
Just saw that someone threw in a comment about Saflok, and I do want to mention that at least one friend who travels quite frequently and is also in possession of a F0 has been able to emulate their cards at least most of the time, and also they've also been passing along an annoying number of Ultralight cards lately! Basically idk man but if they give me another Ultralight card I'm chucking it at their head

iceman1001:
Seems you didnt base your PR on the latest version of the dictionary file.   Something to address.

iceman1001:
Excellent!

PR #1941, opened 2023-03-11
Fix tlv doublefree

leommxj:
`&root->db` and `root` are the same pointer, double free will crash the client when tlv parsing fails


iceman1001:
Thanks,   but I have a fix lined up for it already.

leommxj:
:)

iceman1001:
@leommxj 
if you pull latest it should be fixed

https://github.com/RfidResearchGroup/proxmark3/commit/2e8bacb00f68324396b28b79ef91b034a0e5d006

PR #1926, opened 2023-02-26
Improved "hf mf view" command

Alejandro12120:
Now you can read mifare classic memory without dumping the card. Feel free to contribute as I don't know C :P  

### Screenshots:
‌‌img1​
img2​
img3​‌​​‌​‌‌​‌​​‌​‌‌​‌​‌‌​‌‌‌​​​‌​​‌‌​​​​​‌‌​​‌‌​​‌‌​‌​‌​​​‌​‌‌‌​​‌‌‌​​​​​‌‌​‌‌‌​​‌‌​​‌‌‌

PR #1942, opened 2023-03-13
Hfmfview refactor

DidierA:
Refactor of PR #1926
I added function `mfc_read_tag()`, which is called in  dump and view. 

I tested on a MFC 1k card with MAD, however I don't have data to check the vigik part (starting line 6983)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Add Mifare Classic EV1 signature write support to gen4 magic tag lua script (@augustozanellato)
  - Added XOR key extraction and flag to Guardall G-Prox II (@GuruSteve)
  - Changed verbiage on `hf iclass info` KeyAccess area to be congruent with AA1 and AA2 areas (@GuruSteve)
- - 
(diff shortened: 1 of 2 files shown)


iceman1001:
its an improvement.  I was thinking what if we only make a boolean param for `hf mf dump --nosave`  and guard the  "pm3_save"  call ...    
It would be a much simpler change,  where it wouldn't demand reworking the dumping part and adding a lot of params to view command.

DidierA:
> its an improvement. I was thinking what if we only make a boolean param for `hf mf dump --nosave` and guard the "pm3_save" call ... It would be a much simpler change, where it wouldn't demand reworking the dumping part and adding a lot of params to view command.

Makes sense. And `hf mf view` could call `hf mf dump --nosave` when called without `-f`

iceman1001:
Yeah,    its a little neat trick.

iceman1001:
Lets see,  
I used your refactored function,  and added the `hf mf dump --ns`.   
Now lets see how this can be done in a nicer way

iceman1001:
should print,   shouldn't save...

iceman1001:
ping!

iceman1001:
This needs to be adapted to latest source.

iceman1001:
Closing,  since we have mfc_read_tag and --ns already in the client.     Will push an improvement for the extra parsers soon

PR #1943, opened 2023-03-16
Legic info command for other sources, out of bounds memory

0xdeb:
The `hf legic info` command was only available when reading directly from a card. The command displays deobfuscated and decoded legic data. 

This command is generally really useful and should also be executable on a dump file and on the emulator's memory. 
Changes:
- Extract method `static int decode_and_print_memory(uint16_t card_size, const uint8_t *input_buffer)` from `CmdLegicInfo`
- Call this function from the newly created commands:
  - `CmdLegicEInfo` (`hf legic einfo`): Display deobfuscated and decoded emulator memory
  - `CmdLegicDInfo` (`hf legic dinfo`): Display deobfuscated and decoded content from tag dump file

This PR also implements a couple of checks to prevent out of bounds memory access. Previously when the card contained invalid data (e.g. too large segment length) the client application would perform a out of bounds read and segfault.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Add Mifare Classic EV1 signature write support to gen4 magic tag lua script (@augustozanellato)
  - Added XOR key extraction and flag to Guardall G-Prox II (@GuruSteve)
  - Changed verbiage on `hf iclass info` KeyAccess area to be congruent with AA1 and AA2 areas (@GuruSteve)
- - Added `hf legic info` command for other sources: `hf legic einfo`, `hf legic dinfo`
+ - Added `hf legic info` command for other sources: `hf legic einfo`, `hf legic view` (@0xdeb)
  - 
 
 ## [Nitride.4.16191][2023-01-29]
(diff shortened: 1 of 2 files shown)


iceman1001:
Nice one!  Good catches with length checks too.

It seems to be some confusing names going on in your PR though.
where I agree with these
```
hf legic info 
hf legic einfo 
hf legic view 
```

vs the ones I don't agree with. 
```
view vs dinfo
```


How about you match the names with the former names and I will merge.

0xdeb:
I can't really think of a good name for it without breaking backwards-compatibility. 

We need to come up with a name for the `info` command operating on other sources. 
For the emulator commands, everything is prefixed with e. So `einfo` was a easy chioce.
For the offline commands operating on dumps there is no prefix on the existing command (`view`). And the command without a prefix is already taken. 
So the consistent answer would probably be to add a prefix to the `view` command, but that would change the existing cli. 

What do you think? I will gladly change the command.

iceman1001:
I like info, einfo,    the view command is to look at dumpfiles...  

so   info, einfo and view is good for me.

0xdeb:
So the view command should be replaced by the new one that prints the info? 
On the current master it just prints the dumpfile in hex. Functionality could also be merged, first print the hexdump and then the decoded data.

iceman1001:
Thanks!

PR #1947, opened 2023-03-25
Explain "Unknown Command" behaviour

KurisuWaifu:
Explained what the unknown command actually does. Set on '00' or '02' it enables direct writes to block 0 while setting it on '01' set block 0 on read only mode like a "normal card".


iceman1001:
Nice find!

PR #1958, opened 2023-04-09
`hf_mf_ultimatecard.lua` now supports 10 bytes uid + documentation fixes

Alejandro12120:
Now the param `-u` supports 10 bytes uid only for mifare classic + added `Mifare Mini, 1k, 4k 10 bytes` to tag types + documentation fixes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Added `hf legic info` command for other sources (@0xdeb)
  - Added `hf legic einfo` - views emulator menory (@0xdeb)
  - Changed `hf legic view` - now also print the decoded info of the dump file (@0xdeb)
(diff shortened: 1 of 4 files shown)


Alejandro12120:
image

iceman1001:
Interesting,  well done,
however the change of type param index is making this more complex.  
Now we need to verify that no other place in docs or help text  where this script is referenced that it uses the right type param index.

iceman1001:
line 1232  in the magic note should also be updated.

Alejandro12120:
> line 1232 in the magic note should also be updated.

Sorry, this one? https://github.com/Alejandro12120/proxmark3/blob/c9b02c269fcec98a28093f1e355f10742bd8978a/doc/magic_cards_notes.md?plain=1#L1232

Alejandro12120:
I think there is no more docs affected by this change

iceman1001:
Looks good to go

PR #1959, opened 2023-04-09
Workflow to check for uniq keys

orensbruli:
If any of the .dic files have repeated keys this workflow will fail.
It could be configured as a check for the Pull Requests


iceman1001:
Nice one!
I realize there are many duplicates now .....

iceman1001:
Awesome!

PR #1961, opened 2023-04-11
Add some magic cards notes, fix `hf mfp chk`

team-orangeBlue:
`hf mfp chk` failed on first wrong key, fixed.

Added AKA's for some iKey branded tags, as well as some info.


client/src/cmdhfmfp.c
```diff
@@ -1056,7 +1056,8 @@ static int MFPKeyCheck(uint8_t startSector, uint8_t endSector, uint8_t startKeyA
                 }
 
                 // 5 - auth error (rnd not equal)
-                if (res != 5) {
+                // PM3 client says that RND not equal is -16. Corrected. Seems to work.
+                if (res != -16) {
```
iceman1001, line 1060: Better to use the DEFINE's    Look in Auth4 function which return case you are dealing with.  
-16 ==  PM3_EWRONGANSWER

iceman1001:
Thank you,

By the look at it this is two seperate PR's   
1. fix mfp chk.
2. magic note updates

I suggest you keep track of that in the future.

PR #1962, opened 2023-04-12
Added Siemens Default key

craftbyte:
Added Siemens default key of AES `a0a1a2a3b0b1b2b3c0c1c2c3d0d1d2d3` into the DESfire dictionary


iceman1001:
Excellent!

PR #1963, opened 2023-04-12
Fix the USB enumeration process for large external flash

wh201906:
The `rdv40_spiffs_check()` is a time-consuming process for large external flash. 
When running it between `usb_enable()` and the main loop, the enumeration will excess the timeout, just as the original comment says.


wh201906:
According to the test, the `SPIFFS_check()` takes most of the time in `rdv40_spiffs_check()`
+ For 192KB SPIFFS(173441B available), it takes 220ms
+ For 512KB SPIFFS(474641B available), it takes 811ms
+ For 2MB SPIFFS(1920401B available), it takes 8647ms

iceman1001:
Thanks

PR #1964, opened 2023-04-12
Add the missing change of definition

wh201906:
for commit 79cfa1d8faa56c390a045d779b997633b357c515 
(Although this definition is unused)


iceman1001:
Good catch!

PR #1965, opened 2023-04-13
Fix wrong variable assignment

PiervitoCreanza:
Hi,
I fixed a bug in the `hf mf value` transfer attack that was recently implemented. 
Specifically, there was an issue with the `transferkeytype` being mistakenly assigned to `keytype`. This caused the command to fail when the attacked block had a key of type value `b`.  With this fix the attack should now function correctly.


iceman1001:
Good catch!

PR #1967, opened 2023-04-15
Fix if in mfp chk

AloneLiberty:


iceman1001:
Fixed in #1969

wh201906:
@AloneLiberty Oh sorry I didn't notice you have made a PR for it.

PR #1969, opened 2023-04-16
Fix logic

wh201906:
In commit 1bb78e02a4e788ee27ee72f3208b6e7fe08ec9d9, the code will always break the loop because the `PM3_EWRONGANSWER` is treated as a condition rather than a value, and the value of it is not 0.
I guess the condition there should be `res == PM3_SUCCESS || res == PM3_EWRONGANSWER`


iceman1001:
good catch!

PR #1970, opened 2023-04-19
doc(macos-macports): refactor macports docs

linuxgemini:
Since mid-2022, arm64 builds of stable releases are working. So its been a while not updating this doc :)


iceman1001:
Thanks!

PR #1971, opened 2023-04-20
Fix changelog reminder

wh201906:
The current changelog reminder workflow always comments 
`You are welcome to add an entry to the CHANGELOG.md as well`
even if the PR contains the changelog entry.
This PR fixes it.

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -9,7 +9,7 @@ jobs:
     - name: Changelog Reminder
       uses: peterjgrainger/action-changelog-reminder@v1.2.0
       with:
-        changelog_regex: '/CHANGELOG.md'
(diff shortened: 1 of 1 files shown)


wh201906:
For example, PR #1952 has the changelog entry, but the github-action still commented in it.
***
The GitHub API shows the `CHANGELOG.md` is changed
图片
According to the [code](https://github.com/peterjgrainger/action-changelog-reminder/blob/v1.2.1/src/fileMissing.ts#L9-L14), the `peterjgrainger/action-changelog-reminder@v1.2.0` uses `"/CHANGELOG.md"` for generating a RegExp object to verify it, but it doesn't match.
If I replace the expression from `"/CHANGELOG.md"` to `"CHANGELOG.md"`, it works.
图片

iceman1001:
A nice fix!

PR #1973, opened 2023-04-28
Add text explaining that the format is H10301 for binary

mistial-dev:
and specifying it for non-binary


iceman1001:
Good catch!

mistial-dev:
I went ahead and added it to the ones where the data was in binary, because from a help standpoint I think it helps to understand that those numbers really only exist in the context of a format.

PR #1977, opened 2023-05-04
hf mf sim: reduce 50ms threshold to 6ms for reset to idle

### LINKED ISSUE #1974: hf mf sim fails to respond to Inventory command more than once
**Describe the bug**
Sending Inventory command from RFID reader a second time is unable to find simulated tag.

**To Reproduce**
1. hf mf sim --1k -u 12345678 -i
2. Execute Inventory command from RFID reader. It is able to find the simulated tag.
3. Execute Inventory command from RFID reader again. It is unable to find the simulated tag.

If you explicitly turn off the reader's RF field between steps 2 and 3 for at least 55 ms, it works fine.
Normally the RF reader will turn off the RF field for 6 ms to reset the tags to their reset state which is long enough for proper tags but doesn't work with the simulated tag. The simulated tag seems to need the RF field turned off at least 55 ms in order to change to reset state.

**Expected behavior**
Simulated tag is found every time an Inventory command is executed from RFID reader.

 - OS: Linux x86_64
 - hw version
```
[usb] pm3 --> hw version

 [ Proxmark3 RFID instrument ]

 [ CLIENT ]
  Iceman/master/v4.16191 2023-01-29 20:35:17 3545f4f98
  compiled with............. GCC 12.2.1 20230201
  platform.................. Linux / x86_64
  Readline support.......... present
  QT GUI support............ present
  native BT support......... present
  Python script support..... present
  Lua SWIG support.......... present
  Python SWIG support....... present

 [ PROXMARK3 ]
  firmware.................. PM3 GENERIC

 [ ARM ]
  bootrom: Iceman/master/v4.16191 2023-01-29 20:35:17 3545f4f98
       os: Iceman/master/v4.16191 2023-01-29 20:35:17 3545f4f98
  compiled with GCC 12.2.0

 [ FPGA ] 
  LF image 2s30vq100 2022-03-23 17:21:05
  HF image 2s30vq100 2022-03-23 17:21:16
  HF FeliCa image 2s30vq100 2022-03-23 17:21:27
  HF 15 image 2s30vq100 2022-03-23 17:21:38

 [ Hardware ]
  --= uC: AT91SAM7S512 Rev A
  --= Embedded flash memory 512K bytes ( 60% used )
```
 - hw status
```
[usb] pm3 --> hw status
[#] Memory
[#]   BigBuf_size............. 42608
[#]   Available memory........ 35632
[#] Tracing
[#]   tracing ................ 0
[#]   traceLen ............... 90
[#] Current FPGA image
[#]   mode.................... HF image 2s30vq100 2022-03-23 17:21:16
[#] LF Sampling config
[#]   [q] divisor............. 95 ( 125.00 kHz )
[#]   [b] bits per sample..... 8
[#]   [d] decimation.......... 1
[#]   [a] averaging........... yes
[#]   [t] trigger threshold... 0
[#]   [s] samples to skip..... 0 
[#] 
[#] LF T55XX config
[#]            [r]               [a]   [b]   [c]   [d]   [e]   [f]   [g]
[#]            mode            |start|write|write|write| read|write|write
[#]                            | gap | gap |  0  |  1  | gap |  2  |  3
[#] ---------------------------+-----+-----+-----+-----+-----+-----+------
[#] fixed bit length (default) |  31 |  20 |  18 |  50 |  15 | N/A | N/A | 
[#]     long leading reference |  31 |  20 |  18 |  50 |  15 | N/A | N/A | 
[#]               leading zero |  31 |  20 |  18 |  40 |  15 | N/A | N/A | 
[#]    1 of 4 coding reference |  31 |  20 |  18 |  34 |  15 |  50 |  66 | 
[#] 
[#] HF 14a config
[#]   [a] Anticol override.... std    ( follow standard )
[#]   [b] BCC override........ std    ( follow standard )
[#]   [2] CL2 override........ std    ( follow standard )
[#]   [3] CL3 override........ std    ( follow standard )
[#]   [r] RATS override....... std    ( follow standard )
[#] Transfer Speed
[#]   Sending packets to client...
[#]   Time elapsed................... 500ms
[#]   Bytes transferred.............. 285184
[#]   Transfer Speed PM3 -> Client... 570368 bytes/s
[#] Various
[#]   Max stack usage......... 4808 / 8480 bytes
[#]   Debug log level......... 1 ( error )
[#]   ToSendMax............... 12
[#]   ToSend BUFFERSIZE....... 2308
[#]   Slow clock.............. 31620 Hz
[#] Installed StandAlone Mode
[#]   LF HID26 standalone - aka SamyRun (Samy Kamkar)
[#] 
```
 - data tune
```
[usb] pm3 --> data tune
[=] ---------- Reminder ------------------------
[=] `hw tune` doesn't actively tune your antennas,
[=] it's only informative.
[=] Measuring antenna characteristics, please wait...
 🕛   9
[=] ---------- LF Antenna ----------
[+] LF antenna: 20.68 V - 125.00 kHz
[+] LF antenna: 13.42 V - 134.83 kHz
[+] LF optimal: 26.31 V - 116.50 kHz
[+] Approx. Q factor (*): 6.6 by frequency bandwidth measurement
[+] Approx. Q factor (*): 7.7 by peak voltage measurement
[+] LF antenna is OK
[=] ---------- HF Antenna ----------
[+] HF antenna: 14.31 V - 13.56 MHz
[+] Approx. Q factor (*): 4.2 by peak voltage measurement
[+] HF antenna is OK

(*) Q factor must be measured without tag on the antenna

[+] Displaying LF tuning graph. Divisor 88 (blue) is 134.83 kHz, 95 (red) is 125.00 kHz.
```

net147:
Fixes reader not being able to detect the simulated card on second Inventory command due to the RF field being powered off for only 6ms before being turned on again to reset the card to idle state.

Closes #1974

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Now `script run hf_mf_ultimatecard.lua -u` supports 10bytes UID (@alejandro12120)
  - Update documentation for installation on macOS with MacPorts (@linuxgemini)
  - Added possible Paxton id to hitag2 tag info output
+ - Changed `hf mf sim` - reduce 50ms threshold to 6ms for reset to idle #1974 (@net147)
 
 ## [Nitride.4.16191][2023-01-29]
  - Changed `build_all_firmwares.sh` to fit GENERIC 256kb firmware images (@doegox)
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -2084,8 +2084,8 @@ int EmGetCmd(uint8_t *received, uint16_t *len, uint8_t *par) {
                     if (timer == 0) {
                         timer = GetTickCount();
                     } else {
-                        // 50ms no field --> card to idle state
-                        if (GetTickCountDelta(timer) > 50) {
+                        // 6ms no field --> card to idle state
+                        if (GetTickCountDelta(timer) > 6) {
                             return 2;
                         }
                     }


iceman1001:
Thanks for the PR,

Looking at that code, it was obviously put there.

How about you set the 50ms limit to 6ms instead?    Does it work then too?

net147:
I tested changing the threshold from "> 50" to ">= 1". If disable RF reset during Inventory and manually turn RF field off for 1 ms and then back on, the real card always resets to idle state properly and the simulated card sometimes resets to idle properly and sometimes doesn't. That is why I was thinking of getting rid of this threshold altogether. If I remove the threshold then it always resets to idle state properly when the RF field is off for 1 ms, like the real card. As it is taking average of 32 samples for the field voltage and sample time is 40 us the voltage for the no field check is averaged over 40 x 32 = 1280 us = 1.28 ms already and I am thinking this level of filtering is sufficient that no additional threshold is needed.

If you disagree, I can try with adjusting the threshold instead. I don't know how the RF field off time varies between different readers doing RF reset during Inventory command to find cards that are still in range in addition to cards that have just entered range. If another reader uses a RF field off time of less than 6 ms during RF reset, then whatever threshold works for my reader might not work correctly with another reader.

iceman1001:
lt could be slow readers too,    so its down to to see what the protocol says,    but for now,   reset at 6ms,   and if it works for your setup,  we know where to adapt if someone else has issues.

net147:
Changing from 50ms to 6ms is working okay.

iceman1001:
Excellent!

PR #1982, opened 2023-05-08
Bruteforce engine to crack non-random 32 bit keys

phaseloop:
This PR introduces an integrated and extendable bruteforce engine used to crack keys which were generated using non-random or broken way.

According to various service manuals - a lot of vendors (car washes, vending machines, etc) during installation ask user to provide a PIN or short password which is then used as key for all RFID cards used in particular system/area. We aim to crack those keys by iterating over small key spaces based on ASCII characters. Few dictionaries available in proxmark3 repository already utilize partially this mechanism.

Currently this bruteforce mode is implemented for em4x50 cards - with Mifare classic coming soon + additional "smart" cracking mode.

Examples:

` lf em 4x50 brute --mode range --begin 12330000 --end 12340000` -> tries pwds from 0x12330000 to 0x12340000

this mode was already implemented for em4x50 and was included as "range" mode in the bruteforce engine.

 `lf em 4x50 brute --mode charset --digits --uppercase` -> tries all combinations of keys comprised of ASCII codes for digits and uppercase letters

Performance: about few minutes for searching digits character, about 4h for searching all uppercase ASCII characters. Password cracking for em4x50 is about 27 keys/sec

### DIFF
--- a/armsrc/Makefile
+++ b/armsrc/Makefile
@@ -46,7 +46,6 @@ SRC_CRC = crc.c crc16.c crc32.c
 SRC_ICLASS = iclass.c optimized_cipherutils.c optimized_ikeys.c optimized_elite.c optimized_cipher.c
 SRC_LEGIC = legicrf.c legicrfsim.c legic_prng.c
 SRC_NFCBARCODE = thinfilm.c
-SRC_BRUTEFORCE = bruteforce.c
 
 #   SRC_BEE = bee.c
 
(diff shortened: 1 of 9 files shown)


iceman1001:
A interesting refactoring of the bruteforcer impl.   Nice!

May I suggest you remove the command.json/md  files in this PR,  so I can merge it.   
I will run `make style` anyway after a bunch more merging of PRs.

iceman1001:
ping @phaseloop

phaseloop:
done, thanks :). I'll start working on generator for mifare classic.

iceman1001:
`bzero`   call will need to be addressed,   

Use `memset`  instead

phaseloop:
done, weird that windows build fails on that but ok :)

iceman1001:
Awesome!

PR #1983, opened 2023-05-08
'Fix' writing on hitag2 in password mode

0xdanneh:
Hitag 2 write commands are frequently failing;

Given the following command
`lf hitag write --27 -k 4d494b52 -p 4 -d 11111111`

The context of page 4 may be `11 11 11 11`, `FF FF FF FF`, or a random value.

This needs more investigation, but it seems like a bandage for the issue described above.

### DIFF
--- a/armsrc/hitag2.c
+++ b/armsrc/hitag2.c
@@ -655,6 +655,7 @@ static bool hitag2_password(uint8_t *rx, const size_t rxlen, uint8_t *tx, size_t
     *txlen = 0;
 
     if (bPwd && (bAuthenticating == false) && write) {
+        SpinDelay(2);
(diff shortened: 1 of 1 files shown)


natmchugh:
Just tried this branch out and it does indeed seem to make the write much more reliable. I haven't seen any issues with a write of 0xFFFFFFFF at all.

iceman1001:
Nice one.   Will have to look into the delays even more.

0xdanneh:
Interested to see if you figure anything out if you get a chance to look. I captured some sniffs and trace data but have not had time to analyse it yet. I expect there is a cleaner way to fix this long term.

PR #1984, opened 2023-05-09
fix reading NDEF from Mifare Classic

horrordash:
small MAD processing fix. NDEF record was not read completely

### DIFF
--- a/client/src/cmdhfmf.c
+++ b/client/src/cmdhfmf.c
@@ -5922,10 +5922,10 @@ int CmdHFMFNDEFRead(const char *Cmd) {
     }
 
     PrintAndLogEx(INFO, "reading data from tag");
-    for (int i = 1; i <= madlen; i++) {
+    for (int i = 0; i < madlen; i++) {
         if (ndef_aid == mad[i]) {
             uint8_t vsector[MFBLOCK_SIZE * 4] = {0};
-            if (mfReadSector(i, keyB ? MF_KEY_B : MF_KEY_A, ndefkey, vsector)) {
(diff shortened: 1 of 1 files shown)


iceman1001:
In what way does this PR change the reading?

from
` <= madlen, startindex 1 and no offset` 
to
`< madlen  , and startindex 0,   +1`

horrordash:
The MADDecode function fills the MAD array starting from 0. If we access it starting from 1, then we lose the first sector in "if (ndef_aid == mad[i]) {". So the NDEF record is not fully read

iceman1001:
According to the datasheet the MAD lives in the sector 0. Hence there are no NDEF data in sector 0.   
The command reads sector 0 (and sector 10 if mad2 exists.)   The data is from sector 1 and forward.
the MAD[i] will skip the MAD entry for sector0.    See picture for example. 

Or have you seen MFC cards with NDEF starting from sector 0 and don't use MAD?  In that case I would love to have some more information on that system. 


image

horrordash:
yes, of course, NDEF starts from sector 1 and I have not seen it any other way.
this is a feature of the decoding function (MADDecode). it skips sector 0 and fills the array from sector 1, but from element 0.

iceman1001:
You mean the `MADDecode()` call at line 5898 ,  which is before your change ? 
I fail to understand what you are trying to fix here.   Would you mind explaining a bit more detailed?

horrordash:
the point is that the mad[i] at line 5926 is not a raw MAD sector. this array is the result of the function MADDecode() and contains only application codes (15 codes started at zero element). at some point, the program module was refactored and they simply forgot to change this line. I just did like in other function calls :)

PR #1991, opened 2023-05-19
Add python3 transformation script - nfc to eml

OscarAkaElvis:
Add python3 script to convert amiibo nfc Flipper Zero files to eml files to be used with Proxmark3


OscarAkaElvis:
Is to be run inside the pm3 client

PR #1992, opened 2023-05-21
Auth both key A and key B with default password in `hf mf restore`

wh201906:
For some blank card with ACL `FF 07 80`, you can't write data with key B.


wh201906:
The diff result of this PR is messy. I just apply the `for (int8_t kt = MF_KEY_B; kt > -1; kt--)` to both `if (use_keyfile_for_auth)` part and the opposite part, remove the duplicated code, and change the indent.

<details><summary>Details</summary>
<p>

图片

</p>
</details>

iceman1001:
Nice one,  seems to simplify things!

PR #1996, opened 2023-05-25
Add ability to simulate paradox fobs from fc and cn

jerji:
This pull request cleans up the cmdlfparadox.c class and adds the ability to simulate a fob from facility code and card number


client/src/cmdlfparadox.c
```diff
@@ -230,7 +253,7 @@ static int CmdParadoxClone(const char *Cmd) {
     CLIParserContext *ctx;
     CLIParserInit(&ctx, "lf paradox clone",
                   "clone a paradox tag to a T55x7, Q5/T5555 or EM4305/4469 tag.",
-                  "lf paradox clone --fc 96 --cn 40426                     -> encode for T55x7 tag with fc and cn\n"
+                  "lf paradox clone --fc 96 --cn 40426 [--q5|--em]         -> encode for T55x7 tag with fc and cn\n"
```
iceman1001, line 256: the sample is showing a T55x7 card,     revert this change

client/src/cmdlfparadox.c
```diff
@@ -355,12 +352,15 @@ static int CmdParadoxSim(const char *Cmd) {
     CLIParserInit(&ctx, "lf paradox sim",
                   "Enables simulation of paradox card with specified card number.\n"
                   "Simulation runs until the button is pressed or another USB command is issued.",
-                  "lf paradox sim --raw 0f55555695596a6a9999a59a"
+                  "lf paradox sim --raw 0f55555695596a6a9999a59a           -> simulate tag\n"
+                  "lf paradox clone --fc 96 --cn 40426                     -> simulate tag with fc and cn\n"
```
iceman1001, line 356: has wrong command

iceman1001:
Nice PR with a `make style` .

The are some new claims with this PR.  
I am not sure that FC is less than 999 or rather 256.
nor the removal of old crc calc.  Have you verified that the crc is maxim against multiple paradox systems?

mwalker33:
> 
I believe my CRC comments were based on the supplied data and that CRC matched the data supplied.  As I cant verify if that applies to every tag for this system e.g. older systems may still use the current CRC.

I think the best way forward when we have unknowns like this, is to add the new one as an option and leave the current one as the default.  Then let the users select which one to use.  Once published as an option we can then call on the community to see if we can find more people to test and confirm the findings.  And get more data to confirm how things work.

maybe an option like
--crc <type>

jerji:
The comments I was referring to were not the comments you put but the tag dumps that were already in the file. The three systems I have tested with use the CRC that you figured out

iceman1001:
three systems is a start but I agree with @mwalker33 that we should have old and new crc available until proven its worth.  

My experience of the years tells me to keep things around.

jerji:
Alright, I can add it back, the only thing is that I didn't see a way in the code to create a tag with the old CRC so it would just be for reading?

iceman1001:
yes

iceman1001:
ping @jerji

jerji:
Sorry, was busy, I'll get on that

iceman1001:
Thank you!

jerji:
Thank you for merging :D

PR #2003, opened 2023-06-06
[WIP] Redo magic cards notes

team-orangeBlue:
Documentation about magic cards has now been moved into a separate folder, and split into 3 files.

These new documents should cover magic tags in a more grouped way - country-specific.
Each document now follows its' own naming scheme appropriate for its' region. 
That being said, some modifications may be required.

Brief detailed changelog:
* Made new folder: `/doc/magic_cards`
  * and moved magic cards documents inside
* Original file:
  * Extra (Russian) aka's REMOVED
  * Gen4 GDM REMOVED
* Russian magic:
  * Info about LF tags (H5.5/H7), HF tags (ULtra; UL-5; MF3) ADDED
* Chinese magic:
  * Info about LF tags (ID82xx; ID5200; etc); HF tags (GDM; KUID; etc.) ADDED


team-orangeBlue:
Midway comment: I have noticed active research on Chinese LF 82xx tags. When they will be implemented into PM3, I will update the notes again.

team-orangeBlue:
PR drafted. 

Active research was noticed on QL88 (Mifare Classic EV1 cloning tag - __custom keyed__) and HUID ("FUID++" - __custom keyed__). When there will be more information, PR will be updated and undrafted.

team-orangeBlue:
Hey guys, whoever's keeping track! We got some info on Copykey chips (HUID, QL88), but it is incomplete and has mismatches. We've got to narrow that down for now.
If you're wondering what's special, here you go:

1. HUID
* Okay, it's custom keyed using mystery KDF, but it is same as QL88. But if on QL88 you can crack it, here you can't do much. :P
* Apparently this is just renamed CUID. Observing behavior from Copykey cloner, the sector 0 ACL write protects block 0 and the ACL. *Wow!..*
2. QL88
* So I got the original SAK88-IC tag for "QinLin neighbor technology". And also a CPUK-X (horizontal epoxy). Much to my surprise they showed up as QL88 magic too. So same signature sector madness, same Prng: hard. However no custom keys on the Sak88.
* Very odd, of course. However if you know QL88 you know that the signature sector on it has a header of "SBH111" in ASCII. This is not the case for the rest of the tags. Also sector 16, unlike on real Classic EV1 chips is completely functional. 
  - Also block 0 manufacturer data is always the same: `88980020 000000F8`. Also SAK88 is only used in block 0, not anticollision.

Soo.. fun day - are these QL88 tags some ripoff classic EV1 tags? Well, we're yet to find out.

team-orangeBlue:
I'm required to add a word of warning here and in the .md's later: do NOT run hardnested on your QL88s!

I had 2 chips both of which I have cracked using hardnested, but I could not write block 0. At all. Ever. Even with copykey. It said "write success" when the UID wasn't changed. 
I don't think I was scammed: I paid a normal amount per chip; the enclosure looked fine; sector 17 data seems to be decent, so on, etc. And the KDF keys were correct after all, right?
That being said, if you stil want to poke into your tags and you have a copykey, send a dump with data to the device using a PC and sniff the write twice (emulate a classic 1k tag with a proxmark but dont run `eclr`). You should get around 10.3K of trace - run mfkey32 on (for QL88:) 17th auth attempt; (for HUID:) 23rd auth attempt. *Tested on Copykey X5 m. 10-11 v. 230329-230513*

Oh by the way I read some of the PM3 help text and now know how to detect any 82xx LF chip.

team-orangeBlue:
Okay, added some stuff to the main document.

It includes:
1. Now there are warnings about 06A0 UMC card use. Or - shadow mode on must be "04", FE command doesn't work so use F0, etc.
2. Added Tianaxin TCOS card. Most of the magic is there. The full document can be found [HERE](https://github.com/Tvirus/ebook/blob/main/%E7%A1%AC%E4%BB%B6%E5%8D%8F%E8%AE%AE/rfid/%E5%A4%A9%E5%AE%89%E4%BF%A1TCOS%E7%94%A8%E6%88%B7%E6%89%8B%E5%86%8Cv1.5.pdf).
3. Removed some "aka"'s, "DESFire 4b"'s 0008-20 is now FM1208-9 (but wrong ATS, not mentioned, but still).

iceman1001:
time to merge?

team-orangeBlue:
> time to merge?

Probably is.
I'll add another config for USCUID (FUID), then do an audit of what I made.
If you consider something to be wrong, you're more than welcome to point it out.

*One problem is that I deleted the original magic notes and as such, its' edit history. May be a problem.*

iceman1001:
Yes,

the split between different countries is unwanted.  

We want one document and it should be the original named one.

team-orangeBlue:
PR will be ready by tomorrow.

1, in a different PR. By reverting the deletion I essentially threw all blame on me, which is obviously not something we need.
2, in a single file.

team-orangeBlue:
PR closed because I ~~suck with using git~~ broke the repository.
See #2169.

PR #2008, opened 2023-06-12
Remove duplicated keys

Foxushka:


iceman1001:
The spelling mistake is a good catch.
The keys fidding not so.   I see no reason for it.

PR #2012, opened 2023-06-22
gcc 11.2.1 - missing includes / correction for LEGIC CRC16 calculation

hb9xar:
- fix: missing includes when building with "gcc (GCC) 11.2.1 20220127 (Red Hat 11.2.1-9)"

- correct LEGIC 16bit CRC calculation "hf legic crc" in proxmark3 CLI
  Use init_table(CRC_LEGIC_**_16_**) to be consistent with l_crc16legic in client/src/scripting.c
  With this change, correct Legic Cash CRC16 checksums are calculated.

### DIFF
--- a/client/src/cmdhflegic.c
+++ b/client/src/cmdhflegic.c
@@ -682,7 +682,7 @@ static int CmdLegicCalcCrc(const char *Cmd) {
 
     switch (type) {
         case 16:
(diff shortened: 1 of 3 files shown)


iceman1001:
Nice catches!

Next time make two PR since the code changes isn't related to each other.

PR #2014, opened 2023-06-25
Add ability to compile on iOS

The-SamminAter:
`/tools/hitag2crack/crack5gpu_disabled` and `/tools/hitag2crack/crack5opencl` may need further minor changes to compile on iOS, but this seems to makes everything else work.

I made sure the changes weren't breaking by compiling on Linux and iOS.


iceman1001:
Awesome!

One step closer for iOS running proxmark3 client :)

iceman1001:
@The-SamminAter 
Some issue with our GH Action where it seem to detect `UIKit` instead of the `AppKit`

image

PR #2020, opened 2023-07-02
Added MFC keys to dictionary

noahclements:
Brief changelog:

- Added some more KABA hotel lock keys
- Added a Marriott key 
- Added a LINQ key


iceman1001:
Cool,
are they verified to be static keys for those systems?

noahclements:
very likely not static keys after review

iceman1001:
larger hotels can for example use assa abloy vingcard  (diversified keys)  or kaba dorma saflok (also diversified) 

small hotels uses cheaper systems and they are more likely to have static keys.

PR #2021, opened 2023-07-03
Cleanup arm-side reader-mode mifare classic block read/write code.

nvx:
There was a bunch of duplicate code doing the same thing for reading and writing blocks for mifare classic on the arm side where the only difference was a single byte for the command (normal vs magic vs magic config). This cleans up a lot of that duplication retaining all functionality while culling a few hundred lines of code while making the remaining functions more generic which will be super handy in the near future adding support for some new types of magic cards.

I've avoided changing the RPC interface between the armsrc and client, but with the new functions there is an opportunity to simplify that interface in the future for reading and writing blocks. As hinted by the addition to a define in protocols.h I discovered a GDM read block backdoor command while working on this. Curious to seek more feedback for how we want to do things going forward with the RPC interface if it's worth making more generic interfaces for the client for these sorts of things or if we'd rather keep the RPC commands specific for each combination of regular vs backdoor auth vs magic wakeup, regular vs backdoor vs config read/write (as you can imagine the possibilities for each combination start to go up rather fast - new magic tags mean more of these combinations are being seen in the wild).

While I was cleaning things up, I also took the opportunity to remove the now-unused cuid parameter to a lot of the mifare_classic_* functions.

I also fixed up a few minor typos and replaced some magic numbers with their corresponding `#define`s that I noticed while going through the armsrc mifare code as well.

I also fixed an inconsistency with the colour of decrypted MFC lines in the trace output that I noticed while testing this code (I hope this is minor enough to slip in the same PR else I can carve it out into a separate one if you want).

The new functions I added that replaced the old `Mifare(Read|Write)(Config)?Block(GDM)?` I named `mifare_cmd_(read|write)blocks` rather than following the existing CamelCase as per https://github.com/RfidResearchGroup/proxmark3/blob/master/CONTRIBUTING.md#identifiers - but I'm happy to change this back around if needed. Wasn't sure if the intention was to eventually move all functions to `lower_underscore_case` as they're touched or not.

### DIFF
--- a/armsrc/Standalone/hf_colin.c
+++ b/armsrc/Standalone/hf_colin.c
@@ -821,7 +821,7 @@ int e_MifareECardLoad(uint32_t numofsectors, uint8_t keytype) {
         }
 
         for (uint8_t blockNo = 0; isOK && blockNo < NumBlocksPerSector(s); blockNo++) {
-            if (isOK && mifare_classic_readblock(pcs, colin_cjcuid, FirstBlockOfSector(s) + blockNo, dataoutbuf)) {
+            if (isOK && mifare_classic_readblock(pcs, FirstBlockOfSector(s) + blockNo, dataoutbuf)) {
                 isOK = false;
                 break;
             };
@@ -838,7 +838,7 @@ int e_MifareECardLoad(uint32_t numofsectors, uint8_t keytype) {
         }
     }
 
-    int res = mifare_classic_halt(pcs, colin_cjcuid);
+    int res = mifare_classic_halt(pcs);
     (void)res;
 
     crypto1_deinit(pcs);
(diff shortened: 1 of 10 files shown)


iceman1001:
Looks good.  
The trace line slipped through but its such a small one I don't mind.

I guess you can convert to NG frames too ....   That would be a good thing.

nvx:
> I guess you can convert to NG frames too .... That would be a good thing.

Yeah I almost did this, but then noticed I'd need to update some Lua code too which I thought was getting a bit much for this PR.

iceman1001:
In your next PR then :)

> > I guess you can convert to NG frames too .... That would be a good thing.
> 
> Yeah I almost did this, but then noticed I'd need to update some Lua code too which I thought was getting a bit much for this PR.

PR #2024, opened 2023-07-08
Implement Value Added Services Protocol

gm3197:
A reader implementation of Apple's Value Added Services, used in NFC passes in Apple Wallet. See [my reverse engineered protocol specification](https://gist.github.com/gm3197/ad0959476346cef69b75ea0523214350).

**Testing**
- Download a [demo/test NFC pass](https://pub1.pskt.io/c/gn1v07) and install on an iPhone w/ NFC
- Download the [reader private key](https://github.com/RfidResearchGroup/proxmark3/files/11992593/privkey.zip) for these test passes
- In pm3, `nfc vas reader -p pass.com.passkit.pksamples.nfcdemo -k privkey.der -@`


iceman1001:
Awesome research of yours.   It looks great.   

Looking at it,    this new command would go under  "HF",    not  "NFC"  since its not a NDEF related command, right?
Some minor things with different styles of naming the fcts.   No biggie,
and it seem not to compile on all platforms,  which is the big one.  

Otherwise I will be glad to merge this one.

iceman1001:
I think you forgot to add the new file to the Makefile.

iceman1001:
Alright,  I think I got all the compilation fixes and moved to HF commands.  

Awesome research and findings!

PR #2025, opened 2023-07-09
Fix compiling liblua on iOS

The-SamminAter:
iOS can, in fact, use the system() stdlib call.

If I had to guess, this wasn't an issue before because it didn't know the target was IOS? Not really sure, but removing this makes sense, and I can now successfully compile the client.


iceman1001:
Awesome!

PR #2026, opened 2023-07-09
Create a list of known ECP frames

kormax:
Added a list of most ECP frames known to me. 
I've decided to omit CRC at the end so that they could be easily tested with both NFC-A and NFC-B.

Due to nature of ECPV2 frame format, there could be virtually countless variations of those frames. I am not sure if adding all of them to this list is practical and has any meaning, perhaps this file should be dedicated to frames found IRL, and a separate taxonomy json could be created, for use in parsers and etc.

Note that some of the added examples were found through brute force and even though they are working, real world samples could differ. Added them nevertheless until someone collects/confirms real ones.

To test use `hf 14a raw -akc` command with any of the samples. Easiest to test is a Car Key one.  
Use `hf 14b raw -kc -d 6a02c30209010201` to make your friends think you have a Mercedes.


kormax:
```
{
    "versions": {
        "01": {
            "tci": {
                "000000": {
                    "id": "tci-vas-or-pay",
                    "name": "VAS or payment",
                    "description": "Used when a reader needs a pass or a payment card. Sometimes called VAS over Payment"
                }
            }
        },
    
        "02": {
            "types": {
                "01": {
                    "id": "terminal-type-transit",
                    "name": "Transit",
                    "description": "Used by express-mode enabled transit terminals",
                    
                    "subtypes": {
                        "00": {
                            "id": "terminal-subtype-default",
                            "name": "Default subtype",
                            "description": "",

                            "tci": {
                                "030400": {
                                    "id": "tci-hop-fastpass",
                                    "name": "HOP Fastpass",
                                    "description": ""
                                },
                            },

                            "data": {
                                "length": 5,
                                "name": "Fallback EMV payment networks",
                                "description": "Bit mask of allowed EMV open loop payment cards. First byte is responsible for most popular payment networks"
                            }
                        }
                    }
                }
            }
        }
    }
}
```

PR #2028, opened 2023-07-10
Fix potential memory leak in VAS implementation

kormax:
If I'm not mistaken, premature call to `tlvdb_free ` before `mbedtls_ecp_point_read_binary` caused the latter to get deleted memory, which could have caused a segfault.


client/src/cmdhfvas.c
```diff
@@ -199,6 +198,7 @@ static int LoadReaderPrivateKey(uint8_t *buf, size_t bufLen, mbedtls_ecp_keypair
         return PM3_EINVARG;
```
augustozanellato, line 198: You’re probably missing a `tlvdb_free(derRoot);` here

kormax: Made a change to fix that

iceman1001:
Nice catch!

PR #2029, opened 2023-07-11
Implement uncompressed hardnested tables loader

rdmitr:
This PR implements loading of uncompressed hardnested bitflip tables. It allows us to shave 2-3 seconds off a single hardnested attack, which is quite convenient.

To maintain compatibility with the current behaviour, this implementation tries loading the bz2 files first, and if it fails, proceeds with trying to load the uncompressed .bin files.


iceman1001:
Interesting,  
Time Memory tradeoff.   We have b2zipped files in the repo because of size.
once they are downloaded,  it comes down to the environment.   Not all has that filesystem..

Might be something to consider making it optional.   

Another consideration is that we would like to move all file operations into the fileutils.c..

PR #2030, opened 2023-07-11
[WIP] Use proxmark3 as standard PCSC smartcard reader

gm3197:
- Add `smart relay` command, effectively just passes APDUs through to the OS smartcard driver (pcsc or winsc)
- Uses [Virtual Smartcard](https://frankmorgner.github.io/vsmartcard/virtualsmartcard/README.html) daemon
- Great for using programs like [GlobalPlatform Pro](https://github.com/martinpaljak/GlobalPlatformPro) without needing a whole other smartcard reader
- Snoop in on conversation between software and card with `-v`


iceman1001:
ping @gm3197   

How is this one going?

mistial-dev:
Very much looking forward to this commit.

gm3197:
It currently works for iso 14443A. Support for 14443B is in progress. I'm not as familiar with B, so some research is required, plus I don't have any 14443B cards to test with. Remaining work is:

- The current `atsToEmulatedAtr` method only works for 14A. Another method will need to be written for 14B. See section 3.1.3.2.3.1 of the PC/SC specification
- The 14A specific `SelectCard14443A_4` and `ExchangeAPDU14a` methods will need to be switched for type A/B agnostic methods

iceman1001:
So what is the way forward?    @gm3197  you working on it?   or?

gm3197:
> So what is the way forward? @gm3197 you working on it? or?

I will start an implementation soon. As I mentioned, I don't have any type B smart cards so I'll need someone to test later on. Once I have a working version, I'll reach out on the discord and see if anyone is interested in testing it.

iceman1001:
there has been some improvements with 14b recent.  it should be easier now to add support

iceman1001:
and we will have some improvements to the RDV4 SIM Module soon enough too,  in order to support ISO7816-3 Protocol T0  atleast.   T1 is a different beast.

iceman1001:
@gm3197   ping

gm3197:
@iceman1001 thanks for the ping, I had forgotten about this. Just finished the implementation for iso14b. It is to spec, but I haven't been able to test it as I don't have any 14b cards. If someone could give it a try, that'd be great.

Also `select_card_14443b_4` annoyingly logs failed attempts to select a card unlike it's 14a equivalent. Haven't had the chance to dig in and find a way to silence that when the relay command is polling for cards.

iceman1001:
yeah,  14b needs some more love.  
With simulation,  adapting the logs,  making it more like 14a.  :)

Excellent!

iceman1001:
If you think its ready merge?

gm3197:
@iceman1001 if you're comfortable merging without testing 14b, then it's ready. I don't have permission to merge it, so you'll have to approve it. Thanks.

iceman1001:
Perfect!

doegox:
Tested successfully on a 14b Mobib card :)

iceman1001:
Now we need some documentation of how to use this new function....

doegox:
> Now we need some documentation of how to use this new function....

It's really just `smart relay -h`   
=> https://github.com/RfidResearchGroup/proxmark3/blob/master/client/src/cmdsmartcard.c#L1234   
=> *Requires the virtual smartcard daemon to be installed and running, see https://frankmorgner.github.io/vsmartcard/virtualsmartcard/README.html*  
=> follow these instructions, come back and run `smart relay`...

doegox:
@gm3197 would it be possible to also expose the internal sim module to pcsc? So one can access it without pulling out the sim to another reader. And maybe add some options to restrict which interfaces to poll (14a/14b/contact).

gm3197:
> @gm3197 would it be possible to also expose the internal sim module to pcsc? So one can access it without pulling out the sim to another reader. And maybe add some options to restrict which interfaces to poll (14a/14b/contact).

Yes, this would be pretty easy. Do you think it would make more sense then to break it into `hf 14a relay`, `hf 14b relay`, and `smart relay` for each of the interfaces?

doegox:
> > @gm3197 would it be possible to also expose the internal sim module to pcsc? So one can access it without pulling out the sim to another reader. And maybe add some options to restrict which interfaces to poll (14a/14b/contact).
> 
> Yes, this would be pretty easy. Do you think it would make more sense then to break it into `hf 14a relay`, `hf 14b relay`, and `smart relay` for each of the interfaces?

Hmm it's not so easy to choose... historically "smart" was mostly/entirely for the "sim" and e.g. "hf 14a relay" is not entirely true as it only supports iso14443A-4 (and relay may also be confused with some relay attack setup).

I'd say we can keep a category "smart" that can encompass all iso7816 (T0/T1)/iso14443A-4 (TL)/iso14443B-4 (TL)  activities, so the APDU abstraction layer. Something I discussed a bit with @iceman1001 today: we should also detail precisely which interfaces are supported by the current "smart" commands as some are only supporting contact and some only supporting contactless but many could be extended in the future to support all interfaces.

About "smart relay", again my preference would be to not use "relay" which was a term used so far only for nfc relay attacks. What about "smart pcsc" ? or "smard ifd" which is more correct term but maybe less known.

gm3197:
@doegox Finally got around to implementing this. See #2283

PR #2032, opened 2023-07-12
Eliminate extraneous num_CPUs() calls

rdmitr:
On a server with a large number of CPU cores, we experience weird freezes on hardnested launch. I didn't dig too deep into the core of the issue, but it looks like we get stuck somewhere in the guts of `sysconf()` when it is executed concurently by hundreds of threads.

This PR attempts to exclude extraneous and, frankly, unnecessary calls to the `num_CPUs()` function to work around this issue.


iceman1001:
Interesting find,

server w hundreds of threads.  What are you up to running hardnested on a server farm?

rdmitr:
Haha, not really — my client uses it for different workloads, proxmark happens to be one of them 😄

PR #2033, opened 2023-07-12
Dynamically allocate buckets array for the hardnested bruteforce attack.

rdmitr:
This PR fixes the hardcoded limitation of 128 buckets (and hence 128 threads) in the hardnested attack implementation. Buckets are now allocated dynamically in power-of-two increments.

### DIFF
--- a/client/deps/hardnested/hardnested_bruteforce.c
+++ b/client/deps/hardnested/hardnested_bruteforce.c
@@ -160,6 +160,7 @@ crack_states_thread(void *x) {
         uint8_t *best_first_bytes;
     } *thread_arg;
 
+    const int num_brute_force_threads = NUM_BRUTE_FORCE_THREADS;
     thread_arg = (struct arg *)x;
     const int thread_id = thread_arg->thread_ID;
     uint32_t current_bucket = thread_id;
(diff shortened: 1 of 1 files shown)


iceman1001:
This one feels odd.   When larger than 128 threads,   reserve 256,  512,1024 buckets of memory.. 
quite memory consuming.

rdmitr:
I stuck to the old pattern of reallocating in powers of two. We may want to keep reallocations (which are somewhat expensive) to a minimum. Then again, if someone ever hits e.g. 1024 buckets, that’s only 8 kilobytes of memory — not that much on the scale of the rest of the algorithm.My guess is that since no one ever noticed this issue, the default of 128 buckets will work for 99% cases. The rest are not likely to care about the loss of a few kilobytes ;)On 12 Jul 2023, at 13:57, Iceman ***@***.***> wrote:﻿
This one feels odd.   When larger than 128 threads,   reserve 256,  512,1024 buckets of memory..
quite memory consuming.

iceman1001:
I guess its because no one had access to a setup with more than 128 threads.  

Thanks for the nice additions!

rdmitr:
Yeah, I guess so :)

You're welcome!

PR #2034, opened 2023-07-12
Do not try to detect Mifare classic EV1 signature in offline mode

rdmitr:
This PR improves the startup time in offline mode by not trying to detect the EV1 signature.


iceman1001:
Nice catch!

PR #2039, opened 2023-07-15
Bring LZ4 support for hardnested tables

doegox:
Testing perfs on an old 8-core laptop:
```
BZ2 :      2 Mb, average decompression time: 5514 ms
LZ4 -9:    9 Mb, average decompression time:  837 ms
RAW :    704 Mb, average decompression time:  368 ms (err I mean reading time...)
```

Other LZ4 options don't help: -1 is slower and bigger, --favor-decSpeed does not do anything

As this brings the requirement of an additional library, take time to test on usual platforms...


iceman1001:
Out of the blue,   why are people so interested in the time it takes for hardnested bitflip tables to uncompress and load?

Are you adding a new option of lz4 files as a compliment  or are you removing bz2 ?

doegox:
At the moment it supports and looks first for raw, then lz4 then bz2.
Why ? 
image

iceman1001:
Because, If we are not going to use bzip2 any more then we might just as well remove that dependency from the project

doegox:
b36d5fa04da2ef81763c9a8d808a640d5370fe4f : good catch! I had to change at some point uncompressed_data from uint32_t to char for LZ4F_decompress and forgot that 1

BZip2: maybe we can remove it, depends if we still want the ability to store smaller tables in very memory constrained storage, if that makes sense... (but I doubt such env could have enough RAM to run the attack anyway...)
BTW BZip2 source code was also made available in /deps for cross-compilation with CMake, maybe sth to test for LZ4...

PR #2040, opened 2023-07-16
Make GetATQA/ISO14443-A polling method more flexible

kormax:
Hello.
This change is more of a functioning proposal currently, but in case you're fine with it, it could be merged.

I've refactored the GetATQA method, allowing to pass different polling configurations to it using the newly defined `iso14a_polling_frame` and  `iso14a_polling_parameters` structs.

This allows to remove different proprietary, implementation-specific polling hacks (like ECP, MagSafe) that were once inside this method, and bring them to the outer layers (currently to `iso14443a_select_cardEx` and `iso14443a_fast_select_card` methods, in the future to the client). 
What I assume could be done in next PRs, is to bring the custom-configuration-forming logic (leave WUPA at arm side) to the client side, which could allow more flexibility when someone wants to implement new logic that requires using a custom polling frame or sequence of without cluttering the ARM side.  
An example that comes to mind, is that a VAS reading method could call select card with a VAS-specific ECP frame, which would improve the implementation as it allows to trigger a specific pass to appear on a screen even on a locked device.

I understand that polling method is very important in cases where timing is crucial and my changes might influence other parts of the system so it should be tested. I have tested mifare attacks with this change, seemed to work for me, but I think that a look from a more experienced developer is required to be fully sure.

As for improvements that can be tested already related to this change:
- ECP polling now works every time even without the S-BLOCK-DESELECT hack, thanks to an ability to add additional timeout delay that was needed for Apple Devices.
- Magsafe polling can now be combined with ECP polling using the `hf 14a reader --ecp --mag` making a super-polling loop. I've tested, seemed to work for me for MagSafe, MFC, iPhone detection.

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -142,6 +142,66 @@ Default HF 14a config is set to:
 */
 static hf14a_config hf14aconfig = { 0, 0, 0, 0, 0 } ;
 
+
+// Polling frames and configurations 
+
+/*static iso14a_polling_frame REQA_FRAME = { 
+    { 0x26 }, 1, 7, 0 
+};*/
+
+static const iso14a_polling_frame WUPA_FRAME = { 
+    { 0x52 }, 1, 7, 0,
+};
+
+static const iso14a_polling_frame MAGWUPA1_FRAME = { 
+    { 0x7A }, 1, 7, 0 
+};
+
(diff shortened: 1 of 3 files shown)


iceman1001:
Nice one,

A much needed fixup for the hacky ecp / mag additions.

PR #2041, opened 2023-07-16
Move custom polling frames to client

kormax:
This PR incorporates following changes related to the previous one:
* To save stack memory, passing polling parameters by value was replaced by pointer.
* Reduce polling parameter struct memory footprint by reducing inner array sizes to minimum required;
* Moved ECP, Magsafe polling frame generation to client;
* Remove ECP, Magsafe constants from device;
* Make VAS command emit ECP VAS ONLY frames using new feature, so now a device will display a particular pass for authentication instead of a payment card (express mode does not exist for VAS).

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -143,62 +143,62 @@ Default HF 14a config is set to:
 static hf14a_config hf14aconfig = { 0, 0, 0, 0, 0 } ;
 
 
-// Polling frames and configurations
+// Polling frames and configurations 
 
-/*static iso14a_polling_frame REQA_FRAME = {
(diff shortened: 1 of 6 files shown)


kormax:
Also experienced a problem defining constant structs on client side, as same code working on arm didn’t compile on client side. Is it something caused by build parameters?

iceman1001:
I don't think I like the moving to the client side ...

kormax:
> I don't think I like the moving to the client side ...  

The client-based polling parameter is only sent in case ecp or mag is used, or a non-null parameter pointer is given (look at USE_CUSTOM_POLLING flag).
In other cases I fallback to sending no parameters, as it was before.

PR #2049, opened 2023-07-21
Fix typo in hf_legic_clone.lua

beckerj:
variable "length" was used parameter in function, but later called as "len"


iceman1001:
Nice catch!

PR #2050, opened 2023-07-21
MIFARE Plus traces: Adds MAC, UnMACed to annotations, and code factorisation

DidierA:
This adds `MAC_MAced` `NoMAC_UnMACed`,  and so on to the relevant commands, according to the datasheet.
example:
```
|     |39  ca  9e  2a  4d  03  7a  ae  6e                                       |  ok | WRITE ENCRYPTED(68) MAC
| Tag |0b  07  90  e8  74  62  db  46  e7  72  de  63  bd                       |  ok | 
| Rdr |0a  07  31  48  00  03  be  e0  5c  44  65  7e  3a  1e  8c  d7           |  ok | READ ENCRYPTED(72-74) MAC_MACed
```

This was also the occasion to refactor code by moving more messages to the lookup table.


iceman1001:
Nice!

PR #2057, opened 2023-07-31
Add new DESFire command code and annotation

kormax:
* Added a new DESFire command code `0xEE` definition and annotation.

  This code is not mentioned in any docs. It's used by Apple-ECP-certified DESFire readers at the end of communication to notify the device that the transaction has been successful. It doesn't seem to have any other effect apart from UX, although use this command is mandatory for Apple device to consider a read "successful" (otherwise an error will appear on a screen). Android (Mifare2GO) and physical cards (EV3) don't know this command code.
 
  This seems to be the thing that some reader brochures referred to, by using ECP as an abbreviation for "Enhanced contactless **protocol**" instead of "polling". In this case DESFire protocol has been enhanced to add a command that allows a reader to explicitly tell a device that interaction has been successful, without any assumptions about operations that need to be done. I assume that other protocols (iCLASS, SEOS) may also have new command codes for similar functionality.

* Also updated definition file spacings to be equal. Didn't make any changes to MFP part, not sure why it's marked.

### DIFF
--- a/client/src/cmdhflist.c
+++ b/client/src/cmdhflist.c
@@ -1023,6 +1023,9 @@ void annotateMfDesfire(char *exp, size_t size, uint8_t *cmd, uint8_t cmdsize) {
                             snprintf(exp, size, "CLEAR RECORD FILE");
                         }
                         break;
+                    case MFDES_NOTIFY_TRANSACTION_SUCCESS:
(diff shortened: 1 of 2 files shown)


iceman1001:
Interesting to see how they just keep on polluting protocols

Issue #2058, opened 2023-07-31
lto build failure


ZeroChaos-:
**Describe the bug**
building with `-flto` and some extra noisy cflags to warn about lto issues `-Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing`

```
[=] CXX proxmark3
[...]

PR #2059, opened 2023-07-31
fix lto-type-mismatch in guidummy.cpp

ZeroChaos-:
link time optimization requires all functions named the same to actually operate the same.  I am pretty sure this fixes it without breaking anything.

Fixes: https://github.com/RfidResearchGroup/proxmark3/issues/2058
src/proxgui.h:35:6: error: type of ‘ShowPictureWindow’ does not match original declaration [-Werror=lto-type-mismatch]
   35 | void ShowPictureWindow(uint8_t *data, int len);
      |      ^
src/guidummy.cpp:33:17: note: type mismatch in parameter 2
   33 | extern "C" void ShowPictureWindow(char *fn) {
      |                 ^
src/guidummy.cpp:33:17: note: type ‘void’ should match type ‘int’
src/guidummy.cpp:33:17: note: ‘ShowPictureWindow’ was previously declared here

### DIFF
--- a/client/src/guidummy.cpp
+++ b/client/src/guidummy.cpp
@@ -30,7 +30,7 @@ extern "C" void ShowGraphWindow(void) {
 extern "C" void HideGraphWindow(void) {}
 extern "C" void RepaintGraphWindow(void) {}
 
-extern "C" void ShowPictureWindow(char *fn) {
+extern "C" void ShowPictureWindow(char *fn, int len) {
     static int warned = 0;
 
     if (!warned) {


tcprst:
Tested with `SKIPQT=1` and it's working as expected.

Tested-by:@tcprst

iceman1001:
Great find!
When did this ever trigger?  I haven't seen this error.

ZeroChaos-:
This one is definitely caused by -flto which enforces the C++ One Definition Rule https://en.cppreference.com/w/cpp/language/definition

If you want to test with lto (which gets you smaller and faster binaries) then the Gentoo CXXFLAGS suggestions are `-Werror=odr -Werror=lto-type-mismatch -flto` which basically turns on lto and enables two errors for things that lto doesn't like.

PR #2061, opened 2023-08-01
fix -Wstringop-overread error

ZeroChaos-:
We are reading this 26 bytes into a 32 byte buffer, is this even any worse?
Fixes: https://github.com/RfidResearchGroup/proxmark3/issues/2060


iceman1001:
Interesting catch.  
When would too few bytes into an array be a problem.
usually there is a `snprintf` which is the root problem when copying.

ZeroChaos-:
In this case it's -flto link time optimization complaining I believe.  There are a number of similar things that complain about this stuff so I'm not 100% sure which one caused this warning (I don't specifically pass that cflag).

PR #2069, opened 2023-08-09
Adds nfc to eml conversion tool

celowsky:
The goal of this PR is to provide users of the proxmark3 with a tool to convert files from an nfc format to an eml format.


celowsky:
Already added in https://github.com/RfidResearchGroup/proxmark3/pull/1991

PR #2074, opened 2023-08-10
Added support for changing uid GEN2

TrixyCracker:
now it is possible to change the gen2 uid using the command "hf mf supercard -u <UID>".

### DIFF
--- a/client/src/cmdhfmf.c
+++ b/client/src/cmdhfmf.c
@@ -7040,10 +7040,40 @@ static int CmdHf14AMfSuperCard(const char *Cmd) {
     }
 
     // Super card generation 2
-    if (i == SUPER_MAX_TRACES) {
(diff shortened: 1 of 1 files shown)


AloneLiberty:
As this is Gen4 UMC card, there are already implemented commands under 'hf mf g*' and lua script. And your code won't work with UID length != 4 and with custom password (not zeros). In my opinion better to suggest user already implemented commands.

TrixyCracker:
> As this is Gen4 UMC card, there are already implemented commands under 'hf mf g*' and lua script. And your code won't work with UID length != 4 and with custom password (not zeros). In my opinion better to suggest user already implemented commands.

i'm talking about sniffercard, and code works well with 4bytes UID.

As prove of that, see image below.
Screenshot from 2023-08-11 01-28-26

PR #2077, opened 2023-08-15
Minor (useful) updates

henrygab:
Three items:
1. Extend Little-endian and Big-endian functions to support 5-9 byte data.
2. Update WSL2 instructions for improved user experiences
3. Add hours / minutes / seconds to logfile

<details><summary>LE and BE functions</summary>
<p>

These functions are added because they are useful for em4x70 data structures.
These mirror functionality already existing for 2-4 byte versions:

```C
uint64_t MemLeToUint5byte(const uint8_t *data);
uint64_t MemLeToUint6byte(const uint8_t *data);
uint64_t MemLeToUint7byte(const uint8_t *data);
uint64_t MemLeToUint8byte(const uint8_t *data);
uint64_t MemBeToUint5byte(const uint8_t *data);
uint64_t MemBeToUint6byte(const uint8_t *data);
uint64_t MemBeToUint7byte(const uint8_t *data);
uint64_t MemBeToUint8byte(const uint8_t *data);
void Uint5byteToMemLe(uint8_t *data, uint64_t value);
void Uint6byteToMemLe(uint8_t *data, uint64_t value);
void Uint7byteToMemLe(uint8_t *data, uint64_t value);
void Uint8byteToMemLe(uint8_t *data, uint64_t value);
void Uint5byteToMemBe(uint8_t *data, uint64_t value);
void Uint6byteToMemBe(uint8_t *data, uint64_t value);
void Uint7byteToMemBe(uint8_t *data, uint64_t value);
void Uint8byteToMemBe(uint8_t *data, uint64_t value);
```

In addition, formatting was standardized to make code review of each function easier,
including aligning the columns and treating each byte on a separate code line.

</p>
</details> <details><summary>WSL2 Updates</summary>
<p>

Tracked down the minimum necessary for repeatable, reliably hot-plug support.
Previously, permissions would appear to randomly fail to be set properly, preventing
the client from talking to the device.

Turns out, the root cause is the shutdown of the `WSL2` subsystem.  When it's not in use, it's unloaded (after some undetermined set of requirements are met).  That means if there are no WSL2 terminals open, the next time one is opened might have behavior similar to a reboot, in that the WSL2 state is "fresh" again.

The good news is that the workarounds are simple.

</p>
</details> <details><summary>Logfile update</summary>
<p>

The prior code created a log file based on the date only (`log_YYYYmmdd.txt`).  When running multiple clients (e.g., one live on hardware, one disconnected; or multiple physical proxmark devices), the log files would fail to log everything.

This commit adjusts the log file to include hours, minutes and seconds (`log_YYYYmmddHHMMSS.txt`.  While not a 100% fix, as long as the clients are started at least a full second apart, each client will have its own private log file.  An alternative is to append random characters to the filename.  This may still be recommended later, but this should be good enough for most situations.

As an added bonus, it becomes easier to keep log files in order.

</p>
</details>

### DIFF
--- a/client/src/proxmark3.h
+++ b/client/src/proxmark3.h
@@ -36,7 +36,7 @@
 #define PROXPROMPT_DEV_OFFLINE _RL_BOLD_RED_("offline")
 
 #define PROXHISTORY "history.txt"
-#define PROXLOG "log_%Y%m%d.txt"
+#define PROXLOG "log_%Y%m%d%H%M%S.txt"
 #define MAX_NESTED_CMDSCRIPT 10
 #define MAX_NESTED_LUASCRIPT 10
 
--- a/common/commonutil.c
+++ b/common/commonutil.c
@@ -161,75 +161,255 @@ uint64_t bytes_to_num(uint8_t *src, size_t len) {
 }
 
 uint16_t MemLeToUint2byte(const uint8_t *data) {
-    return (uint16_t)((uint16_t)
-               (data[1] << 8) + data[0]
-           );
+    return (uint16_t)(
+        (((uint16_t)(data[1])) << (8*1)) +
+        (((uint16_t)(data[0])) << (8*0))
+        );
 }
 
 uint32_t MemLeToUint3byte(const uint8_t *data) {
-    return (uint32_t)((uint32_t)
-               (data[2] << 16) + (data[1] << 8) + data[0]
-           );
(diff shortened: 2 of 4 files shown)


iceman1001:
Hm.,..
Not a super fan of adding functions that is not used.  
but yours handles the type casting part ...

PR #2078, opened 2023-08-15
More const

henrygab:
For arm sources, this brings up to `lfzx`.

### DIFF
--- a/armsrc/LCD_disabled.c
+++ b/armsrc/LCD_disabled.c
@@ -53,7 +53,7 @@ void LCDFill(unsigned char xs, unsigned char ys, unsigned char width, unsigned c
     }
 }
 
-void LCDString(char *lcd_string, const char *font_style, unsigned char x, unsigned char y, unsigned char fcolor, unsigned char bcolor) {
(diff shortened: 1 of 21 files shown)


iceman1001:
Is this you fixing cpp warnings?

henrygab:
> Is this you fixing cpp warnings?

Nope.  I'm just walking through the code to understand it more, and choosing to improve it as part of that learning.  `const` is useful in that context by listing the locations where the data structures are not modified by a function.

As you know, `const` is entirely optional ... meaning if it's removed entirely from the code, it would still be functionally the same.  So, I'm not sure if there even _exists_ a warning that would recommend const for parameters (even in C++).

Thanks for the quick merges ... helps keep my local branches clear.  :)

iceman1001:
well... I am sure I seen tons of const warnings in the cpp checker output ...  Too exhausting to fiddle with.   Like you say it doesn't change the code.
I think also too early initialization of vars,  and param names different from header file definition ...    

Yeah,   your local branches get clear... mine gets messed up  ;)

henrygab:
> Yeah, your local branches get clear... mine gets messed up ;)

Just say the word, and I will stop.  :)

iceman1001:
No worries,   I don't mind.

PR #2081, opened 2023-08-17
Fix macports Makefile configuration.

rdmitr:
This PR fixes macports configuration in the client Makefile.


iceman1001:
@linuxgemini   might want to verify it...

PR #2082, opened 2023-08-17
Configurable number of CPU cores.

rdmitr:
This PR adds a new command line argument, `--ncpu`, to override the detected CPU core count. It is intended to allow for better fine-tuning of performance on machines with large numbers of CPUs.

### DIFF
--- a/client/src/proxmark3.c
+++ b/client/src/proxmark3.c
@@ -1004,8 +1004,9 @@ int main(int argc, char *argv[]) {
                 return 1;
             }
             long int ncpus = strtol(argv[i + 1], NULL, 10);
-            if (ncpus < 0 || ncpus >= INT_MAX) {
-                PrintAndLogEx(ERR, _RED_("ERROR:") " invalid number of CPU cores: --ncpu " _YELLOW_("%s") "\n", argv[i + 1]);
+            const int detected_cpus = detect_num_CPUs();
+            if (ncpus < 0 || ncpus >= detected_cpus) {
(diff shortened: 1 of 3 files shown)


client/src/proxmark3.c
```diff
@@ -996,6 +997,22 @@ int main(int argc, char *argv[]) {
...
         }
 
+        if (strcmp(argv[i], "--ncpu") == 0) {
+            if (i + 1 == argc) {
+                PrintAndLogEx(ERR, _RED_("ERROR:") " missing CPU number specification after --ncpu\n");
+                show_help(false, exec_name);
+                return 1;
+            }
+            long int ncpus = strtol(argv[i + 1], NULL, 10);
+            if (ncpus < 0 || ncpus >= INT_MAX) {
```
iceman1001, line 1007: How about for the upper limit use the number of CPU reported by the system.

rdmitr: I considered this but it felt like a bit of an artificial limitation. This is kind of advanced setting, so if one wants to mess with it for whatever reason, I felt I should not get in their way :)

iceman1001: I see no reason why we would allow anyone to set invalid number.   It will only crash the software.  
Change and it can be merged.

rdmitr: It should not crash anything — it will spawn more threads than physical cores, which will probably starve other processes of CPU time but shouldn't crash anything.

Anyways, I will cap this number and submit a change to this PR shortly.

PR #2084, opened 2023-08-18
Update mfc_default_keys.dic

evildaemond:
Added 4 keys extracted from old dumps on Edith Cowan University Smart rider cards


iceman1001:
Thanks Evil!

PR #2085, opened 2023-08-21
Update mfc_default_keys.dic

pickeditmate:
Adding a Dorma Kaba Grand master keycard and two maid cards


iceman1001:
Nice,

are those keys confirmed to be static default keys?    or are they different on other cards from same system installation?

pickeditmate:
Ah sorry I misunderstood the assignment, the only static value common to them is the top in each card ( 0D258FE90296 ) - they're all from the same installation.

I'm not sure if the grand master keys are useful for other grand master cards perhaps.

Shall I close this pr until we know?

iceman1001:
Well,  we usually only want the static default ones across cards.  
the rest are diversified with your UID,  so they are not useful for anyone else.

And we dont want duplicates in the list.

iceman1001:
and we already have `0D258FE90296`  in the dictionary.

PR #2086, opened 2023-08-21
Adds `hf iclass esetblk`, equivalent to `hf mf esetblk`

nvx:
Also some minor tweaks to make commands.json and commands.md properly on Windows. Previously on windows it would cause a difference vs on Linux due to the case of differences of the serial device.

I also noticed a similar "always makes a diff for every line" with the `pm3line_vocabulary.h` due to a fight between the python script adding a space at the end of each line and the rest of make style removing trailing whitespace on lines.


client/src/cmdhficlass.c
```diff
@@ -1222,6 +1222,52 @@ static int CmdHFiClassEView(const char *Cmd) {
     return PM3_SUCCESS;
 }
 
+static int CmdHFiClassESetBlk(const char *Cmd) {
+    CLIParserContext *ctx;
+    CLIParserInit(&ctx, "hf iclass esetblk",
+                  "Display emulator memory.\n"
```
iceman1001, line 1228: I believe you need to update this texts to match the new command

iceman1001:
Nice fixes,   just need the help text to be updated then it can be merged

would actually been better with several PR's...

client/src/cmdhficlass.c
nvx: Whoops missed that one!

nvx:
> would actually been better with several PR's...

yeah I thought that afterwards, wasn't intending to fix the commands build stuff but then ended up fighting it as a result of adding the new command and didn't want a thousand line diff because the tooling was doing weird stuff.

should be right to merge now that I remembered to regenerate the doc/commands.json after fixing the help text now.

iceman1001:
Awesome fixes!
Small but still they were annoying ones.

PR #2087, opened 2023-08-22
Add hf_cardhopper standalone mode

startrekdude:
Adds the hf_cardhopper standalone mode, as one component of my (Sam Haskins) and Trevor Stevado's long-range relay attack on ISO 14443-4 Type A protocols, as presented at DEF CON 31.


iceman1001:
Awesome!

PR #2088, opened 2023-08-22
Improve iClass SIO and legacy credential detection reliability

nvx:
Now relies on the legacy config block for SR detection and the end-of-SIO detection no longer partially cuts off the SIO for any dumps I have.

The previous code would cut off some SIOs mid way through if they happened to contain the wrong bytes mid way through.

Also fixes a couple of super minor display quirks I noticed while running `hf iclass view` on all my dumps - for example if the stars aligned on a SE dump it could in some cases try and decode block 7 as if it were a legacy dump resulting in rather confusing output, and a case where `hf iclass decrypt` would attempt to use a card helper even if the flag wasn't set (again if the stars aligned with the right block data).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Changed iClass SIO and Legacy credential detection to be more reliable (@nvx)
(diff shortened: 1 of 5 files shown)


iceman1001:
Awedsome,  there were some interesting fixes,    I merge as is.

Feedback, 
- The reverse strstr got me confused.  Maybe add a comment about the two functions now in util.c.  
- the SR BIT 0xA0 Should be a MACRO define,  to denote what it is.
- help text for default start 3 and 6,  should be in the general description.   Otherwise the column text becomes too large (save screen space)

I see you must have had some unsecure tags to play with in order to trigger them :)

PR #2090, opened 2023-08-24
Fix macOS errors while building after cryptoRF solvers was added.

p51d78th:
I'm not sure if the count = 0 should have been deleted but it was also being reported as an error with the default make command.

Closes 2089


iceman1001:
Thanks!!

The counter doesn't seem to be used anyway.

PR #2091, opened 2023-08-24
Misc cleanups

nvx:
Cleanups as mentioned in the feedback of PR #2088, as well as a couple of other magic number replacements that I noticed while I was doing it. I also ran "make style" across the codebase, looks like a few other recent PRs hadn't had this done.


iceman1001:
Perfect!

PR #2094, opened 2023-08-24
Remove duplicate MFP definitions. Update AIDlist

kormax:
This commit adds SEOS AIDs to the json database.

It also removes duplicate MFP command definitions created by me on accident in one of the last PR's.


iceman1001:
Nice catch!

PR #2099, opened 2023-08-29
Fix MFU authentication to send PACK correctly.

shallax:


client/src/cmdhfmfu.c
```diff
@@ -2657,7 +2657,7 @@ static int CmdHF14AMfUDump(const char *Cmd) {
 
     //add *special* blocks to dump
     // pack and pwd saved into last pages of dump, if was not partial read
-    dump_file_data.pages = pages - 1;
```
shallax, line 2660: I'm a little concerned that this'll throw off the page offsets for all other code, but without this, the PACK was missing from the emulator memory so it was impossible to send it back when handling a password challenge.

iceman1001:
Thanks!
I like your adding to precompiled responses is a better solution.   Most likely will sort out potential timing issues too.

PR #2102, opened 2023-09-04
dictionaries: Add keys for Vilnius public transit card

ernestask:


iceman1001:
Nice!
Do you know if they are static keys,  ie same keys across different cards for the system?

ernestask:
Good question, I can check and come back to you on that.

ernestask:
I bought a couple of cards and it seems they use different keys, though the thing is that I originally used the “old world” card (pre-rebranding) and these are the “new world” cards (JUDU).

I’ll take a closer look in the evening, maybe the keys are static for the same revision only.

iceman1001:
Usually systems use a key gen algorithm (KDF) to calculate card diversifed keys  but what we want is static keys.  
Look careful if there is one or two keys static across.    

If they made a proper system then its all diversified and they are useless for our dictionary key file. 
but useful for you.

ernestask:
Yup, seems that you’re right, I only see four static keys:

#### A
```
16901cb400bc
8c187e78ee9c
f0fe56621a42
fe2a42e85ca8
```

#### B
```
6a6c80423226
f4ce4af888ae
307448829ebc
c2a0105eb028
```

iceman1001:
So those are the keys we want in this dictionary!
If you update your PR ,  then I can merge

ernestask:
Done!

iceman1001:
Thanks!

if you connect on discord,  I have some questions ...

PR #2103, opened 2023-09-05
Fix crc render position when data len is multiple of 16

piru:
This fixes crc rendering when data len is multiple of 16. Before you would for example get this when the first char of ANSI escape code would get lost:

```
   77369536 |   77392704 | Tag |xx  xx  xx  xx  xx  xx  xx  xx  xx  xx  xx  xx  xx  xx  xx  xx  xx  xx   |     | 
            |            |     |[0m yy  yy  
```


iceman1001:
Awesome!

piru:
@iceman1001 I think this code might not be 100% yet, it likely shifts the output a bit now. The code coloring the output is rather hard to follow. But at least it doesn't omit the ESC code now.

iceman1001:
its an improvement.

PR #2107, opened 2023-09-10
Update hf_mf_ultimatecard.lua

Antiklesys:
Fixed bug, should be 13 instead of 12 as there are two 12 otherwise and this is referring to the 128b one instead


iceman1001:
Good catch!

PR #2109, opened 2023-09-11
Change client/src/cmdhfmf.c to give better failure message

spandox2:
Problem:
"hf mf autopwn" returns with no message if no card can be read
Solution:
Change a DEBUG message to ERR and have it give a hint


iceman1001:
I think we had this message for over ten years... 

For the hints system we don't have that on device side.  Look into adding a hint on the client side?

iceman1001:
Now I see its on client side..    
Here is how you make use of the HINT system instead. 

```
PrintAndLogEx(HINT, "Hint: Try some distance or position of the card");
```

spandox2:
Better Options

PR #2110, opened 2023-09-11
Change in behavior of hf mf autopwn when no card is detected

spandox2:
Problem:
`hf mf autopwn` returns silently when no card is detected. There is a DEBUG message that most people won't see

Proposed Solution:
Change the DEBUG to a FAILED and add a HINT


iceman1001:
Took some time but we got there in the end!

PR #2111, opened 2023-09-12
Fix linking in explicit libatomic devices

lpla:
My Raspberry Pi Zero with latest Raspberry Pi OS thanks this after commit b17a733


lpla:
Hi. I see that macOS is broken now with this change. Probably this should be the way this was fixed, as noted by this other proxmark3/client/Makefile:
```
ifneq ($(platform),Darwin)
    LDLIBS += -Wl,--as-needed -latomic -Wl,--no-as-needed
endif
```

iceman1001:
yeah,  Sorted now!

PR #2116, opened 2023-10-03
Fixing macOS errors.

p51d78th:
ATOMIC_VAR_INIT is now deprecated and is causing build failures for MacOS.

The unused functions were also causing errors. What is the desired approach for dealing with those cases?


iceman1001:
The unused functions was there to validate if the modified multithreaded  version would work or not.
Removing them is ok,   its in the git history if needed.

PR #2117, opened 2023-10-04
Changed hf iclass view/decrypt to detect SIO lengths better and show if legacy credentials are encrypted

nvx:
It now does some super basic parsing of the ASN.1 to determine the length rather than pattern matching. I also feed this information into the block view to correctly highlight the SIO blocks rather than always assuming it's the same number of blocks.

I also added some bounds checking for safety and tweaked the legacy block display to show "HID CFG" for block 6, and "Enc Cred" if the legacy block is encrypted instead of just always showing "Cred".

This fixes the bug mentioned in discord where it would fail to show custom keyed SIOs correctly


iceman1001:
Excellent!

PR #2118, opened 2023-10-05
Update mfc_default_keys.dic

evildaemond:
Added Keys found on old cards plus notes on another key sighting


iceman1001:
Nice!
Do you know if its static default keys across tags from same system?   or is this one card only?

I'm asking since this list is about finding static default keys across tags from same system.

evildaemond:
These are static across the same system from my testing

iceman1001:
Excellent!

PR #2121, opened 2023-10-10
Add documentation for iOS

The-SamminAter:
This patch/pr adds documentation I have written on how to compile and install proxmark3 on iOS. It's not the cleanest process ever, but it's getting there.


iceman1001:
Awesome!

PR #2123, opened 2023-10-10
Added Yale mfc keys

Abbin44:
Added a few keys from Yale NFC door locks


iceman1001:
Nice,
are those keys static across several different card for yale systems?   
or are they from one card?

Abbin44:
After double-checking I actually think all 7 keys are for one tag.

iceman1001:
OK,
This dictionary is for static default keys.    Your diversified keys only work for your tag and isn't useful for others. 

Closing,   but keep up the good work!

PR #2126, opened 2023-10-13
Finished configcard implementation for keyroll cards

Antiklesys:
Finalized implementation of configcard generation for mode 9 & 10 (keyroll) when cardhelper is not present.
Results need to be compared with cardhelper output to confirm blocks 14 & 15 behavior / accuracy.


Antiklesys (commented):
Fixed redundant condition as it was flagged for compiling failures on mac

Antiklesys (commented):
Rolled back changes as using:
`IsCardHelperPresent(false)` instead of `IsCardHelperPresent(false) == false` .

PR #2022, opened 2023-07-04
Fix the timeout of TCP connections

wh201906:
I just found that the timeout of TCP connections is overwritten in TestProxmark()
Here is the test output
<details><summary>Details</summary>

```
[=] Using UART port tcp:192.168.0.1:1234
Current timeout(us):500000
...
Current timeout(us):500000
[=] Communicating with PM3 over USB-CDC over TCP


  8888888b.  888b     d888  .d8888b.   
  888   Y88b 8888b   d8888 d88P  Y88b  
  888    888 88888b.d88888      .d88P  
  888   d88P 888Y88888P888     8888"  
  8888888P"  888 Y888P 888      "Y8b.  
  888        888  Y8P  888 888    888  
  888        888   "   888 Y88b  d88P 
  888        888       888  "Y8888P"    [ ☕ ]


Current timeout(us):20000
...
Current timeout(us):20000
  [ Proxmark3 RFID instrument ]

Current timeout(us):20000
...
Current timeout(us):20000
[!] ⚠️  Received packet frame with variable part too short? 182/412
Current timeout(us):20000
Current timeout(us):20000

```

</details>

PR #2127, opened 2023-10-14
Reduce latency in local TCP connection

wh201906:
Before PR #2022, the default communication timeout for TCP is accidentally set to 20ms, but it works  for Termux.
In PR #2022, I fixed the default timeout of TCP connections, but it increases the latency when running some commands in Termux. To reduce the latency, I added some code to detect whether the host address is `127.0.0.1` or `localhost`, and reduce the timeout if so.


iceman1001:
Great!
Maybe use `strstr`  ?

wh201906:
> Maybe use `strstr` ?

That sounds better. Could you help me to modify that?

PR #2128, opened 2023-10-14
Faster CI

wh201906:
+ Run GitHub Action jobs in parallel for the macOS CI and Ubuntu CI
+ Run multiple `make` jobs


wh201906:
Ubuntu: 9m 20s -> 2m 25s
macOS: 30m 14s -> 11m 12s
Windows: 25m 2s -> 19m 56s

iceman1001:
Looks cool,   if it works then its even better!

PR #2129, opened 2023-10-15
Fixed iclass offline mode commands

Antiklesys:
Fixed iclass offline mode commands


client/src/cmdhficlass.c
```diff
@@ -4472,18 +4472,18 @@ static int CmdHFiClassSAM(const char *Cmd) {
 
 static command_t CommandTable[] = {
     {"help",        CmdHelp,                    AlwaysAvailable, "This help"},
-    {"list",        CmdHFiClassList,            AlwaysAvailable, "List iclass history"},
+    {"list",        CmdHFiClassList,            IfPm3Iclass,     "List iclass history"},
```
iceman1001, line 4475: this can be used offline.

client/src/cmdhficlass.c
```diff
@@ -4498,7 +4498,7 @@ static command_t CommandTable[] = {
     {"-----------", CmdHelp,                    AlwaysAvailable, "---------------------- " _CYAN_("utils") " ----------------------"},
     {"configcard",  CmdHFiClassConfigCard,      IfPm3Iclass,     "Reader configuration card"},
     {"calcnewkey",  CmdHFiClassCalcNewKey,      AlwaysAvailable, "Calc diversified keys (blocks 3 & 4) to write new keys"},
-    {"encode",      CmdHFiClassEncode,          AlwaysAvailable, "Encode binary wiegand to block 7"},
+    {"encode",      CmdHFiClassEncode,          IfPm3Iclass,     "Encode binary wiegand to block 7"},
```
iceman1001, line 4501: users wanted to see the encoded blocks only,  hence it is available in offline

iceman1001:
You have some incorrect assumptions of the commands. 

Only one valid in `hf iclass info`

client/src/cmdhficlass.c
Antiklesys: I think you're confusing this with "hf iclass encrypt" ?
"hf iclass encode" requires the proxmark to be plugged in, doesn't show anything otherwise.
image

client/src/cmdhficlass.c
Antiklesys: Can you provide more insights on this?
On my end this is what happens when using this offline:

image

client/src/cmdhficlass.c
Antiklesys: I've also tried running this offline with the -v option but same outcome

client/src/cmdhficlass.c
iceman1001: Like the output say,   you haven't loaded a trace list so there is nothing to show.

client/src/cmdhficlass.c
iceman1001: ... I dunno,  I have no issue running it with `-v` 

image

client/src/cmdhficlass.c
Antiklesys: Reverted the change for this one, will upload shortly once I finish the checks on the encode function

client/src/cmdhficlass.c
Antiklesys: Unsure why I got an inconsistent behavior, will revert change

iceman1001:
Perfect!

PR #2132, opened 2023-10-16
Increase the speed of uart_receive() in UDP mode

wh201906:
The `read()` can only write data to a continuous block of memory, so theoretically we need a transit buffer to hold the data from `read()` before we can put the data into the ring buffer.
However, we can write data to the internal buffer between `data[rear]` and `data[capacity-1]`, in which the space is continuous.


wh201906:
Added some code for error handling.

PR #2136, opened 2023-10-17
Fix potential stack overflow while making it easier to support more ATRs

jmichelp:
Tested with the only SAM I have (Grace).


iceman1001:
Nice one!

PR #2138, opened 2023-10-18
Change URLs of badge in README.md

wh201906:
The badge of macOS CI always shows failing for a long time, even for successful runs of CI
MacOS Build and Test
`MacOS Build and Test`
It turns out the URL of these badges has changed. So this one works
MacOS Build and Test
`MacOS Build and Test`
The API changes can be found there.
https://github.com/github/docs/pull/3802/files#diff-95d58394efa588c4f92f85274259efb8ca1e10a02bd346549e65b57604c156c4


iceman1001:
Thanks! 
That was an annoying thing...

PR #2139, opened 2023-10-18
Support double-quoted arguments in CLI

jmichelp:
When a quoted command argument is seen, it will take all characters until the next double-quote (no supported escape sequence here for simplicity), skipping spaces (space, tab, etc.) inside.

This is meant to allow copy/pasting of formatted code, especially hex-encoded values. This doesn't add the capability to pass argument containing spaces (e.g. filenames).

This means that the following commands should behave the same:
```
pm3> wiegand encode --fc 101 --cn 1337
pm3> wiegand encode --fc "1 0 1" --cn "1    3    3   7"
```

Or a more realistic/useful example, when copy/pasting hex formatted values from other commands:
```
pm3> hf iclass calcnewkey --old 1122334455667788 --new 2233445566778899
pm3> hf iclass calcnewkey --old "11 22 33 44 55 66 77 88" --new "22 33 44 55 66 77 88 99"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Add support for quoted arguments in the CLI, allowing spaces in them which
+   are removed automatically (@jmichelp)
  - Added UDP support on Windows (@wh201906)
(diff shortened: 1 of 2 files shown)


iceman1001:
Interesting one!

Thanks!

wh201906:
How about paths with spaces? Will the space in it be removed?

wh201906:
I think the code for removing the spaces should be put in the specific commands for the flexibility. For paths we should not touch the spaces in it, and for data/key we can remove the spaces safely.

wh201906:
Ah that's reasonable.

PR #2140, opened 2023-10-18
Add short flag -m for timeout related commands

wh201906:


iceman1001:
I still don't know what this commands with a simple parameter needs three examples how to call it.....

wh201906:
Ok I'll remove one

PR #2143, opened 2023-10-20
Modified hf iclass calcnewkey

Antiklesys:
Added support for a scenario in which the old key is elite and the new key is not elite.


iceman1001:
I remembered doing that ...

then I realized,  if I swap the place with the elite and none elite key... I could do the same without having an extra param....

Antiklesys:
> I remembered doing that ...
> 
> then I realized, if I swap the place with the elite and none elite key... I could do the same without having an extra param....

I’m not that smart. I’ve always been told to run the command manually and then xor the keys.
You’re right iceman, but it’d be handy to have this command to avoid confusion 🙏

PR #2144, opened 2023-10-23
Added bind option for network connections

wh201906:
For some USB serial to UDP bridge apps, the outbound port should be specified.


iceman1001:
I am sure someone will use it.  Maybe termux people will need it.

PR #2146, opened 2023-10-24
cmdscript.c: explicitly include ctype header

s00se:
Brought to attention by preparation for Fedora 41 packaging @ https://bugzilla.redhat.com/show_bug.cgi?id=2245823


iceman1001:
Good catch!

PR #2147, opened 2023-10-25
Clean up the code for timeout

wh201906:
+ Move the local address detection logic from `TestProxmark()` to `uart_open()`, to distinguish the local address in target address and bind address
+ Rename `UART_TCP_CLIENT_RX_TIMEOUT_MS` to `UART_NET_CLIENT_RX_TIMEOUT_MS`
+ Add `UART_UDP_LOCAL_CLIENT_RX_TIMEOUT_MS` for shorter timeout. According to the test results the UDP speed is almost the same as the native USB speed


iceman1001:
Excellent!

PR #2151, opened 2023-10-27
Fixed keyroll mode for hf ic configcard

Antiklesys:
Was able to compare the output of nfc-iclass and I identified two bugs in the generation of blocks 0x0D (was being overwritten with FF, it seems this was an original bug in the original pm3 code?) Fixed the porting for block 15 generation.


iceman1001:
Good catch!

Antiklesys:
Time to deprecate cardhelper support? :P

PR #2152, opened 2023-10-29
Cardhelper Deprecation Proposal

Antiklesys:
Proposed cardhelper deprecation IF we can now confirm all cardhelper functionalites are part of the pm3 repo.
Rebranded cardhelper.c and cardhelper.h to samcard.c and samcard.h as they only contain sam related function currently.


team-orangeBlue:
Out of curiosity... configcard functionality?

Antiklesys:
> Out of curiosity... configcard functionality?

Is fully implemented without the need of the cardhelper.

iceman1001:
No.

PR #2154, opened 2023-10-30
Implemented 'info' command for NTAG424

dankar:
Implemented GetVersion and GetSignature commands for NTAG424, and added getting/printing of them in the "info" command.

I did a dubious thing by making the signature verification from DESFire public by making a wrapper function that didn't need to supply card_type. Seems like that parameter is not needed anyway.

The version information command is the same as for desfire, so might have been possible to use that instead, but the signature was not readable by the desfire commands.

### DIFF
--- a/client/src/cmdhfmfdes.c
+++ b/client/src/cmdhfmfdes.c
@@ -372,7 +372,7 @@ static int mfdes_get_info(mfdes_info_res_t *info) {
 }
 
 // --- GET SIGNATURE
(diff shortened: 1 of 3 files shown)


iceman1001:
I removed that unused param. 
If you just add a function header and remove the static in your PR,    then this becomes a bit cleaner.

dankar:
Awesome, that made it a bit better.

PR #2155, opened 2023-10-30
Fix mf file loading error

henrygab:
The code was incorrectly checking the loop variable `i`,
rather than checking the used data length / current buffer pointer.

This causes dumps with all 64 sectors of data to refuse to load, with false error message that the data buffer size was exceeeded.

Also improved flow for dump files, by providing distinct error message when a data block in the JSON ...
* ... would exceed maximum total data size
* ... has zero data
* ... has fewer than 16 bytes of data

Bonus fix for that third case (< 16 bytes):

The old code would save the partial-block into the buffer ... but all future blocks of data would then be loaded at an incorrect offset.  The fixed code ensures zero-filled data is used, and this is noticable at the client via the warning message.

### DIFF
--- a/client/src/fileutils.c
+++ b/client/src/fileutils.c
@@ -1087,6 +1087,7 @@ int loadFileJSONex(const char *preferredName, void *data, size_t maxdatalen, siz
 
     // Proxmark3 settings file.  No
     if (!strcmp(ctype, "settings")) {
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice catches!

Something makes me believe that more json loading parts could use your fixes...

PR #2161, opened 2023-11-06
Fix uint32_t overflow

wh201906:
If the timeout option of `hw status` is too large, the `1000 * bytes_transferred` will excess the range of `uint32_t` and give a wrong answer.
For USB connection, 20s of timeout can easily cause this bug.


iceman1001:
You are right,  there are more `%llu` at the device side.  We only enforce it on client side. 

I pushed the fix with some minor changes to variable name.

PR #2169, opened 2023-11-09
The fix to the completely fixed magic notes

team-orangeBlue:
The problem with PR 2003 is that all blame for some reason went to me.

I don't keep track of how PRs work, but I'd rather not try my luck with merging #2003 and then making the document essentially explode.
There's now this.
It's ready to be merged.

TLDR: added some country custom chips to cover for the rest of the userbase that could possibly be there.


iceman1001:
Excellent work with identifying and documenting all new different magical tags.

PR #2172, opened 2023-11-13
mfkey: add explicit -lpthread link directive to Makefile for linux builds

ggallagh:
`nested_util.c` requires pthread support. Prior to release 2.34 of glibc (around Aug 2021), libpthread was a separate library that needed to be explicitly supplied to the linker. Add the directive to make all linux builds happy.


iceman1001:
Thanks for the PR!

Now, two PRs ,  two different solutions.  Which to choose.  
Someone mentioned that both works.

SKIPTHREAD seems to be the right path considering its purpose.

If noone else has any objections,  I will close this PR and merge the other.

iceman1001:
Merged the other PR regarding this problem,.

PR #2173, opened 2023-11-14
Real-time LF sampling

wh201906:
#2167 with clear commit history and some minor improvements

Known limitations:
1. Only `bits_per_sample = 8` is supported now.
2. Only USB connection is supported now. The transfer speed should be greater than 300kByte/s to support this feature.
3. When LF image is not loaded into the FPGA, the first 14 bytes could be the response of CMD_WTX rather than real samples.

### DIFF
--- a/armsrc/lfsampling.c
+++ b/armsrc/lfsampling.c
@@ -27,6 +27,7 @@
 #include "lfdemod.h"
 #include "string.h"  // memset
 #include "appmain.h" // print stack
+#include "usb_cdc.h" // real-time sampling
 
 /*
 Default LF config is set to:
@@ -423,6 +424,126 @@ static uint32_t ReadLF(bool reader_field, bool verbose, uint32_t sample_size, bo
 uint32_t SampleLF(bool verbose, uint32_t sample_size, bool ledcontrol) {
     BigBuf_Clear_ext(false);
     return ReadLF(true, verbose, sample_size, ledcontrol);
+}
+
+/**
+ * Do LF sampling and send samples to the USB
+ * 
+ * Uses parameters in config. Only bits_per_sample = 8 is working now
+ * 
+ * @param reader_field - true for reading tags, false for sniffing
+ * @return sampling result
+**/
+int ReadLF_realtime(bool reader_field) {
+    // parameters from config and constants
+    const uint8_t bits_per_sample = config.bits_per_sample;
+    const int16_t trigger_threshold = config.trigger_threshold;
+    int32_t samples_to_skip = config.samples_to_skip;
+    const uint8_t decimation = config.decimation;
+
+    uint32_t sample_size = 64;
+    const int8_t size_threshold_table[9] = {0, 64, 64, 60, 64, 60, 60, 56, 64};
+    const int8_t size_threshold = size_threshold_table[bits_per_sample];
+
+    // DoAcquisition() start
+    uint8_t last_byte = 0;
+    uint8_t curr_byte = 0;
+    int return_value = PM3_SUCCESS;
+
+    initSampleBuffer(&sample_size); // sample size in bytes
+    if (sample_size != 64) {
+        return PM3_EFAILED;
+    }
+
+    sample_size <<= 3; // sample size in bits
+    sample_size /= bits_per_sample; // sample count
+
+    bool trigger_hit = false;
+    int16_t checked = 0;
+
+    return_value = async_usb_write_start();
+    if (return_value != PM3_SUCCESS) {
+        return return_value;
+    }
+
+    BigBuf_Clear_ext(false);
+    LFSetupFPGAForADC(config.divisor, reader_field);
+
+    while (BUTTON_PRESS() == false) {
+        // only every 4000th times, in order to save time when collecting samples.
+        // interruptible only when logging not yet triggered
+        if (unlikely(trigger_hit == false && (checked >= 4000))) {
+            if (data_available()) {
+                checked = -1;
+                break;
+            } else {
+                checked = 0;
+            }
+        }
+        ++checked;
+
+        WDT_HIT();
+
+        if ((AT91C_BASE_SSC->SSC_SR & AT91C_SSC_TXRDY)) {
+            LED_D_ON();
+        }
+
+        if (AT91C_BASE_SSC->SSC_SR & AT91C_SSC_RXRDY) {
+            volatile uint8_t sample = (uint8_t)AT91C_BASE_SSC->SSC_RHR;
+
+            // (RDV4) Test point 8 (TP8) can be used to trigger oscilloscope
+            LED_D_OFF();
+
+            // threshold either high or low values 128 = center 0.  if trigger = 178
+            if (unlikely(trigger_hit == false)) {
+                if ((trigger_threshold > 0) && (sample < (trigger_threshold + 128)) && (sample > (128 - trigger_threshold))) {
+                    continue;
+                }
+                trigger_hit = true;
+            }
+
+            if (unlikely(samples_to_skip > 0)) {
+                samples_to_skip--;
+                continue;
+            }
+
+            logSample(sample, decimation, bits_per_sample, false);
+
+            // write to USB FIFO if byte changed
+            curr_byte = data.numbits >> 3;
+            if (curr_byte > last_byte) {
+                async_usb_write_pushByte(data.buffer[last_byte]);
+            }
+            last_byte = curr_byte;
+
+            if (samples.total_saved == size_threshold) {
+                // request usb transmission and change FIFO bank
+                if (async_usb_write_requestWrite() == false) {
+                    return_value = PM3_EIO;
+                    break;
+                }
+
+                // reset sample
+                last_byte = 0;
+                data.numbits = 0;
+                samples.counter = size_threshold;
+                samples.total_saved = 0;
+
+            } else if (samples.total_saved == 1) {
+                // check if there is any data from client
+                if (data_available_fast()) {
+                    break;
+                }
+            }
+        }
+    }
+    LED_D_OFF();
+    return_value = async_usb_write_stop();
+
+    // DoAcquisition() end
+    StopTicks();
+    FpgaWriteConfWord(FPGA_MAJOR_MODE_OFF);
+    return return_value;
 }
 /**
 * Initializes the FPGA for sniffer-mode (field off), and acquires the samples.
--- a/armsrc/lfsampling.c
+++ b/armsrc/lfsampling.c
@@ -229,7 +229,7 @@ void logSample(uint8_t sample, uint8_t decimation, uint8_t bits_per_sample, bool
 
         // write the current byte
         data.buffer[data.numbits >> 3] |= sample >> bits_offset;
-        int numbits = data.numbits + bits_cap;
+        uint32_t numbits = data.numbits + bits_cap;
 
         // write the remaining bits to the next byte
         data.buffer[numbits >> 3] |= sample << (bits_cap);
@@ -306,7 +306,7 @@ uint32_t DoAcquisition(uint8_t decimation, uint8_t bits_per_sample, bool avg, in
 
         // only every 4000th times, in order to save time when collecting samples.
         // interruptible only when logging not yet triggered
-        if ((checked >= 4000) && trigger_hit == false) {
+        if (unlikely(trigger_hit == false && (checked >= 4000))) {
             if (data_available()) {
                 checked = -1;
                 break;
@@ -325,11 +325,11 @@ uint32_t DoAcquisition(uint8_t decimation, uint8_t bits_per_sample, bool avg, in
         if (AT91C_BASE_SSC->SSC_SR & AT91C_SSC_RXRDY) {
             volatile uint8_t sample = (uint8_t)AT91C_BASE_SSC->SSC_RHR;
 
-            // Test point 8 (TP8) can be used to trigger oscilloscope
+            // (RDV4) Test point 8 (TP8) can be used to trigger oscilloscope
             if (ledcontrol) LED_D_OFF();
 
             // threshold either high or low values 128 = center 0.  if trigger = 178
-            if (trigger_hit == false) {
+            if (unlikely(trigger_hit == false)) {
                 if ((trigger_threshold > 0) && (sample < (trigger_threshold + 128)) && (sample > (128 - trigger_threshold))) {
                     if (cancel_after > 0) {
                         cancel_counter++;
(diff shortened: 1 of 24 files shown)


wh201906:
@henrygab @iceman1001 Could you please review this PR?

client/src/cmdlf.c
```diff
@@ -689,109 +688,182 @@ int CmdLFConfig(const char *Cmd) {
...
+static int lf_read_internal(bool realtime, bool verbose, uint64_t samples) {
     if (!g_session.pm3_present) return PM3_ENOTTY;
 
-    struct p {
-        uint32_t samples : 31;
-        bool     verbose : 1;
-    } PACKED;
-
-    struct p payload;
+    lf_sample_payload_t payload;
```
henrygab, line 700: Could this leave fields of the struct uninitialized?
Even if not true now, still recommend to zero the full `sizeof()` a struct when it is allocated ... else padding bits can be in undefined states, which is technically fine, but annoying when debugging.   Also, longer-term, if the struct changes and an individual use isn't updated, zero'ing the full struct ensures at least repeatable (if not correct) behavior, simplifying tracking bugs down.

<details><summary>Contrast with ... </summary><p/>

Contrast with `sample_config` below, which **_immediately_** calls `lf_getconfig()`, which ensures the struct is fully initialized on success ... that style of constructor-like behavior is likely an allowed exception.

</details>

client/src/cmdlf.c
```diff
@@ -689,109 +688,182 @@ int CmdLFConfig(const char *Cmd) {
...
     clearCommandBuffer();
-    SendCommandNG(CMD_LF_ACQ_RAW_ADC, (uint8_t *)&payload, sizeof(payload));
-    PacketResponseNG resp;
-    if (gs_lf_threshold_set) {
-        WaitForResponse(CMD_LF_ACQ_RAW_ADC, &resp);
+    const uint8_t bits_per_sample = current_config.bits_per_sample;
+    const bool is_trigger_threshold_set = (current_config.trigger_threshold > 0);
+
+    if (realtime) {
+        uint8_t *realtimeBuf = calloc(samples, sizeof(uint8_t));
```
henrygab, line 715: handle allocation failure here also...

client/src/cmdlf.c
```diff
@@ -689,109 +688,182 @@ int CmdLFConfig(const char *Cmd) {
...
-    getSamples(size, verbose);
     return PM3_SUCCESS;
 }
 
+int lf_read(bool verbose, uint64_t samples) {
+    return lf_read_internal(false, verbose, samples);
+}
+
 int CmdLFRead(const char *Cmd) {
+    // In real-time mode, the first few bytes might be the response of CMD_WTX rather than the real samples
```
henrygab, line 762: Can you provide more information about how this edge case occurs?  I ask in case someone with deeper history / understanding of this code base might offer a way to avoid it....

client/src/cmdlf.c
```diff
@@ -689,109 +688,182 @@ int CmdLFConfig(const char *Cmd) {
...
     }
 
-    // response is number of bits read
-    uint32_t size = (resp.data.asDwords[0] / 8);
-    getSamples(size, verbose);
     return PM3_SUCCESS;
 }
 
 int CmdLFSniff(const char *Cmd) {
+    // In real-time mode, the first few bytes might be the response of CMD_WTX rather than the real samples
```
henrygab, line 866: Can you provide more information about how this edge case occurs?  I ask in case someone with deeper history / understanding of this code base might offer a way to avoid it....

armsrc/lfsampling.c
```diff
@@ -228,7 +229,7 @@ void logSample(uint8_t sample, uint8_t decimation, uint8_t bits_per_sample, bool
 
         // write the current byte
         data.buffer[data.numbits >> 3] |= sample >> bits_offset;
-        int numbits = data.numbits + bits_cap;
+        uint32_t numbits = data.numbits + bits_cap;
```
henrygab, line 232: good catch ... all the sources were unsigned.

armsrc/lfsampling.c
```diff
@@ -305,7 +306,7 @@ uint32_t DoAcquisition(uint8_t decimation, uint8_t bits_per_sample, bool avg, in
 
         // only every 4000th times, in order to save time when collecting samples.
         // interruptible only when logging not yet triggered
-        if ((checked >= 4000) && trigger_hit == false) {
+        if (unlikely(trigger_hit == false && (checked >= 4000))) {
```
henrygab, line 309: As per Iceman's comments, use of `unlikely()` is not generally needed.  **_Unless they are needed_**, I support Iceman's position that they should be removed in favor of readability.  This applies to all uses of the `unlikely()` and `likely()` macros....

<details><summary>Details</summary><p/>

In my prior comments during earlier development, there was concern about throughput, and this seemed to be in the time-critical path.  Even then, I couched the comment about GCC extension use with something like, "If timing concerns are critical..." or the like.

The good news is that you seem to have moved away from the very difficult "every processor cycle matters" development.  Well done!   You are now in the happier zone where profiling, hot code paths, etc. can be left behind as bad memories.

BTW ... you did a great job migrating the GCC extensions for this, and wrapping them into clear, easy-to-use macros.  Well done, even if pulled for the final PR!

</details>

armsrc/lfsampling.c
```diff
@@ -424,6 +425,123 @@ uint32_t SampleLF(bool verbose, uint32_t sample_size, bool ledcontrol) {
...
+    uint32_t sample_buffer_len = 64;
+    const int8_t size_threshold_table[9] = {0, 64, 64, 60, 64, 60, 60, 56, 64};
+    const int8_t size_threshold = size_threshold_table[bits_per_sample];
+
+    // DoAcquisition() start
+    uint8_t last_byte = 0;
+    uint8_t curr_byte = 0;
+    int return_value = PM3_SUCCESS;
+
+    initSampleBuffer(&sample_buffer_len);
```
henrygab, line 453: May I suggest moving the local declaration of `sample_buffer_len` to be just above the call to `initSampleBuffer()`.

There's still a duplicated hard-coded magic number (64) used herein.  I don't know what style is preferred (obviously, Iceman's opinion will rule), so here's two strawman examples:

<details><summary>Strawman Alpha</summary><p/>

```C
static const uint32_t sample_buffer_len = 64;
uint32_t actual_buffer_len = sample_buffer_len;
initSampleBuffer(&actual_buffer_len);
if (actual_buffer_len != sample_buffer_len) { //...
```
</details> 

<details><summary>Strawman Omega</summary><p/>

```C
static const uint32_t SAMPLE_BUFFER_LENGTH = 64;
uint32_t sample_buffer_len = SAMPLE_BUFFER_LENGTH;
initSampleBuffer(&sample_buffer_len);
if (SAMPLE_BUFFER_LENGTH != sample_buffer_len) { //...
```

</details>

armsrc/util.c
```diff
@@ -305,3 +305,11 @@ bool data_available(void) {
     return usb_poll_validate_length();
 #endif
 }
+
+bool data_available_fast(void) {
```
henrygab, line 309: Can clarity be improved as to meaning of this new function?

Those not really familiar with `data_available()` might see this new function and think it's a 100% replacement, only faster!  If so, a comment mentioning this is all that is needed.

If it's not a 100% replacement, consider comments that clarify:
* what is meant by "fast"
* are there invariants that are assumed by this function (e.g., what conditions must be true for this to be safe to call?)
* when would callers want to call the non-fast version instaed of this one?

client/src/cmddata.c
```diff
@@ -1707,7 +1708,7 @@ int CmdHpf(const char *Cmd) {
     CLIExecWithReturn(ctx, Cmd, argtable, true);
     CLIParserFree(ctx);
 
-    uint8_t bits[g_GraphTraceLen];
+    uint8_t *bits = malloc(g_GraphTraceLen);
```
henrygab, line 1711: reminder: **_all_** calls to malloc / calloc need to be checked for successful allocation.

henrygab (changes requested):
I reviewed all files EXCEPT `client/src/comms.c`.
Note that I'm just a random user ... so any comments from Iceman can totally overrule my own.  I only noticed a couple bugs (even if one was repeated) ... which is **_amazing_** for the size of this PR.

At a minimum, checking for and handling failed memory allocations is critical.  The rest of my comments are just fit and finish.

Well done!

client/src/cmdlf.c
wh201906: Sure.
The arm side will decompress and load the low frequency bitstream to the FPGA if it is not loaded. In this case, the arm side will send a `CMD_WTX` packet to tell the client to wait for a while. This happens at the first run of LF related commands after power-up the Proxmark3 hardware, or running LF related commands after HF related commands.
The problem is the client cannot detect packets once it goes into raw receive mode. And I cannot simply ignore the first few bytes because they could be the real samples.

client/src/cmdlf.c
wh201906: Let's continue in https://github.com/RfidResearchGroup/proxmark3/pull/2173#discussion_r1393737554

client/src/cmdlfhid.c
```diff
@@ -146,6 +147,7 @@ int demodHID(bool verbose) {
```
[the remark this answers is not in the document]
wh201906: Oh I missed that. Thanks

client/src/graph.h
```diff
@@ -42,7 +42,7 @@ int GetNrzClock(const char *str, bool verbose);
 int GetFskClock(const char *str, bool verbose);
 bool fskClocks(uint8_t *fc1, uint8_t *fc2, uint8_t *rf1, int *firstClockEdge);
 
-#define MAX_GRAPH_TRACE_LEN (40000 * 8)
+#define MAX_GRAPH_TRACE_LEN (40000 * 32)
```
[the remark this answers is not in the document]
wh201906: I guess this is the only limit for obtaining longer traces in the code. I got stack overflows when just increasing this number. And that's why I made commit 8b6a274e289cd96f16616de741f0d02fb10d1e6e to allocate related arrays on the heap.

client/src/cmdlf.c
wh201906: I think `lf_sample_payload_t payload = {0};` is good for it, right?

armsrc/lfsampling.c
wh201906: The magic number there can be `AT91C_EP_IN_SIZE`, but it's defined in `common_arm/usb_cdc.c` so I need to export it.

client/src/cmdlfio.c
```diff
@@ -66,10 +66,11 @@ static int CmdIOProxWatch(const char *Cmd) {
 int demodIOProx(bool verbose) {
     (void) verbose; // unused so far
     int idx = 0, retval = PM3_SUCCESS;
-    uint8_t bits[MAX_GRAPH_TRACE_LEN] = {0};
+    uint8_t *bits = calloc(MAX_GRAPH_TRACE_LEN, sizeof(uint8_t));
```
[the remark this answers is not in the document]
wh201906: Fixed

client/src/cmdlf.c
wh201906: Fixed

client/src/cmdlf.c
wh201906: Fixed

client/src/cmdlf.c
wh201906: I've put the information about when it occurs in this comment

armsrc/lfsampling.c
wh201906: I see. I've tested the code without `unlikely()` and the real-time sampling also works. I've removed them.
It's good to know these GCC extensions. Thank you for your guidance!

armsrc/lfsampling.c
wh201906: Fixed

armsrc/util.c
wh201906: I see. I've put the comments in the `armsrc/util.c`

client/src/cmddata.c
wh201906: I see. I've added the check to all of the memory allocation code in commit 8b6a274e289cd96f16616de741f0d02fb10d1e6e .
I did ignored the check when writing these code. Thanks!

iceman1001:
Gentle reminder that we use _calloc_  instead.   

Some external libs still uses malloc which we don't want to fiddle with.

wh201906:
I replaced declarations like this
`uint8_t bits[XXX]` -> `malloc()`
`uint8_t bits[XXX] = {0}` -> `calloc()`

And all of the allocations using `malloc()` is actually initialized immediately by `getFromGraphBuf()`, so I think it's not necessary to zero-initialize them first.
Anyway, it's up to you. Would you prefer to still use `calloc()` for them?

iceman1001:
I found that using calloc has helped us on different platforms,   so yes,  that is the prefered way.

wh201906:
I've replaced them

iceman1001:
Since I travelling,  I will not be able to look at this PR until I get back.

wh201906:
No problem, have a good time!

client/src/cmdlf.c
henrygab: That was an excellent explanation ... thank you!

iceman1001:
I would say it should be automatic..   
Default sample size or less,  should go the old way,    if a larger sample size,   it should go the alternative way. 
`lf read`   isn't like real time sniffing..  

Hold on with changing demodulation functions.   Otherwise this PR will be become too large.
Better to change that after we get this one merged

wh201906:
OK. Is there anything I need to do for improving this PR?

wh201906:
Any steps to reproduce the memory leak bugs?

iceman1001:
do different lf / data operations?

wh201906:
I got a segmentation fault when running `data modulation` over 132000 samples.
It should be fixed now.

iceman1001:
alright,
doing some more testing,    this text should be adapted to match new samples.

` "data samples -n 10000" `

```
static int CmdSamples(const char *Cmd) {

    CLIParserContext *ctx;
    CLIParserInit(&ctx, "data samples",
                  "Get raw samples for graph window (GraphBuffer) from device.\n"
                  "If 0, then get whole big buffer from device.",
                  "data samples\n"
                  "data samples -n 10000"
                 );
    void *argtable[] = {
        arg_param_begin,
        arg_int0("n", NULL, "<dec>", "num of samples (512 - 40000)"),
        arg_lit0("v", "verbose", "verbose output"),
        arg_param_end
    };
    ```

iceman1001:
Or...
  `data samples`   is a old style...

wh201906:
I'm not sure what I should do now...
It looks like the `data samples` is only for fetching existing data from the device memory, so I guess it has nothing to do with the real-time samples.
As for `lf search -u`, I just tested `lf search -1u` to use the samples from the graphbuffer(300000 samples), and it doesn't crash.

wh201906:
Thank you @iceman1001 and @henrygab for reviewing this PR and #2167

PR #2176, opened 2023-11-14
Fix pthreads on older gcc

phreakmonkey:
- Used $(SKIPPTHREAD) test result from client/CMakeLists.txt
- Tested on Debian Linux 11 (bullseye), build issue resolved

### DIFF
--- a/tools/mfkey/Makefile
+++ b/tools/mfkey/Makefile
@@ -9,6 +9,11 @@ INSTALLTOOLS = $(BINS)
 
 include ../../Makefile.host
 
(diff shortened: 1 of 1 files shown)


phreakmonkey:
(Someone with a newer gcc / distro should test this and make sure my conditional is valid / the build still works there too!)

iceman1001:
Thanks for the PR!

Now, two PRs ,  two different solutions.  Which to choose.  
Someone mentioned that both works.

SKIPTHREAD seems to be the right path considering its purpose.

PR #2178, opened 2023-11-16
Implement Avigilon 56bit format

hamptonmoore:
Implements the basics of the Avigilon 56 bit format https://accessit.acre-co.com/kb/article/531-avigilon-56-bit-card-format/


iceman1001:
Awesome!

PR #2179, opened 2023-11-16
Update mfc_default_keys.dic

JohnELester:
Found keys from two cards each from a different hotel.  That means they are probably diversified, right?  I see other keys in this dictionary already labelled as "diversified," so should I continue to submit found keys that might be diversified or only submit what I know are definitely *not* diversified keys?  I just want to help contribute to this dictionary if I can. Thanks for any tips.

### DIFF
--- a/client/dictionaries/mfc_default_keys.dic
+++ b/client/dictionaries/mfc_default_keys.dic
@@ -2288,3 +2288,10 @@ A71E80EA35E1
 EA0CA627FD06
 80BB8436024C
 5044068C5183
+#
+# Key from Hotel M Montreal (probably diversified)
+7E5E05866ED6
+661ABF99AFAD
+#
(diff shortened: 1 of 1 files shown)


iceman1001:
Well,   if you found same keys as a different person,  then the chances are higher that they are static.

Diversified keys is no useful for a generic dictionary,    it is only useful for you and a dedicated dictionary for your cards. 
but then you would have a key file.

Dont let any of this discourage your expression to share your findings.  We just need a way to make sure its not diversified and those I couldn't verify I put in a section in the end of the dictionary.

I am sure more of the keys in the dictionary is diversifed but verifying all keys is very hard and would need a much larger effort from people all over the world

JohnELester:
Thanks for the tips (and your YT videos are very educational too)!  I appreciate it and will continue to explore and share whatever I can.

Issue #765, opened 2020-06-09
fatal error: bzlib.h: No such file or directory


nico0481:
Hello,
Trying to compile last commit with:
git pull
make clean
make -j 3

I get an error: 
"src/cmdhfmfhard.c:26:10: fatal error: bzlib.h: No such file or directory #include <bzlib.h>"

What I'm doing wrong?

Thank you

Best regards

MadMike7777:
I successfully installed and ran Iceman's Proxmark3 program on an Ubuntu computer but it will not work on a Raspberry Pi.  I get the same fatal error.

doegox:
Please read the updated installation instructions. E.g. https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/Linux-Installation-Instructions.md
Basically you need to install bzip2 developer package, the name depends on the distro you're using.
libbz2-dev on Debian-alike.

PR #2183, opened 2023-11-19
Added common dependency troubleshooting

JaredStemper:
Ran into this issue during my installation process on pm3 easy. I hadn't seen the fix in the troubleshooting guide, but found others with this issue in the Github issues (#765) as well as the Discord, so I thought to add it into the troubleshooting guide to save user's some time searching.


iceman1001:
Thank you!

PR #2189, opened 2023-11-30
Create hf_mf_uid_downgrade.lua

evildaemond:
Adding hf_mf_uid_downgrade, which allows you to add a fc and cn to a UID of a Mifare Classic.


iceman1001:
Awesome!

PR #2192, opened 2023-12-02
Implement hf_unisniff, 14a/14b/15 sniffer with runtime selection and more save options

hazardousvoltage:
Based on existing hf_14/15sniff modules, this combines all 3 into one module.  Select mode at compile time or runtime. 
Offers options to create new files for each session or append to existing file.


iceman1001:
Cool,

I am not sure how the ASK sniffing.. looks like it doesnt do anything.   Maybe a changing name for a "no_operation" ?

hazardousvoltage:
Hi, sorry for the confusion, it's not ASK sniffing, it's "ask the user at runtime which protocol."  It does a prompt loop with the button and LEDs much like dankarmulti.

iceman1001:
I would change the "ASK"  to  "USER" or something else..     ASK mean something different in this world..

PR #2201, opened 2023-12-11
Update documentation

kitsunehunter:
Added documentation for HID downgrades and reader manager as per recommendation from 🧊
Updated note for .eml being deprecated.


iceman1001:
Awesome!

PR #2206, opened 2023-12-12
Update omnikey documentation 

kitsunehunter:
Not sure how links in forked repos work. Might need a edit.


iceman1001:
Cool!

The correct spelling is iCLASS  since its a product and that's how they spell it. 
In pm3 command its all small letters but all other text should be iCLASS.

and for linking,  I suggest you push the file upstream too.  

As a side note,  same goes with MIFARE  as in  `MIFARE Classic (correct)`  vs  `Mifare Classic (wrong)`

iceman1001:
Thanks!

PR #2209, opened 2023-12-14
Clear bss on bootloader start.

nvx:
This fixes a bug with the USB device serial number sometimes not working in bootloader mode.


iceman1001:
Awesome!

henrygab:
Nicely done!  That's a bug that requires deep understanding on compilation and how executables load / initialize.  Well done indeed!

nvx:
> Nicely done! That's a bug that requires deep understanding on compilation and how executables load / initialize. Well done indeed!

I was so confused for so long until it clicked what was wrong :D

PR #2211, opened 2023-12-22
Update mfc_default_keys.dic With more keys

PommyGit:
Just added a bunch of keys to the dictionary


iceman1001:
Nice,

Before any merging we need to sort out if these keys are confirmed static default keys.   Ie:  are these keys to be the same on multiple cards from the same system,    or are they unique on each card.

And from which system do these keys come from?

PommyGit:
These keys are from the Transperth Smartrider system. I generally always have zero success when using hf mf autopwn on such cards but since i started gathering the list using hf mf autopwn -f ./my-key-collection.dic is alwasys a resounding success. I hope that answers your question

iceman1001:
Good,  Lets see what metrodroid project says,    https://github.com/metrodroid/metrodroid/wiki/SmartRider#key-diversification

They concluded that there are some static keys and some diversified.

Now which ones in your list fits their blocks for static keys?   Since those we want in the default key list,

iceman1001:
You also have lots of duplicates in your keys.  That needs to be addressed too.

PommyGit:
Strange how I unique this in bash and this was the output. Sorry for the sloppiness. I'm AFK at the Mo but will address these issues asap including the link to the static keys that @iceman1001  posted.  Sorry guys.

iceman1001:
I will close this one,   since I pushed @s00se  PR and his post of unique keys here.

@PommyGit  you can verify that short list of keys against the link and make a new PR if needed.

Thanks!

iceman1001:
and no... that wasn't unique..  I removed the duplicates...

PR #2215, opened 2023-12-26
Check TCP connection state

wh201906:
The client can detect if the TCP connection is lost and show `offline` in the command line.

图片


iceman1001:
cool!

Does that work on UDP too?

wh201906:
No it doesn't work for UDP.

PR #2217, opened 2023-12-27
SPIFFS: implement dumping from SPIFFS to client trace buffer

hazardousvoltage:
Sometimes one has a bunch of traces in SPIFFS after a long day of standalone work, and wants to look at them and decide what to save.  Adds '-t' option to mem spiffs dump, to download the data into client trace buffer instead of creating a local file which one must then 'trace load'

Not sure I like how I implemented this, or whether it's useful to anyone else.

### DIFF
--- a/client/src/cmdflashmemspiffs.c
+++ b/client/src/cmdflashmemspiffs.c
@@ -16,6 +16,7 @@
 // Proxmark3 RDV40 Flash memory commands
 //-----------------------------------------------------------------------------
 #include "cmdflashmemspiffs.h"
+#include "cmdtrace.h"
 #include <ctype.h>
 #include "cmdparser.h"  // command_t
 #include "pmflash.h"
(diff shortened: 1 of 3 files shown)


iceman1001:
Cool,

Only thing I see is that -t vs -s  is excluding,    when you can use "-t"  and "-s" at the same time.   Save and load graphbuff.

Just check if there was slen and run that part of the code too.

hazardousvoltage:
Hi, thanks for the comments..

-t does not exclude -s, it's excluding -d

Perhaps better, if -t is used, dump only to tracebuffer, but if used with -d, also save to file?

iceman1001:
yeah,   

so to make it perfectly clear how it should work, 

-t  fill trace
-d  save to file

if both params,   do both.  Meaning we can both save to file and load trace.
if only one,  it does what that param intends.

iceman1001:
Ignore my comments,      this way works well.

Thanks!

PR #2225, opened 2024-01-01
Some fixes for the reconnect feature

wh201906:
1. Add `msleep()` to reduce the CPU usage
2. Disable printing error information when calling `uart_open()` in `OpenProxmarkSilent()`

I'm still finding the reason of crash when running `hw connect` to a different port in the offline mode, so I created this as a draft.

### DIFF
--- a/client/src/comms.c
+++ b/client/src/comms.c
@@ -699,7 +699,7 @@ size_t GetCommunicationRawReceiveNum(void) {
 
 bool OpenProxmarkSilent(pm3_device_t **dev, const char *port, uint32_t speed) {
 
-    sp = uart_open(port, speed);
+    sp = uart_open(port, speed, true);
 
     // check result of uart opening
     if (sp == INVALID_SERIAL_PORT) {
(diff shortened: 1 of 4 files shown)


iceman1001:
different port means we can't auto reconnect.  it uses the "serial port"  that was used when first starting the client.
save in the global connection variable `g_conn`

wh201906:
I know, but I thinks it should be possible to connect to a different port by running `hw connect` manually when the original connection is down

iceman1001:
yes,  
   then we would need a way to break the reconnect thread.  ie detect if "pm3 is present and online"

wh201906:
> yes, then we would need a way to break the reconnect thread. ie detect if "pm3 is present and online"

Yes. I'm working on this.

wh201906:
Maybe we should merge this PR first, because the high CPU usage in reconnecting and the infinite `error: Could not connect` really bring a bad user experience, while the `hw connect` to different port is just a feature.

iceman1001:
yes,  the sleep is needed.  

_merged_   :)

iceman1001:
Something tells me the GENERIC build running on devices with no flash,  has the different serial port issue.

wh201906:
> Something tells me the GENERIC build running on devices with no flash, has the different serial port issue.

Do you have the steps to reproduce it?

iceman1001:
I suspect a non-flash enabled proxmark3 device,   will have different serial port enumeration when going from  normal operation - bootloader - flash - normal operation...

wh201906:
Yeah I got the same bug on my EASY before. But I guess it won't happen from the normal operation -> unplug -> plug and reconnect

PR #2226, opened 2024-01-01
hf 14a apdu now uses the FWI and SGFI values from the ATS to determine an appropriate timeout.

nvx:
This fixes issues when using the command against cards that indicate really slow times, such as card emulation by hf_cardhopper.


iceman1001:
Excellent!

PR #2227, opened 2024-01-01
Add hint for iClass in `hf search`

bettse:
It was:

```
[usb] pm3 --> hf search
 🕙  Searching for iCLASS / PicoPass tag...   
[+] iCLASS / Picopass CSN: CE F0 43 01 F7 FF 12 E0 

[+] Valid iCLASS tag / PicoPass tag found
```

And now it is 

```
[usb] pm3 --> hf search
 🕗  Searching for iCLASS / PicoPass tag...   
[+] iCLASS / Picopass CSN: CE F0 43 01 F7 FF 12 E0 

[+] Valid iCLASS tag / PicoPass tag found

[?] Hint: try `hf iclass` commands
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Added hint for iClass to `hf search` (@bettse)
  - Changed `hf 14a apdu` - It now uses the FWI and SGFI values from the ATS to determine an appropriate timeout (@nvx)
  - Added a thread to check when device comes online again. It will connect and update prompt (@iceman1001)
  - Changed CLI offline prompt - replaces the old prompt when offline is detected (@iceman100)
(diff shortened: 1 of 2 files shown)


iceman1001:
This is divider...

either we have hints for all positive identifications,     and should they come direct after each one or all hint goes in the bottom like the rest of the client does. 
or we don't have hints because its one type per identification compared with it is not "unknown" positives like for `hf 14a info`   which can be a range of types.

hm....

Lets go with hints but I want them to be all in the end,     like we do in other places in the client.

bettse:
> Lets go with hints but I want them to be all in the end, like we do in other places in the client.

I've not been in this area before, so I'm uncertain of the implication.  Does that mean I should setup a variable that saves that it was iClass, and then create a conditional after all the checks that would output the hint?

iceman1001:
yes  that is correct.  And add hints for the other checks too,  please.

bettse:
> And add hints for the other checks too, please.

That would be beyond the scope of what I'm looking to achieve: is that a requirement for the PR?

bettse:
😮  I was working on the change to put the hint at the end, I just didn't want to introduce the other hints.

iceman1001:
yeah,  I understand.

PR #2229, opened 2024-01-03
Ignore button jitters in bootloader mode

wh201906:
Commit 51e4df533e79399fef1672325d2ec57fdcc27986 changed the button behavior, but it doesn't work on my Proxmark3 Easy. I think it's caused by the jitter (bounce) when pressing/releasing the button.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Fixed bootloader - Ignore jitters when pressing the button (@wh201906)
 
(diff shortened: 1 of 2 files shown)


wh201906:
I guess RDV4 has a debouncing capacitor so it's a little bit harder to find this bug.

iceman1001:
If this is a hardware design issue,   will it be on all button pressed calls?   if so maybe we would need to fix it,  and guard it with DEFINES.

wh201906:
> If this is a hardware design issue, will it be on all button pressed calls?

I‘m not sure but I guess it will. The `BUTTON_PRESS()` only checks the register state. It cannot handle the jitter.

> and guard it with DEFINES

The software solution of debouncing is compatible with the hardware solution, so I guess it's not necessary.

iceman1001:
You are missing the point,  
all calls to BUTTON_PRESS()  ,   will it suffer from the jitter in the normal device side application?
Your fix is for bootloader only.

PR #2230, opened 2024-01-03
Renamed bool to follow guideline, and a typo

unknown10777:


iceman1001:
we don't use capital letters for bools.

unknown10777:
sorry, I'll correct that in next pr

PR #2231, opened 2024-01-04
Add alt Disney AID for desfire

mazodude:
Noticed the AID for the MagicBand didn't match on my "Key to the world" card from Disney Cruise Line.

Not sure if it is only used on DCL though


iceman1001:
Awesome!

PR #2232, opened 2024-01-04
Rick423n patch 1

rick423n:


iceman1001:
Ok,   so what does this one do?   Besides build on MSBUILD?

1. microsoft/setup-msbuild@v1.0.2    isn't the last version.
2. no make clean,  make check,   and the other normal builds we do
    https://github.com/RfidResearchGroup/proxmark3/blob/master/.github/workflows/windows.yml

iceman1001:
No to mention it fails...

rick423n:
I don’t know what I’m doing still learning 😢

iceman1001:
Why are you making a PR then?

PR #2242, opened 2024-01-09
Update Troubleshooting.md

francis2054:
Added missed entry to linktree at top


francis2054:
Does a small correction like this belong in changelog as well? For future reference, thank you

iceman1001:
I wouldn't add a changelog,   but its up to you.

Issue #2244, opened 2024-01-13
hf 15 raw -k option (keep field on) is broken/removed


Chris-P-Young:
***Description***
The hf 15 cmd has the option to keep RF field on between sending raw commands. This is the -k option.
At some point, this feature has been removed/deliberately broken for some reason.
The option is still present in the command help and is still supported as an option by the command parser.
However, the DirectTag15693Command in iso15693.c now explicitly turns off the field at the end of the command. It also unconditionally resets the field at the start of the command.

***Why is this important?***
Raw mode is useful to explore tag behaviour and custom commands.
Certain tag commands need to remain powered between commands. This means the RF field should stay on.
An example is the NXP ICODE SLIX-L which requires that the tag remains powered during the privacy disable sequence (get random number followed by set password command).
We already have instruction support for SLIX privacy enable/disable, but we should also be able to achieve the same thing using raw commands as Proxmark is meant to be a learning and investigative tool.

Verified with NXP ICODE SLIX-L tag and RRG Field detector fob.
 
***To reproduce***
Using NXP ICODE SLIX-L tag as an example.

Issue raw command GET RANDOM NUMBER, keep field on
`hf 15 raw -k -c -2 -d 02b204`

Tag will respond with 5 octets which includes the random number.

Issue raw command SET PASSWORD (for privacy password)
`hf 15 raw -c -2 -d 02B30404xxxxxxxx (where xxxxxxxx is XOR'D password)`

Client will return 
`! command failed`

The -k option is processed by the client but ignored. The field is removed between raw commands which resets the tag and invalidates the random number. Therefore the set password command will always fail.

***Proposed Solution***
A PR has already been filed (PR #1636) to add a fix to handle the -k option properly.
However, this PR is based on code using the older return code method and has not been picked up for a merge.
I will update and file a new PR.

A second issue is the handling of the timing for Iso15693 raw commands.
The DirectTag15693Command contains some code that parses the raw command string passed to it and uses this to determine whether the command should be transmitted with reader timing or "write alike" timing which has a much longer timeout.
For commands not recognised by this code, the standard (shorter) reader timeout is used which will cause the tag reply to be missed.
Currently this is the case when sending the SLIX-L "set password" and "enable privacy" commands. Both of these commands have "write alike" timing and require the longer timeout.
Without this,  the tag response is missed and PM3 client returns "command failed"

A nasty fix is to add the two extra commands to the switch/case structure to ensure that write alike timing is used.

[...]

PR #2245, opened 2024-01-13
Fix for issue #2244 (hf15 raw -k command)

Chris-P-Young:
***Description***
This PR is submitted to address issue #2244 (hf 15 raw -k option (keep field on) is broken/removed)

Currently, the -k option for hf 15 raw command does not keep the RF field turned on.
The option is parsed, but the DirectTag15693Command in iso15693.c explicitly turns off the field at the end of the command. It also unconditionally resets the field at the start of the command.

***Why is this important?***
This is important for tags that have non standard Iso15693 commands that require the tag to remain powered between commands. An example is the NXP ICODE SLIX-L which requires that the tag remains powered during the privacy disable sequence (get random number followed by set password command). We already have instruction support for SLIX privacy enable/disable, but we should also be able to achieve the same thing using raw commands as Proxmark is meant to be a learning and investigative tool.

***Solution***
This PR updates the code provided in PR #1636 to be compatible with the new code base.
This PR also includes a minor modification to enable "write alike" timing for two ICODE SLIX-L commands (set password / enable privacy". Without this, the timeout is too short and the tag response is missed resulting in the client returning a "command failed" response.

### DIFF
--- a/armsrc/iso15693.c
+++ b/armsrc/iso15693.c
@@ -2434,9 +2434,12 @@ void BruteforceIso15693Afi(uint32_t speed) {
 
 // Allows to directly send commands to the tag via the client
 // OBS:  doesn't turn off rf field afterwards.
-void DirectTag15693Command(uint32_t datalen, uint32_t speed, uint32_t recv, uint8_t *data) {
+void DirectTag15693Command(uint32_t datalen, uint32_t speed, uint32_t flags, uint8_t *data) {
 
     LED_A_ON();
+	
+    bool recv = flags & 1 ? true : false;
+    bool field_already_on = flags & 2 ? true : false;
 
     uint8_t recvbuf[ISO15693_MAX_RESPONSE_LENGTH];
     uint16_t timeout;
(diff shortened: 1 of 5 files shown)


iceman1001:
Great,    
The sending of flags is not good all.  It should use our new prefered way of sending a struct.
Another  suggestion as below,   it would remove the need for a new global var and upside we can prep `hf 15` for APDU support.

add ISODEP_NFCV

```
typedef enum {
    ISODEP_INACTIVE = 0,
    ISODEP_NFCA,
    ISODEP_NFCB,
} isodep_state_t;


void SetISODEPState(isodep_state_t state) {
    isodep_state = state;
    if (APDULogging) {
        PrintAndLogEx(SUCCESS, "Setting ISODEP -> %s%s%s"
                      , isodep_state == ISODEP_INACTIVE ? "inactive" : ""
                      , isodep_state == ISODEP_NFCA ? _GREEN_("NFC-A") : ""
                      , isodep_state == ISODEP_NFCB ? _GREEN_("NFC-B") : ""
                     );
    }
}
```

Chris-P-Young:
Hi Iceman,
Many thanks for your feedback.
Unfortunately, I'm not sure that my coding skills are up to the job of implementing something that will pass review cleanly.
Also, I'm not familiar with APDU and so I will have to leave this to the experts. I raised the issue reports and PR because I wanted to highlight the issues. My initial concern was that the -k flag did not perform as documented. However using a hard coded "command" list to determine response time for a "raw mode" seems an even worse issue.

iceman1001:
No worries,
I have refactored the 15 raw command,    I will need you to test things out.

iceman1001:
try https://github.com/RfidResearchGroup/proxmark3/commit/8d0b41a911ca308eac990cc6bdb21d3fd073ef1c  and see if this works better.

PR #2249, opened 2024-01-15
Multiple hf_cardhopper improvements

nvx:
Allow button presses to break
Handle non-zero CID from reader by relaying RATS to the card and improving PPS and WTX handling More reliably cook ATS values
Ignore packets that look like they're coming from the client (NG packets)


iceman1001:
Nice improvements!

PR #2251, opened 2024-01-16
Add smart bruteforce mode to Mifare Classic and EM4x50

phaseloop:
This is update to bruteforce engine introduced in May 2023. 

This PR adds so called "smart" bruteforce which tries various non-random patterns  which sometimes are used as keys in systems like transport, car wash, vending machines, etc due to poor key management or hardcoded patterns in management software. 

During testing it recovered keys from few problematic Mifare Classic keys that were not recovered using autopwn due to problems with nested attack.

Smart bruteforce is enabled for em4x50 cards and Mifare Classic


client/src/cmdhfmf.c
```diff
@@ -3369,6 +3370,216 @@ static int CmdHF14AMfChk_fast(const char *Cmd) {
...
+        PrintAndLogEx(NORMAL, "");
+        PrintAndLogEx(SUCCESS, _GREEN_("found keys:"));
+
+        printKeyTable(sectorsCnt, e_sector);
+
+        if (use_flashmemory && found_keys == (sectorsCnt << 1)) {
+            PrintAndLogEx(SUCCESS, "Card dumped as well. run " _YELLOW_("`%s %c`"),
+                          "hf mf esave",
+                          GetFormatFromSector(sectorsCnt)
+                         );
```
iceman1001, line 3537: where do you perform the dump?

doc/commands.md
```diff
@@ -754,16 +755,6 @@ Check column "offline" for their availability.
...
-
- { Waveshare NFC ePaper...             }
-
-|command                  |offline |description
-|-------                  |------- |-----------
-|`hf waveshare help      `|Y       |`This help`
-|`hf waveshare loadbmp   `|N       |`Load BMP file to Waveshare NFC ePaper`
-
-
 ### hf xerox
```
iceman1001, line 758: Are your PR based on latest?
```
doc/commands.md
doc/commands.json
```

iceman1001:
Nice!
If  you sort out the conflicting files.  
and then you have a text about dumping from emulator memory,  but I don't see where you trigger the card to be dumped

iceman1001:
Some of your default keys are already in the dictionary.   Remove the duplicates.

client/src/cmdhfmf.c
phaseloop: Later in the file - here:
https://github.com/RfidResearchGroup/proxmark3/pull/2251/files#diff-c5d5158583a2c910bfb303c9489237d06a39e0472c738e1635c07b90b2fd7e8dR3568

I'll move this down to print after the dump.

iceman1001: ok,   fchk does it on device side if all keys are found. 
so the text is ok.

iceman1001: well,  it only does it if the device has a flash memory (RDV4 style)

And now you removed the  "mem"  param,   making the text invalid.

phaseloop: So:

--mem option runs dictionary check from a file stored in flash. In bruteforce mode having --mem option has no sense.

We have those two code blocks in fast check:

```
 if (use_flashmemory && found_keys == (sectorsCnt << 1)) {
            PrintAndLogEx(SUCCESS, "Card dumped as well. run " _YELLOW_("`%s %c`"),
                          "hf mf esave",
                          GetFormatFromSector(sectorsCnt)
                         );
        }

 if (createDumpFile) {

            char *fptr = GenerateFilename("hf-mf-", "-key.bin");
            if (createMfcKeyDump(fptr, sectorsCnt, e_sector) != PM3_SUCCESS) {
                PrintAndLogEx(ERR, "Failed to save keys to file");
            }
            free(fptr);
        }
```

While `createMfcKeyDump` saves files on a PC and nowhere in the flash, printing a message "Card dumped" only when we use `--mem` option IMHO makes no sense.

client/src/cmdhfmf.c
```diff
@@ -3369,6 +3370,216 @@
...
+                num_to_bytes(generator_key, MIFARE_KEY_SIZE, keyBlock + (i * MIFARE_KEY_SIZE));
+                keycnt++;
+
+                if (smart_mode_stage != bctx.smart_mode_stage) {
+                    smart_mode_stage = bctx.smart_mode_stage;
+                    PrintAndLogEx(INFO, "Running bruteforce stage %d", smart_mode_stage);
+
+                    if (msclock() - t1 > 0 && keys_checked > 0) {
+                        PrintAndLogEx(INFO, "Current cracking speed (keys/s): %d",
+                                      keys_checked / ((msclock() - t1) / 1000));
```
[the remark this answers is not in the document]
phaseloop: should be fixed now

client/src/cmdhfmf.c
```diff
@@ -3369,6 +3370,216 @@
...
+            firstChunk = false;
+
+        if (ret == PM3_SUCCESS || ret == 2)
+            goto out;
+
+    }
+
+out:
+    PrintAndLogEx(INFO, "Time in brute mode: " _YELLOW_("%.1fs") "\n", (float)((msclock() - t0) / 1000.0));
+    PrintAndLogEx(INFO, "Total keys checked: " _YELLOW_("%u") "\n", total_keys_checked);
```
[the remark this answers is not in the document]
phaseloop: should be fixed now

client/src/cmdhfmf.c
iceman1001: Its more of an "hidden" extra, if you use fchk with mem.    It will on device side dump the card. 

So  in brute,   you don't use mem and you don't get the hidden extra.   So you can just take away the "card dump" text. 
it is of no use there.

phaseloop: Ok, I was wrong and messed up key dump with full card dump. Removed this message log.

iceman1001:
we tend to use these kinds of defines in string fomatters
```
PRIu64  
PRIu32
```

iceman1001:
Thanks!

PR #2253, opened 2024-01-16
Small Xerox changes, finding correct partnumber additional information

JeroenSteen:
The partnumber additional information was incorrect, because of strncmp seeing "-1" suffix as pattern. It now shows correct information, based on the partnumber example "006R01530". The pattern of partnumbers can be different, so we need to check what the variations can be. I also wonder what "-1" suffix means.

It would be nice if all block data would be placed at the top of the code, to prevent searching.

### DIFF
--- a/client/src/cmdhfxerox.c
+++ b/client/src/cmdhfxerox.c
@@ -469,53 +469,57 @@ static void xerox_print_footer(void) {
 
 // structure and database for uid -> tagtype lookups
 typedef struct {
-    const char *color;
+    const char *color; // cyan, magenta, gold, silver, clear, white, fluo
     const char *partnumber;
-    const char *region;
-    const char *ms;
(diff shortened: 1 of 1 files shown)


JeroenSteen:
We can also look for drum partnumbers, of all the printer models. Untill now I listed only two drums in my gist for "Fuji Xerox Docucentre 236/286/336, Fuji Xerox DocuCentre-II 2005/2055/3005, Fuji Xerox DocuCentre-III 2007/3007 (Mono) " see: https://gist.github.com/JeroenSteen/4b45886b8d87fa0530af9b0364e6b277

For "DCP 550/560/570/C60/C70, PrimeLink C9065/C9070, WC 7965/7975 (Color)" color is 013R00664 and black is 013R00663.

PR #2258, opened 2024-01-18
Show statistical data after tuning

wh201906:
and fixed a bug in `print_progress()`

### DIFF
--- a/client/src/cmdhf.c
+++ b/client/src/cmdhf.c
@@ -288,6 +288,7 @@ int CmdHFTune(const char *Cmd) {
         arg_lit0(NULL, "bar", "bar style"),
         arg_lit0(NULL, "mix", "mixed style"),
         arg_lit0(NULL, "value", "values style"),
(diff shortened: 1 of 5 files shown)


client/src/cmdhf.c
```diff
@@ -288,13 +288,15 @@ int CmdHFTune(const char *Cmd) {
         arg_lit0(NULL, "bar", "bar style"),
         arg_lit0(NULL, "mix", "mixed style"),
         arg_lit0(NULL, "value", "values style"),
+        arg_lit0("s", "stat", "show statistical data after tuning"),
```
iceman1001, line 291: Use NULL,  avoid "-s" since it is reserved for other purposes.

iceman1001:
Nice,

I think you can use  "verbose"   instead of a custom param.    It would fit in the general look and feel of the client.

wh201906:
I don't know how to use the `verbose`. Should I name the option as `--verbose`?

PR #2262, opened 2024-01-21
Made function printlog_xerox_info, for less duplicate code

JeroenSteen:
Made function printlog_xerox_info, for less duplicate code in CmdHFXeroxView and CmdHFXeroxInfo.


iceman1001:
If you pulled latest before,  you would have found my changes already. 

```
git pull
```

PR #2263, opened 2024-01-22
Adding processor flash memory reading, viewing and writing to file.

martian:
Reading memory works when the device is running either osimage or bootloader.

- New memory reading command in osimage and bootloader.
- Extended 'hw readmem' command with length parameter, file writing and hex viewer.
- Introduced '--dumpmem' option to proxmark3 executable to support dumping from bootloader.

Simple interactive examples:
- `hw readmem -f flashdump`
- `hw readmem -l 1024`

CLI example:
- `./pm3 --dumpmem flashdump.bin`

Reading from arbitrary memory ranges can be unlocked using the 'raw' option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Added `--dumpmem` to proxmark3 client for memory dumping to file (@martian01010)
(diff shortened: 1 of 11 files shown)


iceman1001:
Interesting,    a very nice PR.   Clean code.
I have only one issue,  and that is we use fileutils.c  to do all file operations.   There is a long effort to unifiy/refactor all file operations into it. 

Would you mind telling me what the use case for this?   How and when will it be used?

martian:
Thanks. I will have a look at the file operations.

My initial use case was an investigation into what happened with my somewhat-faulty chinese pm3 easy 512KB, where write errors occur in the top half of flash. _Except_ when I overwrite with FF first, it turns out.

I thought being able to dump flash, or read any memory or HW register might come in useful for others in debugging firmware related issues.

PR #2265, opened 2024-01-23
Readmem small fixes

martian:
Some minor cleanup in the new readmem functions.
- number entry should be unsigned and allow hexadecimal
- printing filename to the console twice
- corner case at the end of adressable memory


iceman1001:
Then we all set.   
Its merged

PR #2267, opened 2024-01-23
Fix hf sniff

nvx:
This was broken in commit 17ab86c52 as the forced rounding up of the size to 4-byte alignment in BigBuf_malloc made the size check possibly larger than the buffer size as the check was always +3 on the requested size rather than the rounded size. This was made worse by BigBuf_max_traceLen not taking into account alignment either and the alignmentn check in hfsnoop.c checking to 2 byte alignment instead of 4 byte alignment.

The alignment size check now checks the size after alignment rounding, and BigBuf_max_traceLen takes into account alignment losses too reducing the need for BigBuf consumers to have to care about alignment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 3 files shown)


nvx:
As a side effect this may fix other unrelated uses of BigBuf too (I never tested lf sniff but I wouldn't be surprised if it had the same issue)

iceman1001:
Great catch,

Lets see if its fixing why `hf emrtd`  commands fails.

PR #2269, opened 2024-01-24
iso15 sim & sniff improvments & HF_15SIM standalone

lnv42:
This includes:
- a new `HF_15SIM` standalone mode that dump then simulate iso15 tags
- rework of the iso15 simulator with lot of improvements (more commands answered, more feature supported, more tag info stored in eml memory/dumps (AFI, DSFID, IC, lock status...)  (I made required adaptations for this in dump/esave/eload/eview)
- Some improvement to the hf 15 sniffer code to support low/noisy signal (including a FPGA threshold modification for the iso15 FPGA image)

/!\ Last time I PRed a self compiled FPGA image: it creates a weird bug we lost time sorting out. So it's probably safer if you recompile the FPGA image yourself and commit the recompiled FPGA image(s) right after the merge. /!\


include/iso15.h
```diff
@@ -44,4 +44,31 @@ typedef struct {
     uint8_t raw[];      // First byte in raw,  raw[0] is ISO15693 protocol flag byte
 } PACKED iso15_raw_cmd_t;
 
+#define ISO15693_TAG_MAX_PAGES 64 // in page
+#define ISO15693_TAG_MAX_SIZE 2048 // in byte (64 pages of 256 bits)
```
iceman1001, line 48: 2048 bytes.
2046/64 = 32bytes.    Which blocks are 32 bytes long?!?

iceman1001:
Wow,  this is a monster one.

Standalone,   fpga,   device side...   Would have been easier if this was in smaller PR's  

You did add a fpga bit image btw,    Intentional?

iceman1001:
I will merge anyway :)

PR #2273, opened 2024-01-25
iso15: provide support to load old JSON dump format into iso15_tag_t

lnv42:
/!\ I don't have any old JSON dump file with me, so I did not test much... Sorry for that /!\


iceman1001:
You have some string formatter issue..

PR #2277, opened 2024-01-26
Add support for magic wakeups in GDM/USCUID config commands

nvx:
Changed hf mf gdmcfg/gdmsetcfg commands to support Gen1a and GDM Alt magic wakeups

This was implemented with a new pair of RPCs CMD_HF_MIFARE_READBL_EX and CMD_HF_MIFARE_WRITEBL_EX
these RPCs support all combinations of read/write commands, wakeup, and auth options so
in time can replace the other MFC read/write commands too reduce armsrc code size
and complexity.

Also added config parsing for the gdm cfg block when reading with hf mf gdmcfg and
explicitly with hf mf gdmparsecfg.

image

Also fixes some low hanging differences when running make style on Windows vs Linux, line endings should now always be LF and files should always contain a trailing newline. There's still some other minor annoyances but they're slightly more annoying to fix.


iceman1001:
Nice addition of the wakeup / cfg!

I do notice that the style fixes should have been broken out in a separate PR.

PR #2281, opened 2024-01-29
Fix style error in `udev` rules.

henrygab:
As called out by running `udevadm verify`,
the existing rules violate the requirement
that sections are separated by commas,
(not whitespace).


iceman1001:
hm,  interesting,   my udevadm doesnt have a verify..
For which OS is this?

henrygab:
relative new option.  Here's some related links:
https://github.com/systemd/systemd/pull/26698
https://github.com/systemd/systemd/pull/27871
https://github.com/systemd/systemd/pull/27832

Don't recall which distribution had it; My Ubuntu 20.04 does not appear to have that option, however.

iceman1001:
hm,  I guess the `,` is also a new requirement.   We had the space for years.

PR #2291, opened 2024-02-16
Properly handle "hf mf dump" errors

gsingh93:
I was confused why `hf mf dump` wasn't showing any errors but the dump file was all zeros:
```
[usb|script] pm3 --> hf mf dump
[+] time: 0 seconds

[+] saved 1024 bytes to binary file `/Users/gsgx/hf-mf-D6E6ACDF-dump-001.bin`
[+] saved to json file `/Users/gsgx/hf-mf-D6E6ACDF-dump-001.json`
```

For some reason `PM3_SUCCESS` was being returned from this error path, and the log message was using the `DEBUG` log level. After this PR, it's more clear an error occured:

```
[usb|script] pm3 --> hf mf dump
[-]  iso14443a card select failed
```

### DIFF
--- a/client/src/cmdhfmf.c
+++ b/client/src/cmdhfmf.c
@@ -573,13 +573,13 @@ static int mfc_read_tag(iso14a_card_select_t *card, uint8_t *carddata, uint8_t n
     SendCommandMIX(CMD_HF_ISO14443A_READER, ISO14A_CONNECT, 0, 0, NULL, 0);
     PacketResponseNG resp;
     if (WaitForResponseTimeout(CMD_ACK, &resp, 1500) == false) {
-        PrintAndLogEx(FAILED, "iso14443a card select timeout");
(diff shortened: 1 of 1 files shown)


iceman1001:
Its more of a client design choice.   Not just one command change like you suggested. 

A) Should all commands give a textual feed back if it fails ? 
or 
B) Should it not give out any output and only give output when successful. 


I tend today to prefer B,    so I am making the client act unified towards that way.

gsingh93:
@iceman1001 in that case, I've updated the PR to only change the return value, which will make the command produce no output.

I do prefer option A though. If you prefer option B, is there ever any reason to use "FAILED" or "ERR" log messages? I assume you have some criteria for determining when it's appropriate to use them?

iceman1001:
I am not saying that the pm3 client is a marvelous wonder of unified thought through design concepts.

gsingh93:
@iceman1001 no worries 😄 let me know if the current version of the PR looks good.

PR #2299, opened 2024-02-18
Added led order parameter (for Proxmark3 Easy board)

francesco-scar:
Added parameter `LED_ORDER` in Makefile.platform to correct led GPIO on Proxmark3 Easy board.

When setting `LED_ORDER=PM3EASY`, the LEDs get configured in the right order for this board, allowing correct binary counting of standalone modes and nice animations (SpinUp and SpinDown).

### DIFF
--- a/Makefile.platform.sample
+++ b/Makefile.platform.sample
@@ -10,6 +10,10 @@ PLATFORM=PM3RDV4
 #PLATFORM_EXTRAS=BTADDON FLASH
 #STANDALONE=LF_SAMYRUN
 
+# Uncomment the line below to set the correct LED order on board Proxmark3 Easy
+# Only available with PLATFORM=PM3GENERIC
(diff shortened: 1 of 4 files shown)


iceman1001:
No sure why you swap in two places. 
Wouldn't  armsrc/util.c  be enough to swap the LED B/ D?     

and ignore the  the pin assignments ?

francesco-scar:
I changed the definitions of the colors in armsrc/util.c just for completeness to match the colors on that specific board clone, where they apparently changed both the order and the colors (although they are not used anywhere).

I could have changed the "weights" of each LED in armsrc/util.h lines 45-48 to fix the order during the counting, but not the spin animations.

I think that this is a cleaner solution, so LED_A actually is defined as the LED marked A on the board, and so on

iceman1001:
Alright!
Hope it works on the Pm3 easy.

PR #2300, opened 2024-02-21
Added 14b restore, modified 14b reader,wrbl

Sonic803:
Added the function `hf 14b restore` to restore dump to 14b card.
Modified `hf 14b reader` to make it show the plot like the reader for 14a does.
Modified `hf 14b wrbl` because previously if no flag for the tag size was given it worked even out-of-range but printed that was working with a SRI512 tag.

### DIFF
--- a/client/src/cmdhf14b.c
+++ b/client/src/cmdhf14b.c
@@ -2726,7 +2726,7 @@ int readHF14B(bool loop, bool verbose) {
 plot:
         res = handle_hf_plot();
         if (res != PM3_SUCCESS)
-            break;
+            PrintAndLogEx(DEBUG, "plot failed");
 
     } while (loop && kbd_enter_pressed() == false);
 
--- a/client/src/cmdhf14b.c
+++ b/client/src/cmdhf14b.c
@@ -1646,7 +1646,7 @@ static int CmdHF14BSriWrbl(const char *Cmd) {
                      );
     }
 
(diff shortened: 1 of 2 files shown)


client/src/cmdhf14b.c
```diff
@@ -2573,44 +2694,39 @@ int infoHF14B(bool verbose, bool do_aid_search) {
...
         found |= HF14B_other_reader(verbose);
-        if (found && loop)
-            continue;
-        else if (found)
-            return PM3_SUCCESS;
+        if (found)
+            goto plot;
+plot:
+        res = handle_hf_plot();
+        if (res != PM3_SUCCESS)
```
iceman1001, line 2728: this will break the continues loop if  the hf_plot fails

iceman1001:
Nice!,

This looks like a PR that should be divided into two.  

try use the blocksize define everywhere.

client/src/cmdhf14b.c
Sonic803: I think this is the same behaviour that the 14a reader plot has

iceman1001: then it has the same problem :)

Sonic803: what should it's behaviour be then, continue as if nothing happened or stop trying to plot?

iceman1001:
Thanks!

PR #2302, opened 2024-02-23
WIP: Magic card docs editing

mishamyte:
Some issues were found in the current doc.

Changes in that PR: 
- Updated commands for magic cards (from `hf 14a info` to `hf mf info`)
- Added the usage of commands which were adjusted for GDM (like `hf mf gdmcfg`)
- Added parsed config for well-knows species of `USCUID` tags
- Fixed typos (like `hf mf raw`)
- Links fixes in header (OTP 2.0 link was broken + one "dead" link was present)
- Made a soft linting of MD for style consistency and readability in the editor mode
- Updated info about UL-5, based on the official communication from the manufacturer

The main work is described above. If you have any comments/suggestions - pls, let me know.

### DIFF
--- a/doc/cloner_notes.md
+++ b/doc/cloner_notes.md
@@ -74,7 +74,7 @@ Standard password is normally (for T55xx):  AA55BBBB
 
 ```
 lf t55xx write -b 1 -d E0150A48 --pg1
-If t55xx write -b 2 -d 2D782308 --pg1
+lf t55xx write -b 2 -d 2D782308 --pg1
 ```
 
 # Sniffing the comms
--- a/doc/magic_cards_notes.md
+++ b/doc/magic_cards_notes.md
@@ -318,9 +318,9 @@ UID 7b:
 ^[Top](#top)
 
 ```
-hf 14a info
+hf mf info
 ...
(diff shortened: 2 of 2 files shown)


iceman1001:
Interesting,

I would say you remove too much information.  This document isn't just about latest commands,  but also in documenting the protocol used for magic cards.

mishamyte:
No problem. 

Are you talking about the USCUID section?

Cause I replaced my snippets of raw calls for getting the config for the standard commands.
My point was that all raw commands are described in the general part of USCUID section and I was not sure should we copy them to the concrete one. I could also change the "Commands" section for the types of USCUID and add all commands from that type.

I think that was only replacements I made. Or I'm missing something?

iceman1001:
yeah,   raw snippets and protocol bytes. 
In one case you replaced the config block with the another output in the broken down format.  Copy paste problem I guess.

mishamyte:
Gotcha!

Returned raw commands for corresponding tags and thank you for that copy-paste typo.

Pls, let me know if I should change/add something. 

Thank you!

iceman1001:
All good, 

Thanks!

PR #2303, opened 2024-02-25
Create intertic.py

gentilkiwi:
Basic script to try to interpret Intertic data on ST25TB / SRT512 in french transports


iceman1001:
Nice!
Your choice of CC as a license is a bit problematic.  It isn't really compatible with our GPLv3 license.  

If you could consider swapping license...

gentilkiwi:
> Nice! Your choice of CC as a license is a bit problematic. It isn't really compatible with our GPLv3 license.
> 
> If you could consider swapping license...

Just a copy paste from my usual Licence. Now changed for you :)

iceman1001:
<3

PR #2304, opened 2024-02-26
Update to sample dump part 2 of T5577_Guide.md.

krislyon:
In part 2 of the T5577 guide we create an EM4100 with an ID of 1122334455, and then check that we are successful.  The sample shown to check against does not match and is for an EM4100 with ID of 0F0368568B.

- Removed dump of EM4100 with ID: 0F0368568B
- Added dump of EM4100 with ID: 1122334455


iceman1001:
Good catch!

PR #2308, opened 2024-02-29
hf mf dump print by default

francesco-scar:
Added dump print by default in `hf mf dump` and now supports the noprint flag to allow the same behavior as before.
Also updated the changelog for past commits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,15 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Changed `hf mf dump --np` - now prints dump by default and supports noprint flag (@francesco-scar)
  - Fixed `hf 14a info` - magic detection works again and better hint detection (@iceman1001)
+ - Changed `14b wrbl` - now uses SRIX 4k as default tag (@Sonic803)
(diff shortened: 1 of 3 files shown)


iceman1001:
Why would we have a no-printing ?

francesco-scar:
I don't know, I was considering a scenario where you might only want to save the dump to a file and avoid lengthy printing; I wasn't sure if the previous behavior was desired or not. If you believe it's better, I can simply remove that option.

iceman1001:
Thanks for the interest and effort of making a PR,   however I just don't see the purpose with this.

Closing.

francesco-scar:
Okay, but can we just add a way to view the dumped data without saving it to a file? It's quite annoying to have to repeatedly execute two commands, saving to a file and then displaying its content, just to view the data. Additionally, without printing, I believe the --ns (no save) option is rather useless.

iceman1001:
We have the `hf mf view -h`  command for viewing memory dump files.  
We have `hf mf dump` for saving and print card memory
We have `hf mf dump --ns`  for viewing card memory 

The use-case for  saving only ,  I just don't see we need that.

francesco-scar:
With the current code, the `hf mf dump` command doesn't print the card memory to the screen (correct me if I'm wrong).

iceman1001:
Hm...  
and here I was certain the dump was printed like in the other dump commands.

I pushed a fix for that.

iceman1001:
Alright, 
now I see the reason for your PR,   I already added your stuff,   so I will close this PR,   but you got mentioned in both CHANGELOG and Commit message.

Sorry for the convenience

PR #2316, opened 2024-03-05
Improve reliability of `lf em 4x70 writekey`

henrygab:
After key is written, authentication is attempted using the new key.  This is particularly important for glass modules, where the coupling can be quite weak, to ensure the key is properly written.

Split parameter parsing from client/firmware interaction by adding helper functions.  This enables re-use and building more complex commands easily.

Firmware: Partially-tracked at least one of the places where the notice of extra bits was coming from.  Add a TODO comment to remind where to look for further review.

No changelog entry needed for this, it's primarily internal, and a continuation of the last PR.

### DIFF
--- a/client/src/cmdlfem4x70.c
+++ b/client/src/cmdlfem4x70.c
@@ -29,7 +29,65 @@
 #define LOCKBIT_0 BITMASK(6)
 #define LOCKBIT_1 BITMASK(7)
 
-#define INDEX_TO_BLOCK(x) (((32-x)/2)-1)
+#define BYTE_ARRAY_INDEX_TO_BLOCK(x) ((31-(x))/2)
+
+// result from CMD_LF_EM4X70_INFO
+typedef struct _em4x70_tag_info {
+    /// <summary>
+    /// The full data on an em4x70 the tag.
+    /// [31] == Block 15 MSB == UM2₆₃..UM2₅₆
+    /// [30] == Block 15 LSB == UM2₅₅..UM2₄₈
+    /// [29] == Block 14 MSB == UM2₄₇..UM2₄₀
+    /// [28] == Block 14 LSB == UM2₃₉..UM2₃₂
+    /// [27] == Block 13 MSB == UM2₃₁..UM2₂₄
+    /// [26] == Block 13 LSB == UM2₂₃..UM2₁₆
+    /// [25] == Block 12 MSB == UM2₁₅..UM2₀₈
+    /// [24] == Block 12 LSB == UM2₀₇..UM2₀₀
+    /// [23] == Block 11 MSB == Pin₃₁..Pin₂₄
+    /// [22] == Block 11 LSB == Pin₂₃..Pin₁₆
+    /// [21] == Block 10 MSB == Pin₁₅..Pin₀₈
+    /// [20] == Block 10 LSB == Pin₀₇..Pin₀₀
+    /// [19] == Block  9 MSB == Key₉₅..Key₈₈
+    /// [18] == Block  9 LSB == Key₈₇..Key₈₀
+    /// [17] == Block  8 MSB == Key₇₉..Key₇₂
+    /// [16] == Block  8 LSB == Key₇₁..Key₆₄
+    /// [15] == Block  7 MSB == Key₆₃..Key₅₆
+    /// [14] == Block  7 LSB == Key₅₅..Key₄₈
+    /// [13] == Block  6 MSB == Key₄₇..Key₄₀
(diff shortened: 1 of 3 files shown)


client/src/cmdlfem4x70.c
```diff
@@ -40,100 +98,194 @@ static void fill_buffer_prng_bytes(void* buffer, size_t byte_count) {
...
+
+    PacketResponseNG resp;
+    if (!WaitForResponseTimeout(CMD_LF_EM4X70_WRITE, &resp, TIMEOUT)) {
+        if (!opts.disable_error_output) {
+            PrintAndLogEx(WARNING, "Timeout while waiting for reply.");
+        }
         return PM3_ETIMEOUT;
     }
 
     if (resp.status) {
```
iceman1001, line 184: ...  the PM3_SUCCESS is defined as 0.   If you are pushing your own status codes,  please not. 
To make the understanding of the status code,   always use a comparision,   which allows for easier understanding of what is expected outcome from the device.

client/src/cmdlfem4x70.c
```diff
@@ -29,7 +29,65 @@
...
+} em4x70_tag_info;
+
+// result from CMD_LF_EM4X70_AUTH is defined in ID48LIB (id48.h) as ID48LIB_GRN 
+
+typedef struct _em4x70_cmd_helper_options {
+    // variables named based on default value (zero-init) being false
+    uint32_t use_parity : 1;             // default is false ... no parity
+    uint32_t disable_error_output : 1;   // default is false ... errors/warnings are output using PrintAndLogEx()
+    uint32_t disable_all_output   : 1;   // default is false ... success often his output using PrintAndLogEx()
+    uint32_t : 29; // padding
```
iceman1001, line 88: it almost looks like you are pushing your own printing logic..  Joining a third one to our DEBUG, and VERBOSE patterns...   Please don't.

iceman1001:
I am all for refactoring.   We have noticed over the years that some of the contributions did that a lot in their code parts,  leaving the overall client diversified and gave a nightmare to maintain.  Trying to understand and keep track of those alternative ways.    I am thinking of  HITAG code and ISO15693 code,    so I see similarities here in your refactoring.

Please don't.    Try using the code patterns we have and even if its not how you would like it to be,   it is what it is. 
Otherwise your changes will be need to be implemented across the whole client.   And  I think you are not ready for
that commitment.

client/src/cmdlfem4x70.c
henrygab: This is definitely not creating new status codes.
Rather, the `opts` parameter has a different purpose:  Allowing code re-use without forcing specific outputs to be sent.   For example, being able to re-use the `auth` helper function without a gazillion error messages, if wanting to try a brute force.  Or, periodically looping with a check for a tag, where you want to silently ignore a tag not being present.

This PR was intended to make a cleaner split between client command line parsing, and functions that perform a specific interaction with the firmware.   There are no new error codes.

Does that help explain some of the logic here and address your concern about new error codes?

client/src/cmdlfem4x70.c
henrygab: This is not the intention.  Rather, these helper functions are going to be called from multiple points.   Some of these would expect errors to occur, and some do not.  This is essentially a flag saying, "For this caller, never dump errors to the log" ... which makes sense when the code is expecting the operation will fail most of the time.  There is no reason to clutter the log.   This is a different purpose than DEBUG or VERBOSE ... this is a code path decision.

The simple example is for code that will wait for the presence of a tag.   Why fill the log with thousands of useless TIMEOUT messages?  Of course, such a simple function can be easily duplicated with different printing options.  But then more complex code paths, such as, "Wait for a tag, check if it's writable, then write block X, then auth, then write Block X, then ..." is much more complex, and it can be helpful to avoid duplicating all the helper functions.

I didn't really like the `opts` argument, so I'm open to ideas.   I can generate `function_x_no_output()` and `function_x_only_error_output()` variants of functions instead?   But, code duplication didn't seem to be the right way either?

henrygab:
> Please don't. Try using the code patterns we have and even if its not how you would like it to be, it is what it is.

Can you help me understand which code patterns you are referring to?  I'm generally happy to conform to the patterns of any given project.

I do know of one place where I diverged:  In one place (and only one place), I manually expanded the `CLIGetHexWithReturn()` macro, but only because the macro was failing to compile with the proper size.  And I expanded only one level of the macro ... the minimum necessary to avoid the particular issue.  Is this the divergence you refer to?

PR #2317, opened 2024-03-05
Offline 'hf 14a sniff' not terminated by new usb connection

McEloff:


iceman1001:
We have a reason for this.

The timing is critical and the extra checks will disrupt the sniffs.

McEloff:
That is right. But we could not correctly terminate offline sniffing to receive trace log after usb reconnecting. Push button terminates sniffing but device is not communicates with client. OS resets usb connection and device reboots. After this trace log is missing.

iceman1001:
I must be missing something.

I set up a pm3 w a blueshark (battery)  and installed hf_unisniff.     Disconnected the pm3,  turn it on with battery,  entered standalone mode.   Did the sniffinf,   clicked the button to exit.   Connected the proxmark3 to the client and look at the trace data which was both in the spiffs and tracelog...

iceman1001:
I was curious,  so I made a video about UNISNIFF.   I wanted to try out the not available in ram when reconnecting.

I had no issue what so ever to reconnect and extract the tracelog from ram.   I think you are using the mode in the wrong way.
I am guessing you don't push the button to exit the mode.

McEloff:
You're right! These steps work correctly with the pm3 client for desktop. I was also able to receive the trace after exiting sniffer mode by button. Apparently the problem occurs under Android smartphones and pm3, that connects via OTG with own battary. I tests my light android-client supports some pm3-fuctions. I should close merge requst.

McEloff:
Not actual for generic clients

iceman1001:
You closed it.  Question is if the issue itself is resolved for you and what steps you needed to do?

I don't think we need a change to the source code but curious on what might have gone wrong in your android setup.

McEloff:
I will push this source code to my local repository. In this case my android-client is works and I did not saw any sniffing fails. I think, this malfunction is the hardware Android feature for external devices with own battary. When this device connected to multiply-purpose usb port, OS detects an external power supply, then switched to slave device mode for listen.

PR #2326, opened 2024-03-11
Set max write blocks to 251 in `hf_mf_ultimatecard` before writing version/signature in NTAG 21X

### LINKED ISSUE #2325: Unable to set NTAG version/signature using `hf_mf_ultimatecard`
**Describe the bug**
Both version and signature in NTAG 21X can't be written in magic gen4 cards since they are stored in blocks 242-251, which unless you allow writing in them, it will yield a write error.

**To Reproduce**
1. With a magic card gen4 aka "ultimate card", try changing the version or signature using commands like:
    * `script run hf_mf_ultimatecard -v 0004040201000f03`
    *  `script run hf_mf_ultimatecard -s 8B76052EE42F5567BEB53238B3E3F9950707C0DCC956B5C5EFCFDB709B2D82B3`
    *  `script run hf_mf_ultimatecard -w 1`
2. When reaching those "problematic" blocks, it fails with `wrote block ..XX failed`

**Expected behavior**
It should properly write in those blocks, therefore setting version and signature.

**Screenshots**
N/A

N/A (using latest available version of the script, see https://github.com/RfidResearchGroup/proxmark3/blob/v4.18218/client/luascripts/hf_mf_ultimatecard.lua) 

**Additional context**
I'll be sending a PR fixing it shortly.

antgamdia:
This PR mainly sets the max write blocks to 251 before attempting to write either the version or signature in NTAG chips (see 90e9a0da39bed2ba6cb7d686338abafee208520d).

Besides, it changes the UID and signature used in the CLI example so that they match. Currently, using the example signature returns an error, since the format is invalid.

Finally, a trivial typo (`` ` ``instead of `'`) has been fixed.

_This is a purely personal contribution, not related to my current employer whatsoever._

Fixes #2325

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Set max write blocks to 251 in hf_mf_ultimatecard before writing version/signature in NTAG 21X (@antgamdia)
  - Changed HF_UNISNIFF standalone mode, output and also sniffs iclass too (@iceman1001)
  - Changed `hw standalone` - added `-b` param to control HF_UNISNIFF from client (@iceman1001)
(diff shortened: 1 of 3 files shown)


iceman1001:
Awesome!

PR #2333, opened 2024-03-16
Updated aidlist to include CEPAS AID

zhongfu:
CEPAS is a (Singaporean) standard for e-money smart cards; it's mainly used for transit, road tolls, and parking payments, but it also sees some use for retail payments and access control.


iceman1001:
Thanks!

PR #2337, opened 2024-03-21
Fix MIFARE DESfire DeleteApp command

### LINKED ISSUE #2336: DESfire DeleteApp authenticates to wrong AID
**Describe the bug**
When using `hf mfdes deleteapp`, the client authenticates and selects the AID being deleted instead of the PICC AID, which needs to be selected to manage apps (unless they are delegated). I would suggest either adding another CLI argument for the auth AID or having the default be to have the `aid` parameter specify the AID being deleted instead of the AID used for authentication.

**To Reproduce**
1. Make sure that unauthenticated app management is disabled in the PICC config
2. Create a DESfire app with any AID
3. Try to delete the same app

**Expected behavior**
The client authenticates to the PICC AID and then deletes the app.

craftbyte:
Closes #2336


iceman1001:
Excellent!

PR #2341, opened 2024-04-05
First round of refactoring for Fudan

jlitewski:
Went through and started improving the Fudan code.

- the dump command now can take -v or --verbose
- began refactoring the dump command (Still needs more work but it's stable)
- Added a few other Fudan chips to the list (Not detectable yet because of not having any to get the adqa or sak for them, but once we can, they should just work:tm:)

With these changes, the view command would no longer work since the files that dump make are different than what they used to be. I disabled it with a warning for now until the dump command is fully reimplemented and the view command can take the block size and number of sectors into account.

I would like to keep this open until I can get everything back working 100%, but have it as a PR to get feedback (since I have no clue about anything dealing with C and I'm bound to mess things up)

### DIFF
--- a/client/src/cmdhffudan.c
+++ b/client/src/cmdhffudan.c
@@ -48,19 +48,23 @@
 // iceman:  these types are quite unsure.
 typedef enum {
     FM11RF005M,
-    FM11RF008M,
+    FM11RF005U,
     FM11RF005SH,
+    FM11RF008M,
     FM11RF08SH,
+    FM11RF32M,
+    FM11RF32N,
     FUDAN_NONE,
 } fudan_type_t;
 
-static void fudan_print_blocks(uint16_t n, uint8_t *d) {
+static void fudan_print_blocks(uint8_t sector, uint8_t num_blocks, uint8_t block_size, uint8_t *carddata) {
     PrintAndLogEx(NORMAL, "");
+    PrintAndLogEx(INFO, "Sector %i", sector);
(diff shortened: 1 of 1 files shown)


iceman1001:
Just so that I get this right.  
You added the Fudan MFC clones to the fudan dump command?

We have the `hf mf`  commands for those,  since they aren't any different.

The original purpose for `hf fudan` is to support the non-mfc cards they have.

jlitewski:
> Just so that I get this right. You added the Fudan MFC clones to the fudan dump command?
> 
> We have the `hf mf` commands for those, since they aren't any different.
> 
> The original purpose for `hf fudan` is to support the non-mfc cards they have.

Most of the work was fixing the dump command to attempt to output at least the first sectors full blocks instead of just 8 blocks total. The chip I have also needs a handshake for the last 8 blocks to be read or written, and a lot of the data sheets I found also have that. I'm working on sourcing more of those chips for testing, but that was the true reason I'm putting work into this. I just added any chip I found that supports iso14a from the datasheets I got ahold of, not really considering that they could be read by another group of commands.

iceman1001:
Yeah,   the fudan dump doesn't handle the lasrt 8 blocks since we don't know the authentication needed in order to read / write to them.  

I suggest you focus on figuring out the authentication process first.

Meanwhile I will close this PR since it messes with current impl.
For instance, you should be able to run  `hf 14a info`  and get which type for Fudan you have. 

It should also give you a hint saying which command to use next.

PR #2342, opened 2024-04-10
Plot improvements

jlitewski:
- Added hotkeys to move the yellow/purple cursor around
- Added hotkeys to manually increment/decrement a value from a plot point 
  - This is currently the point where the yellow cursor is at

So, after fighting with the plot graph for _way too long_, I decided to quickly throw this together to make my life easier. Hopefully others have a use for it as well.


iceman1001:
Interesting...     Manual fiddling with values...

PR #2343, opened 2024-04-10
Fix compile error from variable type declaration

p51d78th:
Compiling current code shows this error on macOS & Linux Mint. Reverting the one style line change fixes on both.

`src/proxguiqt.cpp: In member function ‘virtual void Plot::paintEvent(QPaintEvent*)’:
src/proxguiqt.cpp:866:47: error: format ‘%zi’ expects argument of type ‘signed size_t’, but argument 6 has type ‘uint32_t’ {aka ‘unsigned int’} [-Werror=format=]
  866 |     snprintf(str, sizeof(str), "@%u..%u  dt=%zi %szoom=%2.3f  CursorAPos=%zu  CursorBPos=%zu  GridX=%lf  

... 

make[1]: *** [Makefile:1002: obj/proxguiqt.o] Error 1
make: *** [Makefile:177: client/all] Error 2`


iceman1001:
I will merge it now since people seem upset...    :)

PR #2347, opened 2024-04-14
Refactored Plot Annotations and some cleanup

jlitewski:
Not the refactor I _wanted_ to do, but the one I did.

I cleaned up and rewrote how annotations are displayed on the plot window. They are now a lot more flexible and should be easier to add to in the future, since 90% of them are set up and rendered in `Plot::drawAnnotations()`. The only other annotations that are rendered is in the `Plot::PlotGraph()` function, because I really didn't want to redo all the min/max/mean calculations again.

This now has the ability to show up to 4 graphs of min/max/mean numbers, all 4 cursors info (including Operation Buffer changes with CursorA), and has enough space to add future info. Oh, and a lot of the annotations are tied to the state of the objects they track, so if the cursors aren't on screen, those annotations won't show up. Same with the grid, and the three other graphs that can be rendered.

I also moved the overlay buffer (`g_OverlayBuffer`) from `proxguiqt.cpp` to `graph.c` to keep all the graphs together. Once I understand more about how the Demod buffer works, I'll probably move that over as well, but for right now those three are there.

There's also some random cleanup for `graph.c` that touched a _lot_ of files in this PR, since I forgot to remove them before pushing. sorry 😅

image

### DIFF
--- a/client/src/cmddata.c
+++ b/client/src/cmddata.c
@@ -447,7 +447,7 @@ int ASKDemod_ext(int clk, int invert, int maxErr, size_t maxlen, bool amplify, b
         return PM3_EMALLOC;
     }
 
-    size_t bitlen = getFromGraphBuf(bits);
+    size_t bitlen = getFromGraphBuffer(bits);
 
     PrintAndLogEx(DEBUG, "DEBUG: (ASKDemod_ext) #samples from graphbuff: %zu", bitlen);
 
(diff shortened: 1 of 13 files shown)


iceman1001:
OK,  I like the plot changes,   but  renaming functions for the sake of renaming is just annoying. 

I really dislike the renaming part.

PR #2349, opened 2024-04-16
Fix homebrew Qt5 detection on MacOS

philicious:
After install of `brew install qt5` compilation with `make clean && make -j` didnt find Qt5 lib. 

I found [this old PR](https://github.com/RfidResearchGroup/proxmark3/pull/1737) that once fixed same issue but by now homebrew changed the lib path, so here is a fix again

> GUI support:       QT5 found, enabled (Qt version 5.15.13 in /opt/homebrew/Cellar/qt@5/5.15.13/lib)


iceman1001:
Since I don't have MacOS based machine,  I have to take your PR for face value.

PR #2353, opened 2024-04-17
Mostly implement the Operation Buffer

jlitewski:
This should get 90%+ of all the cases that modify data directly to the graph buffer, and instead modify the data to the Operation Buffer instead.

This looks like a big change, but it's basically moving direct GraphBuffer writes and reads to functions I implemented to handle writing to or reading from either the Graph Buffer or the Operation Buffer and keeping those two buffers in sync.

I'm sure I missed a few, and I know I didn't touch some of the direct Graph Buffer modifications because the code looked scary and I didn't want to break something super important without having a solid way to test it locally. Most of the commands I could test all seem to output the same results as before these modifications (minus the obviously broken things with trimming and `hw tune`), so I count that as a win in my book.

Also felt like this was a big enough change to add to the Changelog, so I did.


jlitewski:
So apparently I broke things with this. Going to go through and attempt to fix things so everything is happy.

client/src/cmdhf15.c
```diff
@@ -574,6 +574,7 @@ static int CmdHF15Demod(const char *Cmd) {
     for (i = 0; i < 1000; i++) {
         int corr = 0;
         for (j = 0; j < ARRAYLEN(FrameSOF); j += skip) {
+            //TODO Do I really need to use get_graph_value_at()?
```
iceman1001, line 577: No,   we don't want a "function call" to  make updates to a array in  a loop.    

The solution is a making a local array,   apply changes,   and then copy back to the array we wanted modified...

client/src/cmdhftexkom.c
```diff
@@ -32,10 +32,11 @@
 #define TEXKOM_NOISE_THRESHOLD (10)
 
 static inline uint32_t GetGraphBuffer(uint32_t indx) {
-    if (g_GraphBuffer[indx] < -128)
+    int32_t value = get_graph_value_at(indx, true);
```
iceman1001, line 35: this is just nasty...

client/src/cmdlf.c
```diff
@@ -1007,8 +1008,10 @@ int lfsim_upload_gb(void) {
         clearCommandBuffer();
         payload_up.offset = i;
 
-        for (size_t j = 0; j < len; j++)
-            payload_up.data[j] = g_GraphBuffer[i + j];
+        for (size_t j = 0; j < len; j++) {
+            //payload_up.data[j] = g_GraphBuffer[i + j];
+            payload_up.data[j] = get_graph_value_at(i + j, true);
```
iceman1001, line 1013: again...  nasty

client/src/cmdlfti.c
```diff
@@ -101,35 +101,46 @@ int demodTI(bool verbose) {
...
+
+    //Write the values to the Operation Buffer for storage, to not destroy the Graph Buffer
     for (i = 0; i < g_GraphTraceLen - convLen; i++) {
         lowSum = 0;
         highSum = 0;
 
         for (j = 0; j < lowLen; j++) {
-            lowSum += LowTone[j] * g_GraphBuffer[i + j];
+            lowSum += LowTone[j] * get_graph_value_at(i + j, true);
+            //lowSum += LowTone[j] * g_GraphBuffer[i + j];
```
iceman1001, line 112: see previous comments

client/src/cmdlfti.c
```diff
@@ -101,35 +101,46 @@ int demodTI(bool verbose) {
...
         }
         for (j = 0; j < highLen; j++) {
-            highSum += HighTone[j] * g_GraphBuffer[i + j];
+            highSum += HighTone[j] * get_graph_value_at(i + j, true);
+            //highSum += HighTone[j] * g_GraphBuffer[i + j];
```
iceman1001, line 116: see previous comments

client/src/cmdlfti.c
```diff
@@ -101,35 +101,46 @@ int demodTI(bool verbose) {
...
 
+    //TODO Write to the Demod Buffer from the Operation Buffer and reset the Operation Buffer
     for (i = 0; i < g_GraphTraceLen - convLen - 16; i++) {
         lowTot = 0;
         highTot = 0;
         // 16 and 15 are f_s divided by f_l and f_h, rounded
         for (j = 0; j < 16; j++) {
-            lowTot += (g_GraphBuffer[i + j] & 0xffff);
+            lowTot += (get_graph_value_at(i + j, false) & 0xffff);
+            //lowTot += (g_GraphBuffer[i + j] & 0xffff);
```
iceman1001, line 135: see previous comments

client/src/cmdlfti.c
```diff
@@ -191,10 +208,12 @@ int demodTI(bool verbose) {
     for (i = 0; i < ARRAYLEN(bits) - 1; i++) {
         int high = 0, low = 0;
         for (j = 0; j < lowLen; j++) {
-            low -= g_GraphBuffer[maxPos + j];
+            //low -= g_GraphBuffer[maxPos + j];
+            low -= get_graph_value_at(maxPos + j, false);
```
iceman1001, line 212: see previous comments

client/src/cmdlfti.c
```diff
@@ -191,10 +208,12 @@ int demodTI(bool verbose) {
...
         }
         for (j = 0; j < highLen; j++) {
-            high += g_GraphBuffer[maxPos + j];
+            //high += g_GraphBuffer[maxPos + j];
+            high += get_graph_value_at(maxPos + j, false);
```
iceman1001, line 216: see previous comments

client/src/graph.c
```diff
@@ -109,6 +109,20 @@ void save_restoreGB(uint8_t saveOpt) {
...
+
+    if(useGraphBuffer) {
+        copiedValue = g_GraphBuffer[index];
+    } else {
+        copiedValue = g_OperationBuffer[index];
+    }
+
+    return copiedValue;
+}
+
```
iceman1001, line 125: you are squeezing two arrays into one methodology,      learn to understand the code so it uses  the operations buffer and then applies all changes to right buffer in use.

client/src/graph.c
```diff
@@ -149,16 +163,16 @@ size_t getFromGraphBufferEx(uint8_t *dest, size_t maxLen) {
...
     if (g_GraphTraceLen == 0) return 0;
     if (start >= end) return 0;
 
     size_t i, value;
     end = (end < g_GraphTraceLen) ? end : g_GraphTraceLen;
-    for (i = 0; i < (end - start); i++) {
-        value = g_GraphBuffer[start + i];
+    
+    for (i = start; i < end; i++) {
+        value = get_graph_value_at(i, useGraphBuffer);
```
iceman1001, line 175: see previous comments

iceman1001:
ok, 

Interesting approach but it will not do.   If you going for an operations buffer pattern,  then you need to understand how to correctly use it in the source.  Right now you merging it together and it is mess.

Rethink,  understand,  make a better solution

client/src/cmdhftexkom.c
jlitewski: Can I ask how? I'm genuinely curious about that, mostly because I'm not sure what's nasty about it.

Originally, it was just directly checking the graph buffer at the supplied index. This makes a copy of whatever value is at that index and checks it instead. Wouldn't that be safer to do?

jlitewski:
Okay, pushed the first revision.

I removed the `get_graph_value_at()` function, and I followed your advice and implemented temporary buffers to copy into and write out from in places where it was needed. I mostly just swapped out the function for writing directly to the Operation Buffer, and added text to the commands that use it warning that it does, since it is kinda a new thing

jlitewski:
Okay, I'm going to close this and redo it, because somehow updating my local's master messed up this branch.

PR #2361, opened 2024-04-20
Implemented Buffer SaveStates

jlitewski:
This replaces the implementation of `save_restoreGB` and `save_restoreDB` while allowing more uses in the future.

Also, while working on this implementation, I made a few helper methods that were pushed alongside this. Namely `IfClientDebugIsEnabled()` for the command parser to show debug commands if client debug is enabled, alongside a few testing commands to run the `buffer_savestate_t` implementation through it's paces (only seen if the client is in debug mode)


jlitewski:
If anyone knows of a good VSC extension to check for these errors so I don't have to make 20 force push fixes, that'll be great to know

jlitewski:
Fixed the conflicts the latest version of master introduced

jlitewski:
Second conflict fix

PR #2362, opened 2024-04-22
Update recover_pk.py with first TruST25 (ST) key

gentilkiwi:
TruST25 (ST) - KeyID 0x04?
`curve=secp128r1, hash=sha256` - from block 63 in ST25TV, starting with KeyID?

Tested on multiple `ST25TV02KC` - `K04S` at block 63, signature in ... ASCII...


iceman1001:
Excellent!

PR #2367, opened 2024-04-25
update iclass elite key chk helptext

kitsunehunter:
Since elite keys are in a separate dictionary, the helptext should tell the user to call that elite dictionary when checking elite keys.


iceman1001:
Makes sense

PR #2370, opened 2024-04-26
Fix the issue of SKIPQT builds not working

jlitewski:


iceman1001:
@s00se   does this work now on Termux?

PR #2373, opened 2024-04-27
Fix OOB segfault with markers

jlitewski:
This fixes an issue that was found by `gentilkiwi`. I added bounds checking to the marker placing code. I also modified it to run only when a left/right mouse click was made, instead of doing all the math every time the `Plot::MouseMoveEvent()` function was called and then only applying it if there was a mouse click.


iceman1001:
Nice!
No need to recalc everything with mouse move event indeed

btw
Only need to fix the if-statement style.

PR #2379, opened 2024-05-06
Fix MacOS building

jlitewski:
Pass `--break-system-packages --user` to the python commands to stop it complaining about breaking the externally managed dependencies of a non-virtual environment that's actually a virtual environment.

This **SHOULDN'T** be done locally because it could break things, but it's good enough for github actions


henrygab:
I have a "supported" fix ... no hacks required -- See PR#2380.

jlitewski:
Fair, closing this one in favor of henry's PR

PR #2380, opened 2024-05-06
Fix Github Action builds

henrygab:
* Uses official Github action `actions/setup-python` to install python.
  This avoids the hacks that are otherwise required on MacOS.
* Fix the broken hitag self-test


iceman1001:
Good catches!

PR #2388, opened 2024-05-20
Update mfc_default_keys.dic

CountParadox:
added keys to my apartment building


iceman1001:
You know if they are static keys?   I.e. same keys on all cards in the apartment building?

TomHarkness:
In Aus this is usually the case. Never seen a building where the keys differ.

iceman1001:
Ping,  
can you verify that this PR contains static keys across several cards from the same system?

CountParadox:
Yes; I tested another card and many of these keys also unlocked them. (The remaining were found in default dictionary)

iceman1001:
ok,  many is good indicator,  but it also begs the question,  which keys doesn't exists om multiple cards from same system?

PR #2390, opened 2024-05-21
Add new AID & ECP definitions

kormax:
Added the following:
* AID:
  - Apple Access Key;
  - Aliro;
  - Navigo.
* ECP:
  - Navigo.

* Added ECP taxonomy file, which was missing (I intended to add it when creating ecplist.json initially).


iceman1001:
Better late than never

PR #2395, opened 2024-05-29
Update 4_Advanced-compilation-parameters.md

LupusE:
Typo: Two times HF15SNIFF instead of one HF15SIM in STANDALONE list.


iceman1001:
Good catch!

PR #2399, opened 2024-06-07
Add new t55xx password (002BCFCF) sniffed from cheap cloner

davidbeauchamp:
I have a handheld white cloner and a handful of t5577 fobs that were supposedly re-writable but would not wipe. After some sniffing with the Proxmark3 I was able to uncover the password and to wipe the fobs, subsequently using the fobs with the PM3:

```
[usb] pm3 --> lf t55xx sniff

[=] T55xx command detection
[+] Downlink mode           |  password  |   Data   | blk | page |  0  |  1  | raw
[+] ------------------------+------------+----------+-----+------+-----+-----+-------------------------
[+] Default pwd write       |   44B44CAE | 002BCFCF |  7  |   0  |  19 |  46 | 1001000100101101000100110010101110000000000001010111100111111001111111
[+] ---------------------------------------------------------------------------------------------------
```

Picture of the cloner w/firmware version, was purchased some time in 2019 from eBay:
cloner

~~I couldn't decide if this minor change was worth inclusion in the changelog or not so I left it out.~~ Thank you, bot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 - Added `pm3_tears_for_fears.py` - a ISO14443b tear off script by Pierre Granier
+- Added new t55xx password (002BCFCF) sniffed from cheap cloner (@davidbeauchamp)
 
 ## [Aurora.4.18589][2024-05-28]
 - Fixed the pm3 regressiontests for Hitag2Crack (@iceman1001)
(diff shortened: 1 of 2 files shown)


iceman1001:
Awesome,    
But the first password 44B44CAE ,   when was that set?

davidbeauchamp:
I snipped out a couple lines from the sniff, but it tries to write to blocks 1 and 2 with both `44B44CAE` and `88661858` in a row, and then attempts to write to block 7 with the new password from here (`002BCFCF`). I tried a handful of various uid's and the password didn't change, [unlike a few posts I read on the PM3 forums.](https://www.proxmark.io/www.proxmark.org/forum/viewtopic.php%3Fid=6482.html). Those two passwords appear to be from previous versions of the cloner and already exist in the dictionary.

Issue #1652, opened 2022-04-04
ISO 14443B simulation does not work


jacopo-j:
**Describe the bug**
The `hf 14b sim` feature does not seem to work properly: readers do not detect the simulated tag.

**To Reproduce**
1. Run `hf 14b sim -u 11AA33BB`

**Expected behavior**
External readers detect the presence of a tag.

**Desktop (please complete the following information):**
 - OS: macOS

 **hw version**

```

 [ Proxmark3 RFID instrument ]

 [ CLIENT ]
  Iceman/master/v4.14831-530-g8236de119-dirty-unclean 2022-04-04 00:23:11 917abd9ba
  compiled with............. Clang/LLVM Apple LLVM 13.1.6 (clang-1316.0.21.2)
  platform.................. OSX / aarch64
  Readline support.......... present
  QT GUI support............ present
  native BT support......... absent
  Python script support..... present
  Lua SWIG support.......... present
  Python SWIG support....... present

 [ PROXMARK3 ]
  firmware.................. PM3 GENERIC

 [ ARM ]
  bootrom: Iceman/master/v4.14831-530-g8236de119-dirty-unclean 2022-04-03 15:24:18 e49e4ed9a
       os: Iceman/master/v4.14831-530-g8236de119-dirty-unclean 2022-04-04 00:23:13 917abd9ba
  compiled with GCC 10.3.1 20210824 (release)

 [ FPGA ]
  LF image 2s30vq100 2022-03-23 17:21:05
  HF image 2s30vq100 2022-03-23 17:21:16
  HF FeliCa image 2s30vq100 2022-03-23 17:21:27
  HF 15 image 2s30vq100 2022-03-23 17:21:38

 [ Hardware ]
  --= uC: AT91SAM7S512 Rev B
  --= Embedded Processor: ARM7TDMI
  --= Internal SRAM size: 64K bytes
  --= Architecture identifier: AT91SAM7Sxx Series
  --= Embedded flash memory 512K bytes ( 58% used )
```

 **hw status**

```
[#] Memory
[#]   BigBuf_size............. 42784
[#]   Available memory........ 42784
[#] Tracing
[#]   tracing ................ 1
[#]   traceLen ............... 0
[#] Current FPGA image
[#]   mode.................... HF image 2s30vq100 2022-03-23 17:21:16
[#] LF Sampling config
[#]   [q] divisor............. 95 ( 125.00 kHz )
[#]   [b] bits per sample..... 8
[...]

PR #2402, opened 2024-06-11
Fix ISO 14443-B tag simulation

michi-jung:
See https://github.com/RfidResearchGroup/proxmark3/issues/1652

- Fix Bit Coding PICC -> PCD:  Encoding for 0 and 1 bits were reversed.
- Add a frontend delay for TR0 (No subcarrier) in TransmitFor14443b_AsTag.
- Remove unconditionally prefixing the encoded data with two '1' bits.
- Improve the Type B PICC State Machine implementation.

With these improvements my PCD can read the ISO 14443-B tag emulated by a Proxmark3 Easy.


iceman1001:
Great fix!

PR #2408, opened 2024-06-22
Update aid_desfire.json

ry4000:
**Added a new PACS AID** which was obtained from a NXP TagInfo scan; *however, the person that sent me the screenshot did not show me a photo of the credential in question, leading me to classify it as an unknown MF3DH42*.

*I will follow up with NVX to try and change a blank MFDES card to the above AID to then extract the information from NXP TagInfo on my end and then submit another PR.*

**Corrected minor formatting typo for AID F21150** *to make sure the Country field uses eight [8] spaces; I missed this when submitting my last PR, so apologies for that*.

### DIFF
--- a/client/resources/aid_desfire.json
+++ b/client/resources/aid_desfire.json
@@ -88,6 +88,14 @@
         "Type": "pacs"
     },
     {
+        "AID": "F51BC0",
+        "Vendor": "",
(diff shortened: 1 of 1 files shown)


iceman1001:
Excellent!

PR #2411, opened 2024-06-27
When simulate a ISO15693 tag the pageCount could be zero 

xueliu:
Recently I got a RFID card. Regarding the info below it is produced by EM Microelectronic-Marmin SA Switzerland. It supports three standards: 14a, 15 and Legic Prime.

```
[usb] pm3 --> hf search
[-] Searching for ISO14443-A tag...
[+]  UID: XX XX XX XX XX XX XX
[+] ATQA: 00 41
[+]  SAK: 20 [1]
[+] MANUFACTURER: EM Microelectronic-Marin SA Switzerland
[+]    JCOP 31/41
[=] -------------------------- ATS --------------------------
[+] ATS: 05 77 77 81 02 [ 91 00 ]
[=]      05...............  TL    length is 5 bytes
[=]         77............  T0    TA1 is present, TB1 is present, TC1 is present, FSCI is 7 (FSC = 128)
[=]            77.........  TA1   different divisors are supported, DR: [2, 4, 8], DS: [2, 4, 8]
[=]               81......  TB1   SFGI = 1 (SFGT = 8192/fc), FWI = 8 (FWT = 1048576/fc)
[=]                  02...  TC1   NAD is NOT supported, CID is supported

[?] Hint: try `hf mf` commands


[+] Valid ISO 14443-A tag found

[|] Searching for LEGIC tag...
[+]  MCD: XX
[+]  MSN: XX XX XX
[+] TYPE: MIM1024 card (1002 bytes)

[+] Valid LEGIC Prime tag found

[|] Searching for ISO15693 tag...
[+] UID.... XX XX XX XX XX XX XX XX
[+] TYPE... EM Microelectronic-Marin SA Switzerland (Skidata)

[+] Valid ISO 15693 tag found

[?] Hint: try `hf legic` commands

[?] Hint: try `hf 15` commands

```

I got a message "[=] Tag is empty!", when I dump the card using `hf 15 dump`. The value pageCount in the dump file is 0.
I got an error message "[-] Tag size error: pagesCount=0, bytesPerPage=4" when I tried to load the dump file into the memory.
But I think pagesCount=0 should be accepted since we have an empty ISO 15693 tag.

Using the attached patch I could load the dump into the memory and simulate the empty iso15693 tag properly.

### DIFF
--- a/armsrc/iso15693.c
+++ b/armsrc/iso15693.c
@@ -2173,7 +2173,6 @@ void SimTagIso15693(const uint8_t *uid, uint8_t block_size) {
 
     if ((tag->pagesCount > ISO15693_TAG_MAX_PAGES) ||
             ((tag->pagesCount * tag->bytesPerPage) > ISO15693_TAG_MAX_SIZE) ||
-            (tag->pagesCount == 0) ||
             (tag->bytesPerPage == 0)) {
         Dbprintf("Tag size error: pagesCount = %d, bytesPerPage=%d", tag->pagesCount, tag->bytesPerPage);
         reply_ng(CMD_HF_ISO15693_SIMULATE, PM3_EOPABORTED, NULL, 0);
--- a/client/src/cmdhf15.c
+++ b/client/src/cmdhf15.c
@@ -1246,7 +1246,6 @@ static int CmdHF15ELoad(const char *Cmd) {
 
     if ((tag->pagesCount > ISO15693_TAG_MAX_PAGES) ||
             ((tag->pagesCount * tag->bytesPerPage) > ISO15693_TAG_MAX_SIZE) ||
-            (tag->pagesCount == 0) ||
             (tag->bytesPerPage == 0)) {
         PrintAndLogEx(FAILED, "Tag size error: pagesCount=%d, bytesPerPage=%d",
                       tag->pagesCount, tag->bytesPerPage);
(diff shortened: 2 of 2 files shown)


iceman1001:
Interesting card with those three technologies.  Where did you get that one from?

When it comes to eload an empty dump file  I am doubtful.  An empty dump file like that is an indication of a failed dump.
The only useful information in the file would be UID.  

The preferred way to simulate UID is to use it in the -u parameter.  i.e.  `hf 15 sim -u  E011223344556677`

xueliu:
Beside of these three technologies, the card has also an   ISO/IEC 7816 compatible chip. Normally the card is used for identity authentication and signing documents. So we call it "PKI card". It comes from a famous German automobile manufacturers.

As you said, "An empty dump file like that is an indication of a failed dump". Are we going to keep this design ? If yes. I will close this ticket. Thanks.

iceman1001:
If you can confirm you can simulate with UID and get same results as with your empty dump file patch ?

PKI Card..  I want one,  where can I get one?

xueliu:
> If you can confirm you can simulate with UID and get same results as with your empty dump file patch ?
> 
> PKI Card.. I want one, where can I get one?

I am back. I have simulated  the PKI card with UID `hf 15 sim -u xxxxxx` and loading the empty dump file with my patch. From the reader side (Handy or Flipper Zero), there are no differences. So I will close this pull request.

Regarding the PKI card, as you can see that the supports of ISO 14443-A and ISO15693 are not complete. Internal storage cannot be read except for UID. The card that works in the access control system is Legic Prime. I can pass the access control system by simulating the Legic Prime card.

My PKI card is still in use, and it has privacy information. I guess your only chance to get this card is to join this company. ^_^

PR #2412, opened 2024-06-28
Felica raw command select key fix

RebornedBrain:
### Problem:
Currently when using `hf felica raw` command with flag `-s` it fails because of the CRC check during command reception.
image
But if then we take a look into `hf felica list`, we can see that in fact it receives correct data:
image
The reason of that, is that when `-s` flag is used firmware side calls `felica_select_card` which selects a card and copies all the data to `felica_card_select_t` which then is transmitted to client. 
And then when client tries to check CRC it fails, because it expects raw frame for that check.
### Solution:
This fix changes the logic so `felica_select_card` on the firmware side returns a raw frame to client, which then is able to check CRC correctly. Also it required to change 'read_felica_uid' on the client side, now it copies required data from raw bytes to `felica_card_select_t`.

### DIFF
--- a/armsrc/felica.c
+++ b/armsrc/felica.c
@@ -49,7 +49,7 @@ static uint32_t felica_nexttransfertime;
 static uint32_t felica_lasttime_prox2air_start;
 
 static void iso18092_setup(uint8_t fpga_minor_mode);
-static uint8_t felica_select_card(felica_card_select_t *card);
+static uint8_t felica_select_card(void);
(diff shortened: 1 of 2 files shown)


iceman1001:
Interesting find of the client felica raw -s issue. 
I think a better fix will be to adapt the CRC check on in the client side instead.

I pushed  https://github.com/RfidResearchGroup/proxmark3/commit/dd9bc4d363d8f92978c0d1f28a7b14023f1d69c8 

If you want to test it.

PR #2414, opened 2024-07-04
Fix a few mistakes in Wiegand encodings

jmichelp:


iceman1001:
Nice catches..

interested in how you figured out the indala 27?

jmichelp:
If I remember correctly, this one was discovered by ensuring that `Decoding(Encoding(X)) = X` :)

iceman1001:
hm...
Should make a regression test for that :) :)

PR #2416, opened 2024-07-09
Fix invalid Wiegand format flags.

jmichelp:
Some formats were declared with not adequate flags resulting in the format being filtered out for encoding/decoding unless explictly setting it.


iceman1001:
I am impressed by your keen eye for details.

PR #2419, opened 2024-07-12
Implemented VB6 rng for iclass lookup elite key search

Antiklesys:
Ported @bettse work from the Flipper Zero Picopass repository to use the lookup function with the VB6 rng as an option.
Limited key generation to the first 5000


iceman1001:
Nice!
I would argue the keygen part should be in `common/generator.c`  since cmdhficlass.c is getting way too large.
But its all fine for now.

Antiklesys:
I’ll check to move it perhaps.
I think this code could/should be recycled also for the hficlasschk
function and there could be an optimization to pass as a parameter the
number of keys to generate

mistial-dev:
HID changed the algorithm at some point prior to ICE1750, so there's probably no point in going past that with the VB6 RNG.

PR #2423, opened 2024-07-13
Update CHANGELOG.md

dandri:
Added text for mfc keys for Laugardalslaug in Iceland


iceman1001:
Turned out you didn't check if the key existed before in the dictionary.
I have removed the duplicate but now I wonder if your second key is used on several cards across same system.

dandri:
Skipped my mind to check for duplicates honestly. I've checked multiple cards for that system and they all use the same static keys.

PR #2426, opened 2024-07-15
Lua script for cloning new ELECTRA tags or EM410x to T5577 tag

vasilpetrov:
Script created by: https://github.com/jareckib 
Original topic, and why this exist can be found under: https://forum.dangerousthings.com/t/lua-script-for-proxmark3/19260
Original repo: https://github.com/vasilpetrov/lf_electra


iceman1001:
Interesting for historic reasons,   

For cloning a electra tag with the extra two blocks,  you can use the native client command
`
lf em 410x clone --id 0F0368568B --electra
`

PR #2430, opened 2024-07-19
Iclass Legacy Raw Key Recovery Function

Antiklesys:
Based on the work described in Dismantling iClass whitepaper. 

- `hf iclass legbrute` is tested working on sample data generated accordingly to the hash0 logic.

- `hf iclass legrec ` is partially working: 
1. Logic of operations and sequence are in order 
2. Individual functionalities have been tested on sample/simulated data and was found to be effective and correct.
3. The privilege escalation operation is still NOT working during my tests (may be a silicon issue tho), but the logic in place should be correct.

### DIFF
--- a/armsrc/iclass.c
+++ b/armsrc/iclass.c
@@ -2254,20 +2254,20 @@ Xorring the index of iterations against those decimal numbers allows us to retri
     while (bits_found == -1){
 
     //Step3 Calculate New Key
-    uint8_t GenKeyBlock[PICOPASS_BLOCK_SIZE];
-    uint8_t GenKeyBlock_old[PICOPASS_BLOCK_SIZE];
+    uint8_t genkeyblock[PICOPASS_BLOCK_SIZE];
+    uint8_t genkeyblock_old[PICOPASS_BLOCK_SIZE];
     uint8_t xorkeyblock[PICOPASS_BLOCK_SIZE];
-    generate_single_key_block_inverted(zero_key, index, GenKeyBlock);
+    generate_single_key_block_inverted(zero_key, index, genkeyblock);
 
     //NOTE BEFORE UPDATING THE KEY WE NEED TO KEEP IN MIND KEYS ARE XORRED
     //xor the new key against the previously generated key so that we only update the difference
     if(index != 0){
-        generate_single_key_block_inverted(zero_key, index - 1, GenKeyBlock_old);
+        generate_single_key_block_inverted(zero_key, index - 1, genkeyblock_old);
         for (int i = 0; i < 8 ; i++) {
-            xorkeyblock[i] = GenKeyBlock[i] ^ GenKeyBlock_old[i];
+            xorkeyblock[i] = genkeyblock[i] ^ genkeyblock_old[i];
         }
     }else{
-            memcpy(xorkeyblock, GenKeyBlock, PICOPASS_BLOCK_SIZE);
+            memcpy(xorkeyblock, genkeyblock, PICOPASS_BLOCK_SIZE);
     }
 
     //Step4 Calculate New Mac
--- a/armsrc/iclass.c
+++ b/armsrc/iclass.c
@@ -2293,7 +2293,7 @@ Xorring the index of iterations against those decimal numbers allows us to retri
     for (int i = 0; i < 8 ; ++i) {
         //need to craft the authentication payload accordingly
         memcpy(msg->req.key, iclass_mac_table[i], 8);
-        res = authenticate_iclass_tag(&msg->req, &hdr, &start_time, &eof_time, mac1); //the mac here needs to be changed, mac 2 is a compiling placeholder
+        res = authenticate_iclass_tag(&msg->req, &hdr, &start_time, &eof_time, mac1); //mac1 here shouldn't matter
         if (res == true) {
             bits_found = iclass_mac_table_bit_values[i] ^ index;
             Dbprintf("Found Card Bits Index: " _GREEN_("[%3d]"), index);
--- a/armsrc/iclass.c
+++ b/armsrc/iclass.c
@@ -2153,7 +2153,7 @@ out:
     }
 }
 
-void generate_single_key_block_inverted(const uint8_t startingKey[PICOPASS_BLOCK_SIZE], uint32_t index, uint8_t keyBlock[PICOPASS_BLOCK_SIZE]) {
+void generate_single_key_block_inverted(const uint8_t *startingKey, uint32_t index, uint8_t *keyBlock) {
     uint32_t carry = index;
     memcpy(keyBlock, startingKey, PICOPASS_BLOCK_SIZE);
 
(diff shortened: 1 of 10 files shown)


Antiklesys (commented):
Forgot to add one file

armsrc/iclass.c
```diff
@@ -2152,3 +2152,216 @@ void iClass_Restore(iclass_restore_req_t *msg) {
         reply_ng(CMD_HF_ICLASS_RESTORE, isOK, NULL, 0);
     }
 }
+
+void generate_single_key_block_inverted(const uint8_t startingKey[PICOPASS_BLOCK_SIZE], uint32_t index, uint8_t keyBlock[PICOPASS_BLOCK_SIZE]) {
```
iceman1001, line 2156: use `const uint8_t *starting_key`   instead of arrays with defined sizes.

armsrc/iclass.c
```diff
@@ -2152,3 +2152,216 @@ void iClass_Restore(iclass_restore_req_t *msg) {
...
+
+        carry >>= 3;  // Shift right by 3 bits for the next byte
+        if (carry == 0) {
+            // If no more carry, break early to avoid unnecessary loops
+            break;
+        }
+    }
+}
+// Function to convert an unsigned int to binary string
+void intToBinary(unsigned int num, char *binaryStr, int size) {
```
iceman1001, line 2172: we use  `uintX_t`  to denote unsigned it.  
And functions like these should be reused and located in utils.c

armsrc/iclass.c
```diff
@@ -2152,3 +2152,216 @@ void iClass_Restore(iclass_restore_req_t *msg) {
...
+    binaryStr[size] = '\0';  // Null-terminate the string
+    for (int i = size - 1; i >= 0; i--) {
+        binaryStr[i] = (num % 2) ? '1' : '0';
+        num /= 2;
+    }
+}
+
+// Function to convert a binary string to hexadecimal
+uint8_t binaryToHex(char *binaryStr) {
```
iceman1001, line 2181: see previous comment,   see if you can reuse functions from utils.c  or  commonutils..c

armsrc/iclass.c
```diff
@@ -2152,3 +2152,216 @@ void iClass_Restore(iclass_restore_req_t *msg) {
...
+    return (uint8_t)strtoul(binaryStr, NULL, 2);
+}
+
+// Function to convert an unsigned int to an array of hex values
+void convertToHexArray(unsigned int num, uint8_t *partialKey) {
```
iceman1001, line 2186: see previous comment,   see if you can reuse functions from utils.c  or  commonutils..c

armsrc/iclass.c
```diff
@@ -2152,3 +2152,216 @@ void iClass_Restore(iclass_restore_req_t *msg) {
...
+        { 0x00, 0x00, 0x00, 0x00, 0x10, 0xED, 0x6F, 0x11 }, //Expected mac when last 3 bits of each byte are: 001
+        { 0x00, 0x00, 0x00, 0x00, 0x53, 0x35, 0x42, 0x0F }, //Expected mac when last 3 bits of each byte are: 010
+        { 0x00, 0x00, 0x00, 0x00, 0xAB, 0x47, 0x4D, 0xA0 }, //Expected mac when last 3 bits of each byte are: 011
+        { 0x00, 0x00, 0x00, 0x00, 0xF6, 0xCF, 0x43, 0x36 }, //Expected mac when last 3 bits of each byte are: 100
+        { 0x00, 0x00, 0x00, 0x00, 0x59, 0x7F, 0x4B, 0x58 }, //Expected mac when last 3 bits of each byte are: 101
+        { 0x00, 0x00, 0x00, 0x00, 0x1A, 0xA7, 0x66, 0x46 }, //Expected mac when last 3 bits of each byte are: 110
+        { 0x00, 0x00, 0x00, 0x00, 0xE2, 0xD5, 0x69, 0xE9 }  //Expected mac when last 3 bits of each byte are: 111
+    };
+    //Viewing the weak macs table card 24 bits (3x8) in the form of a 24 bit decimal number
+    static uint32_t iClass_Mac_Table_Bit_Values[8] = {0, 2396745, 4793490, 7190235, 9586980, 11983725, 14380470, 16777215};
```
iceman1001, line 2271: I am just curious how this decimal number goes together with the byte arrays above

armsrc/iclass.c
```diff
@@ -2152,3 +2152,216 @@ void iClass_Restore(iclass_restore_req_t *msg) {
...
+
+    uint8_t zero_key[PICOPASS_BLOCK_SIZE] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+    uint32_t index = 1;
+    int bits_found = -1;
+
+    //START LOOP
+    while (bits_found == -1){
+
+    //Step3 Calculate New Key
+    uint8_t GenKeyBlock[PICOPASS_BLOCK_SIZE];
```
iceman1001, line 2281: please use lower case when choosing variable names.   We don't use mixed case.

client/src/cmdhficlass.c
```diff
@@ -3842,6 +3842,304 @@ void picopass_elite_nextKey(uint8_t* key) {
     memcpy(key, key_state, 8);
 }
 
+static int CmdHFiClassRecover(uint8_t key[8]) {
+
+    uint32_t payload_size = sizeof(iclass_recover_req_t);
+    uint8_t aa2_standard_key[PICOPASS_BLOCK_SIZE] = {0xFD, 0xCB, 0x5A, 0x52, 0xEA, 0x8F, 0x30, 0x90};
```
iceman1001, line 3848: would be better to reuse the keys array in the top of this file for this purpose.

iceman1001:
Nice one!  
The missing attack vector from heart of darkness paper.    I like!

armsrc/iclass.c
Antiklesys: Fixed unsigned int, will check for utils.c

armsrc/iclass.c
Antiklesys: Fixed unsigned int, will check for utils.c

armsrc/iclass.c
Antiklesys: Fixed

armsrc/iclass.c
Antiklesys: The array above is a series of weak macs, those weak macs correspond to the different combinations of the last 3 bits of each key byte.
If we concatenate the last three bits of each key byte, we have a 24 bits long binary string.
If we convert that string to decimal we obtain those decimal numbers.
Xorring the index of iterations against those decimal numbers allows us to retrieve the what was the corresponding sequence of bits of the original key in decimal format.

iceman1001: Aha,   a bit complex.    If you could comment that with like you just wrote then it would make it better

armsrc/util.c
```diff
@@ -410,6 +410,8 @@ uint8_t binaryToHex(char *binaryStr) {
     return (uint8_t)strtoul(binaryStr, NULL, 2);
 }
 
+#define PICOPASS_BLOCK_SIZE 8
```
iceman1001, line 413: This define already exists in: 
`include/iclass_cmd.h`

iceman1001:
image

The log level is `ERR`

Antiklesys:
@iceman1001 whatever is in util.c for now I've just moved there for the sake of compiling.
As you mentioned there may be already some functions that could be reused.
I'll check on those and eventually that'll fix the need to redefine or not PICOPASS_BLOCK_SIZE in there.
I made all the other small fixes in the meantime, thank you for the help with the ERR/ERROR loglevel that was causing compilation outside Proxspace to fail.

iceman1001:
Make sure you apply the  array parameters as pointers everywhere ,  I only pointed it out in one place.

When you feel its ready,  let me know and I will merge it.

armsrc/util.c
Antiklesys: I just removed the definition and replaced it with an 8 for simplicity.

Antiklesys:
@iceman1001 I cleaned up the code and removed a few trailing issues/mistakes/work in progress comments.
I reviewed also the functions to convert bin 2 hex in iclass.c but they're not doing quite the same thing as the ones I implemented are doing, as mines are also dividing a 24 bit binary string in groups of 3 bits while doing the conversion and preparation for the key.
I'm sure there may be further code changes/improvements once I get my hands on a suitable card to test this functionality inside out. I'll definitely continue working on optimizing it and improving it on my end as I come across new findings on this topic.
Also may need to remove/reduce some of the Dbprintf statements in the future depending on "how fast" this goes and to eventually enhance speed, but all of these changes will only be possible by testing the code in production on a sample card.

Happy to merge from my side!

iceman1001:
Excellent work!

PR #2438, opened 2024-08-02
Fix staticnested_0nt on mingw/proxspace

nvx:
Fixes keyCount_mutex[0] and keyCounts[0] not being properly initialised
Fixes progress reporting not properly clearing the line
Switch malloc(a*b) style calls to calloc(a, b)


doegox:
thanks!!

PR #2447, opened 2024-08-15
Corrected an incorrect Readline library URL in client CMakeLists.txt

siowu:
Corrected an incorrect Readline library URL in client and client/experimental_lib CMakeLists.txt
from ftp://ftp.gnu.org/gnu/readline/readline-8.2.tar.gz Change to https://ftp.gnu.org/gnu/readline/readline-8.2.tar.gz


doegox (changes requested):
If to be changed, use https, not http

siowu:
> If to be changed, use https, not http

Yes, I have modified it.

doegox (approved):

PR #2448, opened 2024-08-15
Add --issue capability to hf iclass encode command

micsen:
Added the --issue command flag to the hf iclass encode command as well as implement the custom Defcon RFCTF format that is being used to wiegand


micsen:
Will open again once I also have write to EMU functionality

PR #2449, opened 2024-08-15
Add `--emu` and `--issue` support to `hf iclass encode` 

micsen:
Added the ability to update emulator memory instead of only writing to card as well as encoding issuance level directly from the encode command.


iceman1001:
Awesome!

PR #2456, opened 2024-08-19
Fix the timing issue with HitagS

douniwan5788:
I don't have the NXP HitagS, only tested on 8211.


iceman1001:
Nice!

As I look at the changes it feels like this should have been three PR's   
1.  darvin handling of mkversion
2. hitagS
3. generic lfops / em4x50 fiddling

We dislike mixed PR's since its hard to follow changes. 

If you wouldn't mind splitting up ?

douniwan5788:
Of course, then how should I proceed? Do I have to cherry-pick to different branches?

PR #2461, opened 2024-08-20
Fix Hitags timing

douniwan5788:


iceman1001:
Awesome!

Thanks for splitting it up

PR #2463, opened 2024-08-21
fix: Hitag S read/write in plain mode

douniwan5788:
I don't have a Hitag card reader, so we need more testing on the Simulation part

### DIFF
--- a/armsrc/hitagS.c
+++ b/armsrc/hitagS.c
@@ -1271,7 +1271,7 @@ static int selectHitagS(const lf_hitag_data_t *packet, uint8_t *tx, size_t sizeo
 
     //select uid
     txlen = 0;
-    cmd = 0x00;
+    cmd = 0x00; // 00000 SELECT UID
     txlen = concatbits(tx, txlen, &cmd, 8 - 5, 5);
     txlen = concatbits(tx, txlen, rx, 0, 32);
     uint8_t crc = CRC8Hitag1Bits(tx, txlen);
@@ -1450,7 +1450,7 @@ void ReadHitagS(const lf_hitag_data_t *payload, bool ledcontrol) {
 
         //send read request
         size_t txlen = 0;
-        uint8_t cmd = 0x0c;
+        uint8_t cmd = 0x0c; // 1100 READ PAGE
         txlen = concatbits(tx, txlen, &cmd, 8 - 4, 4);
         uint8_t addr = pageNum;
         txlen = concatbits(tx, txlen, &addr, 0, 8);
@@ -1555,7 +1555,7 @@ void WritePageHitagS(const lf_hitag_data_t *payload, bool ledcontrol) {
     //send write page request
     txlen = 0;
 
-    uint8_t cmd = 0x08;
+    uint8_t cmd = 0x08; // 1000 WRITE PAGE
     txlen = concatbits(tx, txlen, &cmd, 8 - 4, 4);
 
     uint8_t addr = payload->page;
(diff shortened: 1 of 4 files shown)


armsrc/hitagS.c
```diff
@@ -1271,7 +1253,7 @@ static int selectHitagS(const lf_hitag_data_t *packet, uint8_t *tx, size_t sizeo
 
     //select uid
     txlen = 0;
-    cmd = 0x00;
+    cmd = 0x00; // 00000 SELECT UID
```
iceman1001, line 1256: don't we have DEFINE's  for the commands in protocol.h ?

include/hitag.h
```diff
@@ -23,21 +23,25 @@
...
-    RHT2F_TEST_AUTH_ATTEMPTS  = 25,
-    RHT2F_UID_ONLY            = 26,
-    WHT2F_PASSWORD            = 27,
+    RHT2F_AUTHENTICATE,
+    RHT2F_CRYPTO,
+    WHT2F_CRYPTO,
+    RHT2F_TEST_AUTH_ATTEMPTS,
+    RHT2F_UID_ONLY,
+    WHT2F_PASSWORD,
     HT2_LAST_CMD              = WHT2F_PASSWORD,
```
iceman1001, line 45: I think having the values brings more clarity to the code than removing it and let people guess.   
Not every one knows the exact behavior of enum assigned values.

armsrc/hitagS.c
douniwan5788: It seems that hitag2/S commands are missing in protocol.h. Should I create another PR to move all commands define to protocol.h?

include/hitag.h
douniwan5788: I don't think we should encourage people to use the values of the enum directly since they are subject to change over time. Perhaps it's better to make the values inconspicuous.

armsrc/hitagS.c
iceman1001: yeah,  that would be great

iceman1001:
Great love made to the hitag S, 
Since we are changing the clock used,  it will be confusing when looking at the hitag 2 code. 

1. Would you mind writing a bit more about the clock change and why its done. 
2. Look at hitag 2 and see if it benefits from same change?

douniwan5788:
> Great love made to the hitag S, Since we are changing the clock used, it will be confusing when looking at the hitag 2 code.
> 
> 1. Would you mind writing a bit more about the clock change and why its done.
> 2. Look at hitag 2 and see if it benefits from same change?

1. The original clock `TIMER_CLOCK1 = MCK/2` will overflow above `341*T0 (~2.7ms)`, but the Hitag protocol has to handle longer durations (like `t_prog min 716*T0`), so you have to keep track of overflows. `TIMER_CLOCK3 = MCK/32 `gives us a longer overflow time of `5461*T0 (~43ms)`, which is longer than the maximum time (`5000*T0`) used in the protocol. Therefore, normally, we don't have to handle overflows anymore. I can see from the comment that @Philippe Teuwen already tried to switch to `TIMER_CLOCK3` 3 years ago; I just finished his work.
2. I'll do my best. Unfortunately, I don't have a HITAG 2 tag or reader, so I can't test my code directly. I'll start by refactoring and completing HITAG S functions and then see which parts can be reused for HITAG 2.

doegox:
Don't hesitate to complete or adjust https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/clocks.md if needed, thanks!

PR #2466, opened 2024-08-23
Move Hitag cmds to protocols.h

douniwan5788:


iceman1001:
Nice,  
but some issues,    It doesn't like assigning arrays to individual bytes in a array.
`arr[0] = { 0x00 }  ,  should be  arr[0] = 0x00`
or 
`arr[] = { 0x00 }`

image

douniwan5788:
> Nice, but some issues, It doesn't like assigning arrays to individual bytes in a array. `arr[0] = { 0x00 } , should be arr[0] = 0x00` or `arr[] = { 0x00 }`
> 
> image



> Nice, but some issues, It doesn't like assigning arrays to individual bytes in a array. `arr[0] = { 0x00 } , should be arr[0] = 0x00` or `arr[] = { 0x00 }`
> 
> image

Ah, I got it wrong. uint8_t tx[1] defines an array of size 1. I mistakenly thought it assigned to index 1 of tx (replacing too many t[0], which froze my brain) and was wondering why the wrong command could retrieve the UID.

iceman1001:
No worries,    
I am happy Hitag S gets some love.   Thanks!

PR #2467, opened 2024-08-23
add: clone EM410x ID to Hitag S/8211

douniwan5788:


iceman1001:
Thats interesting!

CiRIP:
could this present issues when compiling the device code without hitag support?

iceman1001:
All code which can be turned off and on with compiler directives should be handled in code.   This might have change now,  
@douniwan5788    may have a second look to verify its all guarded correct. 

We should also add this  `hitag S` to  `lf search`  now that it works again

douniwan5788:
> could this present issues when compiling the device code without hitag support?
> 
> edit: never mind, just saw it was handled on the client side, so won't affect compiling the device side. it would mean however that hitag cloning would be unavailable for standalone modes. electra support is also missing.

Thanks for pointing that out! It won't affect during compilation, but it will go wrong when running the command with SKIP_HITAG. Fixed in #2470

douniwan5788:
> All code which can be turned off and on with compiler directives should be handled in code. This might have change now, @douniwan5788 may have a second look to verify its all guarded correct.
> 
> We should also add this `hitag S` to `lf search` now that it works again

I'm still figuring out how to differentiate Hitag series properly. Once I'm done cleaning up the Hitag 1/2/S code, I believe I will find a way to add the Hitag series back to `lf search`.

PR #2468, opened 2024-08-23
Add Sofia public transport system keys (before January 2024)

user890104:
The anonymous multi-use ticket keys are verified to be static between such cards. I was not able to verify the other two (yellow and blue subscription card) as I only have one of each.


iceman1001:
Thanks!
usually you don't have to comment each sector keys,  but its good as is

PR #2469, opened 2024-08-24
USB enumeration / serial number writeup / docs

henrygab:
Created as a strawman response to iceman's
suggestion that this would be useful in the repo.

Feel free to move it elsewhere / edit, as you prefer.


iceman1001:
Great piece of information!

Thanks!

PR #2471, opened 2024-08-24
fix: clone a EM410x ID to Hitag S/8211 with clock

douniwan5788:


client/src/cmdlfem410x.c
```diff
@@ -758,23 +747,50 @@ static int CmdEM410xClone(const char *Cmd) {
...
+                        case 64: packet.data[1] |= 0x20; break;
+                    }
+                    packet.data[2] = 0;
+                    packet.data[3] = 0; //TODO: keep PWDH0?
+                    page = 4;
+                    break;
+                case 4:
+                    memcpy(packet.data, &data[HITAG_BLOCK_SIZE * 0], HITAG_BLOCK_SIZE);
+                    page = 5;
+                    break;
```
iceman1001, line 785: Confusing use of  page variable. 

Needs a better solution.

douniwan5788: Changes to `steps`, is that more clear?

iceman1001: :)
a normal `i` would have been enough ,  but its clearer now. 

thanks!

PR #2472, opened 2024-08-25
Refactor Hitag S operations into separate submenu

CiRIP:
Rationale: Hitag S is sufficiently different from other Hitags that it requires special treatment for every operation. Having a mutually-exclusive set of arguments (--ht2, --hts, etc.) for a command is counter-intuitive, so moving Hitag S, 1 and 2 operations into separate submenus makes sense (see also: `lf em`).


CiRIP:
future work: unify `hitags_config_t` with `struct hitagS_tag`

iceman1001:
Yeah,   the twist everything into same command category "lf hitag"  is ruff.     So seperating it is a good thing. Makes easier to follow and change code.    All good. 
With that said,  the choice of   `lf hitag s` as its own sub category is less good from user perspective.
We use `-s`  as a param for save file.   `lf hitag s dump -s`  would look like a typo.

With `lf em` it was easier since the card family has simple short names. 

A.  New second level category 
```
   lf hitag2 
   lf hitag1 
   lf hitags  
```

B.  longer third level names
   ```
lf hitag ht2
lf hitag ht1
lf hitag hts
   ```

@doegox   @mwalker33    any suggestions?

client/src/cmdlfhitags.c
```diff
@@ -0,0 +1,344 @@
...
+            .ttf_mode       = (config_bytes[1] >> 2) & 0x03,
+            .lock_config    = (config_bytes[1] >> 1) & 0x01,
+            .lock_key       = (config_bytes[1] >> 0) & 0x01
+    };
+    return result;
+}
+
+void hitags_config_print(hitags_config_t config) {
+    PrintAndLogEx(INFO, " Memory type...... " _GREEN_("%s"),
+                  (const char *[]) {"Hitag S 32", "Hitag S 256", "Hitag S 2048"}[config.memory_type]);
```
douniwan5788, line 327: unknown type `11` will overflow the array

CiRIP:
Yeah, `lf hitag s dump -s` looks a bit dumb. My vote would be with `lf hitag hts`.

iceman1001:
ok,   go for alternative two.    `lf hitag ht2`  style

CiRIP:
fixed

PR #2476, opened 2024-08-26
Actually fix ELECTRA parsing

CiRIP:
I hastily pushed the last commit without checking... Here it is _actually_ fixed.


iceman1001:
No worries,   that me merging without testing properly :)

PR #2481, opened 2024-08-30
fix: bitstream_index_map overflow

douniwan5788:


douniwan5788:
> Initial thought is that the enum is handled in a strange a way. You have none, min, max, count. Feels a bit overworked.

When I am searching for `FPGA_BITSTREAM_HF_15` used in the source code, there is some code that doesn't really care about `ISO 15693`, but rather verifies the argument. So, I think `MIN/MAX` should be more appropriate semantically. `FPGA_BITSTREAM_UNKNOWN` is just a placeholder to initialize some variables, used when there is no bitstream loaded. `FPGA_CONFIG_COUNT` can be replaced by `FPGA_BITSTREAM_MAX + 1`, but why bother?

PR #2483, opened 2024-09-02
Update iclass.c legacy recover error logging

Antiklesys:
Improved error logging to better understand what's happening


iceman1001:
Dont forget to remove it when you find the bug.
I doubt that many people uses this command to be annoyed with the extra logging.  So we be fine merging it.

Antiklesys:
This logging shouldn’t interfere with normal operations, just on the
function startup will provide a few messages of what it’s doing or is
failing

PR #2485, opened 2024-09-04
refactor: rename iso1444a Decoding functions

douniwan5788:


iceman1001:
Sorry,  but why are you renaming functions?

douniwan5788:
> Sorry,  but why are you renaming functions?  

These functions are used outside of `iso14443a.c`. I think it's clearer to have '14a' in the name.

PR #2503, opened 2024-09-13
mkversion: Fix readme location

elboulangero:
While building proxmark3 in an environment without `git` installed, I noticed the following errors on stderr:

```
mkversion create test                   env: 'git': No such file or directory
stat: cannot statx '../README.md': No such file or directory
```

The first line "env: 'git': No such file or directory" is expected, as git is not installed.

However, the second line "stat: cannot statx '../README.md': No such file or directory" is a real error: we're running from top-level directory, so the correct path for README.md is just "README.md".


iceman1001:
Nice catch!

PR #2510, opened 2024-09-15
Fix Hitag S get uid

douniwan5788:


armsrc/hitagS.c
```diff
@@ -1226,7 +1211,7 @@ static int hts_select_tag(const lf_hitag_data_t *packet, uint8_t *tx, size_t siz
     hts_send_receive(tx, txlen, rx, sizeofrx, &rxlen, t_wait, ledcontrol, true);
 
     if (rxlen != 32) {
-        DBG DbpString("UID Request failed!");
+        DbpString("UID Request failed!");
```
iceman1001, line 1214: place back the DBG guard

iceman1001:
Looks like you fixing simulation code,   rdbl code ,  em41x0 code  in one PR.   I don't see get_uid being fixed?  

Regarding DBG guard:
We are not using the device side to actively return strings and inform the user since long ago. 
The hitag code base hasn't been updated to match that.  We put the Extended debug around them otherwise they will blurt out in the client like in `lf search` when not a hitag tag is used.   

The preferred way to communicate is by sending status packages to the client, which will handle the printing

douniwan5788:
> Looks like you fixing simulation code, rdbl code , em41x0 code in one PR. I don't see get_uid being fixed?
> 
> Regarding DBG guard: We are not using the device side to actively return strings and inform the user since long ago. The hitag code base hasn't been updated to match that. We put the Extended debug around them otherwise they will blurt out in the client like in `lf search` when not a hitag tag is used.
> 
> The preferred way to communicate is by sending status packages to the client, which will handle the printing

OK, let's fix it for good

PR #2513, opened 2024-09-16
refactor: switch hitagS_tag to union

douniwan5788:


iceman1001:
`reflect64  vs  REV64`
We kept all the REVxx defines in ht2.   Not sure you should replace them with reflect64 call.    Could become a timing issue.

then the uid_le  stuff...    same goes there,  use the same style as in ht2  or we get a very confusing code base

douniwan5788:
> `reflect64 vs REV64` We kept all the REVxx defines in ht2. Not sure you should replace them with reflect64 call. Could become a timing issue.
> 
> then the uid_le stuff... same goes there, use the same style as in ht2 or we get a very confusing code base

Sure, I'm cleaning up the Hitag S code now. All the `REV*` will be replaced with `reflect*` in the end. After that, I'll clean up the Hitag 2 code to reuse most of them.

PR #2520, opened 2024-09-17
refactor: hitag S cleanup

douniwan5788:


iceman1001:
Problem with Hitag device side code is that structure which reserves much ram.  
This needs to be refactored to use Bigbuff_calloc instead.

douniwan5788:
> Problem with Hitag device side code is that structure which reserves much ram. This needs to be refactored to use Bigbuff_calloc instead.

I will look into it.

armsrc/hitagS.c
```diff
@@ -387,21 +402,11 @@ static void hts_stop_clock(void) {
...
-        resp[i] = (rx[i / 8] >> (7 - (i % 8))) & 0x1;
-    }
-
-    for (int i = 0; i < 32; i++) {
-        ans += resp[5 + i] << (31 - i);
-    }
-
     // global var?
-    temp_uid = ans;
+    concatbits((uint8_t *)&reader_selected_uid, 0, rx, 5, 32);
```
n0emis, line 406: I am not entirely sure what is going on here, but when using the simulator, this instruction sets the variable `m` to 0. So maybe some out-of-bounds write or something?

douniwan5788: Appreciate the heads-up, concatbits erased out-of-bounds, i'll fix in a new PR

PR #2524, opened 2024-09-20
Update LUA from v5.2 to v5.4

jmichelp:
Many changes needed to be made which are kept in individual commits:
- Update bundled library to LUA 5.4.7
- Update build scripts and CI workflow to use liblua5.4 (system)
- Update `lualibs/ansicolors.lua` which was written as a lua5.1 module
- Update `lualibs/dkjson.lua` to a more recent version, also adding support to utf-8
- Patch existing LUA scripts to use `table.unpack()` instead of `unpack()`
- Regenerate SWIG wrapper
- Remove use of deprecated `lua_checkunsigned()` and `luaL_pushunsigned()`  in `client/src/scripting.c` to use the APIs
- Backport LUA `bit32` core module which has been deprecated in favor of native support for bitwise operations (but they also moved from 32bit to 64bit integers at the same time!)


iceman1001:
Wow!!!

lets see how much we break!

PR #2526, opened 2024-09-21
fix: `lf search` hangs

douniwan5788:


iceman1001:
Good catch,   
I encountered this hanging yesterday

PR #2534, opened 2024-09-23
Adding hf 14a simaid

evildaemond:
Adding the `hf 14a simaid`, this is heavily based on `hf 14a sim` however, has been pulled out of the original and modified to speed up processing of responses during a simulation. 

### Description
Simulate ISO/IEC 14443 type A tag with 4,7 or 10 byte UID, and filter for AID Values. These AID Values can be responded to and include extra APDU commands on GetData after response. The Simulation will break once the GetData has been replied to, or if the reader sends a Deselect command. 

#### Example 1: 
Simulate a DESFire tag, and if an AID of `a000000000000000000000` appears, respond with `9000`, and on the next GetData request, respond with a `9000`
`hf 14a simaid -t 3 --aid a000000000000000000000 --response 9000 --apdu 9000`

#### Example 2: 
Simulate a DESFire tag with a custom RATS, and if a GetData request is received, respond with a `9000`
`hf 14a simaid -t 3 --rats 05788172220101--apdu 86009000`

#### Example 3: 
Simulate a DESFire tag, and if any AID Values are sent by the reader in a Select command, print them in console
`hf 14a simaid -t 3 -x`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Added detection for FM11NT021 (@iceman1001)
 - Added detection of a magic NTAG 215 (@iceman1001)
 - Fixed hardnested on AVX512F #2410 (@xianglin1998)
+- Added `hf 14a aidsim` - simulates a PICC (like `14a sim`), and allows you to respond to specific AIDs and getData responses
 
 ## [Backdoor.4.18994][2024-09-10]
 - Changed flashing messages to be less scary (@iceman1001)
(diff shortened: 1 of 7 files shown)


iceman1001:
Nice one!  

for future improvements the struct should be somewhere in common.

doegox:
Hello, thank you!

Some remarks for your next PRs:
* In the commit messages, please add a description of your changes right on the first sentence. Because when listing new commits, git tools only show the first sentence. See 
image

* Group commits per added functionality, not one commit per file changed. Sometimes when a bug occurs we track down the bug with `git bisect` and having many non-compilable commits hurts. Each commit should be compilable (unless buggy, that can happen of course)

Thanks!

PR #2536, opened 2024-09-23
refactor: cleanup PacketResponseNG init

douniwan5788:


iceman1001:
I actually deliberately set all the fields to zero manually because the shortcut version of assigning zero doesn't always clear out all memory.

iceman1001:
You could do a  "memset(payload,  0,  sizeof, "  instead.

douniwan5788:
> I actually deliberately set all the fields to zero manually because the shortcut version of assigning zero doesn't always clear out all memory.

According to the C99 Standard 6.7.8.21, all remaining fields will be initialized. Do you have a reproducible case? Could it be an old compiler bug?
> If there are fewer initializers in a brace-enclosed list than there are elements or members of an aggregate, or fewer characters in a string literal used to initialize an array of known size than there are elements in the array, the remainder of the aggregate shall be initialized implicitly the same as objects that have static storage duration.

iceman1001:
Hard core experience.   So we are not doing this way.

douniwan5788:
> memset(payload, 0, sizeof

ok, I'll replace it with the `memset(payload, 0, sizeof)` form. Will that reopen the PR?

PR #2540, opened 2024-09-25
Fixed argument count for SimulateIso14443aTag() in hf_young.c

jakkpotts:
SimulateIso14443aTag() now requires five arguments, added NULL to all calls in hf_young.c


iceman1001:
Good catch!

PR #2543, opened 2024-09-30
Completed hf iclass unhash pre-image generation

Antiklesys:
Finished the pre-image generation algorithm to take in account edge scenarios based on special values of z0..z7 that return the same modulo values even tho they have different original values. This allows the generation of more than 8 pre-images when reversing hash0, showing the full set of valid pre-images per each diversified key.


iceman1001:
Awesome!

PR #2544, opened 2024-09-30
fix hf mfu pwdgen for 7 byte uid

ANTodorov:
The last 3 bytes of the 7 byte uid are zeroed out
Tested with the example from the forum http://www.proxmark.org/forum/viewtopic.php?pid=44238#p44238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Fixed `hf mfu pwdgen` for the 7 byte UID (@ANTodorov)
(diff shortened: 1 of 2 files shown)


iceman1001:
I believe the correct flow would be  if  uidlen == 4 then zero out and adapt uidlen.

ANTodorov:
It was my initial fix, but then I've saw that the uuid is initialized with zeroes at the creation
So I thought there is no need to zero-out something that is already zero. tested with a 4 byte uid on the command line:
```
[usb] pm3 --> hf mfu pwdgen --uid 04CAB22A
[=] -----------------------------------
[=]  UID 4b... 04 CA B2 2A 
[=]  UID 7b... 04 CA B2 2A 00 00 00 
[=] -----------------------------------
```
If you prefer the exact check against 4-byte I'll push an update asap.

ANTodorov:
ah, uid len is not updated... I see yep. updating

iceman1001:
Good catch!

PR #2546, opened 2024-10-01
Fix missing line return when viewing dumps

jmichelp:
For some reason, `isprint()` claims that `0xff` is printable. But it's used by `PrintAndLog` functions as a marker value to suppress the line return.

So when viewing a dump where the last byte of a block/sector is `0xff`, it was suppressing the new line between blocks/sectors. This fixes it by excluding `0xff` from printable values.


iceman1001:
Great catch!

I wouldn't have thought that one.   the 0xFF as a marker might not be documented.

PR #2549, opened 2024-10-01
unhash hints for next steps

Antiklesys:
Added hints for next steps after unhashing


iceman1001:
Well,  maybe some better instructions than that.  

like how to run `hashcat` since I doubt there is a known plain text DES mode for it.

PR #2556, opened 2024-10-04
fix: `lf hitag hts rdbl` read by page

douniwan5788:


iceman1001:
Nice!

How about you use #defines  for  the reasons codes.   

This is the preferred style

`#define    PM3_REASON_xxx   -4`

douniwan5788:
That's my plan! But I think we should only set `#define PM_REASON_UNKNOWN -1` in pm3_cmd.h. Let each command decide on their own headers for other reason values.

iceman1001:
I prefer having one place where you define it.   If all commands going to have their own,  its gonna be a nightmare maintaining.

PR #2564, opened 2024-10-07
Fix SimulateIso14443aInit usage

archi:
Added NULL as `uint8_t *iRats` to all invokations in armsrc/Standalone.

Not sure if that is the right thing to do: The actual code checks for the presence of a special flag, but does not contain a check for NULL. IMHO adding a NULL check in `iso14443a.c` line 1259 wouldn't hurt, and could catch programming errors (flag set by accident, but NULL passed).

Maybe @evildaemond wants to verify whether my fix does the right thing or if I missed something (they added the iRATs to `iso14443a.h` with 56324b16b2).


evildaemond:
Hey @archi,

No issues on my end, that's fine, catches that issue, had a chat with nvx regarding this during a conference, and identified the same issue, just adding a NULL should cover that in this case.

PR #2566, opened 2024-10-09
Add mf_backdoor_dump script

Aptimex:
Simple python script that uses the backdoor keys from `fm11rf08s_recovery.py` script to quickly dump the tag contents that those keys can directly read. 

Really handy if you just want to read the tag data without waiting 15 minutes to crack the real keys.


doegox:
@Aptimex you're testing the 3 keys but key 518B3354E760 is for a 4k card, so you should adapt TOTAL_SECTORS when this key is found, or based on SAK fingerprinting (18)

Aptimex:
> @Aptimex you're testing the 3 keys but key 518B3354E760 is for a 4k card, so you should adapt TOTAL_SECTORS when this key is found, or based on SAK fingerprinting (18)

Ah thanks for pointing that out, I'll do that. Hadn't gotten that far in your (awesome) paper yet :)

PR #2571, opened 2024-10-14
fm11rf08s_recovery.py: option to keep generated dictionaries after processing; save dump and key to the path shown in `prefs show`

Eltrick:
Both of these annoyed me way too much so we're here now


doegox:
thanks :)

PR #2576, opened 2024-10-16
fix em4x50_read function

ANTodorov:
affected `lf search` and `lf em 4x50 rdbl -b <blk>`

```
[usb] pm3 --> lf em 4x50 rdbl -b 3

Thread 10 "WorkerThread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffdefff700 (LWP 2592)]
em4x50_prepare_result (words=0x7fffdeffe310, lwr=3, fwr=<optimized out>, data=<optimized out>)
```

```
[usb] pm3 --> lf search

Thread 10 "WorkerThread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffdefff700 (LWP 3091)]
em4x50_prepare_result (words=0x7fffdeffe340, lwr=32, fwr=<optimized out>, data=<optimized out>)
    at /home/ant/src/RF/Proxmark3/pm3/client/src/cmdlfem4x50.c:40
40	            words[i].byte[j] = data[i * 4 + (3 - j)];
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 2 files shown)


iceman1001:
Question is why we don't use the struct any longer...

PR #2577, opened 2024-10-17
feat: detect march and mcpu

Cryolitia:
The compiler don't support `march=native` / `mcpu=native` on risvc64, either. This fix the compile failure on riscv64.

Reported-by: https://archriscv.felixc.at/.status/log.htm?url=logs/proxmark3/proxmark3-4.18589-1.log

Refactor the way that detect whether `march=native` / `mcpu=native` are supported on the platform, following the comment.

Add an option `DONT_BUILD_NATIVE` to suppress building with `march=native` / `mcpu=native`. For building it someone-self, it is helpful to generate the fastest binary, but for Linux or some other operating system distribution package maintainer, it is over optimized. We may build and package this software on a machine that is `x86_64-v3` but distribute it to a user that using an old hardware like `Intel Atom`, and result in a `SIGILL`.

Already tested it on `x86_64` and `riscv64`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added an Makefile variable `DONT_BUILD_NATIVE` in mfd_aes_brute Makefile to easify downstream package
+- Auto detect whether compile option `march=native` is supported for mfd_aes_brute Makefile
 - Changed `hf mf sim` - support data-first and nested reader attacks (@doegox)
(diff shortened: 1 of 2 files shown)


iceman1001:
Awesome!    It sure will help out the package maintainers a lot.

nekorouter:
Since I just have a Proxmark3-Easy and a RISC-V device (LicheePi 4A) on hand, I thought I can check if pm3 can run on riscv environment, and it runs good.

baesd on commit https://github.com/RfidResearchGroup/proxmark3/commit/61a993de826335db2d92d1f60c3ef4ae64439b73 ,build with ```PLATFORM=PM3GENERIC```, ```LED_ORDER=PM3EASY``` settings.

Screenshot 2025-06-23 at 15 15 46

Screenshot 2025-06-23 at 15 12 36

Screenshot 2025-06-23 at 15 11 47


I did some simple check, ```pm3-flash-fullimage``` work; ```mf autopwn``` M1 card and ```15 dump``` SLIX card work; ```15 sim``` 15693 card and read using flipper zero work.

I hope those info may help, have a good day!

PR #2579, opened 2024-10-17
improve Hitag S output

douniwan5788:


client/src/cmdlfhitaghts.c
```diff
@@ -37,28 +37,45 @@
 static int CmdHelp(const char *Cmd);
 
 static const char *hts_get_type_str(uint32_t uid) {
+    // source 1: https://www.scorpio-lk.com/downloads/Tango/HITAG_Classification.pdf
+    // IDE Mark
+    // Each HITAG chip contains an unique Device Identifier (IDE ) so called a Serial Number.
```
iceman1001, line 42: Unique Identifier  ( UID )  4 bytes.   its not IDE..

image

client/src/cmdlfhitaghts.c
```diff
@@ -309,12 +326,93 @@ static int CmdLFHitagSRead(const char *Cmd) {
...
         }
-        if (card->pages_reason[i] >= 0)
-            PrintAndLogEx(SUCCESS, "%02u | %s", page_addr, sprint_hex_ascii(card->pages[i], HITAGS_PAGE_SIZE));
-        else
+        if (card->pages_reason[i] >= 0) {
+            PrintAndLogEx(SUCCESS, "% 3u | %s  | " NOLF, page_addr, sprint_hex_ascii(card->pages[i], HITAGS_PAGE_SIZE));
+
+            // access right
+            if (page_addr == HITAGS_UID_PADR) {
+                PrintAndLogEx(NORMAL, _RED_("RO  ")NOLF);
```
iceman1001, line 334: _RED_  for  `read only`  might be a bit too harsh.    How about we just let it be normal white?

iceman1001:
Good improvements,     this needs to be a bit adjusted to fit into the pm3 client style.

client/src/cmdlfhitaghts.c
douniwan5788: Simply copy/paste from the unofficial HITAG_Classification.pdf. I'm confused about the strange abbreviation "IDE" too. I can't find anything on the internet about the "IDE mark" except in the PDF itself. I don't even know if it's really applicable to Hitag S.

client/src/cmdlfhitaghts.c
douniwan5788: I'm try to following the lf hitag read --ht2 color scheme. Read-only is usually normal for the UID page. However, for other pages, it means they are locked.

client/src/cmdlfhitaghts.c
```diff
@@ -37,28 +37,45 @@
...
-            return "PCF 7941";
-        case 7:
-            return "PCF 7952";
-        case 9:
-            return "PCF 7945";
-        default:
-            return "";
-    }
+    uint8_t pid0 = uid & 0xF;
+    uint8_t pid1 = (uid >> 4) & 0xF;
```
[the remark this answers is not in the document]
douniwan5788: Sweet, I didn't notice that. Handy!

client/src/cmdlfhitaghts.c
iceman1001: I never heard of IDE mark too.

client/src/cmdlfhitaghts.c
iceman1001: We need a better unified color scheme.

PR #2580, opened 2024-10-19
lf hitag hts: use correct byteorder for authentication

n0emis:
tested with a hitagS transponder.

one could also reverse rnd when initializing the state, but this way it is in the same order as previously

### DIFF
--- a/armsrc/hitagS.c
+++ b/armsrc/hitagS.c
@@ -70,7 +70,7 @@ static uint32_t reader_selected_uid;
 static int rotate_uid = 0;
 static int sof_bits;                                // number of start-of-frame bits
 static uint8_t pwdh0, pwdl0, pwdl1;                 // password bytes
-static uint8_t rnd[] = {0x74, 0x12, 0x44, 0x85};    // random number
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice,
now @douniwan5788   just need to verify it and we can merge

douniwan5788:
It is indeed in the same order as previously, except for `rnd`. Please reverse the init value of `rnd` instead to maintain the exact same `ks` as before.
https://github.com/RfidResearchGroup/proxmark3/blob/db1ce081ab0e396f0e6d7bf221e6ef15778d8e00/armsrc/hitagS.c#L73

douniwan5788:
@n0emis Indeed, you're correct. I overswapped the byte order. In addition to swapping the `rnd` byte order. What do you think about modifying all calls to `ht2_hitag2_init` to match this pattern instead?
`state = ht2_hitag2_init(REV64(tag.data.s.key_le), REV32(tag.data.s.uid_le), REV32(*(uint32_t *)rx));`

n0emis:
@douniwan5788 done.

works with a normal hitagS tag and in simulation mode (though a further patch is needed for that. will submit in a seperate pr)

douniwan5788:
LGTM

PR #2581, opened 2024-10-20
Updated hf iclass legrec arm side and added custom key to dictionary

Antiklesys:
Updated hf iclass legrec arm functionality
Added new custom standard key to the repository

Todo: Improve keygen algorithm efficiency

### DIFF
--- a/armsrc/iclass.c
+++ b/armsrc/iclass.c
@@ -2176,77 +2176,20 @@ void generate_single_key_block_inverted(const uint8_t *startingKey, uint32_t ind
 void iClass_Recover(iclass_recover_req_t *msg) {
 
     bool shallow_mod = false;
-
-    LED_A_ON();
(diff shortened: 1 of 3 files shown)


iceman1001:
14b reader call inside iclass info...      I am quite certain iclass loads the hf15 image which doesn't have support 14b.
Now we need to look at arm side, context switching inside the device during a command is not optimal at all. 
Needs to be verified before merging.

Great stuff with the legbrute improvements!

Antiklesys:
> 14b reader call inside iclass info... I am quite certain iclass loads the hf15 image which doesn't have support 14b. Now we need to look at arm side, context switching inside the device during a command is not optimal at all. Needs to be verified before merging.
> 
> Great stuff with the legbrute improvements!

Thanks! Feel free to do a partial merging if you want to take more time to check on the other part.
No change to legbrute, this is legrec :) More code to come in the next days.

PR #2591, opened 2024-10-26
hitags sim: apply correct coding after select(uid)

n0emis:
in the current sim, the anticoll coding is still used after the tag has been selected. this patch ensures, that the corresponding mc coding is used after select(uid)

### DIFF
--- a/armsrc/hitagS.c
+++ b/armsrc/hitagS.c
@@ -526,6 +526,20 @@ static void hts_handle_reader_command(uint8_t *rx, const size_t rxlen,
 
                     tx[4] = crc;
                 }
+                switch (tag.mode) {
+                    case HT_STANDARD:
+                    case HT_ADVANCED: {
+                        m = MC4K;
+                        break;
+                    }
+                    case HT_FAST_ADVANCED: {
+                        m = MC8K;
+                        break;
+                    }
+                    default: {
+                        break;
+                    }
+                }
             }
             break;
         }


douniwan5788:
The issue should be fixed in https://github.com/RfidResearchGroup/proxmark3/pull/2587. Could you retest in the latest master branch?

n0emis:
Hmm, not sure what is going on here. Either this breaks the simulator fully or I'm holding it wrong

n0emis:
okay, there was an issue with the timing, see #2595

but the coding is still not working right.

```
[#] hts_set_frame_modulation: set m to 2 (mode 200)
[#] UID request: length: 5 first byte: c0
[#] HT_ADVANCED
[#] hts_set_frame_modulation: set m to 0 (mode 192)
[#] hitag_send_frame: coding 0
[#] hitag_send_frame: (32) 5F C2 11 84
[#] hts_set_frame_modulation: set m to 2 (mode 192)
[#] SELECT
[#] SELECT match
[#] hitag_send_frame: coding 0
[#] hitag_send_frame: (40) CA 00 00 FF
```

somewhere between `hts_set_frame_modulation` and `hitag_send_frame` the modulation changes, but I can't find any reference to `m`. not sure what is happening here

n0emis:
see https://github.com/RfidResearchGroup/proxmark3/pull/2520#discussion_r1818103250

PR #2596, opened 2024-10-27
Delete extra whitespace character in mf_backdoor_dump.py

xueliu:
In this patch an extra whitespace character in mf_backdoor_dump.py will be deleted. 
This blank character prevents the for loop from trying the next backdoor key.


iceman1001:
Never good to depend your loop statement to long output strings from the client since we have a tendency to change it

PR #2602, opened 2024-10-29
Add shower card keys used by Seijsener

stuiterveer:


iceman1001:
Nice,  are these verified to be static across cards or system?

stuiterveer:
For sure! 3 cards from the same location all with the same keys, for a system that's intended to be used by Seijsener across multiple harbors! :)

iceman1001:
Excellent!

PR #2606, opened 2024-11-02
Add JAKCOM R5 smart ring t55x default passwords

shellster:
Bought one of these cool little rings.  It has two integrated t55x clone chips (among others), however they come pre-password protected and the tradition test mode wipes didn't seem to unlock them.  I was able to unlock them with passwords in their manual (http://www.jakcom.com/ins/r5/r5en.html) and so I thought they would be good to add to the dictionary.


iceman1001:
Excellent!

PR #2607, opened 2024-11-02
improve cardhopper to improve reliability when parts of the packet are buffered

nvx:
if packets arrive in quick succession part of the next packet can sit in the buffer which isn't checked by cardhopper resulting in the standalone hanging. this makes it check for data in the internal buffer too preventing that condition.

I expect the same benefit may occur in `recieve_ng` (sic) by changing the `usb_poll_validate_length()` check to be `usb_read_ng_has_buffered_data() || usb_poll_validate_length()` as this does in the cardhopper standalone but it'd need more testing. I've been testing this change with cardhopper for a few months now and I'm confident it's good but haven't tested the same change in `recieve_ng` yet


iceman1001:
Excellent!

PR #2608, opened 2024-11-02
add gdm flag to hf mf c* commands

nvx:
this enables the use of hf mf c* commands with gdm/uscuid cards when the alt wake up mode 20(7)/23 is enabled rather than gen1 wake up


iceman1001:
This is a good one

PR #2610, opened 2024-11-03
clear the prompt using ANSI "clear entire line" escape

nvx:
this replaces the old behaviour of using a fixed length of spaces which can sometimes fail to clear the entire line if it's too long


iceman1001:
Looks better for sure!

Just make sure it works on systems where they disabled ANSI  (ie  ` pref set color --off` )
and I recommend moving the clear line ANSI code to  the `/include/ansi.h`

nvx:
> Just make sure it works on systems where they disabled ANSI (ie ` pref set color --off` ) and I recommend moving the clear line ANSI code to the `/include/ansi.h`

I expect it should, it's used in a few other places so would be worth rejigging the other places to use that define too if moved there

PR #2613, opened 2024-11-04
Fudan FM11R08* full-card recovery script; with Bambu tag support

csBlueChip:
Beyond the original recovery script (used by this script) this script performs a number of tasks to aid us teaching NFC at our local defcon meet.
qv. https://eprint.iacr.org/2024/1275.pdf [Philippe Teuwen]

+ Full breakdown of Block 0 (esp. Fudan specific details)
+ Verify UID/BCC
+ SAK decode
+ Optionally perform online Fudan validation check
+ [[key recovery]]
+ Dump all 64+8 blocks
+ ...with all 32+2 keys
+ Decode ACL [clearer output]
+ [[MAD decode]]
+ Bambu tag decode
+ Create full 64+8 block dump (incl keys) ...for the CTF flag

### DIFF
--- a/client/pyscripts/fm11rf08_full.py
+++ b/client/pyscripts/fm11rf08_full.py
@@ -53,7 +53,7 @@ except ModuleNotFoundError:
 # >> "logfile"
 #==============================================================================
 def startlog(uid,  append = False):
-	global logfile
+	global  logfile
 
 	logfile = f"{dpath}hf-mf-{uid:08X}-log.txt"
 	if append == False:
@@ -74,8 +74,8 @@ def lprint(s,  end='\n', flush=False):
 # >> "keyfile"
 #==============================================================================
 def main():
-	global prompt
-	global p
+	global  prompt
+	global  p
 
 	prompt = "[bc]"
 	p      = pm3.pm3()  # console interface
@@ -94,8 +94,10 @@ def main():
 	getDarkKey()
 	decodeBlock0()
 
-	global keyfile
+	global  keyfile
+	global  mad
 
+	mad     = False
 	keyfile = f"{dpath}hf-mf-{uid:08X}-key.bin"
 	keyok   = False
 
@@ -137,7 +139,7 @@ def main():
 # >> "dpath"
 #==============================================================================
 def getPrefs():
-	global dpath
+	global  dpath
 
 	p.console("prefs show --json")
 	prefs = json.loads(p.grabbed_output)
@@ -158,7 +160,7 @@ def checkVer():
 # >> args.
 #==============================================================================
 def parseCli():
-	global args
+	global  args
 
 	parser = argparse.ArgumentParser(description='Full recovery of Fudan FM11RF08* cards.')
 
@@ -183,8 +185,8 @@ def parseCli():
 '''
 #==============================================================================
 def getDarkKey():
-	global dkey
-	global blk0
+	global  dkey
+	global  blk0
 
 	#          FM11RF08S        FM11RF08        FM11RF32
 	dklist = ["A396EFA4E24F", "A31667A8CEC1", "518b3354E760"]
@@ -196,13 +198,12 @@ def getDarkKey():
 	for k in dklist:
 		cmd = f"hf mf rdbl -c 4 --key {k} --blk 0"
 		print(f"{prompt} `{cmd}`", end='', flush=True)
-		res = p.console(f"{cmd}")
+		res = p.console(f"{cmd}", capture=False)
 		if res == 0:
 			print(" - success")
 			dkey = k;
 			break;
 		print(f" - fail [{res}]")
-		_ = p.grabbed_output
 
 	if dkey == "":
 		print(f"{prompt}")
@@ -219,8 +220,8 @@ def getDarkKey():
 # >> "uids"
 #==============================================================================
 def decodeBlock0():
-	global uid
-	global uids
+	global  uid
+	global  uids
 
 	# We do this early so we can name the logfile!
 	uids = blk0[0:11]                            # UID string  : "11 22 33 44"
@@ -230,7 +231,7 @@ def decodeBlock0():
 	lprint(prompt)
 	lprint(f"{prompt}              UID         BCC         ++----- RF08 ID -----++")
 	lprint(f"{prompt}              !           !  SAK      !!                   !!")
-	lprint(f"{prompt}              !           !  !  ATQA  !!     RF08 Hash     !!")
+	lprint(f"{prompt}              !           !  !  ATQA  !!     Fudan Sig     !!")
 	lprint(f"{prompt}              !---------. !. !. !---. VV .---------------. VV")
 	#                              0           12 15 18    24 27                45
 	#                              !           !  !  !     !  !                 !
@@ -333,12 +334,10 @@ def fudanValidate():
 # If keys cannot be loaded AND --recover is specified, then run key recovery
 # >> "keyfile"
 # >> "key[17][2]"
-# >> mad!
 #==============================================================================
 def loadKeys():
-	global keyfile
-	global key
-	global mad
+	global  keyfile
+	global  key
 
 	key = [[0 for _ in range(2)] for _ in range(17)]  # create a fresh array
 
@@ -361,7 +360,7 @@ def loadKeys():
 # >> "keyfile"
 #==============================================================================
 def recoverKeys():
-	global keyfile
+	global  keyfile
 
 	badrk   = 0     # 'bad recovered key' count (ie. not recovered)
 
@@ -402,7 +401,6 @@ def verifyKeys():
 	global  mad
 
 	badk = 0
-	mad  = False
 
 	lprint(f"{prompt} Check keys..")
 
@@ -417,7 +415,7 @@ def verifyKeys():
 			cmd = f"hf mf rdbl -c {ab} --key {key[sec][ab].hex()} --blk {bn}"
 			lprint(f"{prompt}   `{cmd}`", end='', flush=True)
 
-			res = p.console(f"{cmd}")
+			res = p.console(f"{cmd}", capture=False)
 			lprint(" " * (3-len(str(bn))), end="")
 			if res == 0:
 				lprint(" ... PASS", end="")
@@ -433,9 +431,6 @@ def verifyKeys():
 				lprint(" - MAD Key")
 			else:
 				lprint("")
-
-	# We need to flush all the output we just collected <shrug>
-	_ = p.grabbed_output
 
 	if badk > 0:
 		lprint(f"{prompt} ! {badk} bad key", end='')
--- a/client/pyscripts/fm11rf08_full.py
+++ b/client/pyscripts/fm11rf08_full.py
@@ -1,28 +1,34 @@
 #!/usr/bin/env python3
 
 #------------------------------------------------------------------------------
-# Revision log:
+# Revision log & Licence
 #------------------------------------------------------------------------------
 '''
-1.1.0 - BC  - Major refactor
-1.0.0 - BC  - Initial release
+1.2.0 - BC  - Proxmark3 Submission
 '''
-script_ver = "1.1.0"
+script_ver = "1.2.0"
 
 '''
-This was originally created for my local defcon chapter to aid teaching NFC.
-We wanted a realistic challenge which could NOT be resolved by typing `autopwn`.
-Enter FM11RF08S tags ... which (currently) don't get autopwn'ed.
+Copyright @csBlueChip
 
-We decided the flag would be: The MD5 of the card data,
-but this had to include all 8 documented Dark blocks AND all 32+2=34 Keys.
-Not rocket surgery if you know what you're doing, but non-trivial for someone
-who is still struggling to spell `autoporn` <-- yes, this happened!
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
 
-Ultimately we needed a tool by which to demo everything.
-This is it; and this is me proffering it to the wider community.
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


iceman1001:
Nice one,
We prefer to use GPL3.0 or later as a license in this project,   so if you wouldn't mind?

csBlueChip:
I totally failed to spot the GPL3 licence - clearly monoscopically focused
on getting the script working - <facepalm>

If I change my script to be dual licence (GPL/MIT), will that satisfy your
requirements? ...Or is it a case of "GPL3 or nothing"?

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
+#!/usr/bin/env python3
```
doegox, line 1: I find the script name confusing.
fm11rf08s_recovery provides also a full dump and a key dump, like autopwn
so call this one "demo" or "verbose" but not  "full"

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
...
+	for line in p.grabbed_output.split('\n'):
+		if " | " in line and "# | s" not in line:
+			blk0 = line[10:56+1]
+
+#+=============================================================================
+# Extract data from block 0
+# >> "uid"
+# >> "uids"
+#==============================================================================
+def decodeBlock0():
```
doegox, line 221: it would be interesting to isolate this function in a separate file to be imported, so other ppl can use it in their scripts s well, or call the script with a provided block0

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
...
+	lprint(f"{prompt}   ATQA     : {atqa:04X}")   # show ATQA
+	lprint(f"{prompt}   Fudan ID : {type}")       # show type
+	lprint(f"{prompt}   Fudan Sig: {hash}")       # show ?Partial HMAC?
+	lprint(f"{prompt}   Dark Key : {dkey}")       # show key
+
+#+=============================================================================
+# Fudan validation
+# >> "blk0"
+#==============================================================================
+def fudanValidate():
```
doegox, line 296: it would be interesting to isolate this function in a separate file to be imported, so other ppl can use it in their scripts s well, or call the script with a provided block0

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
...
+# Read all block data - INCLUDING Dark blocks
+# >> blkn
+# >> "data[]"
+'''
+[=]   # | sector 00 / 0x00                                | ascii
+[=] ----+-------------------------------------------------+-----------------
+[=]   0 | 5C B4 9C A6 D2 08 04 00 04 59 92 25 BF 5F 70 90 | \........Y.%._p.
+'''
+#==============================================================================
+def readBlocks():
```
doegox, line 464: redundant with the dump file already provided by recovery()
or call it verifyDump as you have the verifyKeys...

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
...
+#+=============================================================================
+# Patch keys in to data
+# >> "key[][]"
+# >> "data[]"
+# >> keyok!
+'''
+  3 | 00 00 00 00 00 00 87 87 87 69 00 00 00 00 00 00 | .........i......
+'''
+#==============================================================================
+def patchKeys(keyok):
```
doegox, line 511: note that dump from recovery is already patched

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
...
+			lprint(f"{prompt}   :{data[cnt]}")
+
+		cnt += 1
+		if (cnt % 4 == 0) and (n != blkn[-1]):  # Space between sectors
+			lprint(prompt)
+
+#+=============================================================================
+# Let's try to detect a Bambu card by the date strings...
+#==============================================================================
+def detectBambu():
```
doegox, line 570: it would be interesting to isolate this function in a separate file to be imported, so other ppl can use it in their scripts s well, or call the script with a provided block0

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
...
+# Dump bambu details
+# https://github.com/Bambu-Research-Group/RFID-Tag-Guide/blob/main/README.md
+# >> "data[]"
+'''
+      6           18          30          42         53
+      |           |           |           |          |
+  3 | 00 00 00 00 00 00 87 87 87 69 00 00 00 00 00 00 | .........i......
+'''
+#+=============================================================================
+def dumpBambu():
```
doegox, line 602: it would be interesting to isolate this function in a separate file to be imported, so other ppl can use it in their scripts s well, or call the script with a provided block0

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
...
+		Both Key A and Key B can READ the Block
+		Only Key B can WRITE to the Block
+		The block cannot be used as a "counter" because:
+			Neither key can perform increment nor decrement commands
+
+	WARNING: 
+		IF YOU PLAN TO CHANGE ACCESS BITS, RTFM, THERE IS MUCH TO CONSIDER !
+'''
+#==============================================================================
+def dumpAcl():
```
doegox, line 793: it would be interesting to isolate this function in a separate file to be imported, so other ppl can use it in their scripts s well, or call the script with a provided block0.
Note that we already have `hf mf acl`

client/pyscripts/fm11rf08_full.py
csBlueChip: if you dump the card using `recovery` and then dump the card using `full` and then do an `ls` you will see that the (so called) "full" file is 128 bytes longer because (as discussed in the verbose intro) it includes "all 8 documented Dark blocks"
```
proxmark@linux:~$ ls hf-mf-5CB49CA6-dump*.bin -l
-rw-r--r-- 1 prox prox    1152    Nov  4 00:26 hf-mf-5CB49CA6-dump18.bin
-rw-r--r-- 1 prox prox    1024    Oct 30 18:59 hf-mf-5CB49CA6-dump.bin
```
As for the name ...It was called `bc.py` until the last minute, and I just thought "the full 18 blocks" is what I was going after, and grabbed the word "full" ...If you think that name is misleading, then yes, let's pick something better.

client/pyscripts/fm11rf08_full.py
```diff
@@ -0,0 +1,942 @@
...
+	global p
+
+	prompt = "[bc]"
+	p      = pm3.pm3()  # console interface
+
+	getPrefs()
+	checkVer()
+	parseCli()
+
+	print(f"{prompt} Fudan FM11RF08[S] full card recovery")
```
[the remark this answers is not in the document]
csBlueChip: Again I just wanted a way to highlight that this script retrieves the 8 Dark blocks, I am happy to accept a better/clearer message that conveys that.

client/pyscripts/fm11rf08_full.py
csBlueChip: If/when this script gets pulled, I would be flattered if functions were yanked to a library. If/when that happens it would probably be wise to pick a more "API" style name for the function. I dunno, maybe `hf_mf_block0_decode(binaryData[16])` ...but I'm sure you guys probably have a naming convention - so "whatever fits" is good by me

Same goes for the following 3 similar comments :)

client/pyscripts/fm11rf08_full.py
csBlueChip: qv. https://github.com/RfidResearchGroup/proxmark3/pull/2613#discussion_r1830270895

client/pyscripts/fm11rf08_full.py
csBlueChip: qv. https://github.com/RfidResearchGroup/proxmark3/pull/2613#discussion_r1830270895

client/pyscripts/fm11rf08_full.py
csBlueChip: qv. https://github.com/RfidResearchGroup/proxmark3/pull/2613#discussion_r1830270895

client/pyscripts/fm11rf08_full.py
csBlueChip: Considerations:
~ This function reads the 8 Dark blocks in addition to the 64 "normal" blocks (which were read by `recovery()`)
~ `verifyKeys` takes the known keys and verifies that they are still valid. ...because we would like to avoid waiting "<30mins" if possible
~ `readBlocks` deliberately re-reads the card each time it is run, so as to detect edits to the card ...it only takes a few seconds ...I have to read it all to verify nothing has changed, at which point, well, I've just read it all ...There may well be credit to writing a `verifyBlocks()` function to detect changes - but that would need to call `readblock()` and `loadOldFile()`
~ The feature bloat had already crept in when I started detecting and decoding various tags - I just had to draw the line somewhere (for now)

client/pyscripts/fm11rf08_full.py
csBlueChip: qv. https://github.com/RfidResearchGroup/proxmark3/pull/2613#discussion_r1830270895

this function requires data other than block 0

I was originally going to call `hf mf acl` (as I say, this script is initially to support a defcon workshop) and parse the output (if I had to), but in the end I found the output difficult to decipher, and tricky to parse - in the end, some bit twiddling just seemed easier.

client/pyscripts/fm11rf08_full.py
csBlueChip: For reasons discussed above, I do not load the existing `dump.bin` so this action is 'required'

doegox:
> Thank you for taking time to explain your position. I really hadn't considered the potential nightmare of maintaining multiple licences in one repository. I totally understand.

Just to say, I second Iceman comment on the license with this example: I discussed many parts of this script as better being moved into some common MFC helper functions Python module. Now if each helper function had to retain its original license, it becomes a nightmare to constitute such module mixing licenses.

> In case you're wondering: I have both personal and professional reasons to dislike what GPL3 has done to hinder development of/with cross-platform support, and the damage it has done to the security landscape; and I just want to make sure anything I personally release can be used by anybody for any reason and in any way they so desire ...yourself (and this fantastic project) included :)

Fair enough. On our side, we tend to dislike the several manufacturers who stole the proxmark3 code to make their closed-source variants...
 
> How about this?: ~ remove the `print("Licence....")` line from the code ~ add this comment to the top of the script "To minimise administrative overheads on the Proxmark code, this copy of the script is released under the GPLv3 licence. If you require/desire a less restrictive licence, this script may instead be obtained from [a link on my github page] under the MIT licence" ...Does that work for you? :)

IMHO the first sentence is useless.
"The original version of this script is also kept at your disposal under the MIT license at the following link..."

"Original version" is important because once merged, over time, this code will evolve in the repo.
E.g. we can merge it now and tomorrow I'll do the code refactoring into Python module(s).

client/pyscripts/fm11rf08_full.py
doegox: Ha! Initially I didn't make the link when you say "dark".
You use the term "dark" both for the backdoor key (see getDarkKey) and for the advanced verification method blocks.
In MFC, dark refers to darkside (and now the variant darknested which replaces darkside+nested)
and personally I would avoid reusing the same term in other contexts.
* dark key => backdoor key
* dark blocks => advanced verification method blocks (or some shorter avblocks for example)
If the extra blocks are useful to get, better to change the existing commands than adding an entire new script.
Right now I refrained to do so because the existing commands cannot handle dumps of unexpected sizes, so your dump cannot be used by e.g. hf mf restore. We already have the problem with the keys file at the moment.
This is planned to handle properly these oversized dumps and keyfiles, but it takes time.
So we can merge your code but don't be surprised if tomorrow it becomes redundant.

client/pyscripts/fm11rf08_full.py
doegox: if you need to read the card, using "hf mf ecfill -c 4 -k ..." is wayyy faster.
Then, yes, for now, you've to dump manually the 8 avblocks because as said above, it is not yet supported in ecfill or any other mfc emulation command. (I think even letting a "hf mf ecfill --4k -c 4 -k..." dumping and failing over unexisting sectors is still faster)

doegox (approved):

iceman1001:
Here is what I suggest you put on top of your script file.   
And remove the line in Main().

> #
> #  Copyright @csBlueChip
> #
> #   This program is free software: you can redistribute it and/or modify
> #   it under the terms of the GNU General Public License as published by
> #   the Free Software Foundation, either version 3 of the License, or
> #   (at your option) any later version.
> #
> #   This program is distributed in the hope that it will be useful,
> #   but WITHOUT ANY WARRANTY; without even the implied warranty of
> #   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
> #   GNU General Public License for more details.
> #
> #   See LICENSE.txt for the text of the license.
> #
> #  The original version of this script is also kept at your disposal under the MIT license at the following link   <enter link>
> #   or reach out to author for alternative licenses. 
> 

and since we use git,  we don't need a "internal" change log.


When @doegox   is happy,   we merge

doegox:
We can merge, then I'll make the discussed changes (module etc)

doegox:
but yeah change the licensing terms before merge, else it looks like maintainers are rewriting your terms...

client/pyscripts/fm11rf08_full.py
csBlueChip: As I say, the script as it stands solved a very specific problem. And, as with the "library functions" comments, if this script become a part of the (metaphoric) Borg Collective, that's no bad thing.

Re. the word "dark": We needed a word so we could discuss it. Your paper used the phrase "dark keys" and extended nested to "dark nested" ...So we took inspiration, and called the additional blocks (which we discovered from the same paper) "dark blocks" ...Again, if it's confusing - this is exactly the right moment to change it :)

csBlueChip:
> but yeah change the licensing terms before merge, else it looks like maintainers are rewriting your terms...

it is done

iceman1001:
Excellent!
I like the idea for your ctf challenge

Issue #2500, opened 2024-09-11
Unable to read tag with hf_legic.lua


fprackwieser:
When I try to read a tag with the hf_legic script I always get this error.

```
> rt
[+] TYPE: MIM256 card (234 bytes)
[+] Reading tag memory.
[+] Saved 256 bytes to binary file `C:\ProxSpace\pm3/hf-legic-temp-********.bin`
[+] Saved to json file `C:\ProxSpace\pm3/hf-legic-temp-********.json`

256 bytes from C:\ProxSpace\pm3/hf-legic-temp-AdKqeYSu.bin loaded
create virtual tag from 256 bytes

Tag-Type: SAM
1 Segment(s) found
256 bytes for Tag processed
[-] Error - can't find `hf-legic-temp-********.eml`
[-] error - C:\ProxSpace\pm3\proxmark3\client\luascripts/hf_legic.lua:769: bad argument #1 to 'os.remove' (string expected, got nil)

[+] finished hf_legic

[usb] pm3 -->
```

iceman1001:
Aha,   we don't generate `.eml` files any more.   

The script needs to be updated

jmichelp:
@fprackwieser can you confirm the script works for you now that PR #2615 has been merged?

You may surface other bugs but reading should work :)

iceman1001:
Ping,   but closing meanwhile

PR #2615, opened 2024-11-04
Start fixing legic script to work with lu5.4

jmichelp:
I haven't tried all the code yet (mainly because I was initially lacking blank Legic that I could mess with)

Should at least fix issue #2500


iceman1001:
Perfect!

PR #2616, opened 2024-11-04
Fix potential overflow in readblock

jmichelp:
Also replaced hard-coded buffer sizes with the correct macro.


iceman1001:
Nice catch,   definitely better code style

PR #2627, opened 2024-11-11
Update mfulc_default_keys.dic

Septimus4:
Bangkok hotel


iceman1001:
A Mifare Ultralight-C key is 16 bytes long.

Septimus4:
My bad it was a password for a ev1, do we care adding it to [4_byte_words.dic](https://github.com/RfidResearchGroup/proxmark3/blob/master/client/dictionaries/extras/4_byte_words.dic) ?
I see the list was not updated in a long time

iceman1001:
We would add it in  https://github.com/RfidResearchGroup/proxmark3/blob/master/client/dictionaries/t55xx_default_pwds.dic
I know its ambiguous but its because we reuse it with T55XX.

When it comes to your pwd,  do you know its a default one?   ie same pwd on more than one hotel room key?

Septimus4:
Okay noted, thank you.
I should be able to obtain keys for other rooms. I'll investigate further.

iceman1001:
Looking forward to your next PR :)

PR #2632, opened 2024-11-13
add option to build against non-default python3 package

ANTodorov:
Tested in the following docker containers
* opensuse-leap - just `make` is failing, `make PYTHON3_PKGCONFIG=python-3.11` succeed
* opensuse-tumbleweed - just `make` succeed, as expected
* debian-13-trixie - just `make` succeed, as expected

So, the patch is properly activated only when the make variable is defined

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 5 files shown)


doegox:
Thanks but this was due to some recent code of mine that needed to be limited to Python 3.10
This should be fixed now, tested in the docker Opensuse Leap as well.

ANTodorov:
Great!
Anyway, I believe it is a good idea to have an option to select which python3 to use if there are several available on the system.
What do you think?

In case of approval I think only the docs/description needs to be reworded accordingly...

doegox:
fair enough, thanks.

PR #2634, opened 2024-11-14
revert opensuse-leap docker to default Python 3.6

ANTodorov:
@doegox  forgot to restore the opensuse-leap/Dockerfile changes when reworking commit https://github.com/RfidResearchGroup/proxmark3/commit/39c846ada4832a874a1744bc45851b385857f6cb


doegox:
ha yes I didn't notice...

PR #2638, opened 2024-11-17
use all spi flash v0.1

ANTodorov:
* introduced a new communication command CMD_FLASHMEM_PAGES64K to get the number of 64k pages
* "the last page" is a special, holding the dicts and some other stuff, relocated to there
* raised timeout when wiping a mem page (W25Q16 looks a bit slower than W25X20BV)
* loop all pages in Flash_WipeMemory()


common_arm/flashmem.c
```diff
@@ -457,6 +454,38 @@ void Flashmem_print_info(void) {
...
 }
 
+//read spi flash JEDEC ID and fill the global variable spi_flash_p64k
+bool FlashDetect(bool flash_init) {
+    flash_device_type_t flash_device = {0};
+
+    if (flash_init) {
+        if (!FlashInit()) {
+            if (g_dbglevel > 3) Dbprintf("FlashDetect() FlashInit fail");
+            return 0;
```
iceman1001, line 464: don't mix return codes.

common_arm/flashmem.c
```diff
@@ -471,6 +500,12 @@ bool FlashInit(void) {
         return false;
     }
 
+#ifndef AS_BOOTROM
+    if (spi_flash_p64k == 0){
+        if (!FlashDetect(false)) return 0;
```
iceman1001, line 505: use same style with Curly Brackets

include/pmflash.h
```diff
@@ -37,10 +37,16 @@
 #ifndef FLASH_MEM_MAX_SIZE
 # define FLASH_MEM_MAX_SIZE     0x40000  // (262144)
 #endif
+#ifndef FLASH_MEM_MAX_SIZE_P
+# define FLASH_MEM_MAX_SIZE_P(p64k) (1024 * 64 * p64k)
```
iceman1001, line 41: use extra parenthesis around params in defines.   ie:  `(1024 * 64 * (p64k))`

iceman1001:
Nice PR!  with some minor things to sort out.

iceman1001:
the only thing left is how we transition from old to new seamless...

Everyone with a rdv4 would need to "re-wipe" their spiffs and run init script to make it work?

ANTodorov:
Just checked with my RDV4. It is with 256k spi flash, so no change - all offsets are same even the signature check is OK:
image
(tested t55xx with password too, no problem using the dict from the flash)

For the PM3 Easy's with the 2MB spi flash there will be a need to wipe page 0..3 - the "old" format and re-upload the dictionaries. But I believe the owners that had built PM3GENERIC+FLASH are enough tech competent to do the migration...

ANTodorov:
Hmm in the updated version the re is no "protection". I think I should update the page wiping boundaries in the code... just e second
> * to erase it:
>   * **Beware** it will erase your flash signature so better to back it up first as you won't be able to regenerate it by yourself!
>   * edit the source code to enable Page 3 as a valid input in the `mem wipe` command.
>   * Updating keys dictionaries doesn't require to erase page 3.
>

PR #2639, opened 2024-11-18
fix failed spi flash detection

ANTodorov:
fallback to 4 pages when SPI IC is unknown
some code styling
remove some unused code

### DIFF
--- a/common_arm/flashmem.c
+++ b/common_arm/flashmem.c
@@ -455,34 +455,23 @@ void Flashmem_print_info(void) {
 }
 
 //read spi flash JEDEC ID and fill the global variable spi_flash_p64k
-bool FlashDetect(bool flash_init) {
+bool FlashDetect(void) {
     flash_device_type_t flash_device = {0};
-
-    if (flash_init) {
-        if (!FlashInit()) {
-            if (g_dbglevel > 3) Dbprintf("FlashDetect() FlashInit fail");
-            return false;
-        }
-    }
 
     if (!Flash_ReadID(&flash_device, true)) {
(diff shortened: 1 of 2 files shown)


iceman1001:
Hm....
bootrom running spiffs when we need to get the ID out.  
If I remember it correct it was when you have an older bootrom and we flash a newer.    Have you tried that scenario?

ANTodorov:
The flash ID is dedicated SPI request via Flash_UniqueID() (0x4B) https://github.com/RfidResearchGroup/proxmark3/blob/master/common_arm/flashmem.c#L516
The SPI is initialized, the UID is red via 0x4B and then stopped.
There is no spiffs or signature red via the bootloader - these functions are in the firmware context, not the bootloader.
Or we are talking about an old bootloader... Any hint how far back in the logs should I go to check?

ANTodorov:
re-checked, there is no spiffs in the bootloader only the stuff to read the unique ID, used for the USB serial
image

ANTodorov:
I think I am close to v2 - two global variables, one with the jedec identifier, the second one - a pointer to the spiFlashTable array
should I push the change to this PR or make a new one based on this?

This PR is just fixing a bug - out of bound index. here https://github.com/RfidResearchGroup/proxmark3/pull/2639/commits/68e354d7ef22eb260cd9f038ccfc1381546e55f3#diff-e8edff7301ceb63dec5c330434c2b1001aef706b32fab2120852f0170fd11996R468

the refactored code I am working at the moment is making things more clearer

iceman1001:
Make a new PR

PR #2643, opened 2024-11-19
refactor spi flash detection

ANTodorov:
`spi_flash_data` is a structure of the data, red from the flash IC
`*spi_flash_p` is a pointer to the detected flash IC, points to the default '256k/4 pages/unknown' device


common_arm/flashmem.c
```diff
@@ -43,7 +43,34 @@ static uint32_t FLASHMEM_SPIBAUDRATE = FLASH_BAUD;
...
+
+    { 0xEF, 0x21, 0x7022,  4, "W25Q02JV" },
+    // identified by Manufacturer /Device ID
+//    { 0xEF, 0x05, 0x0000,  1, "Winbond!!!" },
+    { 0xEF, 0x10, 0x0000,  2, "W25*10BV!!!" }, //  128k
+    { 0xEF, 0x11, 0x0000,  4, "W25*20BV" },    //  256k
+    { 0xEF, 0x12, 0x0000,  8, "W25*40BV" },    //  512k
+    { 0xEF, 0x13, 0x0000, 16, "W25*80BV" }     // 1024k
+};
+const spi_flash_t *spi_flash_p = SpiFlashTable;
```
iceman1001, line 71: If you move this table to the header file where you declare it,  you can remove the extern reference to it. 
Not sure why you complicate it with pointers.  I would have a static extern for table_idx,   and a "get_item" function.

ANTodorov: You are right. 
Actually only the number of the 64k pages is needed outside of the flashmem.c file.
rework in progress...

iceman1001:
Awesome!

PR #2648, opened 2024-11-21
add KDF functionality to rf08s 'full' card recovery script

csBlueChip:


iceman1001:
One of these days the integration will become better.

PR #2652, opened 2024-11-21
Cherry pick POC of emv simulation

n-hutton:
Hello,

I have here a proof of concept that allows the proxmark3 RDV4 to use the attached card reader for signing contactless transactions, turning the proxmark into a payment device (only works with Visa cards for now and I think is hardcoded for UK based payments, if anyone wants to try it out, I can fix that).

I think it is pretty interesting feature to enable because it allows you to MITM/play around with contactless flow.

I think that this is only possible due to security lapses in the way Visa differentiates between contactless and contact transactions, you are kindof tricking the card into performing this signature. I suspect that it bypasses PIN protections in certain circumstances. I can write long boring technical explanation on this later. Visa was not interested in this bug report so here we are.

I assume that there is interest in having this in the codebase - if not it would be good to know so I don't spend time cleaning the PR up! Note that this PR is messy and not in its final form (but does work every time).

If there is interest, it would be good to get some answers to these questions:

- The invokation/usage is currently `emv smart2nfc` because it doesn't 'sit' anywhere. Is this desirable?
- The timing of responding to NFC is difficult and so I had to re-write I2C for various reasons (seen in this pr as i2c_direct), should I try to integrate into i2c.c or keep it standalone?
- The main part of the code is in emvsim.c - this is basically ripped from mifaresim.c because all we care about is communicating once the handshaking is completed. Should emvsim stand alone, or should it be a special case in mifaresim?
- Should I split this PR into smaller PRs (for i2c for example)
- Anything else anyone can think of


iceman1001:
Interesting PR,  I will need to have a better look at it before merging.   Need to understand :)    Sorry for the delay.

nneul:
#2718 fyi.

PR #2645, opened 2024-11-19
Add new Mifare Classic keys from MifareClassicTool project

onovy:

PR #2655, opened 2024-11-22
Add new Mifare Classic keys from Momentum firmware project.

onovy:


iceman1001:
Nice,
We seen some minor issues with flipper keys since they have a tendency to add diversified keys (calculated based on UID usually) to their dictionaries.  Derived keys are only useful for that one specific card. 

Are these keys verified to be static across multiple cards from the same system ?

onovy:
That's right, but TBH I'm not sure.

It looks like there are 5 sources of their keys:
- their users via MR
- PM3 itself
- https://github.com/Stepzor11/NFC_keys
- https://github.com/ikarus23/MifareClassicTool
- https://github.com/flipperdevices/flipperzero-firmware/

MFC+Flipper keys are already merged into PM3 (https://github.com/RfidResearchGroup/proxmark3/pull/2645/commits/d6c966ac4b4e6a9c7618508989700c2ebaf2de50 https://github.com/RfidResearchGroup/proxmark3/pull/2645/commits/e205829af2f621d90a0994f1488f40d34af177e5)

PR #2657, opened 2024-11-23
add new exes to .gitignore

csBlueChip:
make is creating exes which are being caught by `git add -A` (and other commands)

```
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        tools/mf_nonce_brute/
        tools/mfkey/
        tools/nonce2key/
```


iceman1001:
Those folders shouldn't exist any more since @doegox  moved them under `tools\mfc` folder.

If you have something left inside you should just be able to delete those folders.

PR #2658, opened 2024-11-24
Update t55xx_default_pwds.dic

Akury83:
Update to include Chinese cloner ZX-COPY10 t5577 write password

Sniffed using the pm3 client:

image


iceman1001:
Awesome!

PR #2664, opened 2024-11-28
Add option to set and get maximum read/write block number using hf_mf_ultimatecard script

piotrva:
As this was not covered before in the script mentioned I made few changes:

1. Added display of maximum read/write block to read magic configuration option
2. Added `-b` argument to set maximum read/write block
3. Needed to split argument string because reached lua maximum string length
4. Refactored help for `-t` option to increase readability, because already reached the limit explained above

Tested with two exactly same unbranded UMC's I own.


piotrva:
A demo of the feature:
image

iceman1001:
Thanks!

PR #2665, opened 2024-11-28
added JEDEC information for SPI flash W25Q64JV

ANTodorov:
added some "extrapolated" but logical matches by Manufacturer/Device ID only

some reordering


iceman1001:
Thanks!

PR #2666, opened 2024-11-29
Automatically set maximum r/w block for preset configurations of Mifare Classic in hf_mf_ultimatecard.lua, add presets for 2k variants.

piotrva:
- Added automatic setting of maximum read/write block when setting preset configurations (templates) for  Mifare Classic 1k and 4k.
- Added preset configurations for 2k variants


iceman1001:
Since there are no 2k MIFARE Classic cards,  I think you need to change text to point out its a MIFARE Plus in 2K

piotrva:
> Since there are no 2k MIFARE Classic cards, I think you need to change text to point out its a MIFARE Plus in 2K

Decided for now to drop this support, as this will be better to add altogether with other MIFARE Plus tags in Classic mode (SL1).

PR #2671, opened 2024-12-01
Fix Gen 3 APDU block 0 SAK not being written correctly

Eltrick:


iceman1001:
Interesting catch.  The solution is hard to read.  Should make it clearer.

PR #2672, opened 2024-12-02
add static laundry card keys

kitsunehunter:
these cards are stored value and manipulating the data is useful


iceman1001:
Excellent!

PR #2673, opened 2024-12-02
fix pacs data in example trace filename

nvx:
noticed the filename had a slightly wrong fc and cn


iceman1001:
interesting... so you can extract the SIO PACS from a trace...

PR #2686, opened 2024-12-10
Remove duplicate keys

Willy-JL:
i imagine these were missed due to different capitalization, if there is any automated checking it may be worth making it case insensitive when looking for duplicates


piotrva:
Verified that all removed keys are duplicates.

iceman1001:
Interesting that the uniqe workflow isn't case insensitive

iceman1001:
fixed with workflow to be case insensitive now. https://github.com/RfidResearchGroup/proxmark3/commit/9cdef9ceb4a8db209611f801dc0884ac58a589c8

PR #2688, opened 2024-12-12
Added hf iclass trbl

Antiklesys:
Added functionality to perform tearoff attacks on iclass blocks.
The logic seems correct but I haven't yet had much luck in succeeding the tear-off attacks


client/src/cmdhficlass.c
```diff
@@ -2927,6 +2928,199 @@ static int CmdHFiClass_ReadBlock(const char *Cmd) {
...
+        PrintAndLogEx(ERR, "Tearoff end delay must be bigger than the start delay.");
+        return PM3_EINVARG;
+    }
+
+    if(tearoff_start < 0 || tearoff_end <= 0){
+        PrintAndLogEx(ERR, "Tearoff start/end delays should be bigger than 0.");
+        return PM3_EINVARG;
+    }
+
+    CLIParserFree(ctx);
```
iceman1001, line 3035: You would need to move this one a bit earlier...

nvx:
I'm confused I thought this was already possible with the `hw tearoff` command without needing a dedicated command for it. I didn't think other technologies had dedicated tearoff write commands?

Antiklesys:
Other technologies don't have dedicated commands, the `hw tearoff` command doesn't loop and test different tearoff delays, is very manual and has to be configured and enabled for every interaction. This automates that process and creates a delay range.
It also checks the card values and read response to assess whether the tearoff was successful or not.

@iceman1001 any additional blockers on this?

nvx:
Ah that makes sense

PR #2694, opened 2024-12-20
Add annotations for FMCOS2.0 CPU Card

klks:
Adding annotations for the FMCOS 2.0 CPU Card that is used/sold in China.


iceman1001:
That was a nice PR.   Just they way I like 'em

PR #2697, opened 2024-12-26
Move keys library to spiffs

piotrva:
This change moves flash-stored key dictionaries (Mifare, iClass and T55XX) from statically allocated area at the end of FLASH memory to a file located under SPIFFS file system.
This also implements automatic writing of the keys to SPIFFS using legacy methods.

Things to note:
- Currently there seem to be no implementation/use of iClass keys stored in flash, so this change only transfers all the keys to SPIFFS file
- In the future it will be a good idea to consider extension of SPIFFS into the last page of FLASH, as currently most of the page is left blank.
- The change was tested using Proxmark3 RDV4.01 and some Mifare Classic 1k cards and some T5577 tags.
- Cross-testing by other users would be more than welcome.

image
image
image
image


iceman1001:
Nice and it removed some redundant code.

PR #2700, opened 2024-12-29
Extend SPIFFS partition to last SPI FLASH page

piotrva:
This change extends area of flash available for SPIFFS into the last page of FLASH.
Two last sectors are still reserved - the last sector for the signature, the one before the last sector - simply reserved for future use.

The change was tested with Proxmark RDV4.01 while filling memory up to 80% - the signature remained intact.

It was noted that after reaching 80% of memory fill SPIFFS operation started to be erroneous, but the same behawior was observed for 825dea27a0b1631b1325a86f9c0bc110b440f907 so the change introduced by the pull request is not causing this issue.


iceman1001:
Awesome!

PR #2703, opened 2025-01-04
Add custom suffix for hf mf autopwn to -o argument

zxkmm:
This is because hf mf autopwn is the most used command and this can be used as "taking notes", to prevent causing a not sortable file list after long time use. 

usage:

`hf mf autopwn -o front_door`
`hf mf autopwn -o "front_door"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Added simulation function to `hf iclass legrec` (@antiklesys)
 - Added keys from Momentum firmware projects. (@onovy)
 - Added Dutch Statistics Agency default key (@eagle00789)
-- Added cuxtom suffix for `hf mf autopwn` command. (@zxkmm)
+- Changed hf mf autopwn - now allows for custom suffix (@zxkmm)
 
 ## [Orca.4.19552][2024-11-22]
 - Fixed `hf_legic.lua` - removed bit32 commands from the script (@diorch1968)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Added simulation function to `hf iclass legrec` (@antiklesys)
 - Added keys from Momentum firmware projects. (@onovy)
 - Added Dutch Statistics Agency default key (@eagle00789)
+- Added cuxtom suffix for `hf mf autopwn` command. (@zxkmm)
 
 ## [Orca.4.19552][2024-11-22]
 - Fixed `hf_legic.lua` - removed bit32 commands from the script (@diorch1968)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Added simulation function to `hf iclass legrec` (@antiklesys)
 - Added keys from Momentum firmware projects. (@onovy)
 - Added Dutch Statistics Agency default key (@eagle00789)
+- Added cuxtom suffix for `hf mf autopwn` command. (@zxkmm)
```
iceman1001, line 29: Change to: 

- Changed `hf mf autopwn`  - now allows for custom suffix (@zxkmm)

client/src/cmdhfmf.c
```diff
@@ -2454,6 +2454,7 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
         arg_lit0("a",   NULL,             "Input key A (def)"),
         arg_lit0("b",   NULL,             "Input key B"),
         arg_str0("f", "file",    "<fn>",  "filename of dictionary"),
+        arg_str0("o", "output",  "<fn>",  "filename suffix for dump and key files"),
```
iceman1001, line 2457: replace    "output"  with  NULL

client/src/cmdhfmf.c
```diff
@@ -2691,7 +2697,13 @@ static int CmdHF14AMfAutoPWN(const char *Cmd) {
     }
 
     // read uid to generate a filename for the key file
-    char *fptr = GenerateFilename("hf-mf-", "-key.bin");
+    char suffix[FILE_PATH_SIZE];
+    if (outfilename[0] != '\0') {
```
iceman1001, line 2701: easier to check  the outfnlen  variable...    `if (outfnlen) `

zxkmm:
Thank you for the review!
- fixed changelog
- for the file name buffer i thought check itself is safer but i think check len is safe enough too, so I changed it to yours.
- for the long argument, would you mind tell me why it's better to not define long argument? I already do that as your request but just curious why.

doc/cheatsheet.md
```diff
@@ -294,6 +294,7 @@ Options:
 -a                             Input key A (def)                                                                
 -b                             Input key B                                                                      
 -f, --file <fn>                filename of dictionary                                                           
+-o, --output <fn>              filename suffix for dump and key files
```
iceman1001, line 297: need also to update the text here

iceman1001:
We are trying not to have "long_parameter_names"  since it will fill up out screen space.  Think people using it on phones.   So even if you need a "long" parameter,  try making as short as possible.  In this case  "output"  doesn't add anything.  We could have used  "out" but feels a bit odd.   just making a short -o  is enough since its one of the "reserved" parameter names like -f ..

iceman1001:
Thanks for the fast changes!

zxkmm:
> We are trying not to have "long_parameter_names" since it will fill up out screen space. Think people using it on phones. So even if you need a "long" parameter, try making as short as possible. In this case "output" doesn't add anything. We could have used "out" but feels a bit odd. just making a short -o is enough since its one of the "reserved" parameter names like -f ..

Thanks for the explanation!

PR #2706, opened 2025-01-05
Revise "full" recovery script

csBlueChip:
Revise 'RF08S_full' script to NOT attempt to recover a card it does not support (v1.4)

I tried to add support for other cards to it, but the code just isn't up to it.... Different block counts, variable sector sizes, bespoke code for certain blocks ...it would probably be prudent to abstract the KDF & decode functionality too.

I plan to develop a more structured approach that will present a more adaptable solution to the problem.

I now have a modest selection of card/chips/tags, so I need to start reading datasheets and work out what other quirks are hiding in the bushes - eg. I never anticipated variable sector sizes!

### DIFF
--- a/client/pyscripts/fm11rf08s_full.py
+++ b/client/pyscripts/fm11rf08s_full.py
@@ -14,7 +14,7 @@ import json
 from fm11rf08s_recovery import recovery
 
 author = "@csBlueChip"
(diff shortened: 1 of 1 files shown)


iceman1001:
Having a nice identification of tags is nice but it tends to become spaghetti code fast

PR #2707, opened 2025-01-06
Added SEOS Support

evildaemond:
Added generic SEOS Support with the following commands;
- `hf seos info` (slightly improved from the original)
- `hf seos pacs` (return the SIO value from the default ADF and default tag)
- `hf seos adf ` (send a get data request to an ADF with a custom tag list)
- `hf seos gdf` (make a request for a GDF based upon a command found in their docs)
- `hf seos managekeys` (manage your own keys used for SEOS commands)

You still need keys to use this commandset, this does not have any keys and is theoretical at this stage until someone tests it with the keys.


iceman1001:
Awesome!

PR #2715, opened 2025-01-14
fix hf mf csetuid --gdm

nvx:


iceman1001:
is there any more of those in the source left?

nvx:
> is there any more of those in the source left?

I don't think so, but then I thought I had them all when I added the feature in the first place so 😅

iceman1001:
Seems to be one in 
```
armsrc/Standalone/hf_young.c
58:    uint8_t params = (MAGIC_SINGLE | MAGIC_WUPC | MAGIC_DATAIN);
187:                    -Use mf_chinese_set_block(0, block0, oldUID, wantWipe, MAGIC_SINGLE | MAGIC_WUPC) to write it
```

nvx:
> Seems to be one in
> 
> ```
> armsrc/Standalone/hf_young.c
> 58:    uint8_t params = (MAGIC_SINGLE | MAGIC_WUPC | MAGIC_DATAIN);
> 187:                    -Use mf_chinese_set_block(0, block0, oldUID, wantWipe, MAGIC_SINGLE | MAGIC_WUPC) to write it
> ```

ah yeah I didn't look at standalones tbh

PR #2717, opened 2025-01-14
Possible fix for MFC EV1 signature dumping in 'hf mf autopwn'

Eltrick:


iceman1001:
Lets see if it works :)

PR #2720, opened 2025-01-15
CodeQL -- Actual fixes

henrygab:
Many of the CodeQL bugs of type "Comparison between A of TypeA and B of wider type TypeB" were caused by having arithmetic within a `for()` loop's comparison clause.  Not only were the operations redundant, but they could cause unintended consequences (e.g., integer promotion rules).  Relatively simple to fix (use larger of two sizes for loop variable, compare unsigned vs. unsigned, etc.).

The other reason for many of these was that spiffs, in the definition of `struct spiffs_t`, defined field `block_count` of type `uint32_t`, but defined field `free_cursor_block_ix` to be of type `spiffs_block_ix`.  In PM3, `spiffs_block_ix` is defined as `uint16_t`.  Thus, a variable of type `spiffs_block_ix` cannot directly loop against `block_count` (potential infinite loops).

Inline "conversation"s added below, both to explain some changes, and to mark code that should have a second set of eyes review.

### DIFF
--- a/armsrc/spiffs_check.c
+++ b/armsrc/spiffs_check.c
@@ -536,6 +536,13 @@ static s32_t spiffs_page_consistency_check_i(spiffs *fs) {
     s32_t res = SPIFFS_OK;
     spiffs_page_ix pix_offset = 0;
 
+    // this _should_ never happen, but prefer to see debug message / error
+    // rather than silently entering infinite loop.
+    if (fs->block_count > ((spiffs_block_ix)(-1))) {
+        SPIFFS_DBG("Avoiding infinite loop, block_count "_SPIPRIbl" too large for spiffs_block_ix type\n", fs->block_count);
+        SPIFFS_API_CHECK_RES(fs, SPIFFS_ERR_INTERNAL);
+    }
+
     // for each range of pages fitting into work memory
     while (pix_offset < SPIFFS_PAGES_PER_BLOCK(fs) * fs->block_count) {
         // set this flag to abort all checks and rescan the page range
--- a/armsrc/spiffs_check.c
+++ b/armsrc/spiffs_check.c
@@ -586,7 +586,7 @@ static s32_t spiffs_page_consistency_check_i(spiffs *fs) {
                      0);
             // traverse each page except for lookup pages
             spiffs_page_ix cur_pix = SPIFFS_OBJ_LOOKUP_PAGES(fs) + SPIFFS_PAGES_PER_BLOCK(fs) * cur_block;
-            while (!restart && cur_pix < total_blocks_plus_one_page) {
+            while (!restart && cur_pix < SPIFFS_PAGES_PER_BLOCK(fs) * (cur_block+1)) {
                 //if ((cur_pix & 0xff) == 0)
                 //  SPIFFS_CHECK_DBG("PA: processing pix "_SPIPRIpg", block "_SPIPRIbl" of pix "_SPIPRIpg", block "_SPIPRIbl"\n",
                 //      cur_pix, cur_block, total_blocks, block_count);
--- a/armsrc/spiffs_check.c
+++ b/armsrc/spiffs_check.c
@@ -540,7 +540,7 @@ static s32_t spiffs_page_consistency_check_i(spiffs *fs) {
     uint32_t total_blocks = SPIFFS_PAGES_PER_BLOCK(fs) * block_count;
     uint32_t total_blocks_plus_one_page = total_blocks + SPIFFS_PAGES_PER_BLOCK(fs);
 
-#pragma region    // check for overflow once, before looping
+//#pragma region    // check for overflow once, before looping
     // this _should_ never happen, but prefer to see debug message / error
     // rather than silently entering infinite loop.
     if (block_count > ((spiffs_block_ix)(-1))) {
@@ -568,7 +568,7 @@ static s32_t spiffs_page_consistency_check_i(spiffs *fs) {
     }
     // RESULT: spiffs_page_ix can safely be used for loop index vs. each of
     //         block_count, total_blocks, and total_blocks_plus_one_page
-#pragma endregion // check for overflow once, before looping
+//#pragma endregion // check for overflow once, before looping
 
 
     // for each range of pages fitting into work memory
--- a/armsrc/spiffs_check.c
+++ b/armsrc/spiffs_check.c
@@ -535,33 +535,61 @@ static s32_t spiffs_page_consistency_check_i(spiffs *fs) {
 
     s32_t res = SPIFFS_OK;
     spiffs_page_ix pix_offset = 0;
+    // Avoid arithmetic in loop conditions (integer promotion rules can cause unintended consequences)
+    uint32_t block_count = fs->block_count;
+    uint32_t total_blocks = SPIFFS_PAGES_PER_BLOCK(fs) * block_count;
+    uint32_t total_blocks_plus_one_page = total_blocks + SPIFFS_PAGES_PER_BLOCK(fs);
 
+#pragma region    // check for overflow once, before looping
     // this _should_ never happen, but prefer to see debug message / error
     // rather than silently entering infinite loop.
-    if (fs->block_count > ((spiffs_block_ix)(-1))) {
-        SPIFFS_DBG("Avoiding infinite loop, block_count "_SPIPRIbl" too large for spiffs_block_ix type\n", fs->block_count);
+    if (block_count > ((spiffs_block_ix)(-1))) {
+        SPIFFS_DBG("Avoiding infinite loop, block_count "_SPIPRIbl" too large for spiffs_block_ix type\n", block_count);
         SPIFFS_API_CHECK_RES(fs, SPIFFS_ERR_INTERNAL);
     }
+    // this checks for overflow of the multiplication of block_count+1 with SPIFFS_PAGES_PER_BLOCK(fs)
+    if (((uint32_t)(-1)) / SPIFFS_PAGES_PER_BLOCK(fs) > (block_count+1)) {
+        // checking with +1 block count to avoid overflow also in inner loop, which adds one page...
+        // would exceed value storable in uint32_t
+        SPIFFS_DBG("Overflow: pages per block %04x with block count "_SPIPRIbl" results in overflow\n", SPIFFS_PAGES_PER_BLOCK(fs), block_count);
+        SPIFFS_API_CHECK_RES(fs, SPIFFS_ERR_INTERNAL);
+    }
+    // because loop indices are using spiffs_page_ix type,
+    // that type can hold a large enough value
+    if (total_blocks > ((spiffs_page_ix)-1)) {
+        SPIFFS_DBG("Avoiding infinite loop, total_blocks "_SPIPRIpg" too large for spiffs_page_ix type\n", total_blocks);
+        SPIFFS_CHECK_RES(SPIFFS_ERR_INTERNAL);
+    }
+    // because loop indices are using spiffs_page_ix type,
+    // that type can hold a large enough value
+    if (total_blocks_plus_one_page > ((spiffs_page_ix)-1) || total_blocks_plus_one_page < total_blocks) {
+        SPIFFS_DBG("Avoiding infinite loop, total_blocks_plus_one_page "_SPIPRIpg" too large for spiffs_page_ix type\n", total_blocks_plus_one_page);
+        SPIFFS_CHECK_RES(SPIFFS_ERR_INTERNAL);
+    }
+    // RESULT: spiffs_page_ix can safely be used for loop index vs. each of
+    //         block_count, total_blocks, and total_blocks_plus_one_page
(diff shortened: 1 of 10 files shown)


armsrc/spiffs_check.c
```diff
@@ -535,26 +535,61 @@ static s32_t spiffs_page_consistency_check_i(spiffs *fs) {
 
     s32_t res = SPIFFS_OK;
     spiffs_page_ix pix_offset = 0;
+    // Avoid arithmetic in loop conditions (integer promotion rules can cause unintended consequences)
```
henrygab, line 538: Please have second set of eyes review these checks.

armsrc/thinfilm.c
```diff
@@ -114,8 +114,10 @@ static int EmSendCmdThinfilmRaw(const uint8_t *resp, uint16_t respLen) {
     }
 
     // Ensure that the FPGA Delay Queue is empty
-    uint8_t fpga_queued_bits = FpgaSendQueueDelay >> 3;
-    for (i = 0; i <= (fpga_queued_bits >> 3) + 1;) {
+    uint16_t fpga_queued_bits = FpgaSendQueueDelay >> 3;
```
henrygab, line 117: explicitly use unsigned values for the maths, and store back as `uint16_t`.  This allows the `for()` loop to index on a variable of size `uint16_t`.

client/src/cmdhf14a.c
```diff
@@ -3226,13 +3226,22 @@ int CmdHF14ANdefRead(const char *Cmd) {
...
+    for (size_t i = offset; i < ndef_size + offset; i += max_rapdu_size) {
+        size_t segment_size = max_rapdu_size < ndef_size + offset - i ? max_rapdu_size : ndef_size + offset - i;
+
         keep_field_on = i < ndef_size + offset - max_rapdu_size;
         aREAD_NDEF_n = 0;
         param_gethex_to_eol("00b00000", 0, aREAD_NDEF, sizeof(aREAD_NDEF), &aREAD_NDEF_n);
         aREAD_NDEF[2] = i >> 8;
         aREAD_NDEF[3] = i & 0xFF;
+
+        // Segment_size is stuffed into a single-byte field below ... so error out if overflows
```
henrygab, line 3238: If previous code ever had a `segment_size` greater than `0xFF`, then the old code would have silently truncated the value.  After review, this appeared to be incorrect behavior, and that there was an implicit assumption that `segment_size` would always be single-byte value.

client/src/cmdhfmf.c
```diff
@@ -117,17 +117,31 @@ static char *GenerateFilename(const char *prefix, const char *suffix) {
...
+    if (*src == NULL) {
         return PM3_EMALLOC;
+    }
 
     // empty e_sector
     for (size_t i = 0; i < items; i++) {
         for (uint8_t j = 0; j < 2; j++) {
             (*src)[i].Key[j] = 0xffffffffffff;
-            (*src)[i].foundKey[j] = 0;
+            // (*src)[i].foundKey[j] = 0; // calloc zero's these already
```
henrygab, line 144: No, I cannot explain why CodeQL alerted on this line, and therefore cannot explain why its removal makes CodeQL happier.  But, it's redundant since `calloc()` ensures memory is zero at allocation, so ... removal is OK.

client/src/cmdlfem410x.c
```diff
@@ -245,7 +245,7 @@ static int ask_em410x_binary_decode(bool verbose, uint32_t *hi, uint64_t *lo, ui
         else if (ans == -4)
             PrintAndLogEx(DEBUG, "DEBUG: Error - Em410x preamble not found");
         else if (ans == -5)
-            PrintAndLogEx(DEBUG, "DEBUG: Error - Em410x Size not correct: %zu", size);
+            PrintAndLogEx(DEBUG, "DEBUG: Error - Em410x Size not correct: %zu", *size);
```
henrygab, line 248: `size` was a pointer to the size, not the size itself.

armsrc/spiffs_check.c
```diff
@@ -535,26 +535,61 @@ static s32_t spiffs_page_consistency_check_i(spiffs *fs) {
...
             CHECK_CB(fs, SPIFFS_CHECK_PAGE, SPIFFS_CHECK_PROGRESS,
-                     (pix_offset * 256) / (SPIFFS_PAGES_PER_BLOCK(fs) * fs->block_count) +
-                     ((((cur_block * pages_per_scan * 256) / (SPIFFS_PAGES_PER_BLOCK(fs) * fs->block_count))) / fs->block_count),
+                     (pix_offset * 256) / total_blocks +
+                     ((((cur_block * pages_per_scan * 256) / total_blocks)) / block_count),
                      0);
             // traverse each page except for lookup pages
             spiffs_page_ix cur_pix = SPIFFS_OBJ_LOOKUP_PAGES(fs) + SPIFFS_PAGES_PER_BLOCK(fs) * cur_block;
-            while (!restart && cur_pix < SPIFFS_PAGES_PER_BLOCK(fs) * (cur_block + 1)) {
+            while (!restart && cur_pix < total_blocks_plus_one_page) {
```
iceman1001, line 589: this one says "cur_block + 1"   vs  "total_blocks_plus_one_page"   ???

armsrc/spiffs_hydrogen.c
```diff
@@ -52,8 +52,16 @@ s32_t SPIFFS_format(spiffs *fs) {
...
+    // this _should_ never happen, but prefer to see debug message / error
+    // rather than silently entering infinite loop.
+    if (block_count > ((spiffs_block_ix)(-1))) {
+        SPIFFS_DBG("Avoiding infinite loop, block_count "_SPIPRIbl" too large for spiffs_block_ix type\n", block_count);
+        SPIFFS_API_CHECK_RES_UNLOCK(fs, SPIFFS_ERR_INTERNAL);
+    }
+
     spiffs_block_ix bix = 0;
-    while (bix < fs->block_count) {
+    while (bix < block_count) {
```
iceman1001, line 64: is it really that block_count is large,  or ?
block_count = 0,  nothing happens,
if large number,  the  spiff_ok check should fail when trying to erase a block outside of existing once..

not sure how this would be a infinite loop?

armsrc/thinfilm.c
```diff
@@ -114,8 +114,10 @@ static int EmSendCmdThinfilmRaw(const uint8_t *resp, uint16_t respLen) {
...
+    fpga_queued_bits /= 8u;
```
iceman1001, line 118: use the `>>> 3`   for consistency.    Once upon a time we didn't want to trigger floats inside the arm code because of size.

client/src/cmdhf14a.c
iceman1001: Proper solution would not to use size_t but u32 and look into the ndef standard to handle larger than 0xFF sized segments?

client/src/cmdhfmf.c
```diff
@@ -117,17 +117,31 @@ static char *GenerateFilename(const char *prefix, const char *suffix) {
...
 
+
+    // typedef struct {
+    //     uint64_t Key[2];
+    //     uint8_t foundKey[2];
+    // } sector_t;
+
+    // This allocates based on the size of a single item
+    _Static_assert(sizeof(sector_t) >= 18, "Unexpectedly small sector_t"); // if packed, would be 18
+    _Static_assert(sizeof(sector_t) == 24, "Sector_t used to be padded to 24 bytes?"); // not packed, so each entry must be 24 bytes
```
iceman1001, line 133: This code looks more like debug code.

```
// ISSUE: C99 has no _Static_assert() ... was added in C11
```

client/src/cmdlfem410x.c
iceman1001: good catch

client/src/cmdlfem4x50.c
```diff
@@ -779,6 +781,12 @@ static int CmdEM4x50Reader(const char *Cmd) {
         // iceman,  misuse of return status code.
         int now = resp.status;
 
+        // prevent massive stack corruption if unexpected results from device.
+        if (now > EM4X50_NO_WORDS) {
+            PrintAndLogEx(WARNING, "word count was: %d, limiting to %d", now, EM4X50_NO_WORDS);
+            now = EM4X50_NO_WORDS;
+        }
+
         if (now > 0) {
```
iceman1001, line 791: this one can be "   if (now) {

client/src/cmdlfem4x50.c
```diff
@@ -779,6 +781,12 @@ static int CmdEM4x50Reader(const char *Cmd) {
         // iceman,  misuse of return status code.
         int now = resp.status;
 
+        // prevent massive stack corruption if unexpected results from device.
+        if (now > EM4X50_NO_WORDS) {
+            PrintAndLogEx(WARNING, "word count was: %d, limiting to %d", now, EM4X50_NO_WORDS);
+            now = EM4X50_NO_WORDS;
+        }
+
```
iceman1001, line 789: the proper fix would be not misusing the return status code and check it accordingly

armsrc/thinfilm.c
iceman1001: good catch

armsrc/spiffs_hydrogen.c
henrygab: This is just another location where spiffs's choice to not use type `spiffs_block_ix` for `fs->block_count` causes headaches.

The original issue is, of course, that `uint32_t` for `fs->block_count` could store a value `>= UINT16_MAX`.  If so, this would enter an infinite loop, because no amount of incrementing the variable `spiffs_block_ix bix` would ever cause the loop to exit.   The later call to `spiffs_erase_block()` would see no problem, because it's getting a `uint16_t` value.   Thus ... infinite loop.

Anyways, by moving `fs->block_count` to a local variable, CodeQL can reason that the variable value does not change.  Thus, the earlier check of `block_count <= ((spiffs_block_ix)-1)` ensures the infinite loop never occurs.  Although I'm not positive, this change may also allow the compiler to better optimize use of `block_count`.

Does that make more sense?

armsrc/spiffs_check.c
henrygab: Excellent catch!  This is my error.   I have reverted this line as I cannot find an easy way to make CodeQL happy here.

armsrc/thinfilm.c
henrygab: ok

client/src/cmdhfmf.c
henrygab: removed newly added static assertions.

client/src/cmdlfem4x50.c
henrygab: This PR is focused on CodeQL changes.  The existing code was as shown, and the change is stylistic in nature.  Thus, it is outside the scope of this PR.

client/src/cmdlfem4x50.c
henrygab: Yes, that would be the proper fix.
This PR is focused on CodeQL fixes.  It is not intended to fix other problems.  Thus, this is outside the scope of this PR.

armsrc/spiffs_hydrogen.c
iceman1001: aha,   so erase_block takes u16...   rgr that.    yeah,   make everything u16?

Thanks, excellent explanation btw

armsrc/spiffs_check.c
```diff
@@ -535,26 +535,61 @@
...
             CHECK_CB(fs, SPIFFS_CHECK_PAGE, SPIFFS_CHECK_PROGRESS,
-                     (pix_offset * 256) / (SPIFFS_PAGES_PER_BLOCK(fs) * fs->block_count) +
-                     ((((cur_block * pages_per_scan * 256) / (SPIFFS_PAGES_PER_BLOCK(fs) * fs->block_count))) / fs->block_count),
+                     (pix_offset * 256) / total_blocks +
+                     ((((cur_block * pages_per_scan * 256) / total_blocks)) / block_count),
                      0);
             // traverse each page except for lookup pages
             spiffs_page_ix cur_pix = SPIFFS_OBJ_LOOKUP_PAGES(fs) + SPIFFS_PAGES_PER_BLOCK(fs) * cur_block;
-            while (!restart && cur_pix < SPIFFS_PAGES_PER_BLOCK(fs) * (cur_block + 1)) {
+            while (!restart && cur_pix < SPIFFS_PAGES_PER_BLOCK(fs) * (cur_block+1)) {
```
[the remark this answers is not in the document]
henrygab: This is the one fix I reverted, as my fix was actually a regression that you caught via code review.

This is therefore not a failure that should prevent approval of the PR.

iceman1001:
So we stuck at that stupid `spiffs_page_ix`   again.   Can we use that instead of u16?

armsrc/spiffs_hydrogen.c
henrygab: Yes, make everything u16, ... **_and_** add explicit checks that the (locally copied) value will never overflow when cast to u16.👍

PR #2728, opened 2025-01-24
fix bigbuf allocators (tracing + malloc) overwriting each other

shuffle2:
* BigBuf.c: use s_ prefix for statics
* BigBuf_Clear_ext already calls clear_trace, so remove extra calls
* add some sanity checking of allocator args
* dont compare PDC_RNCR to false

### DIFF
--- a/armsrc/iso14443b.c
+++ b/armsrc/iso14443b.c
@@ -980,7 +980,6 @@ void Simulate_iso14443b_srx_tag(uint8_t *uid) {
     // allocate command receive buffer
     BigBuf_free();
     BigBuf_Clear_ext(false);
-    clear_trace();
     set_tracing(true);
 
     uint16_t len, cmdsReceived = 0;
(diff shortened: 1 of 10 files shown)


armsrc/iso14443b.c
```diff
@@ -1381,12 +1380,12 @@ static int Get14443bAnswerFromTag(uint8_t *response, uint16_t max_len, uint32_t
             if (AT91C_BASE_SSC->SSC_SR & (AT91C_SSC_ENDRX)) {
 
                 // primary buffer was stopped
-                if (AT91C_BASE_PDC_SSC->PDC_RCR == false) {
+                if (!AT91C_BASE_PDC_SSC->PDC_RCR) {
```
iceman1001, line 1383: Code style,   we are moving away from !expression  since its easy to miss its a negation.  
if (expression)
and 
if (expression == false)

is our current code style.

iceman1001:
Nice finding this bug!  

And yes, I know our contributing.md file hasn't been updated with the `if (exp)` and `if (exp ==false)` comparison style.  
We still moving to it slowly.

armsrc/iso14443b.c
iceman1001: Sure.

PR #2731, opened 2025-01-26
support for HID SAM communication with Picopass cards

jkramarz:
This merge request swaps Picopass handling implementation to a similar to current SEOS handling procedures.


client/src/cmdhficlass.c
```diff
@@ -5403,18 +5404,66 @@ static int CmdHFiClassSAM(const char *Cmd) {
...
+        arg_strx0("d", "data",     "<hex>", "DER encoded command to send to SAM"),
         arg_param_end
     };
     CLIExecWithReturn(ctx, Cmd, argtable, true);
-    bool verbose = arg_get_lit(ctx, 1);
+
+    bool verbose = false;
+    if (arg_get_lit(ctx, 1)) {
+        verbose = true;
+    }
```
iceman1001, line 5421: you can simplify all these if-statements,  to just an assignment.
Like this. 
`bool verbose = arg_get_lit(ctx, 1);`

client/src/cmdhficlass.c
```diff
@@ -5403,18 +5404,66 @@ static int CmdHFiClassSAM(const char *Cmd) {
...
+
     CLIParserFree(ctx);
 
     if (IsHIDSamPresent(verbose) == false) {
         return PM3_ESOFT;
     }
 
     clearCommandBuffer();
-    SendCommandNG(CMD_HF_SAM_PICOPASS, NULL, 0);
+    SendCommandMIX(CMD_HF_SAM_PICOPASS, command_flags, 0, datalen, data, datalen);
```
iceman1001, line 5466: Please rework to use structs and the SendCommandNG() call.   MIX is to be considered deprecated.

client/Makefile
```diff
@@ -759,6 +759,7 @@ SRCS =  mifare/aiddesfire.c \
 		scripting.c \
 		ui.c \
 		util.c \
+		util_hidsio.c \
```
iceman1001, line 762: consider renaming to just  `hidsio.c`

armsrc/sam_picopass.c
```diff
@@ -30,6 +30,213 @@
...
+ * @param response_len Pointer to the variable where the length of the retreived data will be stored.
+ * @return Status code indicating success or failure of the operation.
+ */
+static int sam_send_request_iso15(const uint8_t *const request, const uint8_t request_len, uint8_t *response, uint8_t *response_len, const bool shallow_mod, const bool break_on_nr_mac, const bool prevent_epurse_update) {
+    int res = PM3_SUCCESS;
+    if (g_dbglevel >= DBG_DEBUG)
+        DbpString("start sam_send_request_iso14a");
+
+    uint8_t buf1[ISO7816_MAX_FRAME] = {0};
+    uint8_t buf2[ISO7816_MAX_FRAME] = {0};
```
iceman1001, line 53: consider making these "bigbuf_malloc"  and free the memory once the whole command is ran.

client/src/cmdhficlass.c
jkramarz: Changed here and in SEOS.

client/Makefile
jkramarz: OK, it is `hidsio` now.

armsrc/sam_picopass.c
jkramarz: Changed here and in SEOS. I think I carried this static allocation from some previous version of my code, where I tried using BigBuf for more temporary buffers.

iceman1001:
Your OSX build is failing.

jkramarz:
Is OSX build the only one compiling client with clang instead of gcc? It catched a very nasty bug: I've accidentally limited max command to the size of pointer, by using the macro in a wrong way.

iceman1001:
> Is OSX build the only one compiling client with clang instead of gcc? It catched a very nasty bug: I've accidentally limited max command length to the size of pointer, by using the macro in a wrong way.

Yes,  Only OSX builds with clang in our GH Actions.

We only use  `CLIGetHexBLessWithReturn`  in total five places,  maybe we should refactor away its use.

iceman1001:
If its all ready to merge,  let me know.

PR #2733, opened 2025-01-28
fm11rf08s_recovery.py - Option to not save the sec32 keys to the generated keyfile

Eltrick:


iceman1001:
Feels like a complicated solution.  Maybe just call it with the right amount of sectors to save?

Eltrick:
I initially thought of that, but I didn't find any reason to have any other options than to save with/without the extra sector that the chip in question has
It's only a MFC with the extra weird sector at index 32 after all

iceman1001:
odd design choice.

PR #2734, opened 2025-01-28
Added spain transport keys

Larision:


iceman1001:
Have you verified its static keys across multiple cards from same system?

Larision:
yes, same keys to all cards.

PR #2737, opened 2025-02-02
Update t55_chk_date.lua

jareckib:
A simple script for searching the password for T5577. The script creates a
    dictionary starting from the entered starting year to the entered ending year.
    There are two search methods - DDMMYYYY or YYYYMMDD. Checking the entire year
    takes about 1 minute and 50 seconds. Date from 1900 to 2100. The script may be
    useful if the password is, for example, a date of birth.

### DIFF
--- a/client/luascripts/t55_chk_date.lua
+++ b/client/luascripts/t55_chk_date.lua
@@ -5,15 +5,14 @@ local getopt = require('getopt')
 local dash = string.rep('--', 32)
 
 author = '    Author: jareckib - created 01.02.2025'
-version = '    version v1.00'
+version = '    version v1.01'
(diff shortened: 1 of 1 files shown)


jareckib:
I corrected the error when selecting 1 or 2

iceman1001:
Nice,
you could also make the "start" and "end"  range input parameters to the script.   This way you can automate calls to the script.  There is no reason for the script to try to read user input. 

`script run lf_t55xx_chkdate -s 1900 -e 2020` 
if you have default values for start and end,   it would be even easier.

`script run lf_t55xx_chkdate -s 1910`   
where end would default to current year.

PR #2741, opened 2025-02-03
Fix facility code and card number checking in LF HID Brute

Donny-Guo:
In  `lf hid brute` command, there's a checking on the range of facility code and card number to determine when it should stop. The original code (`card_hi.FacilityCode < 0xFF` and `card_hi.CardNumber < 0xFFFF`) does not account for different card formats.

Fix it by pulling the corresponding card format limit from FormatTable.

### DIFF
--- a/client/src/cmdlfhid.c
+++ b/client/src/cmdlfhid.c
@@ -544,7 +544,7 @@ static int CmdHIDBrute(const char *Cmd) {
     }
 
     wiegand_card_t card_hi, card_low;
-    cardformatlimit_t limit = get_card_format_limit(format_idx);
+    cardformatdescriptor_t card_descriptor = HIDGetCardFormat(format_idx).Fields;
     memset(&card_hi, 0, sizeof(wiegand_card_t));
 
     char field[3] = {0};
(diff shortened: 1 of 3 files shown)


iceman1001:
1.  Don't remove a section code without any comment.  I don't think you meant to. 
2. This solution feels wrong.   Where I agree that the format in use should need a better struct,  I don't see adding the extra struct meaningful.  It can easily be added to the first struct. I would use names like  "MaxFC",  "MaxCN",  etc.   
3. This should then be modified in all the unpack and pack functions.    And other places where code like this is being used.

iceman1001:
Now its a better solution. 

When I mentioned pack/unpack functions it has nothing to do with being static or not.
It has to do with the checks that are inside each function.   They should get its designated format from the table of formats and then implement range checks accordingly.   You follow me?

```
static bool Pack_Tecom27(wiegand_card_t *card, wiegand_message_t *packed, bool preamble) {
    memset(packed, 0, sizeof(wiegand_message_t));

    if (card->FacilityCode > 0x7FF) return false; // Can't encode FC.
    if (card->CardNumber > 0xFFFF) return false; // Can't encode CN.
    if (card->IssueLevel > 0) return false; // Not used in this format
    if (card->OEM > 0) return false; // Not used in this format

```

PR #2743, opened 2025-02-03
Update range checkings in all pack functions for LF HID

Donny-Guo:
Per suggestions from @iceman1001 ->[PR](https://github.com/RfidResearchGroup/proxmark3/pull/2741)
1. Update the following range checking of all pack functions
```
static bool Pack_Tecom27(wiegand_card_t *card, wiegand_message_t *packed, bool preamble) {
    ...

    if (card->FacilityCode > 0x7FF) return false; // Can't encode FC.
    if (card->CardNumber > 0xFFFF) return false; // Can't encode CN.
    if (card->IssueLevel > 0) return false; // Not used in this format
    if (card->OEM > 0) return false; // Not used in this format

    ...
```
with
```
static bool Pack_Tecom27(int format_idx, wiegand_card_t *card, wiegand_message_t *packed, bool preamble) {
    ...

    if (!validate_card_limit(format_idx, card)) return false;

    ...
```

2. No changes to unpack functions (considering code readability + not necessary for all), for example:
```
static bool Unpack_H10301(wiegand_message_t *packed, wiegand_card_t *card) {
    memset(card, 0, sizeof(wiegand_card_t));
    if (packed->Length != 26) return false; // Wrong length? Stop here.

    card->CardNumber = (packed->Bot >> 1) & 0xFFFF;
    card->FacilityCode = (packed->Bot >> 17) & 0xFF;
    card->ParityValid =
        (oddparity32((packed->Bot >> 1) & 0xFFF) == (packed->Bot & 1)) &&
        ((evenparity32((packed->Bot >> 13) & 0xFFF)) == ((packed->Bot >> 25) & 1));
    return true;
}
```

### DIFF
--- a/client/src/wiegand_formats.c
+++ b/client/src/wiegand_formats.c
@@ -19,13 +19,12 @@
 #include <stdlib.h>
 #include "commonutil.h"
 
+static bool validate_card_limit(int format_idx, wiegand_card_t *card);
 
-static bool Pack_Defcon32(wiegand_card_t *card, wiegand_message_t *packed, bool preamble) {
+static bool Pack_Defcon32(int format_idx, wiegand_card_t *card, wiegand_message_t *packed, bool preamble) {
     memset(packed, 0, sizeof(wiegand_message_t));
-    if (card->FacilityCode > 0x00FFFF) return false; // Can't encode FC.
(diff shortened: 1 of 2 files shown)


iceman1001:
A nice refactor,   less clutter and easier to understand.

PR #2748, opened 2025-02-07
force bitfiles to be identical for same source code

n-hutton:
The bitfile generated by `xilinx webpack ISE 10.1` was not deterministic - that is, for a given set of input files, the bitfile was always different.

This is due to bitgen inserting an ascii string of datetime, at the start of the file. You can see this with xxd.

There is no option to override or disable this, so this solution is just to overwrite that portion of the binary file. This way, you're not constantly committing essentially the same bitfile to the repo again and again.

Place and route is another step that is random and could produce different bitfiles. The fpga is so empty that differing starting seeds produced the same result currently, but I have added a fixed seed so that it will not be a source of differing bitfiles in the future.

I don't know if you used a different version than `Xilinx WebPack ISE 10.1` you'd still get a different bitfile initially I think.


iceman1001:
How about .gitignore and then only certain people would be able to commit the generated fgpa binaries?

n-hutton:
I don't think the .gitignore really solves the issue - once the bitfile is checked into the repo (which I assume we do want), changes to it subsequently will still be flagged and added by git as usual. If you are suggesting fully not tracking the bitfiles, that would be a solution, although I personally like deterministic bitfiles in any case

iceman1001:
I guess looking at commits,  its not many people who is actually compiling the fpga images.   You are an exception.

n-hutton:
Actually, now I look at it again, this PR is incorrect - the location of the date string isn't fixed because it depends on the file name :sweat_smile: 

I can fix this if you think the PR has merit, otherwise I'll close it and just use the solution locally; it is fairly minor after all.

iceman1001:
I am thinking that we are reading that date when we present the `hw status` 

If its not used,  then we merge,   but I have a sneaky suspicion 

image

iceman1001:
```
 fpgaloader.c
623:    Dbprintf("  mode.................... %s", g_fpga_version_information[bitstream_target_to_index(downloaded_bitstream)]);
```

iceman1001:
@n-hutton   let me know when you feel this PR is ready to merge

iceman1001:
image

n-hutton:
I am happy to merge this

PR #2752, opened 2025-02-14
update cmdhfemrtd.c to support JPEG2000 code stream photo

leommxj:
Some passport (eg. Chinese passport) save facial image in DG2 using JPEG2000 code stream, which magic is `FF 4F FF 51`.

https://en.wikipedia.org/wiki/JPEG_2000
https://github.com/RfidResearchGroup/proxmark3/issues/1117#issuecomment-2556661621

### DIFF
--- a/client/src/cmdhfemrtd.c
+++ b/client/src/cmdhfemrtd.c
@@ -767,23 +767,35 @@ static bool emrtd_select_and_read(uint8_t *dataout, size_t *dataoutlen, uint16_t
 
 static const uint8_t jpeg_header[4] = { 0xFF, 0xD8, 0xFF, 0xE0 };
 static const uint8_t jpeg2k_header[6] = { 0x00, 0x00, 0x00, 0x0C, 0x6A, 0x50 };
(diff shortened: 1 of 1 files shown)


iceman1001:
Cool,
why did you change the file suffix from ".jp2"  ->  ".jpc"  ?

leommxj:
> Cool, why did you change the file suffix from ".jp2" -> ".jpc" ?

Only the JPEG2000 code stream (which magic is `FF 4F FF 51`) is using `.jpc`.
The JPEG2000 (which magic is `00 00 00 0C 6A 50 ....`) is still using `.jp2` .
Following the wikipedia's content:
> JPEG 2000 images should—if stored as files—be boxed in the JPEG 2000 file format, where they get the .jp2 extension. 

and 

> There is no standardized extension for codestream data because codestream data is not to be considered to be stored in files in the first place, though when done for testing purposes, the extension .jpc, .j2k or .j2c is commonly used.

iceman1001:
Thanks for the clarification!

PR #2758, opened 2025-02-17
Fixes to DESFire product type recognition

kormax:
* Add known Apple Wallet DESFire hardware version info to known variants in `getCardType` function to prevent `CmdHF14ADesInfo` from bailing out on it.
* Fix `getProductType` and `getProductTypeStr`, which incorrectly interpreted documentation referenced above the function, accessing ProductInfo at "byte 2" via `[2]` instead of `[1]`.


iceman1001:
Nice catch!

PR #2759, opened 2025-02-18
add helpful note for macos users failing linking

n-hutton:
Fairly explanatory


iceman1001:
Thank heavens for your deep insights in MacOS

PR #2772, opened 2025-03-05
Fix wiegand InnerRange56

jmichelp:


iceman1001:
Not sure that fixes that format..  Could be some parity too..

PR #2782, opened 2025-03-14
Refactor Hitag low-level functions into hitag_common

douniwan5788:


iceman1001:
Something is breaking all the GH Actions.

douniwan5788:
fixed

PR #2785, opened 2025-03-16
Em4x70 - Major improvements to tracing, stability, and debugability

henrygab:
1. Rework how communications with tag occur.
  * Data to send to the tag is fully pre-generated as a bitstream
  * Bits sent and received are logged, including start / end times

2. Support built-in `hw dbg` for controlling verbosity of debug output

This change provides the framework for more easily supporting a variety of tags.
It is the foundation for work underway to support V4070 tags (e.g., Porsche Boxster)

Because of the new bitstream generation and logging, a surprising characteristic of the legacy code's behavior was brought to light... each of the commands that sent additional data (AUTH, PIN, WRITE) were:
* inserting an extra RM zero bit
* forcibly-enabling the use of command parity

While this was not expected, this PR maintains the legacy behavior, as removal of that behavior resulted in failures of those three commands.  This change provides a starting point which will simplify discovering the root cause of the hack.

Future TODO:
* Root-cause why the third RM bit is needed.  If it's a coding error (e.g., LIW timing), fix that code to remove the hack.
* Change the arm/client interface for em4x70 to ONLY use arrays of bytes, with well-defined content endianness, to reduce the repeated reversing of byte arrays in the code.


henrygab:
If you'd prefer smaller changes, I could create a PR similar to https://github.com/henrygab/proxmark3/pull/4.   However, not all the commits in that working branch worked (and some did not compile).   On the flip side, that would provide commit history, with commits that explicitly log the old and new methods ... showing they create equivalent outputs.

PR #2794, opened 2025-03-19
Enhance Hitag S annotation and debugging

douniwan5788:


client/src/cmdlfhitaghts.c
```diff
@@ -36,7 +36,83 @@
...
+            } else if (command == HITAGS_WRITE_BLOCK) {
+                exp_len += snprintf(exp + exp_len, size - exp_len, "WRITE_BLOCK");
+            } else if (command == HITAGS_QUIET) {
+                exp_len += snprintf(exp + exp_len, size - exp_len, "QUIET");
+            }
+            // Hitag 1 commands
+            else if (command == HITAG1_RDCPAGE) {
+                exp_len += snprintf(exp + exp_len, size - exp_len, "RDCPAGE");
+            } else if (command == HITAG1_RDCBLK) {
+                exp_len += snprintf(exp + exp_len, size - exp_len, "RDCBLK");
```
iceman1001, line 88: maybe make the text a bit like the rest... "BLK -> BLOCK"

douniwan5788: These are actually Hitag 1 commands. I need to find a better way to distinguish between Hitag 1 and Hitag S.

PR #2795, opened 2025-03-19
Hitagµ

douniwan5788:


iceman1001:
Nice!

but  GH Actions is failing.

iceman1001:
the special characters is a no-go.     Uni-codes acts strange on all the different platforms we support.  So we do not use them.

iceman1001:
I have done this mistake,  and others with me,  over the years.   So just avoid it.

douniwan5788:
> the special characters is a no-go. Uni-codes acts strange on all the different platforms we support. So we do not use them.

Yes, the Hitag µ is annoying. I can’t find any official documentation that clearly explains what µ means—they just use the symbol µ everywhere, not 'mu' or 'micro.' I’m compromising by using the letter 'u' in the code, but I can’t find a proper way to display it for users.

iceman1001:
I think its related to mu/micro as a smaller product than 2/1/S...    but I dunno.

PR #2803, opened 2025-03-22
Correct typos in the compilation configuration item for PM3MAX

siowu:
Correct typos in the compilation configuration item for PM3MAX


iceman1001:
Good catch!

PR #2812, opened 2025-03-27
ZUID - Clarification on usable config bytes

Eltrick:


iceman1001:
maybe make it extra clear which bytes we mean with eight and seven?

Eltrick:
that was what the extra clarifications in parentheses was meant to be
though it's understandable it's not quite clear enough

iceman1001:
Sometimes it is better to visualise it to remove all kinds of doubts. 

```
                xx  zz
7AFF00000000000000
```

PR #2814, opened 2025-03-29
Standalone mattyrun spiffs

johndekroon:
Greetings,

I did some fieldwork experiments and I noticed there was no reliable and practical way to clone Mifare classic cards in a stand alone mode. Mattyrun comes close to this. It works fine when connected to the computer, but on standalone mode I wasn't able to access the memory again and dumping the contents. Apart from that, if you finally managed to clone a card by copying it from someone, it would be highly unpractical if you lose the contents because of your battery runs flat.

The new model contains spiffs, which is a useful way of storing the contents, which I implemented in this PR. I tested it, and it works fine for the MFC cards I encountered. I tested the cards on 5 different locations.

Before we can merge this, I need to get answers on a few questions:
- What happens if you run this on a proxmark without spiffs memory?
- What happens if the spiffs memory is full?
- For now, I only implemented the 1k cards, because I only encountered those. What would be the best way to add support for 2k and 4k cards, and how do I test it without having access to a system that uses cards like that?
- Most of the times, you first need to get the keys that are being used. There Is currently no easy way to add your keys. The only way is to add your keys to the source code, compile it and upload the new firmware. I thought of using the spiff memory and just add a text file with custom keys. I was wondering if that's the most simple way, or if there are other implementations worth to consider.

Thank you!

### DIFF
--- a/armsrc/Standalone/hf_mattyrun.c
+++ b/armsrc/Standalone/hf_mattyrun.c
@@ -33,6 +33,7 @@
 #include "mifaresim.h"  // mifare1ksim
 #include "mifareutil.h"
 #include "proxmark3_arm.h"
+#include "spiffs.h"
 #include "standalone.h" // standalone definitions
 #include "string.h"
 #include "ticks.h"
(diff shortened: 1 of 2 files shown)


iceman1001:
Nice improvements!

There is already a default keys dictionary file stored on spiffs if you init the rdv4 correct.  Maybe something to use?

iceman1001:
You would also need put in guards if people uses this standalone mode on non-spiffs enabled devices.

iceman1001:
ping

PR #2817, opened 2025-04-02
Improve Tesla NFC card reader command

elafargue:
This improves the `hf tesla info` command:

- Compatibility with more javacard variants
- Read the full certificate properly, not just the first 256 bytes
- Optionally parse the certificate (ASN.1)
- Don't bail at each error and try to read what it can
- Better form factor parsing
- Read all four public keys, not the first three

### DIFF
--- a/client/src/cmdhftesla.c
+++ b/client/src/cmdhftesla.c
@@ -24,6 +24,7 @@
 #include "cmdtrace.h"
 #include "cliparser.h"
 #include "cmdhf14a.h"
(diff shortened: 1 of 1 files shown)


iceman1001:
failing GH actions
bild

elafargue:
Updated the PR - the warning didn't appear on MacOS, it should be OK on all platforms now.

iceman1001:
Great!

PR #2821, opened 2025-04-07
Add support for setting polling loop annotations in `hf 14a config`

kormax:
This PR adds support for [Polling Loop Annotations](https://developer.android.com/reference/android/nfc/cardemulation/CardEmulation#registerPollingLoopFilterForService(android.content.ComponentName,%20java.lang.String,%20boolean)), a new NFC primitive added in Android 15.

Polling Loop Annotation (PLA) is a special NFC frame that is sent during the polling loop, which allows mobile device to perform special logic upon detecting one, for instance, for automatic selection of emulated credential, or for requesting authentication before communication is allowed to begin.

PLA is configured via `hf 14a config` and is persisted on PM3 side. If configured, it is automatically used in all instances where HF14A reader polling is performed, regardless of the command set, without the need to add special --pla argument on per-command basis.

Considering that PLA acts as a superset of ECP, if this change gets merged, ECP-specific parameters and functions can be removed in a next PR, as config-based option serves as a more powerful replacement.


iceman1001:
interesting PR,  never heard of PLA.  
However the :nocrc doesnt follow style.  
I suggest you add a literal `--pla-nocrc`  instead.

kormax:
> interesting PR, never heard of PLA. However the :nocrc doesnt follow style. I suggest you add a literal `--pla-nocrc` instead.

Thinking about it, I'd rather remove the NOCRC option altogether, as neither IOS nor Android work especially well with frames that lack CRC.

kormax:
If you'd like to test this functionality on your device, but you don't have a new Android device (Like Pixel 8 or newer) with an [appropriate app](https://github.com/kormax/android-observe-mode-demo), you can use one of the added examples with an iPhone, as it also works with ECP. For instance:
1)  Indicate that reader is "Transit" compatible:
```
hf 14a config --pla 6a02c801000300027900000000
```
2) Attempt EMV transaction. Your device should respond without auth:
```
emv exec -sa
```

PR #2824, opened 2025-04-15
Move `mag`safe polling argument to `hf 14a config`

kormax:
Additions:
* Move `mag` argument to `hf 14a config` option, allowing to enable/disable magsafe support for all places where HF14A polling is performed, instead of needing to add support on per-command basis;
* ~~Add `m` and `p` shorthands for `mag` and `pla` arguments in `hf 14a config`, allowing to set those options via shorthands like `-mon`, `-moff`, `-poff`~~;
* Mark `frame` argument in `ReaderTransmit*` functions as const to make it useable with const array references;
* Add more aliases for disabling `mag`/`pla` config options (`disable`/`off`/`skip`).

Removals:
* Remove `ecp` argument from all commands where it was present, as this functionality was replaced by more powerful `pla` in `hf 14a config`.
* Remove `mag` argument from all commands where it was present.

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -133,6 +133,36 @@ static uint32_t LastProxToAirDuration;
 #define SEC_Y 0x00
 #define SEC_Z 0xc0
 
+
+static const iso14a_polling_frame_t WUPA_CMD = {
+    { ISO14443A_CMD_WUPA }, 1, 7, 0
+};
+
+static const iso14a_polling_frame_t REQA_CMD = {
+    {ISO14443A_CMD_REQA }, 1, 7, 0
+};
+
+static const iso14a_polling_frame_t MAGWUPA_COMMANDS[4] = {
+    {{ 0x7A }, 1, 7, 0},
+    {{ 0x7B }, 1, 7, 0},
+    {{ 0x7C }, 1, 7, 0},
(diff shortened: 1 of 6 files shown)


iceman1001:
I don't see these in the changes,  what am I missing?

`
Add m and p shorthands for mag and pla arguments in hf 14a config, allowing to set those options via shorthands like -mon, -moff, -poff;
`

iceman1001:
clever little trick,   hard to follow.  I dislike those since it will make it hard for people who comes afterwards. 
Not to mention the nightmare to understand  why  `-moff` isn't  `-m -o -f -f` which would be the normal client parsing.
and `--moff`  doesn't work. 

I suggest you remove and simplify.

kormax:
> clever little trick

This is the default behavior for 90+% of cli parsers.

kormax:
After looking up why it works this way, I think most argparse implementations were inspired by [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) spec, but there are definitely differences on per-implementation basis.

iceman1001:
we have simpler arg parsing than POSIX,    and that is our code style.

kormax:
Shorthand argument aliases "p" and "m" were removed to prevent misuse and confusion.

PR #2826, opened 2025-04-16
fix offset transferred in ntag424_read_data

shuffle2:
havent seen a tag with such large space, but this looks wrong


iceman1001:
We could also use the value to bytes function and memcpy if that would be easier to read

PR #2832, opened 2025-04-29
Update downgrade guide with support for elatec reader

kitsunehunter:


iceman1001:
Interesting update,   
but the exe and bix...   If to be included in the project then we would need to know the software licenses and it should be under `tools\twn`  folder.

kitsunehunter:
> Interesting update, but the exe and bix... If to be included in the project then we would need to know the software licenses and it should be under `tools\twn` folder.

kitsunehunter:
> Interesting update, but the exe and bix... If to be included in the project then we would need to know the software licenses and it should be under `tools\twn` folder.

The bix is my own config files. The exe is from elatec. Just hard to find on the site.

iceman1001:
ok, 
move it to `tools\twn`

iceman1001:
Excellent

iceman1001:
Might comment that the exe is for windows and where to find for maos/linux equivalent

kitsunehunter:
> Might comment that the exe is for windows and where to find for maos/linux equivalent

will update the omnikey and twn4 directions for mac/linux at some point once I get a chance to test those platforms

PR #2839, opened 2025-05-05
Add Andalusian public transport card keys

imhexp:
Added 16 keys (A) from Consorcio de Transportes cards and 32 keys from EMT Málaga cards, making a total of 48 new keys.
The other 16 keys from Consorcio de Transportes are diversificated and need a hardnested attack to be recovered.


iceman1001:
Interesting,  but why does  the keys from # Card keys from EMT Malaga (Spain) bus system
feel like a private message you are sending?

imhexp:
> Interesting,  but why does  the keys from # Card keys from EMT Malaga (Spain) bus system
> feel like a private message you are sending?

They do? Interesting. I've never looked at them that way, but they are indeed real keys recovered from one of the readers.

PR #2853, opened 2025-05-22
Fixed SimulateIso14443aTag() to make MFU counter increments persistent in emulator memory.

sup3rgiu:
- Fixed `SimulateIso14443aTag()` to make MFU counter increments persistent in emulator memory:
Previously, the counters were copied to a local variable that was later used in read/write operations. I replaced this local variable with the actual blocks used to store the counters in the emulator memory.

- Fixed arguments for `SimulateIso14443aInit` in `hf_msdsal.c`, `hf_cardhopper.c`, `hf_reblay.c` and `hf_tcprst.c`.

### DIFF
--- a/armsrc/Standalone/hf_cardhopper.c
+++ b/armsrc/Standalone/hf_cardhopper.c
@@ -233,10 +233,8 @@ static void become_card(void) {
 
     tag_response_info_t *canned;
     uint32_t cuid;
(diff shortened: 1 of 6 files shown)


sup3rgiu:
The failed check does not appear to be related to the code changes, but rather to an error in the Windows environment setup.

PR #2868, opened 2025-05-29
update hf iclass legrec

Antiklesys:
Removed unused code/values from hf iclass legrec


Antiklesys (commented):
Done nothing, seeing if github stops erroring

PR #2869, opened 2025-05-30
Modified iclass recover and tear operations

Antiklesys:
1- Renamed legreclookup to legbrute to be in line with the command name 2- Updated estimate values with speed increase gains 3- Improved some if statements readability in iclass.c and added start_time = eof_time + DELAY_ICLASS_VICC_TO_VCD_READER; to increase speed by ~8% (1.86 loops per second to 2.01 loops per second = ~560 more loops per hour).

Tried disabling some arm communications/comments but the speed increase was negligible (~1 sec / 1000 updates).


iceman1001:
More speed!

PR #2871, opened 2025-05-31
Updated hf iclass legrec

Antiklesys:
Updated hf iclass legrec to support pm3 button interrupt for user abort
Cleaned up messaging to look more neat and tidy during the process


iceman1001:
Nice!

If the loop allows it,   we also usually add a  "dataavailable"  call since that will allow the client programmatically break the loop.  
or the "hw break"  command also to work.

Issue #2873, opened 2025-06-02
hf iclass sim broken since Blue Ice


Antiklesys:
***Things to try before submitting bug report***
read the [troubleshooting guide](https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/Troubleshooting.md)

***Compilation problems***
Try compiling with verbose.  `make VERBOSE=1` with main makefile or `make V=1` with cmake.

***flashing problems***
Have you followed the instructions properly?  ie,  flashed bootrom seperately first if you are going from Offical repo to Iceman repo.

-

**Describe the bug**
Something broke hf iclass sim -t 3 (not sure about the other modes) between the Orca and BlueIce release.

In Orca the simulation happens correctly, but with blueice it's throwing errors. I'm trying to read it with a flipper and i can get all the card details with Orca but with BlueIce the CSN is getting zeroed and there's clearly some issues with the simulation/emulation of the card.

This issue seems to be affecting the current version too.

**To Reproduce**

1. Compile BlueIce release or later
2. Run hf iclass eload -f <imagename>
3. Run hf iclass sim -t 3
4. Scan the pm3 with an omnikey or flipper zero
5. It will error and the flipper will report a CSN of all zeros

1. Compile Orca release or earlier
2. Run hf iclass eload -f <imagename>
3. Run hf iclass sim -t 3
4. Scan the pm3 with an omnikey or flipper zero
5. No errors all ok

**Expected behavior**
The simulation should happen correctly like it was happening in Orca or earlier releases.

iceman1001:
@n-hutton   it seem to be some kind of issue related to your changes to the fpga images.

Antiklesys:
For reference, this is the PR: https://github.com/RfidResearchGroup/proxmark3/commit/280b3301ee27be5b815b39f45f702ab5bd3306ae

Antiklesys:
Fixed now

PR #2876, opened 2025-06-02
revert date/time strip because of possible error with blue ice release

n-hutton:
Bitfiles are built in my environment but do not strip datetime from the image.

#2873


Antiklesys:
Tested working from my end

Antiklesys:
Re-testing with a fresh pull. My bad something was incorrect with my previous tests.

Antiklesys:
@iceman1001 & @n-hutton my "everything works" seems it was a fluke. Unfortunately the issue persists with these changes.

PR #2879, opened 2025-06-03
R&Y: Updated MAD TTP AID in `aid_desfire.json`

ry4000:
### Updated
- `000001` is not the MAD Tarjeta Transporte Público AID; it is actually `010000`, so removed references to it on that AID.
- `010000` removed the `(Alternative Endian)` designation.

Many thanks in advance, and kind regards

-R&Y.


iceman1001:
well,..  we have been keeping alternative endian since .. confusion.. ?

PR #2880, opened 2025-06-04
Update mfc_default_keys.dic

dandri:
add key for Waferlock shadow programming card and shadow user card


iceman1001:
Looked manually enough to believe its used across cards...
Now what is waferlock shadow ?..

PR #2882, opened 2025-06-04
change seed, making new iclass builds work again

n-hutton:
I found that changing the seed will allow iclass to work as before. reference issue: #2873 

Reverting to how the makefile was before, that is, not providing a seed, still does not work (internally defaults to 1 anyway).

I tested seeds 1 to 10 - of these, 1, 3 and 8 did not work.

It seems like it is not my environment, this would happen to others' builds. I would guess that there is a timing constraints issue that can cause this, perhaps to the pwr_oe* pins (which modulate the reader's field). Could also be some sort of metastability when crossing clock domains I suppose.


iceman1001:
Excellent!

PR #2885, opened 2025-06-06
Implemented working multithreading support in hf iclass legbrute

Antiklesys:
Implemented key nibble based multithreading support for hf iclass legbrute. It takes the whole iclass keyspace based on the first 4 bits of the key value (from 0x0 to 0xF) and divides it across the number of available threads.

E.g. on a 8 threads implementation:
- thread 1 will test keys starting with 0x0 onwards
- thread 2 will test keys starting with 0x2 onwards
- thread 3 will test keys starting with 0x4 onwards
- thread 4 will test keys starting with 0x6 onwards
- thread 5 will test keys starting with 0x8 onwards
- thread 6 will test keys starting with 0xA onwards
- thread 7 will test keys starting with 0xC onwards
- thread 8 will test keys starting with 0XE onwards

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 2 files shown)


Antiklesys:
I made a few edits from the browser and it wasn't a good idea :P Hopefully fixed now

PR #2899, opened 2025-06-11
LF HITAG: Adjust timing for LF ADC measurements to increase stability when the tag is in public mode/TTF mode

### LINKED ISSUE #2897: LF HITAG 2: Bug with initial LF ADC measurements when the tag is configured in public mode A
**Describe the bug**
Hello, I'm playing a bit with HITAG 2 CR80 cards using Proxmark 3 RDV4.01 and I've found something extremely strange here. When the tag is configured in the default "HITAG 2" operating mode it works perfectly and I can interact with it using `lf hitag dump` or `rdbl`, `wrbl` commands.

However, I've found out that if the tag is configured in "public mode A" with Manchester coding, all the abovementioned commands start to fail and the tag is almost uninteractable (still a few % chance that the command will go through successfully).

Here is an example of the problem with the tag configured in public mode A:

```
[usb] pm3 --> lf hitag dump --pwd
[=] Authenticating to Hitag 2 in Password mode
[#] List identifier in password mode
[#] Configured for Hitag 2 reader
[#] LF ADC average 89
[#] Detected timeout after [77] nrz samples
[#] Error in Manchester decoding, abort
[#] Unknown frame length: 21
[#] DEBUG: Error - hitag failed
```

```
[usb] pm3 --> trace list -t ht2
[#] downloading tracelog data from device
[+] Recorded activity ( 22 bytes )
[=] start = start of start frame. end = end of frame. src = source of transfer.
[=] Hitag 1 / Hitag 2 / Hitag S / Hitag µ - Timings in ETU (8us)
[#] Searching implicit relative paths
[#] Searching preferences paths
[#] Searching user .proxmark3 paths
[#] Searching current workdir paths
[+] Loaded 2 keys from dictionary file `/home/vm/Desktop/proxmark3/client/dictionaries/ht2_default.dic`

      Start |        End | Src | Data (! denotes parity error)                                           | CRC | Annotation
------------+------------+-----+-------------------------------------------------------------------------+-----+--------------------
          0 |        162 | Rdr | 5: 18                                                                   |     | START AUTH
        354 |       2355 | Tag |21: 16  4C  38                                                           |     | ?
```

The transponder's UID indeed starts with `164C3...` so it seems like it figured out the initial part correctly and then it got entirely lost with demodulation. The bug occurs in various stages - sometimes proxmark wouldn't see any response from the transponder, sometimes just a few bits etc.

Which is more interesting, I've found that this correlates with the actual contents of pages 4-5. The bug would very likely show up when page 4 is written with some low value like `0x00000004`. Though some other values are much less likely to trigger the bug and proxmark's commands will function properly.

I've also found out that in my case, the "LF ADC average" is around 120 for the cards that work correctly, and much lower (for instance 89) for cards that don't work.

The problematic card starts to work fully correctly after I had hardcoded the following in `lfadc.c:lf_sample_mean()`, thus bypassing the actual measurements:
```
    adc_avg = 120;
    previous_adc_val = adc_avg;
```

**Summary:** I belive what happens is that those LF ADC measurements were originally intended to be carried out under "radio silence" conditions. This makes perfect sense since a factory clean tag would not output anything by default. However, when the tag is configured in "public mode", it will continously transfer a static sequence of bits like an EM4100 transponder, and I belive that it somehow messes up with the initial LF ADC measurements, which then affects other calculations and causes proxmark getting lost.

I would be grateful if somebody could look at the issue and suggest a way to fix the issue.

**To Reproduce**

1. Initially, test `lf hitag dump` command with a factory clean HITAG 2 to ensure that it works correctly.
2. Configure the tag in public mode A and write low values to pages 4-5 (like `0x00000004`).
3. The `lf hitag dump` command should now be very likely to throw errors without completing succesfully.
4. The issue fades away if public mode A is disabled again.

**Expected behavior**

The `lf hitag dump` command succeeding regardless if any public mode is configured or not.

 - OS: [e.g. iOS] Ubuntu Noble
 - inside proxmark3 client run the following commands and paste the output here.
 - hw version
 - hw status
 - hw tune

```
[usb] pm3 --> hw version

 [ Proxmark3 ]

 [ Client ]
  Iceman/HEAD/v4.20142-240-gf41d6fad5-dirty-suspect 2025-06-11 03:15:03 0e3b986d1
  Compiler.................. GCC 13.3.0
  Platform.................. Linux / x86_64
  Readline support.......... present
  QT GUI support............ present
  Native BT support......... present
  Python script support..... present ( 3.12.3 )
  Python SWIG support....... present
  Lua script support........ present ( 5.4.7 )
  Lua SWIG support.......... present

 [ Model ]
  Device.................... RDV4
  Firmware.................. RDV4
  External flash............ present
  Smartcard reader.......... present
  FPC USART for BT add-on... absent

 [ ARM ]
  Bootrom.... Iceman/HEAD/v4.20142-240-gf41d6fad5-dirty-suspect 2025-06-11 03:14:50 0e3b986d1
  OS......... Iceman/HEAD/v4.20142-240-gf41d6fad5-dirty-suspect 2025-06-11 03:15:08 0e3b986d1
  Compiler... GCC 13.2.1 20231009

 [ FPGA ] 
 fpga_pm3_hf.ncd image 2s30vq100 10-06-2025 19:32:18
 fpga_pm3_lf.ncd image 2s30vq100 10-06-2025 19:32:18
 fpga_pm3_felica.ncd image 2s30vq100 10-06-2025 19:32:18
 fpga_pm3_hf_15.ncd image 2s30vq100 10-06-2025 19:32:18

 [ Hardware ]
  --= uC: AT91SAM7S512 Rev B
  --= Embedded flash memory 512K bytes ( 77% used )
```

```
[usb] pm3 --> hw status
[#] Memory
[#]   BigBuf_size............. 37988
[#]   Available memory........ 37988
[#] Tracing
[#]   tracing ................ 1
[#]   traceLen ............... 0
[#] Current FPGA image
[#]   mode.................... fpga_pm3_hf.ncd image 2s30vq100 10-06-2025 19:32:18
[#] Flash memory
[#]   Baudrate................ 24 MHz
[#]   Init.................... ok
[#]   Mfr ID / Dev ID......... EF / 11
[#]   JEDEC Mfr ID / Dev ID... EF / 3012
[#]   Memory size............. 256 kB (4 pages * 64k)
[#]   Unique ID (be).......... 0x25B98297307C69D5
[#] Smart card module (ISO 7816)
[#]   version................. v3.10 ( Outdated )
[#] LF Sampling config
[#]   [q] divisor............. 95 ( 125.00 kHz )
[#]   [b] bits per sample..... 8
[#]   [d] decimation.......... 1
[#]   [a] averaging........... yes
[#]   [t] trigger threshold... 0
[#]   [s] samples to skip..... 0 
[#] 
[#] LF T55XX config
[#]            [r]               [a]   [b]   [c]   [d]   [e]   [f]   [g]
[#]            mode            |start|write|write|write| read|write|write
[#]                            | gap | gap |  0  |  1  | gap |  2  |  3
[#] ---------------------------+-----+-----+-----+-----+-----+-----+------
[#] fixed bit length (default) |  29 |  17 |  15 |  47 |  15 | n/a | n/a | 
[#]     long leading reference |  29 |  17 |  15 |  47 |  15 | n/a | n/a | 
[#]               leading zero |  29 |  17 |  15 |  40 |  15 | n/a | n/a | 
[#]    1 of 4 coding reference |  29 |  17 |  15 |  31 |  15 |  47 |  63 | 
[#] 
[#] HF 14a config
[#]   [a] Anticol override........... std    ( follow standard )
[#]   [b] BCC override............... std    ( follow standard )
[#]   [2] CL2 override............... std    ( follow standard )
[#]   [3] CL3 override............... std    ( follow standard )
[#]   [r] RATS override.............. std    ( follow standard )
[#]   [m] Magsafe polling............ disabled
[#]   [p] Polling loop annotation.... disabled 00000000000000000000000000000000
[#] Transfer Speed
[#]   Sending packets to client...
[#]   Time elapsed................... 500ms
[#]   Bytes transferred.............. 348160
[#]   Transfer Speed PM3 -> Client... 696320 bytes/s
[#] Various
[#]   Max stack usage..... 3520 / 8480 bytes
[#]   Debug log level..... 1 ( error )
[#]   ToSendMax........... -1
[#]   ToSend BUFFERSIZE... 2308
[#]   Slow clock.......... 30981 Hz
[#] Installed StandAlone Mode
[#]   LF HID26 standalone - aka SamyRun (Samy Kamkar)
[#] Flash memory dictionary loaded
[#]   Mifare... 0 keys - dict_mf.bin
[#]   T55xx.... 0 keys - dict_t55xx.bin
[#]   iClass... 0 keys - dict_iclass.bin
```

```
[usb] pm3 --> hw tune

[=] -------- Reminder ----------------------------
[=] `hw tune` doesn't actively tune your antennas.
[=] It's only informative.
[=] Measuring antenna characteristics...
[#] DEBUG: (setClockGrid) clear settings
 🕑  19[#] Got Waiting Time eXtension request 1500 ms
 🕐   8[#] Got Waiting Time eXtension request 1500 ms


[=] -------- LF Antenna ----------
[+] 125.00 kHz ........... 71.96 V
[+] 134.83 kHz ........... 31.06 V
[+] 123.71 kHz optimal.... 73.17 V
[+] 
[+] Approx. Q factor measurement
[+] Frequency bandwidth... 12.2
[+] Peak voltage.......... 12.8
[+] LF antenna............ ok

[=] -------- HF Antenna ----------
[+] 13.56 MHz............. 46.95 V
[+] 
[+] Approx. Q factor measurement
[+] Peak voltage.......... 8.2
[+] HF antenna ( ok )

[=] -------- LF tuning graph ------------
[+] Orange line - divisor 95 / 125.00 kHz
[+] Blue line - divisor   88 / 134.83 kHz


[=] Q factor must be measured without tag on the antenna
```

rfidgeek1337:
resolve #2897

**WIP until the change is tested with different HITAG fobs/cards**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Added `hf_mfu_ultra.lua` script enables restoring dump to ULTRA/UL-5 tags and clearing previously written ULTRA tags (@mak-42)
 - Fixed `hf mfu sim` to make persistent the counter increases in the emulator memory (@sup3rgiu)
 - Fixed `hf mf mad` to correctly display MAD version 2 card publisher sector (@BIOS9)
+- Fixed `lf hitag dump` and related commands stability when tag is configured in public mode/TTF mode (@rfidgeek1337)
 
 ## [Blue Ice.4.20142][2025-03-25]
 - Added `des_talk.py` script for easier MIFARE DESFire handling (@trigat)
--- a/armsrc/lfadc.c
+++ b/armsrc/lfadc.c
@@ -240,7 +240,7 @@ void lf_init(bool reader, bool simulate, bool ledcontrol) {
     // minimum: 313T0 = 313 * 8us = 2504us = 2.50ms - Hitag2 tag internal powerup time
     //          280T0 = 280 * 8us = 2240us = 2.24ms - HitagS minimum time before the first command (powerup time)
     // maximum: 545T0 = 545 * 8us = 4360us = 4.36ms - Hitag2 command waiting time before it starts transmitting in public mode (if configured so)
-    //          565T0 = 565 * 8us = 4520us = 4.52ms - HitagS waiting time before entering TTF mode
+    //          565T0 = 565 * 8us = 4520us = 4.52ms - HitagS waiting time before entering TTF mode (if configured so)
     // Thus (2.50 ms + 4.36 ms) / 2 ~= 3 ms, should be a good timing for both tag models
     SpinDelay(3);
 
--- a/armsrc/lfadc.c
+++ b/armsrc/lfadc.c
@@ -241,7 +241,7 @@ void lf_init(bool reader, bool simulate, bool ledcontrol) {
     //          280T0 = 280 * 8us = 2240us = 2.24ms - HitagS minimum time before the first command (powerup time)
     // maximum: 545T0 = 545 * 8us = 4360us = 4.36ms - Hitag2 command waiting time before it starts transmitting in public mode (if configured so)
     //          565T0 = 565 * 8us = 4520us = 4.52ms - HitagS waiting time before entering TTF mode (if configured so)
-    // Thus (2.50 ms + 4.36 ms) / 2 ~= 3 ms, should be a good timing for both tag models
+    // Thus (2.50 ms + 4.36 ms) / 2 ~= 3 ms (rounded down to integer), should be a good timing for both tag models
     SpinDelay(3);
 
     // Steal this pin from the SSP (SPI communication channel with fpga) and use it to control the modulation
--- a/armsrc/lfadc.c
+++ b/armsrc/lfadc.c
@@ -236,8 +236,13 @@ void lf_init(bool reader, bool simulate, bool ledcontrol) {
     FpgaSetupSsc(FPGA_MAJOR_MODE_LF_READER);
 
     // When in reader mode, give the field a bit of time to settle.
-    // 313T0 = 313 * 8us = 2504us = 2.5ms  Hitag2 tags needs to be fully powered.
-    SpinDelay(10);
+    // Optimal timing window for LF ADC measurements to be performed:
+    // minimum: 313T0 = 313 * 8us = 2504us = 2.50ms - Hitag2 tag internal powerup time
+    //          280T0 = 280 * 8us = 2240us = 2.24ms - HitagS minimum time before the first command (powerup time)
+    // maximum: 545T0 = 545 * 8us = 4360us = 4.36ms - Hitag2 command waiting time before it starts transmitting in public mode (if configured so)
+    //          565T0 = 565 * 8us = 4520us = 4.52ms - HitagS waiting time before entering TTF mode
+    // Thus (2.50 ms + 4.36 ms) / 2 ~= 3 ms, should be a good timing for both tag models
+    SpinDelay(3);
 
     // Steal this pin from the SSP (SPI communication channel with fpga) and use it to control the modulation
     AT91C_BASE_PIOA->PIO_PER = GPIO_SSC_DOUT;


iceman1001:
Got verified by @en4rab that this patch works as expected!

PR #2902, opened 2025-06-13
Poland Warsaw public transport card keys

EpicPlayerA10:
All 32 keys extracted from the card reader. All cards have the same set of keys.


iceman1001:
Nice,  if you clean out the duplicates first ,   then we can merge

EpicPlayerA10:
> Nice, if you clean out the duplicates first , then we can merge

Done

PR #2914, opened 2025-06-22
fix card limits for S12906 format

TeCHiScy:
- As S12906 pack function suggests, the issue level field takes 2 bits, thus ranging from 0 to 3. The card number takes 24 bits, ranging from 0 to 0xffffff. Reference from [1](https://acre.my.site.com/knowledgearticles/s/article/x107), [2](https://www.everythingid.com.au/hid-card-formats-i-15?srsltid=AfmBOor2UAGlvB7R6Zmj7B7rtL-LExfBjh7I3ZEyoLLbg7Pk7UbC1za-) also confirms the CN should be ranging from 0 to 0xffffff. https://github.com/RfidResearchGroup/proxmark3/blob/61a993de826335db2d92d1f60c3ef4ae64439b73/client/src/wiegand_formats.c#L684-L685
- That shows an inconsistent with the limits defined in the format table. Specifically, the IL/CN limits in the table was reversed, accidentally.
https://github.com/RfidResearchGroup/proxmark3/blob/61a993de826335db2d92d1f60c3ef4ae64439b73/client/src/wiegand_formats.c#L1480
- I propose to fix the limits of S12906 format.


iceman1001:
Good catch!

PR #2925, opened 2025-07-04
Updated iclass restore to support privilege escalation

Antiklesys:
Updated hf iclass restore to support privilege escalation to restore card's content using a single AA1 --nr mac value.
This allows to write cards for which the debit key is not known.


Antiklesys:
image

PR #2927, opened 2025-07-05
Updated hf iclass wrbl replay

Antiklesys:
replay behavior to use privilege escalation instead of having to generate specific block/content macs for hf iclass wrbl


iceman1001:
Now you are breaking old functionality.   You need to rethink this one.

Antiklesys:
@iceman1001  should be fixed now

Antiklesys:
Running more tests as it now broke the privilege escalation as I'm checking the length of the mac, but the mac is always 4 0 bytes and the only actual check happens on client side. I'll have to check for the mac value to be != from 00000000

Antiklesys:
issue_fix_final_forreal_latest_v3 :D 
Tested both behaviors (original and privilege escalation) and both working now.

PR #2929, opened 2025-07-05
Add Gen2 Magic Card Detection method responds

0x6r1an0y:
1. **Expanded ATS Signature Dictionary**
- NT2H1611G0DU and NT2H1511G0DU ATS responds

2. **Improved Detection Method Classification**
- Distinguished between ATS-based detection (`MAGIC_FLAG_GEN_2_ATS`) and invasive detection (`MAGIC_FLAG_GEN_2_INVASIVE`)

ATS S
invasive

### DIFF
--- a/armsrc/mifarecmd.c
+++ b/armsrc/mifarecmd.c
@@ -3051,6 +3051,12 @@ void MifareCIdent(bool is_mfc, uint8_t keytype, uint8_t *key) {
         } else if (memcmp(card->ats, "\x85\x00\x00\xA0\x00\x00\x0A\xA5\x00\x04\x04\x02\x01\x00\x0F\x03\x79\x0C", 18) == 0) {
             // test for NTAG213 magic gen2
             isGen2 = true;
+        } else if (memcmp(card->ats, "\x85\x00\x00\xA0\x00\x00\x0A\xAA\x00\x04\x04\x02\x01\x00\x13\x03\x60\xA6", 18) == 0) {
+            // test for non-standard NTAG216 magic gen2 888bytes (NT2H1611G0DU) ( USCUID-UL )
+            isGen2 = true;
+        } else if (memcmp(card->ats, "\x85\x00\x85\xA0\x00\x00\x0A\x5A\x00\x04\x04\x02\x01\x00\x11\x03\x44\xB4", 18) == 0) {
+            // test for non-standard NTAG215 magic gen2 504bytes (NT2H1511G0DU) ( USCUID-UL )
+            isGen2 = true;
         }
 
         if (isGen2) {
--- a/armsrc/mifarecmd.c
+++ b/armsrc/mifarecmd.c
@@ -3033,33 +3033,30 @@ void MifareCIdent(bool is_mfc, uint8_t keytype, uint8_t *key) {
         if (memcmp(card->ats, "\x09\x78\x00\x91\x02\xDA\xBC\x19\x10", 9) == 0) {
             // test for some MFC gen2
             isGen2 = true;
-            flag |= MAGIC_FLAG_GEN_2_ATS;
         } else if (memcmp(card->ats, "\x0D\x78\x00\x71\x02\x88\x49\xA1\x30\x20\x15\x06\x08\x56\x3D", 15) == 0) {
             // test for some MFC 7b gen2
             isGen2 = true;
-            flag |= MAGIC_FLAG_GEN_2_ATS;
         } else if (memcmp(card->ats, "\x0A\x78\x00\x81\x02\xDB\xA0\xC1\x19\x40\x2A\xB5", 12) == 0) {
             // test for Ultralight magic gen2
             isGen2 = true;
-            flag |= MAGIC_FLAG_GEN_2_ATS;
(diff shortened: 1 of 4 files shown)


iceman1001:
Its alway fun to improve things.  I totally agree. 
When it comes to screen space,  the pm3 is trying to be "clean" and "crisp",
this PR to identify between which way was used to determine a gen2 card,   I don't see it bring any useful information to be printed.  The only needed output is "gen2",  how that was determined is not very interesting for a end user. 

Not to mention It clutters up the output.   

I think you can find ever more parts to improve with the proxmark3 project,  and I welcome it.  
but think functionality and usability  when figuring out what to improve.

0x6r1an0y:
Hi, thanks for your feedback. I totally understand the point about keeping the output clean. Let me explain the background a bit more, so you see where this PR came from.
I have a NTAG 216 card (NT2H1611G0DU), which is classified under Ultralight by `hf 14a info`.
```
[usb] pm3 --> hf 14a info

[+]  UID: 04 71 F6 0F 32 02 89   ( double )
[+] ATQA: 00 44
[+]  SAK: 00 [4]
[+]       NXP Semiconductors Germany
[+] Possible types:
[+]    MIFARE Ultralight
[=] -------------------------- ATS ----------------------------------
[!] ATS may be corrupted.
[=] Length of ATS (18 bytes incl. 2 Bytes CRC) doesn't match TL
[+] ATS: 85 00 00 A0 00 00 0A AA 00 04 04 02 01 00 13 03 [ 60 A6 ]
[=]      85...............  TL    length is 133 bytes

[?] Hint: Use `hf mfu *` magic commands
[?] Hint: Try `hf mfu info`
```
However, this card is a [USCUID-UL](https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/magic_cards_notes.md#uscuid-ul-configuration-guide) magic card with configurable features 
```
[usb] pm3 --> hf mfu info

[=] --- Tag Information --------------------------
[+]       TYPE: NTAG 216 888bytes (NT2H1611G0DU) ( USCUID-UL )
[+]        UID: 04 71 F6 0F 32 02 89
[+]     UID[0]: 04, NXP Semiconductors Germany
[+]       BCC0: 0B ( ok )
[+]       BCC1: B6 ( ok )
[+]   Internal: 48 ( default )
[+]       Lock: 00 00  - 0000000000000000
[+]        OTP: E1 10 6D 00  - 11100001000100000110110100000000

(truncate)
```
In my case, this card is configured to allow Gen2 commands, but hf 14a info cannot detect that with the current logic, unless more ATS patterns are manually added to the code.
Meanwhile, `hf mfu info` does detect its USCUID-UL and magic capabilities properly. The issue is, for some NTAG cards, the current detection logic in `hf 14a info` has a blind spot. The output part — showing "(detected via ATS signature)" — was meant for user awareness and debugging. I just wanted to share the background so the reasoning behind the PR is clearer.

0x6r1an0y:
Also, I hope you don't mind me asking for a bit more clarification regarding the detection logic and flag handling. 

1. I figure out that `hf 14a info` already includes some basic detection for Ultralight Gen2 magic cards, based on hardcoded ATS signatures like this:

```c
if (memcmp(card->ats, "\x0A\x78\x00\x81\x02\xDB\xA0\xC1\x19\x40\x2A\xB5", 12) == 0) {
    isGen2 = true;
    flag |= MAGIC_FLAG_GEN_2;
}
```

However, based on the USCUID-UL documentation, the ATS structure provides a more convenient way to detect gen2 behavior. `byte 7` indicates whether the card is in CUID mode. Maybe adjusting the logic to check this specific byte would make detection more robust.

```
85 00 00 A0 00 00 0A AA 00 04 04 02 01 00 13 03 60 A6 
                  ^^ >> 0A for enable CUID mode
```

I completely respect your emphasis on keeping `hf 14a info` clean, both in logic and output. My question is:

- Since `hf 14a info` already performs some Ultralight Gen2 detection,
  Would it be acceptable to refine the **internal logic**, based on this more reliable ATS structure?

- Alternatively, if you prefer that all Ultralight Gen2 detection be strictly isolated to `hf mfu info`, I can respect that too. I just want to make sure I don't accidentally work in the wrong direction.

---

2. For NTAG 216 (NT2H1611G0DU) cards, especially when configured with USCUID-UL to enable Gen2-like behavior:
Should these cards be assigned: Only `MAGIC_FLAG_NTAG21X` or`MAGIC_FLAG_GEN_2` or both flags at the same time?

---

I truly appreciate the opportunity to contribute to the project. Sometimes it's a bit hard for me to fully grasp where the boundaries working in the big project, your clarification would really help me align my work with your vision.

iceman1001:
Yes,  refine internal logic is totally fine.   I merge that instantly  (well,  after memory checks etc  ;)
If you can refactor the gen2 interal logic to handle gen2 in same way on both MFU / 14a,    then that is a good PR. 

When things going to be printed,  we work with colors,  to point out important parts,   so a user quickly can identify the critical information.   If you need DEBUG things,   or a  "verbose" output ,  we support that too. 

For DEBUG,  its usually for debugging critical functions which doesn't make so much sense for anyone else then a developer.
for verbose,   its when its good information but not important enough to always be seen.

PR #2932, opened 2025-07-09
fixes to mqtt (win sockets)

virtyvoid:
- proper pthread cancellation (segfault).
- closesocket instead of close & using SOCKET type instead of int (gdb warning).


iceman1001:
Awesome!

PR #2934, opened 2025-07-12
Add recognition of Swissbit iShield Key Mifare as MFD EV3

ah01:
I got my hands on [iShield Key 2 Pro Mifare](https://www.swissbit.com/en/products/security-products/ishield-key/). This authentication token supports MIFARE DESFire EV3 (info from NXP's TagInfo is _DESFire EV3C implementation on P71D600_). 

It has different version info than typical DESFire, so here it is.

PS: Info from pm3 client before this change:

```
[=] HW Version.. 04810043011A05
[=] SW Version.. 04814603001A05
[=] Version data identification failed. Report to Iceman!
```


iceman1001:
Awesome!

PR #2937, opened 2025-07-16
ModemManager solution 2 Updated

grugnoymeme:
Just updated the Solution number 2 for the ModemManager linux's users problem.
I think that to mask the service is gonna be a better choice than just disable it.
This because some external services or some softwares coud be re-enable it or just re-activate it without asking it, so this way is clean and safe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Changed `mem load` - now handles UL-C and UL-AES dictionary files (@iceman1001)
 - Changed `hf mfu sim` - now support UL-C simulation (@iceman1001)
 - Added `!` - run system commands from inside the client. Potentially dangerous if running client as SUDO, SU, ROOT (@iceman1001)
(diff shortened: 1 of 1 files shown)


grugnoymeme:
yup, sorry i forgot it.. now should be ok.

iceman1001:
Ok,  Since I don't know,   what is masking doing  ?

grugnoymeme:
it's keeping the service installed but totally inactive, like i'd be dead..
it's stronger than just a disable option, So this way is very safe to avoid mistakes..

"When you disable a service with sudo systemctl disable, all symlinks for the service are removed. When masking a service, the symlinks are moved and then point to /dev/null. When you simply disable a service, it can still be started manually. When you mask a service, it cannot be started manually. In other words, masking a service renders the service permanently unusable until it’s unmasked."

iceman1001:
Ok,  thanks for the explanation!

grugnoymeme:
thank u for ur super work!!!

iceman1001:
> thank u for ur super work!!!

Its a community effort!

PR #2940, opened 2025-07-19
Improved DESFire support and added emulation

mistial-dev:
- Added DESFire emulator in armsrc/desfiresim.c with all major commands
- Fixed value operations (credit/debit/limited credit) to work correctly in MAC mode
- Removed value operations from EV1D40TransmitMAC array to match real card behavior
- Added auto-detection fallback for MAC mode compatibility with older cards
- Added emulator with 28 applications (kinda EV1 standard)
- Enhanced test suite with comprehensive value operation and multi-app testing
- Added emulator commands: sim, ereset, eload, eview, test
- Consolidated user documentation in doc/desfire.md with emulation guide
- Value file operations work on both emulator and real DESFire cards
- Proper limit checking and transaction commitment for value operations
- Support for all file types: Standard, Backup, Value, Linear/Cyclic Record
- Authentication with DES, 2TDEA/3DES, 3TDEA, and AES keys
- ISO file ID support and access rights enforcement

### DIFF
--- a/armsrc/Makefile
+++ b/armsrc/Makefile
@@ -37,7 +37,7 @@ APP_CFLAGS = $(PLATFORM_DEFS) \
 SRC_LF = lfops.c lfsampling.c pcf7931.c lfdemod.c lfadc.c
 SRC_HF = hfops.c
 SRC_ISO15693 = iso15693.c iso15693tools.c
-SRC_ISO14443a = iso14443a.c mifareutil.c mifarecmd.c epa.c mifaresim.c sam_common.c sam_mfc.c sam_seos.c
+SRC_ISO14443a = iso14443a.c mifareutil.c mifarecmd.c epa.c mifaresim.c sam_common.c sam_mfc.c sam_seos.c desfiresim.c
 
 #UNUSED: mifaresniff.c
 SRC_ISO14443b = iso14443b.c
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -55,6 +55,7 @@
 #include "mifarecmd.h"
 #include "mifaredesfire.h"
 #include "mifaresim.h"
+#include "desfiresim.h"
 #include "emvsim.h"
 #include "pcf7931.h"
 #include "Standalone/standalone.h"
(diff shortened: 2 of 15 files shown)


iceman1001:
This is an interesting PR.    

Usually its easier with smaller pull requests and change it gradually to easier find and keep a history changes.
With that said,   the build breaks at

image

mistial-dev:
> This is an interesting PR.
> 
> Usually its easier with smaller pull requests and change it gradually to easier find and keep a history changes. With that said, the build breaks at

Thanks.  I was able to replicate.  Large commits is a bad habit, I don't like to commit broken code, and I often find things mid-refactor.  6-8 hours later, I come up for air and commit.  

I've found some CLI weirdness, and a couple bugs while doing my own review, so I'll get those fixed and committed.

iceman1001:
The PR could be split up with  smaller ones,   here is a simple suggestion.

1.     Removed value operations from EV1D40TransmitMAC array to match real card behavior
2.     Added auto-detection fallback for MAC mode compatibility with older cards
3.     Limit the EV1 26 apps
4.     Added emulator with 28 applications (kinda EV1 standard),  plus the rest
5.     Enhanced test suite with comprehensive value operation and multi-app testing
6.     Consolidated user documentation in doc/desfire.md with emulation guide


I had a look at the changes.   Some other feedback,  I noticed you put the command dispatcher code inside  `arm/iso14443a.c`  and that file is too large as it is.  I think its better if you move the dispatcher part into  `arm/desfiresim.c` .   but I see this code uses "14a sim" as base.  Hm...... decisions, decisions 


The reused defines and protocol commands should in in  `include/desfire.h`
The new structs should reside somewhere in  `common\`   but it could also in inside the aforementioned include/desfire.h for a single place.


Since we don't have a dump format for desfire yet,  there are some suggestions for json file layouts. 
The file extention should be `.json` and  not  `.mfdes`,    the normal way is to put the targeted card type inside the file name part.
ie: `hf-mfdes-<UID>-dump.json`  


Regarding the help text changes, 
The help text didn't follow the `client` pattern before,   but making the text shorter and concise is preferred.  
`"Key type (default: 2TDEA for factory cards)")`   -> `"Key type (deft: 2TDEA)"),` 
If more words are needed then it should be in the help text description part explaining the parameter better. 


I haven't look at the emulator / arm side memory usage but I figured you use the bigbuf correct.   No "large" allocations of arrays etc. 


I am impressed with the code quality,   looks like you worked quite a bit with it. 
The tests added is a nice touch!

iceman1001:
And I don't expect there to be any blockers on this one.  Will be merged quite fast.   Getting a file format design for desfire dumps is the major thing missing.  Which can be added later, so nothing there either.

mistial-dev:
> The PR could be split up with smaller ones, here is a simple suggestion.

Ok.  I'll refactor this into smaller PRs.  Closing this for now.

PR #2956, opened 2025-08-17
Add HID® Crescendo® AIDs to AID list

jkramarz:


iceman1001:
Why did you remove one entry?

jkramarz:
Lines 1243 and 1251 currently contains exactly the same entries for the same AID A000000308000010000100, so I assumed it was a mistake. Shall I remove the first commit from the PR?

iceman1001:
Thanks!  Now we need to modify the unique check in order for this not to happen again

PR #2960, opened 2025-08-28
added hf saflok command

STIEBELJOSHUA:
added hf saflok commands to:
Read and decode mfc saflok cards
Provision new saflok cards
Encode new saflok data blocks
Modify saflok data blocks
interrogate variable keys from saflok cards

I Still need to add support for log parsing and ULC support and clean up the code a bit but this is what I have so far


iceman1001:
Excellent!

PR #2962, opened 2025-08-30
add generated keys to mfc_default_keys.dic

Hargriv:
add generated keys to mfc_default_keys.dic


iceman1001:
The default keys dictionary is a curated list with known static default keys across cards and system.

What is your Generated keys coming from?

Hargriv:
These are the keys generated by the script for my intercom key, some of the keys that are already in the file I did not add so as not to duplicate the keys and some that are not I added to the file

iceman1001:
Is there generator that is able to generate these keys?
and is the keys same across multiple cards on the intercom system?

Hargriv:
Here is my script for generating keys from one known key, The first key was obtained using the vulnerability of mfkey32

As for other cards from this system, I can't say whether the same keys will be used or whether there will be some kind of shift, since I don't have another such card to check

[mifare_keys.zip](https://github.com/user-attachments/files/22064408/mifare_keys.zip)
[mifare_keys.txt](https://github.com/user-attachments/files/22064409/mifare_keys.txt)

PR #2963, opened 2025-08-30
Add zero initizlier for DesfireContext_t to prevent UB

kormax:
This PR adds zero-initialization to all uses of `DesfireContext_t` in order to prevent issues, such as with automatic DF Name selection, which happened for users if DFNameLen was initialized to non-zero value (which happens on some pre-C23 compilers).


iceman1001:
Perfect!

PR #2968, opened 2025-09-01
Improve output checks in mf_backdoor_dump.py

robo-w:
Apparently the number of whitespaces in the parsed output changed over time. This fix should be able to handle all variations of it.


iceman1001:
Thank you!

PR #2973, opened 2025-09-03
Wiegand simplex36 update

henrygab:
* more `const` correctness for utility functions, which allows compiler to optimize more, and allows caller to choose to store the data in ROM (instead of copying it to RAM before calling these functions).
* Add alternate decoding for HID Simplex 36-bit 

<details><summary>Source images from Discord</summary>
<p>

image
image
image

</p>
</details>


client/src/wiegand_formats.c
```diff
@@ -704,6 +704,59 @@ static bool Unpack_S12906(wiegand_message_t *packed, wiegand_card_t *card) {
     return true;
 }
 
+
+// if (card->FacilityCode != 217) return false; // Must be 0b11011001 aka 217 for Grinnell?
```
iceman1001: There is a few of these comments,   they don't belong in a PR

iceman1001: hm, looking at the Source images there is a Fix data part in the algo.   Hm....

henrygab: You are right.  I removed the extra comments.
I also uncommented the requirement for the fixed facility code, so it's now required to match that fixed value on both encode and decode.

PR #2979, opened 2025-09-10
Add --page flag to hf iclass dump

R4H33M:
PicoPass / IClass cards can have multiple pages but right now the Proxmark only allows inspecting and modifying the first page. This is a first step to adding some functionality to work with multi-page Picopass cards. I'm hoping to add some other features as well like being able to simulate an entire 32K picopass card (or just some specific pages if needed).

This is just a rough first attempt that doesn't completely work yet, due to how the iclass dump command reads the card "twice". I wanted to get the code out there and get some feedback first.

### DIFF
--- a/armsrc/iclass.c
+++ b/armsrc/iclass.c
@@ -1306,7 +1306,7 @@ static bool iclass_send_cmd_with_retries(uint8_t *cmd, size_t cmdsize, uint8_t *
  * @return false = fail
  *         true = Got all.
  */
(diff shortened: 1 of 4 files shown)


iceman1001:
Interesting addition,   the "rough first PR"...    In what state do you consider this to be?

R4H33M:
Yea it's not in a acceptable state right now. I think I'll shelve this until I have something more complete.

iceman1001:
Ping!

Tine to get this in better shape.

PR #2991, opened 2025-09-22
remove unused wiegand files from armsrc

jkramarz:
Not referenced by any other files in armsrc.


iceman1001:
Those belongs to something that hasn't gotten merged.   
I will keep them in the repo for now.  If they bother you ,   we could rename them  "xxx_disabled"..

jkramarz:
No worries, it looked like something stale from "the original firmware", that duplicates code currently implemented in `client`.

iceman1001:
Yeah,   the whole wiegand decode/encode should be moved to common files between ARM and CLIENT

PR #2999, opened 2025-10-09
Adding Verkada 40-bit format

aaronmaxlevy:
Adding support for the 40-bit Wiegand format used by Verkada for their DESFire cards (and possibly other things?).


iceman1001:
Cool,   do you have some raw samples to use with regression testing?

aaronmaxlevy:
> Cool, do you have some raw samples to use with regression testing?

Here are some samples:
`86400007d3` : FC: 50  CN: 1001
`86400007d9`: FC: 50  CN: 1004
`8a20002721`: FC: 81  CN: 5008

Where would be the best place to add these — I see a couple of test cases in `tools/pm3_tests.sh` but not sure if there is a better place?

Thanks!

iceman1001:
`tools/pm3_tests.sh` is a good place to run some odd tests.

aaronmaxlevy:
> `tools/pm3_tests.sh` is a good place to run some odd tests.

Sounds good — I just pushed a commit that adds these test cases there.

aaronmaxlevy:
Thanks!

PR #3007, opened 2025-10-23
R&Y: Corrected `Jett's` Keys in `mfc_default_keys.dic`

ry4000:
# Updates
- `#` was removed from the `Jett's` keys making them valid keys once again.

Many thanks in advance, and kind regards,

-randy.


iceman1001:
since that line is no more,  maybe update first post to match the PR better?

ry4000:
> since that line is no more, maybe update first post to match the PR better?

Updated with apologies.

PR #3011, opened 2025-11-02
Remove direct use of strndup to support older compilers

kormax:


iceman1001:
We have our own `str_ndup`  function in utils.c

PR #3015, opened 2025-11-06
Add dialout fix instructions for WSL2

jwidess:
Added a small section with instructions for fixing USB device access in WSL2, due to the user not having access to dialout. I've run into this a few times over the last year or two of using Proxmark when installing on a new machine with WSL, thought adding this might help out newbies unfamiliar with Linux.

### DIFF
--- a/doc/md/Installation_Instructions/Windows-WSL2-Installation-Instructions.md
+++ b/doc/md/Installation_Instructions/Windows-WSL2-Installation-Instructions.md
@@ -393,6 +393,26 @@ sudo service udev restart
 sudo udevadm trigger --action=change
 ```
 
+#### USB device access in WSL2 (tty permissions)
+
+When you attach the Proxmark3 to WSL2 via `usbipd`, the Linux device node (for example `/dev/ttyACM0`) is created inside the WSL distro. That device is usually owned by `root:dialout` and is only readable/writable by `root` and members of the `dialout` group.
+
+If you are seeing `Could not find Proxmark3 on /dev/ttyACM0`, it usually means your user doesn't have permission to open the serial device.
+
+Fix:
+```bash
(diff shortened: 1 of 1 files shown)


iceman1001:
I think its mentioned in the documentation already.    If you run  `make accessrights`   it should add it. 

As seen here:
https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/Windows-WSL2-Installation-Instructions.md#install-the-udev-rules-1


Feel free to improve the text there instead.

jwidess:
> I think its mentioned in the documentation already.    If you run  `make accessrights`   it should add it. 
> 
> As seen here:
> https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/md/Installation_Instructions/Windows-WSL2-Installation-Instructions.md#install-the-udev-rules-1
> 
> 
> Feel free to improve the text there instead. 

Apologies, I should have made it more clear that this was a fix on top of running `make accessrights` as that still didn't give my user access to dialout. Could be a problem with make or my WSL2 instance, but this did fix it. Thanks again @iceman1001 for your work!

iceman1001:
I think I would love to see our  "make accessrights" command handle this execption aswell.

jwidess:
> I think I would love to see our "make accessrights" command handle this execption aswell.

Apologies, forgot to respond! Thanks for the merge, but yes I will take a look at the `accessrights:` section of the Makefile to see if adjusting that will fix this issue, I'm guessing this might be an issue unique to WSL2 and the environment variables used.

iceman1001:
Whilst documentation is great also having the commands handle the exceptions is crucial.    Having to explain for end users that they have to look into documentation and follow a fix ,  rather than fix the command  makes the whole user experience not optimal.

PR #3017, opened 2025-11-07
Fix invalid  country code len for lf fdxb reader command

VarChar42:
When running "lf fdxb reader" in continuous mode, the country code was formatted as 4 chars long instead of 3.


iceman1001:
Nice catch,  it is a zero padded string with one extra '0' char....    Not quite the dramatic invalid country code len title ...

Issue #3018, opened 2025-11-10
QR code encoding error


BullyDeveloper:
1. Go to `client/src/qrcode`
2. When running command `make -j8 all` the file **qrcode.c** is returning an error

Changed to the previous commit where this file is not even commited and everything works fine

iceman1001:
Missing a bunch details,  however the actions shows that compilation fails at Proxspace / Mingw env.

Alex1602:
```
pm3 ~/proxmark3$ make clean && make -j10 all

===================================================================
Version info:      Iceman/master/v4.20728-129-gfd06c38a8
Platform name:     Proxmark3 generic target
PLATFORM:          PM3GENERIC
PLATFORM_FPGA:     xc2s30
PLATFORM_SIZE:     512
Platform extras:   No extra selected
Included options:  -DLED_ORDER_PM3EASY LF HITAG EM4x50 EM4x70 ZX8211 GENERAL_HF ISO15693 LEGICRF ISO14443b ISO14443a ICLASS FELICA NFCBARCODE HFSNIFF HFPLOT COMPRESSION
Standalone mode:   LF_SAMYRUN
===================================================================
[*] MAKE hitag2crack/clean
[*] MAKE crack2/clean
[*] MAKE crack3/clean
[*] MAKE crack4/clean
[*] MAKE crack5/clean
[*] MAKE crack5opencl/clean
[*] MAKE client/clean
===================================================================
Version info:      Iceman/master/v4.20728-129-gfd06c38a8
Client platform:   MINGW64_NT-10.0-26100
GUI support:       QT5 found, enabled (Qt version 5.15.10 in C:/ProxSpace/msys2/mingw64/lib)
native BT support: Bluez not found, disabled
Jansson library:   system library found
Lua library:       system library found
Python3 library:   Python3 v3.11 found, enabled
GD library:        GD not found, disabled
Readline library:  enabled
Whereami library:  system library not found, using local library
Lua SWIG:          wrapper found
Python SWIG:       wrapper found
compiler version:  cc.exe (Rev2, Built by MSYS2 project) 13.2.0
===================================================================
PLATFORM MINGW64_NT-10.0-26100
[*] MAKE bootrom/clean
[*] MAKE fpga_compress/clean
[*] MAKE armsrc/clean
[*] MAKE recovery/clean
[*] MAKE mfc_card_only/clean
[*] MAKE mfc_card_reader/clean
[...]

PR #3019, opened 2025-11-10
Added explicit utf-8 encoding to solve build problems

Alex1602:
This solves:
QR code encoding error     #3018


Alex1602:
Ah I see you added the hex values :) will close this one

iceman1001:
Nice "u8"   I didn't know of that notation before.    @nvx   solved it with hex instead.

nvx:
TIL, strange some OS/compiler versions didn't care, oddly even mingw didn't seem to care on some of the values but did care with others

iceman1001:
🤷🏼     but now we have a fix !

PR #3021, opened 2025-11-10
Replaced hf mf bambukeys with hf mf keygen with multiple KDFs support + Snapmaker U1 filament spool KDF

Foxushka:
Also replaced output with table for better experience


iceman1001:
Interesting PR,   I have some improvements to the kdf table concept

PR #3024, opened 2025-11-15
[hf seos] Fix buffer overflow

aaronjamt:
Fixes a crash on some systems due to a buffer overflow. There are 2 parts:
1. The `padToBlockSize` method is passed a buffer of length `block_size - ARRAYLEN(command_trailer)`, but told to pad it to `block_size` bytes, resulting in a buffer overflow of `ARRAYLEN(command_trailer)` bytes.
3. Instead of allocating the buffer with size=`command_len * 2 + 1`, the `command_len` int is passed to `sizeof()`, which returns the size of an `int` rather than the actual `command_len` value.
2.1. There is also no reason to calculate `sizeof(command_convert)` when `command_convert` is defined above as being `command_len` bytes.


iceman1001:
Great catch!

PR #3025, opened 2025-11-15
[hf seos] Support variable length SIO

aaronjamt:
Currently, Seos credentials are expected to produce a 64-byte cryptogram, but not all SIOs fit in 64 bytes. This PR allows the Proxmark to decode an arbitrary length cryptogram.


iceman1001:
Excellent!

PR #3027, opened 2025-11-18
[hf seos] Correct command padding size

aaronjamt:
Currently, the unencrypted command is truncated at `block_size` bytes. This PR calculates the actual size it should be, rounded up to the nearest multiple of `block_size`.


iceman1001:
Interesting,   I believe a better way is to move this patch inside the fct `padToBlockSize()`   instead of having it outside every where.

Meaning you make the input length a pointer and modify it.

PR #3031, opened 2025-11-21
[hf seos] Updated key storage and usage

aaronjamt:
This PR splits the Privacy and Authentication keys into separate lists. This allows for more control over which keys are used, as well as hopefully clarifying the purpose of each. Additionally, since there is no fixed "read" vs "write" key slot for authentication, this allows selecting the keyslot each key applies to, which should make it possible to use with custom ADF configurations.

### DIFF
--- a/client/src/cmdhfseos.c
+++ b/client/src/cmdhfseos.c
@@ -47,46 +47,55 @@ static uint8_t zeros[16] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
 static int CmdHelp(const char *Cmd);
 
 typedef struct {
-    uint8_t nonce[8];
-    uint8_t privEncKey[16];
-    uint8_t privMacKey[16];
-    uint8_t readKey[16];
-    uint8_t writeKey[16];
-    uint8_t adminKey[16];
-} keyset_t;
+    uint8_t keyslot: 4;
+    uint8_t encKey[16];
+    uint8_t macKey[16];
+} priv_keyset_t;
 
-keyset_t keys[] = {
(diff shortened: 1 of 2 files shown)


iceman1001:
I don't really see any reason for this.   I will close this one.

aaronjamt:
> I don't really see any reason for this. I will close this one.

This was discussed in the #seos channel on your Discord server. The same privacy keys can be used with multiple sets of authentication keys, so splitting them like this makes them easier to manage. Additionally, each credential may only use one set of privacy keys, but have multiple sets of authentication keys in different key slots for different purposes, like reading vs writing. In PR #3032, for example, this is used to provide a separate key for writing, since that operation auths against keyslot 02 (rather than 01, used for reading).

iceman1001:
And I don't agree with that discussion hence this closed.

aaronjamt:
> And I don't agree with that discussion hence this closed.

Understood. I'm fine dropping the majority of this, but I'm not sure how you'd like me to implement selecting the keyslot for writing. The three options I can think of are:
1. Only add `readKeyslot`/`writeKeyslot`/`adminKeyslot` values to the key storage, then have it select that keyslot depending on the operation
2. Add an additional `--keyslot` parameter to `hf seos adf`/`hf seos pacs`/`hf seos write` and pass that as an additional argument through the call stack to where it's used
3. Just hardcode an `if` or `switch` statement to use a fixed keyslot depending on the operation

Do any of these work for you, or do you have any better ideas?

iceman1001:
See the keys as a set with key slots.    Make the keyslot modifiable with parameter if needed otherwise use the assigned key slot to the key set.

PR #3028, opened 2025-11-18
[hf seos] Add keyslot parameter to `seos_mutual_auth`

aaronjamt:
This PR adds a new parameter to the `seos_mutual_auth` function to provide the keyslot to authenticate against, rather than hardcoding keyslot 1. This means that other keyslots can be used if desired.

PR #3032, opened 2025-11-21
[hf seos] Add write support

aaronjamt:
This PR adds a new `hf seos write` command, which allows writing data to a Seos ADF. It builds on top of the changes in PRs #3027, #3028, #3029, #3030, and #3031 so I'm marking it as a Draft PR until those are merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Fix `hf mfdes value --op clear` commands for clearing more than 0x80000000 values and getfilesettings mac mode (@merlokk)
 - Added ATR fingerprinting to `hf 14a/14b info` (@doegox)
 - Added `Verkada 40-bit` format (@aaronmaxlevy)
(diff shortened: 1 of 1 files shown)


iceman1001:
Better to close this one too and rebuild a PR on top of what is getting merged.   The key list split isn't gonna happen

aaronjamt:
> Better to close this one too and rebuild a PR on top of what is getting merged.

Why not merge master or rebase?

> The key list split isn't gonna happen

In that case I'm not sure how to proceed with this PR. The keyslot used for writing is different than the one used for reading. I could hardcode a check for read vs write and change the keyslot inside of `seos_pacs_adf_select`, but then the same auth keys used for reading would be attempted for writing, and it wouldn't be obvious to the user what's happening.

iceman1001:
Yeah,   this PR is problematic and I agree you better make a new one based on latest master.

aaronjamt:
> Yeah,   this PR is problematic and I agree you better make a new one based on latest master. 

Understood, will do.

PR #3036, opened 2025-11-29
Band-aid fix to strip erroneous length byte and CRC from client payload

kormax:
A lot of code in `cmdhffelica.c` was doing one of the following:
1) Calling `AddCrc` to the frame data, while also setting `FELICA_APPEND_CRC` which causes CRC to be appended by the ARM side.
2) Provisioning an empty byte for length, even though ARM side calculates and adds it automatically in `felica_sendraw`. This issue also compounded if extra CRC was sent due to 1).

This PR does the following to fix the issue:
1) Removes wrong invocations (which are all of them) of `AddCrc` in client code, instead using FELICA_ADD_CRC.
2) Add a band-aid inside of `clear_and_send_command` to check if first byte of the payload looks like length, removing it if it is.

The fix was tested with the following commands (with client and ARM built on macOS using clang):
- hf felica litedump
- hf felica info
- hf felica reader
- hf felica dump
- hf felica scsvcode

### DIFF
--- a/client/src/cmdhffelica.c
+++ b/client/src/cmdhffelica.c
@@ -51,8 +51,6 @@
 #define FELICA_SERVICE_ATTRIBUTE_PURSE_SUBFIELD (0b000110)
 
 
-#define AddCrc(data, len) compute_crc(CRC_FELICA, (data), (len), (data)+(len)+1, (data)+(len))
-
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice!

We need to push the crc and len additions to device side.

PR #3043, opened 2025-12-08
[hf seos] Carry when incrementing

aaronjamt:
Currently we only increment the last byte of the input, even if it overflows from 0xFF back to 0x00. I tested this behavior against a real Seos card by setting the RND.IFD to all-0xFFs and the card failed to read. After adding this patch and testing again, the card was successfully read.

### DIFF
--- a/client/src/cmdhfseos.c
+++ b/client/src/cmdhfseos.c
@@ -164,9 +164,10 @@ static int decrypt_cryptogram(uint8_t *key, uint8_t *input, uint8_t *out, int in
 
 static void increment_command_wrapper(uint8_t *input, int input_len) {
     // Increment the end of the header by 1
-    uint32_t value = MemBeToUint4byte(&input[input_len - sizeof(uint32_t)]);
+    uint8_t* offset = &input[input_len - sizeof(uint32_t)];
+    uint32_t value = MemBeToUint4byte(offset);
     value++;
-    Uint4byteToMemBe(&input[input_len - sizeof(uint32_t)], value);
+    Uint4byteToMemBe(offset, value);
 }
 
 static void padToBlockSize(const uint8_t *input, int *inputSize, int blockSize, uint8_t *output) {


iceman1001:
Maybe make it a u16 reference and increase instead of this logic

aaronjamt:
> Maybe make it a u16 reference and increase instead of this logic

This wouldn't fully solve it, as then if the last 2 bytes are 0xFFFF it would still have the same issue it does now. It would need to be at least a u128 to be able to handle AES-128 properly (since the passed array of bytes is a full block size of the selected algoithm).

iceman1001:
highly unlikely scenario for the device to send the full u128 integer space.     Make it a u32 ref and I promise you the command iteration never will go above it.   You will most likely burnt out your proxmark device before.

aaronjamt:
> highly unlikely scenario for the device to send the full u128 integer space.     Make it a u32 ref and I promise you the command iteration never will go above it.   You will most likely burnt out your proxmark device before.

Will do. As it stands, the Proxmark always uses a RND.IFD value of all 0x00s so this issue won't ever appear under normal use, but it did appear while testing with custom provided values.

iceman1001:
We love using bad random numbers.

aaronjamt:
Just tried to make it a `uint32_t*` instead, but (at least on my machine) the endianness is incorrect. For instance, `[0x01, 0x01, 0x01, 0x01]` is incremented to `[0x02, 0x01, 0x01, 0x01]`, rather than `[0x01, 0x01, 0x01, 0x02]`, and I'm not sure how to fix this without writing a bunch of unreadable C pointer magic.

iceman1001:
in `common/commonutil.c`  there is  functions like
```
u32 a = MemBeToUint4byte(d);

Uint4byteToMemBe(d, a);
```

client/src/cmdhfseos.c
```diff
@@ -163,7 +163,10 @@ static int decrypt_cryptogram(uint8_t *key, uint8_t *input, uint8_t *out, int in
 }
 
 static void increment_command_wrapper(uint8_t *input, int input_len) {
-    input[input_len - 1]++; // Increment the last element of the header by 1
+    // Increment the end of the header by 1
```
aaronjamt, line 166: I could also split the `&input[input_len - sizeof(uint32_t)]` out onto a separate line, if you'd like. Maybe something like:
```c
    uint8_t* input_end = &input[input_len - sizeof(uint32_t)];
    uint32_t value = MemBeToUint4byte(input_end);
    value++;
    Uint4byteToMemBe(input_end, value);
```

iceman1001: yeah ,  that will make it even more readable.  if you rename it from `input_end` to `offset`   I merge :)

iceman1001:
Excellent!

PR #3044, opened 2025-12-11
[hf seos] Make keyslot configurable

aaronjamt:
(AKA #3031 attempt 2)

Replaces unused `nonce` value with a new `keyslot` value. By default, key indexes 0 and 1 use keyslot 1, key index 2 uses keyslot 3, and key index 3 uses keyslot 9 respectively. This means that `hf seos adf` and `hf seos pacs` work the same as before, and `hf seos gdf` was updated to default to key index 3 to also maintain compatibility. Key index 2 is set to keyslot 3 to prepare for future write support.

### DIFF
--- a/client/src/cmdhfseos.c
+++ b/client/src/cmdhfseos.c
@@ -47,6 +47,7 @@ static uint8_t zeros[16] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
 static int CmdHelp(const char *Cmd);
 
 typedef struct {
+    uint8_t keyslot;
     uint8_t nonce[8];
     uint8_t privEncKey[16];
     uint8_t privMacKey[16];
@@ -57,6 +58,7 @@ typedef struct {
 
 keyset_t keys[] = {
     {
+        0x01,                                                                                               // Keyslot
         { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 },                                         // Nonce
         { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // privEncKey
         { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // privMacKey
@@ -65,6 +67,7 @@ keyset_t keys[] = {
         { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 } // adminKey
     },
     {
+        0x01,                                                                                               // Keyslot
         { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 },                                         // Nonce
         { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // privEncKey
         { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, // privMacKey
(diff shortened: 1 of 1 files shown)


client/src/cmdhfseos.c
```diff
@@ -1300,15 +1300,15 @@ static int seos_print_keys(bool verbose) {
...
         for (uint8_t i = 0; i < ARRAYLEN(keys); i++) {
             if (memcmp(keys[i].privEncKey, zeros, sizeof(zeros)) == 0)
-                PrintAndLogEx(INFO, " %u |", i);
+                PrintAndLogEx(INFO, " %u |    |", i);
             else
-                PrintAndLogEx(INFO, " %u | " _YELLOW_("%s"), i, sprint_hex(keys[i].nonce, 8));
+                PrintAndLogEx(INFO, " %u | " _YELLOW_("% 2u") " | " _YELLOW_("%s"), i, keys[i].keyslot, sprint_hex(keys[i].readKey, 8));
         }
-        PrintAndLogEx(INFO, "---+------------------------");
+        PrintAndLogEx(INFO, "---+----+------------------------");
```
aaronjamt: I figured also showing an actual key would be useful, rather than just the slot number. Should there be logic here to show a different key based on the selected keyslot, or maybe show one of the privacy keys instead?

client/src/cmdhfseos.c
```diff
@@ -1335,12 +1335,12 @@ static int seos_load_keys(char *filename) {
 
     size_t i = 0;
     for (; i < bytes_read / kn; i++) {
-        memcpy(keys[i].nonce, dump + (i * kn), 8);
```
aaronjamt, line 1342: Could also be done like:
`keys[i].nonce = *(dump + (i * kn));`
I did it this way to keep it more similar to the rest of the block, but let me know if you'd rather change it.

client/src/cmdhfseos.c
```diff
@@ -1335,12 +1335,12 @@ static int seos_load_keys(char *filename) {
...
-        memcpy(keys[i].privMacKey, dump + ((i * kn) + 24), 16);
-        memcpy(keys[i].readKey, dump + ((i * kn) + 40), 16);
-        memcpy(keys[i].writeKey, dump + ((i * kn) + 56), 16);
-        memcpy(keys[i].adminKey, dump + ((i * kn) + 72), 16);
+        memcpy(&keys[i].keyslot, dump + (i * kn), 1);
+        memcpy(keys[i].privEncKey, dump + ((i * kn) + 1), 16);
+        memcpy(keys[i].privMacKey, dump + ((i * kn) + 17), 16);
+        memcpy(keys[i].readKey, dump + ((i * kn) + 33), 16);
+        memcpy(keys[i].writeKey, dump + ((i * kn) + 49), 16);
+        memcpy(keys[i].adminKey, dump + ((i * kn) + 65), 16);
```
aaronjamt: Not as relevant to this PR, but maybe it would make sense to have the offsets instead stored as a variable and incremented by each memcpy since these odd numbers look weird?

Alternatively, does it make more sense to move the keyslot to the end of the struct so all the rest of the values are nice even multiples of 8?

client/src/cmdhfseos.c
```diff
@@ -1513,15 +1513,16 @@ static int CmdHfSeosManageKeys(const char *Cmd) {
                   "Manage SEOS Keys in client memory, keys are required to authenticate with SEOS cards\n",
                   "hf seos managekeys -p\n"
                   "hf seos managekeys -p -v\n"
-                  "hf seos managekeys --ki 0 --nonce 0102030405060708  -> Set nonce value at key index 0\n"
-                  "hf seos managekeys --load -f mykeys.bin -p          -> load from file and prints keys\n"
-                  "hf seos managekeys --save -f mykeys.bin             -> saves keys to file\n"
+                  "hf seos managekeys --ki 0 --keyslot 1              -> Set keyslot value at key index 0\n"
+                  "hf seos managekeys --ki 1 --read 0102030405060708  -> Set read key value at key index 1\n"
```
aaronjamt: I figured I'd switch the line over to the read key, since it's more similar to the original message (showing how you can input a hex string) and it applies to more values. I added an additional line for the keyslot, but I'm not sure if this is actually necessary.

client/src/cmdhfseos.c
```diff
@@ -1569,8 +1568,8 @@ static int CmdHfSeosManageKeys(const char *Cmd) {
     if (key_index >= 0) {
         operation += 3;
         if (key_index < 4) {
-            if (nonce_len != 0) {
-                PrintAndLogEx(SUCCESS, "Current value for nonce[%d] " _GREEN_("%s"), key_index, sprint_hex_inrow(keys[key_index].nonce, 8));
+            if (keyslot != -1) {
+                PrintAndLogEx(SUCCESS, "Current value for keyslot[%d] " _GREEN_("%u"), key_index, keys[key_index].keyslot);
```
aaronjamt: Everywhere the keyslot is used internally, it's fetched with a `%02X` format string, but I think it's better displayed to the user in decimal. Is this actually true or should I swap it back to hex everywhere?

client/src/cmdhfseos.c
iceman1001: We will try not to show any keys for SEOS output.   This lessen the possibility for users to mistakenly post keys from screen dumps.

client/src/cmdhfseos.c
iceman1001: or make a struct that covers the expected items.

client/src/cmdhfseos.c
iceman1001: help texts ain't wrong to have samples.   How many sample of usage is a good question.  As long as user understand how to operate the command

client/src/cmdhfseos.c
iceman1001: the hex is used since that is what you will see when sniffing traffic from a seos card & reader communication.

If in decimal you will need to understand it's not hex, vs when it's in hex.  

Hence we use hex for all output and inputs besides from  index values,  like "keyslot index"

iceman1001:
We had nonce in there because of what we observed and found it useful.    Removing it lessen the understandings.

aaronjamt:
I reverted the `nonce` removal and rebased, now this PR just adds the `keyslot` value alongside it.

PR #3047, opened 2025-12-18
Add qrcode sources to CMakeLists in the experimental lib

Sanduuz:
The experimental_lib CMakeLists.txt file is missing the source file for the qrcode functionality, which includes symbols for e.g. `qrcode_print_matrix_utf8`. The lib builds without errors, but when trying to use it for example in python an error is presented:
`undefined symbol: qrcode_print_matrix_utf8`. This is fixed by adding the qrcode sources to CMakeLists.txt.


iceman1001:
Good catch!

PR #3048, opened 2025-12-18
Fix iCLASS block 7 decoder to strip sentinel bit

tweathers-sec:
When decoding iCLASS block 7 Wiegand data, the decoder was showing incorrect bit counts (27 for 26-bit, 36 for 35-bit, etc.) because it wasn't stripping the sentinel bit that's added during encoding.

This fix:
- Detects when binary string has sentinel bit (leading '1' making length one more than common formats)
- Strips the sentinel bit and recreates u96 values from corrected binary string
- Uses corrected length for wiegand decoding

Fixes decoding for 26-bit, 35-bit, and other formats where sentinel bit causes incorrect format detection.

Pre and post edit output:

sentinel_bit_fix (1)

### DIFF
--- a/client/src/cmdhficlass.c
+++ b/client/src/cmdhficlass.c
@@ -1511,11 +1511,36 @@ static void iclass_decode_credentials(uint8_t *data) {
             char binstr[64 + 1];
             hextobinstring(binstr, hexstr);
             char *pbin = binstr;
+            // Strip leading zeros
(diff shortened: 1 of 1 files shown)


iceman1001:
Cool cool.  
If you have a look in the Cmdwiegand.c  I believe you can add same fix.

PR #3051, opened 2025-12-20
[hf seos] Add write support

aaronjamt:
This should be a better version of #3032. Most of the code in here is just duplicated and modified from existing code (`seos_write_data` is based on `seos_get_data`, `seos_write` is nearly identical to `seos_pacs`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Fix `hf mfdes value --op clear` commands for clearing more than 0x80000000 values and getfilesettings mac mode (@merlokk)
 - Added ATR fingerprinting to `hf 14a/14b info` (@doegox)
 - Added `Verkada 40-bit` format (@aaronmaxlevy)
+- Added `hf seos write` command (@aaronjamt)
 
 ## [Phrack.4.20728][2025-09-11]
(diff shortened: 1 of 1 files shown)


client/src/cmdhfseos.c
```diff
@@ -1456,6 +1557,69 @@ static int CmdHfSeosPACS(const char *Cmd) {
...
+    if (oid_len == 0) {
+        oid_len = 17;
+    }
+
+    // convert OID hex to literal string
+
+    char oid_buffer[256] = "";
+    for (int i = 0; i < oid_len; i++) {
+        sprintf(oid_buffer + (i * 2), "%02X", oid_hex[i]);
+    }
```
iceman1001: there should be some hex to string functions to call instead.    param_hex_eol etc

aaronjamt: Should that also be changed in `CmdHfSeosPACS` and `CmdHfSeosADF` then for consistency?

iceman1001: yes please

aaronjamt: Will do. Should those be in a separate PR or just do that all in here?

iceman1001: separate PR please,  since its more easy to follow changes that way. 

By now, I notice you are getting into the preferred code style.   It might not be the best code style,  but slowly we will get nice code in this old project.

PR #3055, opened 2025-12-21
Fix CMAC padding for ARM

aaronjamt:
As requested [here](https://github.com/RfidResearchGroup/proxmark3/pull/3054#discussion_r2637710559)


iceman1001:
just a heads up,  both padding was using 16 bytes but you mentioned 8 byte for the 3des version.   
Double check so there is no mistake in lengths.

aaronjamt:
I know the 3DES implementation was working with the same padding method so even if it should be changed, it's not doing any harm currently. I'll take a look.

aaronjamt:
Actually, looks like the length is set from the calling function, and that's already been changed. No worries there!

PR #3058, opened 2025-12-21
[hf seos] Simulate support (1/2)

aaronjamt:
This is the first part of the Seos emulation/simulation support, specifically adding the necessary code to the actual ARM MCU for simulating. I've tested this against an HID reader and a Flipper Zero with the Seos Compatible app.

### DIFF
--- a/armsrc/seos.c
+++ b/armsrc/seos.c
@@ -264,12 +264,8 @@ void SimulateSeos(seos_emulate_req_t *msg) {
     memcpy(data, msg->uid, msg->uid_len);
     FLAG_SET_UID_IN_DATA(flags, msg->uid_len);
 
-    // TODO: Is this required?
-    uint8_t ats[] = { 0x05, 0x78, 0x77, 0x80, 0x02 };
-    flags |= FLAG_ATS_IN_DATA;
-
     // 12 = HID Seos 4K card
-    if (SimulateIso14443aInit(12, flags, data, ats, sizeof(ats), &responses, &cuid, NULL, NULL) == false) {
+    if (SimulateIso14443aInit(12, flags, data, NULL, 0, &responses, &cuid, NULL, NULL) == false) {
         BigBuf_free_keep_EM();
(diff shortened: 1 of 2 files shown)


armsrc/seos.c
```diff
@@ -0,0 +1,837 @@
...
+                                        break;
+                                    }
+                                }
+                                tlv_offset += length;
+                            }
+
+                            if (selected_oid) {
+                                // Synthesized IV: half a block of random data followed by half of the CMAC of that data
+                                uint8_t synthesized_iv[max_bs];
+                                memset(synthesized_iv, 0, bs/2); // TODO: Maybe actually use random data?
```
iceman1001: avoid division on arm side.   use  `bs >> 1`  to make a div by 2.

iceman1001:
I believe it's two folded.   
1.  flash size,  meh,  doesn't quite as long as we aint in 90% range but I think division adds 3k just like that since it introduces floats. 
2. RAM usage / stack issues -  the use of arrays in the PR is extensive.  I think it has to be rewritten to use two bigbuf allocated spaces and when using clear it before.

aaronjamt:
>     1. flash size,  meh,  doesn't quite as long as we aint in 90% range but I think division adds 3k just like that since it introduces floats.

I did take out the division but that didn't change flash size. If you say it's fine then I won't worry about it.

>     2. RAM usage / stack issues -  the use of arrays in the PR is extensive.  I think it has to be rewritten to use two bigbuf allocated spaces and when using clear it before.

Done, using 2 "work" buffers allocated once at the beginning. There's still a few arrays created here and there but the largest ones are avoided now.

iceman1001:
Awesome!
the division most likely still is in the arm code in other places.   We still good with memory.

Good thing with the big buff,  try work it in even more.   I see there are many arrays still. 

I will merge this but once you done,   please put the extra effort and look at arrays.   They steal a lot of ram and we have stack issues as is.

PR #3059, opened 2025-12-22
Update docs

aaronjamt:
Noticed the documentation hasn't been updated in a little while.


iceman1001:
it's one of those things that gets updated when we run a `make style` PR

PR #3060, opened 2025-12-22
[hf seos] Simulate support (2/2)

aaronjamt:
This is the second part of the Seos emulation/simulation support, specifically adding the client-side command to start simulation on the Proxmark. #3058 must be merged first as it adds `seos_cmd.h`.

### DIFF
--- a/client/src/cmdhfseos.c
+++ b/client/src/cmdhfseos.c
@@ -1698,8 +1698,9 @@ static int CmdHfSeosSim(const char *Cmd) {
         arg_str0("t", "tag", "<hex>", "<0-100> hex bytes for tag to simulate (Default: FF00)"),
         arg_str0("o", "oid", "<hex>", "<0-100> hex bytes for OID (Default: 2B0601040181E438010102011801010202)"),
         arg_int0(NULL, "ki", "<dec>", "Specify key index to set key in memory"),
-        arg_str0("u", "diversifier", "<hex>", "<0-16> hex bytes for diversifier (Equivalent of UID)"),
+        arg_str0(NULL, "div", "<hex>", "<0-16> hex bytes for diversifier (Equivalent of UID)"),
         arg_str0("d", "data", "<hex>", "<0-128> hex bytes for data (Must be valid BER-TLV)"),
+        arg_str0("u", "uid", "<hex>", "<0-10> hex bytes for UID (Must be a RID i.e. [0]=0x08)"),
         arg_lit0("l", "legacy", "Use legacy algorithms (3DES/SHA1)"),
         arg_param_end
     };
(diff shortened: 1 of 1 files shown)


client/src/cmdhfseos.c
```diff
@@ -1676,6 +1676,118 @@ static int CmdHfSeosADF(const char *Cmd) {
...
+        PrintAndLogEx(ERR, "Data to simulate must be supplied");
+        return PM3_ESOFT;
+    }
+
+    srand(time(NULL));
+    seos_emulate_req_t request = {
+        .encr_alg = encryption_algorithm,
+        .hash_alg = hashing_algorithm,
+
+        .uid = {0x08, rand()%255, rand()%255, rand()%255},
```
aaronjamt: This could also be hard-coded to something like {0x08, 0x01, 0x02, 0x03}

iceman1001: give the user the option to supply a UID as parameter. 
and default to a simple boring UID like 8,1,2,3,      when doing research having a fixed simple UID helps sometimes.  You never know when

client/src/cmdhfseos.c
```diff
@@ -1676,6 +1676,118 @@ static int CmdHfSeosADF(const char *Cmd) {
...
+                  "hf seos sim -o 2B0601040181E438010102011801010202 -u 01020304050607 --ki 2 -d 12345678\n"
+                  "hf seos sim -o 2B0601040181E438010102011801010202 --legacy -t FF41 -d 12345678\n"
+                 );
+
+    void *argtable[] = {
+        arg_param_begin,
+        arg_str0("t", "tag", "<hex>", "<0-100> hex bytes for tag to simulate (Default: FF00)"),
+        arg_str0("o", "oid", "<hex>", "<0-100> hex bytes for OID (Default: 2B0601040181E438010102011801010202)"),
+        arg_int0(NULL, "ki", "<dec>", "Specify key index to set key in memory"),
+        arg_str0("u", "diversifier", "<hex>", "<0-16> hex bytes for diversifier (Equivalent of UID)"),
```
iceman1001: I figure we have a parameter clash.  
-u  == UID
I suggest --div  and no short param,   ie:   NULL, "div",  "<hex>"

iceman1001:
as a suggestion for future improvements,    let this command accept a "-f" for a seos dump file,  which is then uploaded to the device side.   

I know it most likely will need to rework current sim implementation,  but....

aaronjamt:
> as a suggestion for future improvements, let this command accept a "-f" for a seos dump file, which is then uploaded to the device side.
> 
> I know it most likely will need to rework current sim implementation, but....

I don't think that would be required, if you just parse the dump file on the client side and format the relevant data into the `seos_emulate_req_t` to send over.

iceman1001:
I think your simulation is limited.  It assumes PACS adf only.   A SEOS card can do much more and has much more.

aaronjamt:
> I think your simulation is limited. It assumes PACS adf only. A SEOS card can do much more and has much more.

Sure, but the current client-side code and the format of the dump file dont't support any more than PACS ADFs anyways. I'd be happy to add more to the simulation but I'd need to see at least a working client-side implementation to get started on that.

PR #3061, opened 2025-12-22
[hf seos] Validate synthesized IV

aaronjamt:
The "synthesized IV" is made of half a block of random data, plus half a block of the CMAC of that data. This PR adds a check to verify that the CMAC matches. It also adds some missing return value checks so that we don't continue trying to read the card if we encounter an invalid CMAC.

### DIFF
--- a/client/src/cmdhfseos.c
+++ b/client/src/cmdhfseos.c
@@ -1195,11 +1195,17 @@ static int seos_pacs_adf_select(char *oid, int oid_len, uint8_t *data_tag, int d
 
     uint8_t keyslot = keys[key_index].keyslot;
     res = seos_challenge_get(RNDICC, sizeof(RNDICC), keyslot);
-    if (res != PM3_SUCCESS) return res;
+    if (res != PM3_SUCCESS) {
+        return res;
+    }
     res = select_df_decode(response, resplen, &ALGORITHM_INFO_value1, &ALGORITHM_INFO_value2, CRYPTOGRAM_encrypted_data, MAC_value);
-    if (res != PM3_SUCCESS) return res;
+    if (res != PM3_SUCCESS) {
+        return res;
+    }
     res = select_DF_verify(response, resplen, MAC_value, sizeof(MAC_value), ALGORITHM_INFO_value1, key_index);
-    if (res != PM3_SUCCESS) return res;
+    if (res != PM3_SUCCESS) {
+        return res;
+    }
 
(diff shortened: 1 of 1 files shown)


client/src/cmdhfseos.c
```diff
@@ -1186,13 +1199,12 @@ static int seos_pacs_adf_select(char *oid, int oid_len, uint8_t *data_tag, int d
     resplen -= 2;
 
     uint8_t keyslot = keys[key_index].keyslot;
-    seos_challenge_get(RNDICC, sizeof(RNDICC), keyslot);
-    select_df_decode(response, resplen, &ALGORITHM_INFO_value1, &ALGORITHM_INFO_value2, CRYPTOGRAM_encrypted_data, MAC_value);
+    res = seos_challenge_get(RNDICC, sizeof(RNDICC), keyslot);
+    if (res != PM3_SUCCESS) return res;
```
iceman1001: I really don't like the one-liners any longer.
I understand why people love them,  and so did I before. 
now I prefer seeing it multilined,   since I find it much easier to identify the "return".

client/src/cmdhfseos.c
```diff
@@ -858,6 +858,19 @@ static int select_ADF_decrypt(const char *selectADFOID, uint8_t *CRYPTOGRAM_encr
...
+            uint8_t iv_size = i/2;
+            create_cmac(keys[key_index].privMacKey, CRYPTOGRAM_encrypted_data_raw, mac, iv_size, sizeof(mac), encryption_algorithm);
+            mac[0]++;
+            if (memcmp(CRYPTOGRAM_encrypted_data_raw+iv_size, mac, iv_size) != 0) {
+                PrintAndLogEx(ERR, "MAC Verification Failed");
+                
+                // PrintAndLogEx(ERR, "Synthesized IV................... "_YELLOW_("%s"), sprint_hex_inrow(CRYPTOGRAM_encrypted_data_raw,iv_size));
+                // PrintAndLogEx(ERR, "Synthesized IV MAC............... "_YELLOW_("%s"), sprint_hex_inrow(CRYPTOGRAM_encrypted_data_raw+iv_size,iv_size));
+                // PrintAndLogEx(ERR, "Synthesized IV Computed MAC...... "_YELLOW_("%s"), sprint_hex_inrow(mac,iv_size));
+
```
iceman1001: either you print the messages,  or you remove them.   
I do understand having debug messages in the code,  but lets make things clean.

PR #3062, opened 2025-12-23
[hf seos] Reduce array usages

aaronjamt:
Removed almost every array usage in the Seos simulation code, using BigBuf instead. Hopefully this should be a good "final" place to leave the Seos simulation code until people find bugs.

### DIFF
--- a/armsrc/seos.c
+++ b/armsrc/seos.c
@@ -63,97 +63,66 @@ static uint8_t round_to_next(uint8_t value, uint8_t step) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


armsrc/seos.c
```diff
@@ -208,7 +167,6 @@ void SimulateSeos(seos_emulate_req_t *msg) {
     uint8_t diver_cmac_key[16];
```
aaronjamt, line 167: These should be the only arrays left. Do you want these also switched to BigBuf or is that excessive?

iceman1001: those are small enough,  let them be as is

PR #3065, opened 2025-12-23
[hf seos] Minor cleanup

aaronjamt:
There's no need to create a 128-byte array when the largest amount of data we actually need to store are two 16-byte keys.

### DIFF
--- a/client/src/cmdhfseos.c
+++ b/client/src/cmdhfseos.c
@@ -512,7 +512,6 @@ static void create_mutual_auth_key(uint8_t *KEYIFD, uint8_t *KEYICC, uint8_t *RN
     // PrintAndLogEx(SUCCESS, "hash Input....................... " _YELLOW_("%s"), sprint_hex_inrow(hash_in,ARRAYLEN(hash_in)));
 
     uint8_t output[32]; // Buffer to store the two 16-byte keys
-    uint8_t hashedOutput[32];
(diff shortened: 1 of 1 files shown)


client/src/cmdhfseos.c
```diff
@@ -526,7 +526,7 @@ static void create_mutual_auth_key(uint8_t *KEYIFD, uint8_t *KEYICC, uint8_t *RN
         counter++;
         set_counter_big_endian(hash_in, counter);
         sha1hash(hash_in, sizeof(hash_in), hashedOutput);
-        memcpy(output + 20, hashedOutput, 20);
+        memcpy(output + 20, hashedOutput, 12);
```
iceman1001, line 528: 12 ?    you sure you didn't mean 16?

aaronjamt: Yes, only 32 bytes are used total. SHA1 outputs 20 bytes each time, 40 total, but it just takes the first 32 and ignores the rest.

iceman1001: good, 
so `hashedOutput[32]`  could be [20]  ?

aaronjamt: For SHA1, yes, but it's also used for SHA256 which directly outputs 32 bytes. Hmm actually SHA256 could also be refactored to copy directly to output, in which case yes. Good call!

PR #3066, opened 2025-12-24
[hf seos] Another minor bugfix

aaronjamt:
Fix another "may be used uninitialized" error reported on the Discord server.


iceman1001:
for fixed array,   it's better to use: 
```
uint8_t mac[8] = {0};

instead of
memset(mac, 0, sizeof(mac));
```

We use memset for dynamic allocated memory from either bigbuf_calloc or  variable sized  (mac[ my_len )

PR #3077, opened 2026-01-05
Add simple card/tag cloner script

daddyparodz:
This PR adds a single file, pm3-clone.sh, a Bash command wrapper for Proxmark3 that simplifies cloning Ultralight/NTAG and MIFARE Classic tags.


iceman1001:
Interesting choice to shell script it.  I see more python scripts.   Anyway,  it should reside under `tools`  folder.   We want to keep our root folder clean

PR #3082, opened 2026-02-02
Add Kaba evolo AIDs

xtruan:
Adding Kaba evolo AIDs. Note that `F52100` "Kaba evolo - Access" has the same AID as `F52100` "Lucid Motors Valet Card". Let me know if any changes need to be made to deal with this.


iceman1001:
excellent!

Issue #2766, opened 2025-03-01
Broken fm11rf08_recovery.py


Steve707-AE:
**Describe the bug**
After commit [f5650a5](https://github.com/RfidResearchGroup/proxmark3/commit/f5650a53afec108f8c3f3dd3a30128ba36faa46a), when the command `script run fm11rf08s.py` is executed it gets error because the single backtick is not printed and fails on doing the split to find the path.

**To Reproduce**
1. Have the last commit
2. Run `script run fm11rf08s.py` 

**Expected behavior**
Execution of the script.

**Screenshots**

![Image](https://github.com/user-attachments/assets/6365d601-6c78-41db-88c3-82cb620d3ae8)

![Image](https://github.com/user-attachments/assets/3ebe0d9c-78a5-47f3-8339-ff5a2ddcab4f)

**Desktop (please complete the following information):**
 - OS: Windows 11

iceman1001:
pull latest and see if it still exist

Issue #2689, opened 2024-12-16
Some scripts use hardcoded tools path


CoelacanthusHex:
**Describe the bug**

Some scripts use hardcoded tools path, if packager changes INSTALLTOOLSRELPATH when installing the proxmark3 client (Arch Linux packager did it, see https://gitlab.archlinux.org/archlinux/packaging/packages/proxmark3/-/blob/main/PKGBUILD?ref_type=heads ), it will try to find it in the default path, and fail to find it.

https://github.com/RfidResearchGroup/proxmark3/blob/59ae5d22850542892bdde4af0b844543268604e0/client/pyscripts/fm11rf08s_recovery.py#L60-L61

**To Reproduce**

1. Install Proxmark3 client with `INSTALLTOOLSRELPATH=bin` and `PREFIX=/usr` (so tools will in /usr/bin)
2. `script run fm11rf08_full`
3. See error

```
[usb] pm3 --> script run fm11rf08_full
[+] executing python /usr/bin/../share/proxmark3/pyscripts/fm11rf08_full.py
[+] args ''
Cannot find /usr/share/proxmark3/tools/staticnested_1nt, abort!
[!] ⚠  Script terminated by SystemExit -1
[!] ⚠  finished fm11rf08_full with exception
```

**Expected behavior**

The script runs normally.

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - inside proxmark3 client run the following commands and paste the output here.
 - hw version
<details>
<summary>Details</summary>
```
[usb] pm3 --> hw version

 [ Proxmark3 RFID instrument ]

 [ Client ]
  Iceman/master/v4.19552-suspect 2024-11-22 12:15:19 a3319015f
  compiled with............. GCC 14.2.1 20240910
  platform.................. Linux / x86_64
  Readline support.......... present
  QT GUI support............ present
  native BT support......... present
  Python script support..... present ( 3.12.7 )
  Python SWIG support....... present
  Lua script support........ present ( 5.4.7 )
  Lua SWIG support.......... present

 [ Proxmark3 ]
  device.................... device / fw mismatch
  firmware.................. RDV4
  external flash............ present
  smartcard reader.......... present
  FPC USART for BT add-on... absent

 [ ARM ]
  bootrom: Iceman/master/v4.19552-suspect 2024-11-22 12:15:19 a3319015f
       os: Iceman/master/v4.19552-suspect 2024-11-22 12:15:19 a3319015f
[...]

Issue #2565, opened 2024-10-07
[idea] Extra help text when autopwn fails on "Static encrypted nonce"


Lerold:
**Is your feature request related to a problem? Please describe.**
When running `hf mf autopwn` and it fails on
_[!!] 🚨 Error: Static encrypted nonce detected. Aborted_

**Describe the solution you'd like**
Propose to run `script run fm11rf08s_recovery` to get the keys.
Or even better would be to include it as one of the things autopwn runs when it detects it.

iceman1001:
fixed

Issue #2553, opened 2024-10-02
hf mf isen --collect_fm11rf08s - AcquireStaticEncryptedNonces: Auth1 error


robshee81:
This is what I get:

[usb] pm3 --> hf mf isen --collect_fm11rf08s
[#] AcquireStaticEncryptedNonces: Auth1 error
[
  [
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"]
  ],
  [
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"],
    ["00000000", "00000000"]
  ],
  [
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"],
    ["0000", "0000"]
  ]
]

So it stops me from being able to use the fm11rf08s_recovery.py script, as it fails at this point or run on its own and I don't know why or how to fix it.

hf mf info

[+] Sector 0 key A... FFFFFFFFFFFF
[+] Sector 0 key B... FFFFFFFFFFFF
[+] Sector 1 key A... FFFFFFFFFFFF
[#] failed reading block
[...]

PR #3090, opened 2026-02-07
doc: add FM11RF08S manual key recovery guide

joetomasone:
## Summary

Step-by-step guide for recovering all sector keys from Fudan FM11RF08S MIFARE Classic 1K cards when the automated `fm11rf08s_recovery.py` script fails.

## Problem

The FM11RF08S has a hybrid nonce behavior (normal first-auth, static encrypted nested-auth) that causes all three standard attacks to fail with contradictory error messages:
- `hf mf nested` → "PRNG is not predictable"
- `hf mf staticnested` → "Normal nonce detected"  
- `hf mf hardnested` → "Static encrypted nonce detected. Aborted"

The official `fm11rf08s_recovery.py` script handles this correctly in theory, but has documented breakage across platforms:
- Missing `_pm3` SWIG module (#2553, common on Homebrew/macOS)
- Hardcoded tool paths (#2689)
- Backdoor auth failures (#2553)
- Unicode crashes (#2838)
- Path parsing breakage (#2766)

There's also an open request (#2565) to have `autopwn` detect FM11RF08S and suggest the recovery script.

## Solution

This guide documents a **manual pipeline** that bypasses all script dependencies:

1. `hf mf info` — identify FM11RF08S
2. `hf mf autopwn` — recover known keys via dictionary
3. `hf mf isen --collect_fm11rf08s_without_backdoor` — collect static nonces
4. `staticnested_1nt` / `staticnested_2x1nt_rf08s` — offline candidate generation
5. `hf mf fchk` — brute force candidates against card
6. `hf mf autopwn -f` — full dump with all recovered keys
7. `hf mf mad` — analyze dump contents

Includes troubleshooting section, timing expectations (~15 min for 3 unknown keys), and a table of related known issues.

## Testing

Tested on RDV4 with Iceman firmware v4.20728 on macOS aarch64 (Apple Silicon). Successfully recovered all 32 keys from a hotel card (Vingcard + Timelox + ENKOA) where `autopwn` left 3 keys unsolved and the recovery script crashed.

### DIFF
--- a/doc/fm11rf08s_manual_recovery.md
+++ b/doc/fm11rf08s_manual_recovery.md
@@ -1,0 +1,240 @@
+# Cracking FM11RF08S Hotel Cards with Proxmark3
+
+A step-by-step guide for recovering all sector keys from Fudan FM11RF08S MIFARE Classic 1K cards when standard attacks fail.
+
+## Background
+
+The FM11RF08S is a Chinese clone of the NXP MIFARE Classic 1K. Hotels use them because they're cheap. The chip has a well-documented backdoor (see [Doegox's paper](https://eprint.iacr.org/2024/1275)) and a quirk that breaks every standard Proxmark3 attack: its first-auth nonce behaves like a weak PRNG, but its nested-auth nonce is **static and encrypted**. This causes:
+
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice!

PR #3095, opened 2026-02-14
hf felica dump: vastly improve speed; fix traces being empty

kormax:
* Vastly improve execution speed for `hf felica dump` and `hf felica scsvcode`commands by preventing the cyclic tear-down and activation of the RF field on each command-response cycle.
* Fix `hf felica dump` not saving logs by cleaning the buffer and creating new "connection" only on start of the command.
* Add length validation to `hf felica raw`.

I've tested the following `hf felica` commands after the change:
`reader`, `dump`, `litedump`, `rqservice`, `rqresponse`, `scsvcode`, `rqsyscode`, `rqspecver`, `resetmode`, `raw`.

### DIFF
--- a/armsrc/felica.c
+++ b/armsrc/felica.c
@@ -47,6 +47,7 @@
 static uint32_t felica_timeout;
 static uint32_t felica_nexttransfertime;
 static uint32_t felica_lasttime_prox2air_start;
+static bool felica_field_active;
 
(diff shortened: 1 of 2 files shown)


iceman1001:
the felica commands should be converted to use NG packages instead too.

Nice PR!

PR #3096, opened 2026-02-15
felica: use NG frames for commands

kormax:
* Convert client code felica to NG command format;
* Handle ng frames in felica_sendraw;
* Fix length validation issue in `hf felica raw`.

I've performed a sanity check by running the following `hf felica` commands after the change: 
- reader
- info
- dump
- scsvcode
- litedump
- sniff
- raw

I wasn't sure what to do with sub-command parsing in `appmain.c`, so I've decided to leave it as in there.

As for other improvements: I'm thinking about getting rid of `hf felica litedump` command on the arm side, instead merging that logic right into `hf felica dump` as a special case for systems 0x12fc and 0x88B4.
That change, though, may require some prerequisite ones (add support for specifying request code and system code when CONNECTing, allow commands to work without invoking prior `reader`).

### DIFF
--- a/armsrc/felica.c
+++ b/armsrc/felica.c
@@ -62,14 +62,19 @@ static uint32_t iso18092_get_timeout(void) {
     return felica_timeout - (DELAY_AIR2ARM_AS_READER + DELAY_ARM2AIR_AS_READER) / (16 * 8) - 2;
 }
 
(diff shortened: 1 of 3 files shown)


iceman1001:
yeah,   one dump command is what we want.   
And awesome with NG conversion!

PR #3097, opened 2026-02-17
Add BASHAUTOTRANS Russian Transport card keys

lux0id:
Add Ufanet key


iceman1001:
```

./client/dictionaries/mfc_default_keys.dic
      2 F4086EA40D79

```

remove the duplicate key from your PR,  and is these keys static?   ie,  same across different cards from same system?

PR #3099, opened 2026-02-19
New "hw decay" cmd monitors hf tuning cap voltage decay (pm3 easy booster board detection)

dangerous-tac0s:
If you have the Midwest Gadgets Booster board installed in a proxmark3 easy it says the HF antenna is unusable. This is due to it coupling with an lc tank circuit. There was too much variation in the booster boards for simple voltage detection to work reliably so we stumbled into this. It became quite easy to spot three things: regular pm3 easy, pm3 with booster board, and pm3 on surface that interferes.

```
CLIParserInit(&ctx, "hw decay",
                  "Measure HF antenna decay after field-off.\n"
                  "Captures how quickly the peak-detect capacitor voltage drops\n"
                  "after the 13.56 MHz field is turned off. Different antenna loading\n"
                  "(unloaded, booster board, damaged) produces different decay profiles.",
                  "hw decay\n"
                  "hw decay --ms 100  --> stabilize for 100ms before measurement\n"
                  "hw decay --us 5000 --> measure 5ms decay window\n");

    void *argtable[] = {
        arg_param_begin,
        arg_int0(NULL, "ms", "<dec>", "Field stabilization time in ms (default: 50)"),
        arg_int0(NULL, "us", "<dec>", "Measurement window in us (default: 2000)"),
        arg_param_end};

```

And "hw tune" was updated

```
    // If HF is unusable or marginal, run a quick decay measurement to check
    // for booster board. With a booster, the first fast-ADC decay sample reads
    // 50-500 (rapid discharge). Without a booster, it reads >1000.
    bool hf_booster_detected = false;
    if (!IfPm3Rdv4Fw() && package->v_hf < HF_MARGINAL_V)
```

We check the first measure during our 5us interval read from "hw decay"

```
            if (decay_num > 0 && decay_samples[0] >= 50 && decay_samples[0] <= 500)
            {
                hf_booster_detected = true;
            }
```
```
 if (hf_booster_detected)
    {
        PrintAndLogEx(SUCCESS, "");
        PrintAndLogEx(SUCCESS, "Your HF antenna measurement shows");
        PrintAndLogEx(SUCCESS, "low voltage that is consistent");
        PrintAndLogEx(SUCCESS, "with the installation of a booster");
        PrintAndLogEx(SUCCESS, "board. If you do not have a");
        PrintAndLogEx(SUCCESS, "booster board installed, either");
        PrintAndLogEx(SUCCESS, "your antenna is malfunctioning or");
        PrintAndLogEx(SUCCESS, "you have a tag on the HF antenna.");
    }
    else
    {
        PrintAndLogEx(SUCCESS, "");
        PrintAndLogEx(SUCCESS, "Approx. Q factor measurement");
```

A similar pattern is followed to detect interference--it's just in the middle of the road decay value wise.

### DIFF
--- a/client/src/cmdhw.c
+++ b/client/src/cmdhw.c
@@ -29,7 +29,7 @@
 #include <Python.h>
 #endif
 
-#include "cmdparser.h"      // command_t
+#include "cmdparser.h" // command_t
 #include "cliparser.h"
 #include "comms.h"
 #include "usart_defs.h"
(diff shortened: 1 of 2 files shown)


iceman1001:
Nice,

 I have no way to test this one,   need to merge it blind. 
 Does this still also show potential Q factor despite there is a booster board?

dangerous-tac0s:
@iceman1001 It does not show potential Q factor when we think there is a booster board. But of course we could if you prefer. It also sounds like we should ship you a booster board--does Amal have an address for you?

iceman1001:
Thanks for the offer,   Amal does have an address.   Not particular need though,   but its very kind of you.

The Q factor would be nice to have also.    you make a new PR with it?

PR #3102, opened 2026-02-26
[14a sim] Stop sim with Enter key

aaronjamt:
This PR allows the user to stop an in-progress `hf 14a sim` by pressing the Enter key. This can be useful in situations where the Proxmark's button is inconvenient to access, such as when connected wirelessly, or for automated applications.

### DIFF
--- a/client/src/cmdhf14a.c
+++ b/client/src/cmdhf14a.c
@@ -1040,9 +1040,10 @@ int CmdHF14ASim(const char *Cmd) {
     sector_t *k_sector = NULL;
     size_t k_sectors_cnt = MIFARE_4K_MAXSECTOR;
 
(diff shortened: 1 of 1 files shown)


iceman1001:
does  the client side handling the new broader scope for sending a noticifcation breakloop command?

aaronjamt:
I'm not entirely sure what you mean. This PR moves the enter keypress check to the top of the loop, so it isn't skipped when `WaitForResponseTimeout` returns `false`, and then ensures that `CMD_BREAK_LOOP` will always be sent on enter keypress, stopping the simulation on the Proxmark itself.

iceman1001:
my bad,  I confusing it for being on device side

PR #3103, opened 2026-02-26
[14a sniff] Stop sniff with Enter key

aaronjamt:
This PR allows the user to stop an in-progress `hf 14a sniff -i` by pressing the Enter key. This can be useful in situations where the Proxmark's button is inconvenient to access, such as when sandwiched between a tag and a reader.

The value of `checker` is based on the timing [here](https://github.com/RfidResearchGroup/proxmark3/blob/478371222510de5bf02bb71f46e9f6281158c98b/armsrc/iso14443a.c#L1110). I've verified that it still sniffs correctly, but I'm happy to increase this further if necessary.

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -852,10 +852,19 @@ void RAMFUNC SniffIso14443a(uint8_t param) {
 
     uint32_t rx_samples = 0;
 
+    uint16_t checker = 12000;
+
     // loop and listen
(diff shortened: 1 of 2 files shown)


iceman1001:
check your logic with the counter.   you are increasing and checking against 0.

aaronjamt:
Good catch. Retested and it still sniffs correctly.

PR #3107, opened 2026-03-02
Add 'hf.field.timeout' to prefs

kormax:
This pull request allows configuring HF field activity timeout on PM3 through `param set hf.field.timeout`.
When that option is set, if Proxmark3 does not sense any user-driven activity under the timeout duration while waiting for next command, and it has an ON HF field, it turns it off.

This option is intended for hardware safety purposes, based on my recent experience with closing the PM3 client while on a long-running command, which seemingly caused the field to remain ON even after the command interrupted and client closed.
After coming back in an hour, I found the device and the card that was laid on it were boiling-hot. Thankfully, there seems to be no hardware damage, but this experience motivated me to try adding something to prevent this in the future.

A couple of considerations:
* To prevent affecting existing users, this option is OFF by default;  Perhaps, it's worth making this option ON by default, but making the timeout big enough not to cause any surprises (say, 30 min);
* Ideally, this feature could use `g_hf_field_active`, but it's not used by modules aside from `hf 14a`, and `felica` has its own version;
* To avoid changing a lot of code in all of the modules, I've decided to hook into FPGA configuration instead, and use a custom `g_hf_field_timeout_active` variable instead.

### DIFF
--- a/client/src/cmdparser.c
+++ b/client/src/cmdparser.c
@@ -212,14 +212,26 @@ void CmdsHelp(const command_t Commands[]) {
     PrintAndLogEx(NORMAL, "");
 
     int i = 0;
+    size_t max_name_len = 16; // minimum width for command name column
+    while (Commands[i].Name) {
+        if (Commands[i].IsAvailable()) {
+            size_t name_len = strlen(Commands[i].Name);
+            if (name_len > max_name_len) {
+                max_name_len = name_len;
+            }
+        }
+        ++i;
+    }
+
+    i = 0;
     while (Commands[i].Name) {
         if (Commands[i].IsAvailable()) {
             uint8_t old_printAndLog = g_printAndLog;
             g_printAndLog &= PRINTANDLOG_PRINT;
             if (Commands[i].Name[0] == '-' || Commands[i].Name[0] == ' ') {
-                PrintAndLogEx(NORMAL, "%-16s %s", Commands[i].Name, Commands[i].Help);
+                PrintAndLogEx(NORMAL, "%-*s %s", (int)max_name_len, Commands[i].Name, Commands[i].Help);
             } else {
-                PrintAndLogEx(NORMAL, _GREEN_("%-16s")" %s", Commands[i].Name, Commands[i].Help);
+                PrintAndLogEx(NORMAL, _GREEN_("%-*s") " %s", (int)max_name_len, Commands[i].Name, Commands[i].Help);
             }
             g_printAndLog = old_printAndLog;
         }


iceman1001:
interesting..      but the default use of a clock might be problematic.     the clocks are thrown up and down in the arm src for different functions.    

I see the reason for this PR,   we did a big code review and made sure all commands turned off the field ,  also added the `hw break`  to be able on client side to trigger exits from loops,   which then should be turning of field and clear out stuff after running.    It all depends on each command who has the field on,  to be gracefully handling it. 
We found this to be "easiest" way,   enforcing a device command to proper shutdown.   

if  you experience a case when your device kept the field on,  then that command you were running isn't handling itself as according to our preferred way and should be fixed.

kormax:
I don’t think that current client-side safeguards actually work. 

For instance, even though commands like “emv search”, “mfdes bruteaid”, “felica dump”, etc. do have DropField at the end of their code, if user doesn’t wait until completion and exits via CTRL+C, the client exists without dropping the field.

It may be possible to try adding other safeguards to the client, like sending a DropField command if PM3 client is killed via a signal as a last ditch effort.

But there are also cases where commands don’t turn off the field after completion, and it’s by design, those are commands with “keep” flag. I’m not sure it’s possible to design around them without any sort of device-side timers.

Nevertheless, your note about the clocks being meddled with by other commands isn't something I considered. I need to think if something can be done about it.

iceman1001:
I tried once or twice to add ctrl-c  signal hooking,   in order to detect and make client side try to tell device to shutoff field etc.

but then as you say we have the keep params, and long running simulations,    having one good solution feels not so easy.  

the client side with keep param commands,  maybe we should have a warning if we exit or have a way to inform the client that the field is up.

kormax:
IMO, the warning message won't be effective enough as it depends on user actually noticing it (before closing the window or clearing output) and acting upon this.

I'd like to look into the exit signal method instead of timers or messages. Perhaps it could be made an option instead of the one I proposed initially, so that it doesn't affect users that haven't subscribed to it.

iceman1001:
but also remember for standalone runs,   people wanna disconnect from client and keep the pm3 running,  then after long time,  hook it up and connect the device again.  Dont wanna loose all data

kormax:
That's why this could be an option that the user subscribes to ~~implicitly~~ explicitly.

If we talk about the originally suggested timer logic, it won't fire with standalone mode, as standalone script consumes the execution loop where the timer check is.

On that note, I've made a POC where client sends a field off command upon reception of an exit signal, and it seems to work fine, but I'm not feeling right at the moment about it delaying the closure of the client by a noticeable amount (100-200ms). Also, we may need to ignore sending the command if standalone mode was started while connected.

iceman1001:
100-200ms  is quit long.   

Did you regression test...   we have a lot of functionality on device side

iceman1001:
So,   where do we wanna go with this one?   @doegox   @mwalker33    @merlokk     some ideas how to make it work ?

doegox:
Hmm, I'm also concerned by monopolizing a timer.

Food for thought: since years, I'm using a simple "NFC flash nail sticker" on my antenna and it helped several times to detect bugs (HF being not turned off when it should) and fix them. RF consumption is really negligible. 
image

Another option if we only consider regular cases (hf 14a raw -k) we could add a timer and warning on the client side (dropping a warning message every 10s in the client when the client thinks the HF field is still on). Maybe even something in the prompt to indicate the last command maintains the HF field on. At least for playing with commands like 14a raw it would be a helpful indicator.

kormax:
Frankly, a warning message feels like a feel-good-non-solution that does not address the underlying issue of protecting people from damaging their hardware if they forget to disable the field on the device. Hence why I wanted to avoid it to the last moment.

That said, since I've been struggling with trying to make client-side field-kill work acceptably quickly without broad changes to the whole client, and since I have no understanding of what would need to be done to satisfy timer use concerns, I would agree that adding a "fat" warning when client exits knowing that the field is ON has some value, so I'm going to do that so we can close this issue for now.

kormax:
Dunno if to go with "fat" as in ASCII 

```log
[=]  /$$$$$$$$ /$$$$$$ /$$$$$$$$ /$$       /$$$$$$$         /$$$$$$  /$$   /$$
[=] | $$        | $$  | $$      | $$      | $$  \ $$      | $$  \ $$| $$$$| $$
[=] | $$$$$     | $$  | $$$$$   | $$      | $$  | $$      | $$  | $$| $$ $$ $$
[=] | $$__/     | $$  | $$__/   | $$      | $$  | $$      | $$  | $$| $$  $$$$
[=] | $$       /$$$$$$| $$$$$$$$| $$$$$$$$| $$$$$$$/      |  $$$$$$/| $$ \  $$
[=] |__/      |______/|________/|________/|_______/        \______/ |__/  \__/
```

Or use emojis, and color coding to attract attention
```log
[=] ################################
[=] ⚠️ FIELD IS ON      
[=] ⚠️
[=] ⚠️ Keeping the device with a field powered ON                    
[=] ⚠️ for prolonged amounts of time
[=] ⚠️ may or may not cause damage to the hardware.
[=] ################################
```

iceman1001:
the last one and with "Antenna field is ON"  in  RED 
the rest of the text as white.  

Would be a big ass warning.   

there is the  "field_on"  global param tend to use with 14a apu,  emv  , etc   which could be a way to detect it.
query it and when exiting the client,  send the fpga off...

kormax:
I’ve attempted implementing a client-driven warning, but considering that we’re dealing with an undetermined state, since we can’t be sure whether the device received or processed the last sent command when exiting, the “warn on exit” solution works exceptionally poorly, with around 15% false positives or negatives for commands like hf emv.
Pouring salt into the wound, getting to that "success rate" required adding a client-side state handler for all commands and responses and their flags, to "guess" the state of the device. This adds a big maintenance surface as one needs to sync the logic for determining the device state on the client with any device-side code changes.

At least from what I can see, there does not seem to be an acceptably viable solution apart from the device-side one.

Returning back to the original topic, could you explain concern with the original timer-driven solution?

iceman1001:
cant have something hogging a timer,  I think is our main concern

kormax:
I interpret “hogging” as “using too much”. You’re talking about GetTickCountDelta invocation in the PR? 

I also remember the “throwing the timer up and down” in your original comment.
Do I understand correctly that other commands using the timer could throw the tick count off. Would that be in absolute terms, or relatively? I wouldn't think that the timer sliding a bit into either direction would affect the end goal too much, and excessive jumps (or rollbacks) could be detected.

I think it's worth mentioning that we don't configure or update any timers in this PR, only use existing ones.

iceman1001:
it means the clock are setup and torn down in most commands.  if you use one,  you can be certain the last value you had stored ,  the next one will not be a nice increase from there and you would need to setup the clock once again every time any command touched the clock. 

Hogging a clock,  was meant as  "having a dedicated clock for this purpose"   is not feasible.

doegox:
if things like hf emv are not turning properly the field, these are bugs to track down and to fix, or any other situation where the client (and user) thinks the field is off while it's maintained on.
automatically turing off the field by the pm3 itself in these situations is hiding actual issues.

kormax:
Hello.

First, I'd like to thank you for your patience and taking part in the conversation, trying to find a solution.

> if things like hf emv are not turning properly the field, these are bugs to track down and to fix

> 
> automatically turing off the field by the pm3 itself in these situations is hiding actual issues.
> 
> 

As I see it currently, the proxmark3 client cannot make reliable guarantees that the field is turned off at the end of the command execution, regardless of command being implemented correctly or not.

While there may be commands that can be improved, and EMV module is one of them, the core of the issue is that the client application has limited control over when it's going to be killed off by the OS, which may happen before DropField command is issued, or as it is being sent to the device. For the same reason, field state tracking for warnings has less than desired correctness rate.

Attempting to interfere via exit signals may work for dropping the field or guarding command cleanup sections, but it introduces noticeable delays to the exit process, and would require sizable changes to the codebase. Perhaps exit signal method could be made fast enough, but I'm not sure on how to do it in a non-intrusive and fast way.

kormax:
If this information is of any context, doing Ctrl+C on macOS, I am reliably able to cause PM3 field to hang after client exit for any command which isn't "one shot", especially long running ones, including commands that attempt to cleanup through DropField.

Just to name a few example modules where I'm able to do that: Desfire/EMV/Felica/VAS/GST etc.

doegox:
What about implementing proper interruption routines for each command that takes a long time by nature?
There are only a very few of them, better to tackle them than risking impacting all commands with the timer approach.
There are already a few commands like that, grep "Press.*to abort"
And if the user dismisses the instructions and kills the client without much consideration, well, bad for him.
Ctrl-C on pm3 client (and then just leaving it plugged and unattended) is never a good idea.

kormax:
Good evening.

> better to tackle them than risking impacting all commands with the timer approach

I understand now that this is the main concern behind the discussion.

I would really appreciate if a person which knows this hardware much better than I do could point to the issues in my thinking here, to make this a learning moment.

As I see the code base on branch currently, the `RTTC` timer is updated and configured only in the `StartTickCount` function, declared in `ticks.c`, line `100`, though assignment to `AT91C_BASE_RTTC->RTTC_RTMR`.

The `StartTickCount` function is called from 2 locations:
1. `AppMain`, before the main processing loop starts.
2. `TimingIntervalAcquisition`, invoked when handling `CMD_TIA`, called via `hw tia`.

As for reading the `RTTC` timer, which happens through access to `AT91C_BASE_RTTC->RTTC_RTVR`, this happens in 2 functions:
1. `GetTickCount` in `ticks.c`, line `113`;
2. `GetTickCountDelta` in `ticks.c`, line `117.

Counting all the places those functions are called from, would amount to 37 and 17 respectively, used as a means of generating RNG and/or measuring execution speed or inner command time out.

Investigating any other uses of uses of registers related to `RTTC_` peripheral, the codebase contains no other references, apart from the headers declared in `at91sam...h`.

Given that, I'm struggling to understand how an additional call to `GetTickCount` and `GetTickCountDelta` could cause any issues for any other users of `RTTC`.

Perhaps there are chip or hardware platform related quirks I am failing to take into account, like if any of `AT91C_BASE_PWMC` or `AT91C_BASE_TC(0/1/2)` timers, which the underlying commands actively reconfigure, could indirectly affect the behavior of `RTTC`.

Many thanks for your time
- Max

doegox:
Hello @kormax ,

When I jumped in the conversation, I didn't realize you never touch the counter, only watch it.
And indeed the RTC is running on its own and don't risk being messed up, beside resets via TIA.

So yeah, second thought, about the question of using scarce counters, I think I'm ok with that.

There is still the risk that someone will set it to be safe, forget about it and days later, when he'll use the proxmark in scenarios where it is normal to stay on for long time, it will just stop working and he'll think the device is buggy...
I'm not sure of the full logic, maybe all "normally long" commands won't jump to the main loop anyway, and the watchdog will never be triggered...
To be tested...
EDIT: the warning message sent when triggered should avoid that confusion, so that seems ok.

About
```c
StartTickCount();
g_last_activity_ms = GetTickCount();
```
I'd move it in `StartTickCount()` so it's always called when the tick count is reset. (e.g. in TIA as well)
And it can probably simplified into `g_last_activity_ms = 0;`

About `prefs set hf.field.timeout`, given the typical timescale of this watchdog, I'd express it in seconds rather than milliseconds.
it can be called `hf.field.timeout_sec` and the code can convert it to ms internally, so it's more user friendly.

I think we can give it a try.
@iceman1001 what's your opinion?

iceman1001:
if you fine with the concept and nothing else is touching it,    sure,   

I dont get the  "I'f move it into StartTickCount()"    ,  I belong to the school of having functions do one thing and not magically change a global var....    Regardless,  a proper documentation somewhere about how this flow is working / suppose to work,   and why / what things it touches,     and we haven't  made it worse from before. 

No,  I don't think a GH issue comments are documentation enough either.. _sorry_   
but make sure the commit message is to point and add a section in the clocks.md  about this.   Or somewhere else where it may fit.

doegox:
Doc: good point! best place is in the https://github.com/RfidResearchGroup/proxmark3/blob/master/doc/clocks.md @kormax 

move it into StartTickCount() => g_last_activity_ms must be reset every time StartTickCount is called, so either g_last_activity_ms must be added everywhere StartTickCount is called (right now it's missing in TIA) or StartTickCount takes care of it and there is no risk that one day for X reason, someone adds a StartTickCount elsewhere and ignores he has to care about g_last_activity_ms manually...

iceman1001:
aha,  that makes total sense. 

hm,  well,  its down to document why its there so nobody gets the idea of removing it.

kormax:
As an additional suggestion, we could maintain an `epoch` or `cookie` value related to configuration of the clock, which could be used by the callers to verify that the clock wasn't touched since they last looked at it. (This value will be updated by method(s) that configure the registers in any manner).

doegox:
I'm sorry @kormax I didn't understand your suggestion. Can you detail?

kormax:
Instead of relying on `StartTickCount` to reset `g_last_activity_ms` value to 0, and a caller noticing it before that value is mutated, a more universal solution could instead be to add a global variable, like, `rttc_epoch` which is updated with a random value (or incremented) upon repeat `StartTickCount` invocations.
Any callers that rely on measured `GetTickCount` could, if they wish to, save the local `rttc_epoch ` locally, and check if it hasn't changed in between calls, thus indicating that their measurements can't be relied upon anymore.

doegox:
hmm why not. I'd provide a getter like `GetTickCountLabel`. No need to be a global variable, it could be scoped to `ticks.c` and the mechanism is more visible along the other fcts for some one discovering the API.

kormax:
Sorry for taking so long, finally had some time to get back to the topic (we're reaching a month mark for a PR, probably one of the longest discussions for a while here, hehe). 

* Add RTTC clock label variable retrievable through `GetTickCountLabel`, used to check if clock config changed since last observation.
* Added document mentions on use of the clock and label variable;
* Updated parameter name to `hf.field.timeout_sec`, now configured in seconds `--sec 5`.
* Added a log message when field times out while the user is in the client:
```
[usb] pm3 --> prefs set hf.field.timeout_sec --sec 5
[=]     HF field timeout........ 5 s
[usb] pm3 --> hf 14a raw -akc 55

[#] HF field auto-off: inactivity timeout (5000 ms). To disable, use 'prefs set hf.field.timeout_sec --sec 0'
```

doegox:
Tested locally, sounds good to me. @iceman1001 ok for merging ?

iceman1001:
Awesome!

It has be discussed,   its been documented,  its been understood.    Thank you @kormax    
The project has been improved regards to device safety.    

even if I am not sure why the max column length is doing in this pr but now I am just tried of this one ;)  and it is merged.

doegox:
@iceman1001 column: the new pref setting is super long and broke the fixed columns when doing e.g. `pref set`

PR #3108, opened 2026-03-03
Add nix.shell, nixos installation documentation

CaffeinatedOpe:
I was struggling with dependencies on my system, so I figure others will or are as well. Adding a shell.nix file and documentation to go with it will help simplify that process down the line. The shell file pulls in all the same requirements as the nix package for proxmark3.


iceman1001:
hm,   interesting,   I have to let @doegox decide this one.   He is better on *nix stuff.

From my perspective,  I try to avoid custom dedicated OS related files in the root folder.  It too many scripts and docs already there.  Besides if we gonna add a custom shell file for each distro or OS and maintain it,     meh.

CaffeinatedOpe:
If it would make things cleaner, the file could be moved to another folder with little to no impact. I put it in the root folder because that's the usual for most of the projects I've seen, but the location doesn't change functionality.

iceman1001:
if any,  it should be under `tools` folder.    maybe its own sub folder too.

doegox:
Under tools seems fine. I'm currently trying to deploy a docker with nixos to see how this works.

doegox:
tested in a debian docker, thanks

PR #3109, opened 2026-03-03
Handle inverted polarity RX frames in 'hf felica'

kormax:
During my experiments dedicated to improving FeliCa performance on PM3, I've noticed that sometimes the data was received successfully, but it was bit-to-bit inverted from the expected response.

Doing further research, I've stumbled into some documents, which gave a clue along the following lines: "SYNC should be used to determine polarity". This implied that it was expected that data could come in either polarity, and hence SYNC byte could be used to determine polarity the data came in.

Ideally, something like this would better be implemented on an FPGA, but to avoid touching things I'm not versed in, I've decided to modify the ARM side instead.

Thanks to those changes, response reception success rate for my Apple iPhone and Google Pixel samples rose from 3% and 10% to 80% each. While this is still not acceptable for normal use, it's now at least possible to detect mobile devices with "hf felica info", and other command that do re-tries.

### DIFF
--- a/armsrc/felica.c
+++ b/armsrc/felica.c
@@ -77,7 +77,7 @@ static uint32_t iso18092_get_timeout(void) {
 static uint8_t frameSpace[FELICA_MAX_RF_FRAME_SIZE];
 
 //structure to hold incoming NFC frame, used for ISO/IEC 18092-compatible frames
-static struct {
+typedef struct {
     enum {
         STATE_UNSYNCD,
         STATE_TRYING_SYNC,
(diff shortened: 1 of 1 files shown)


kormax:
Github is experiencing issues, win build failed because of that and may need to be re-run.

armsrc/felica.c
```diff
@@ -88,127 +88,164 @@ static struct {
...
-    switch (FelicaFrame.state) {
+    switch (f->state) {
 
         case STATE_UNSYNCD: {
             // almost any nonzero byte can be start of SYNC. SYNC should be preceded by zeros, but that is not always the case
             if (bt > 0) {
-                FelicaFrame.shiftReg = reflect8(bt);
-                FelicaFrame.state = STATE_TRYING_SYNC;
+                f->shiftReg = reflect8(bt);
+                f->shiftRegInv = reflect8((uint8_t)~bt);
```
iceman1001: why not take  `~(f->shiftReg)`  and save some cycles?

iceman1001:
Great find!

I had no clue that the sync byte determined polarity in FeliCa.

armsrc/felica.c
kormax: Makes sense, haven't thought about reflect8 result being invertible, fixed.

PR #3113, opened 2026-03-06
Improve `hf mfdes bruteaid`

kormax:
This PR makes the following changes to `hf mfdes bruteaid` command;

* Fix the command not detecting the tag not being present after the initial poll, continuing to iterate over options "into the void". 
This issue was caused by downstream DESFire communication methods doing one of the following:
 a) Not checking the actual response length;
 b) Not validating missing status from an invalid one, conflating lack of response with an error response;
* Add re-connection retries in case a tag is lost during brute force. 
Apart from being useful for detecting connection issues with normal cards, this is particularly helpful for working with Apple Wallet, as it activates the NFC interface only for a brief grace period (30-60 seconds), and you have to re-poll the card with ECP in order to continue, which this feature does (If `hf 14a config --pla ecp.*` was used).
* Add `--presets` option. 
Presets specify the way the attempted values are generated, and the following ones were implemented:
 a) `dict` - uses the `aid_desfire.json` file as the bruteforce dictionary;
b) `mad` - replaces separate `--mad` argument, brute forces MAD range `FxxxF`;
c) `ascii` - various subtypes, with `letters`, `numbers`, etc;
d) `full` - the original behavior.

Thanks to those changes, I've been able to discover a hidden AID used by WaveLynx only in their LEAF-formatted Apple Wallet badges `F51CDF` through the upgraded bruteforce command.

Retries demo (you can see that AIDs increment through errors, so we progress even after losing connection):
```log
[usb] pm3 --> hf mfdes bruteaid
[=] Bruteforcing with preset full range 000000-ffffff step 1 candidates 16777216
[=] Enumerating through all AIDs manually, this will take a while!
[+] Found New DESFire AID 000000
[!] ⚠️  No card response while checking AID 00005F. Reselecting card (1/5)...
[!] ⚠️  No card response while checking AID 000093. Reselecting card (1/5)...
[!] ⚠️  No ISO14443-A Card in field
[!] ⚠️  No card response while checking AID 000093. Reselecting card (2/5)...
[!] ⚠️  No ISO14443-A Card in field
[!] ⚠️  No card response while checking AID 000093. Reselecting card (3/5)...
[!] ⚠️  No ISO14443-A Card in field
[!] ⚠️  No card response while checking AID 000093. Reselecting card (4/5)...
[!] ⚠️  No card response while checking AID 000168. Reselecting card (1/5)...
[!] ⚠️  No ISO14443-A Card in field
[!] ⚠️  No card response while checking AID 000168. Reselecting card (2/5)...
```

Demo of using `dict` preset for checking for already known AID values on a card:
```log
[usb] pm3 --> hf mfdes bruteaid --preset dict
[=] Bruteforcing with preset dictionary candidates 404
[=] Dictionary source: `aid_desfire` (direct + inverted byte order)
[+] Found New DESFire AID F51CDF
 🕘 Brute DESFire AID Progress 66.5 %   current AID: 422206
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added `dict`, `ascii`, `mad` presets for `hf mfdes bruteaid` (@kormax)
 - Added tag loss detection & recovery into `hf mfdes bruteaid` (@kormax)
 - Added --credit option for `hf iclass legrec` command to perform a credit key recovery. This is experimental and unfinished as it only partially works.(@antiklesys)
 - Added hardening for all host binaries. Exact level of hardening depends on the OS (@doegox)
--- a/client/src/cmdhfmfdes.c
+++ b/client/src/cmdhfmfdes.c
@@ -18,8 +18,10 @@
 // Code heavily modified by B.Kerler :)
 
 #include "cmdhfmfdes.h"
+#include <ctype.h>
 #include <stdio.h>
 #include <string.h>
+#include "jansson.h"
 #include "commonutil.h"             // ARRAYLEN
 #include "cmdparser.h"              // command_t
 #include "comms.h"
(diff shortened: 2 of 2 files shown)


iceman1001:
Amazing fixes and new functionality!

PR #3116, opened 2026-03-07
hf mfp dump implementation

apply-science:
A early version for performing dumps of MIFARE Plus credentials that supports both SL1, SL3 and mixed mode credentials.

### DIFF
--- a/client/src/cmdhfmfp.c
+++ b/client/src/cmdhfmfp.c
@@ -1901,16 +1901,16 @@ static int CmdHFMFPChk(const char *Cmd) {
     return PM3_SUCCESS;
 }
 
-static int mfp_load_keys_from_json(const char *filename, uint8_t foundKeys[2][64][AES_KEY_LEN + 1]) {
+static int mfp_load_keys_from_json(const char *filename, uint8_t foundKeys[2][64][AES_KEY_LEN + 1], bool verbose) {
 
-    // loadFileJSON handles "mfpkeys" file type via loadFileJSONex.
+    // loadFileJSONex handles "mfpkeys" file type.
     // Buffer layout: UID(7) + pad(3) + SAK(1) + ATQA(2) + ATSlen(1) + ATS(atslen)
     //   then flat keys: KeyA0(16) KeyB0(16) KeyA1(16) KeyB1(16) ...
     uint8_t data[14 + 256 + (2 * 64 * AES_KEY_LEN)];
     memset(data, 0, sizeof(data));
     size_t datalen = 0;
 
-    int res = loadFileJSON(filename, data, sizeof(data), &datalen, NULL);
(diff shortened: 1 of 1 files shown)


client/src/cmdhfmfp.c
```diff
@@ -1915,53 +2082,513 @@ static int CmdHFMFPDump(const char *Cmd) {
...
+            PrintAndLogEx(INFO, "Probing " _YELLOW_("%u") " MFC keys against %d SL1 sectors...", mfcProbeKeyCnt, sl1_need_keys);
+            for (uint8_t s = 0; s < numSectors; s++) {
+                if (kbd_enter_pressed()) {
+                    PrintAndLogEx(WARNING, "\naborted via keyboard");
+                    break;
+                }
+                if (sectorSL[s] != MFP_SL_1) {
+                    continue;
+                }
+                if (mfcFoundKeys[0][s][0] && mfcFoundKeys[1][s][0]) {
```
iceman1001: maybe use the defines  KEY_A,  KEY_B  instead of 0,1    which would be easier to follow

client/src/cmdhfmfp.c
```diff
@@ -1915,53 +2082,513 @@ static int CmdHFMFPDump(const char *Cmd) {
...
+                // Timeout or other error -> SL3
+                sectorSL[s] = MFP_SL_3;
+            }
         }
+        DropField();
 
+        int pre_sl3 = 0, pre_sl1 = 0;
+        for (uint8_t s = 0; s < numSectors; s++) {
+            if (sectorSL[s] == MFP_SL_3) pre_sl3++;
+            else if (sectorSL[s] == MFP_SL_1) pre_sl1++;
```
iceman1001: We would need to run `make style`   since the code style is kind of mixed.

iceman1001:
Cool new functionality!

But why do I get a feeling you are mixing check keys (or find all needed keys)   and dump  in the same go?

I suggest you have a `hf mfp chk`  which gives you a file with all keys needed for  sl1, sl3  (mixedin mode)  

and the reading of a json file,   belongs in the  fileutils.c

client/src/cmdhfmfp.c
apply-science: Renamed accordingly.

client/src/cmdhfmfp.c
apply-science: Adjusted the style accordingly.

iceman1001:
the dump command should support the `nosave`  parameter as well.

apply-science:
> the dump command should support the `nosave` parameter as well.

It already does. 
  - Line 2000: arg_lit0(NULL, "ns", "No save to file")                                                                                      
  - Line 2022: bool nosave = arg_get_lit(ctx, 5);                                                                                           
  - Line 2332-2337: Skips saving when --ns is set

iceman1001:
Execellent!   Great PR!

PR #3117, opened 2026-03-08
Introduce 'hf felica seacinfo' command

kormax:
There exists a special type of FeliCa cards, called FeliCa SEAC, which are "generation zero" cards, deployed before FeliCa was fully released, using Sony proprietary crypto instead of DES/3DES.

Those cards use the same RF and SOF/EOF as regular FeliCa, but on application level, their commands and responses are different in format.

I'm still in the process of fuzzing for potential commands, but meanwhile, I've decided to introduce `hf felica seacinfo` command that POLLs for and detects this special type of cards, printing its counterpart of what may be IDM + shortened PMM (printed as IDSEAC until further notice).

At first I wanted to add it directly as a fallback to `hf felica info`, but due to still relatively poor FeliCa performance  in PM3, and longer than usual response times for SEAC (those cards could take ~150ms to answer), this added an extra second or more of delay to that command, so I've decided to make it separate for now.

```log
[usb] pm3 --> hf felica seacinfo

[=] --- Tag Information ---------------------------
[=] Type........... FeliCa SEAC
[=] IDSEAC......... 0600************0707
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added `hf felica seacinfo` command (@kormax)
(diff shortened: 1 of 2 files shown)


iceman1001:
Awesome!

PR #3118, opened 2026-03-09
Add universal node discovery for FeliCa

kormax:
This PR adds `hf felica discnodes` method for performing universal node discovery for FeliCa cards, as inspired by some of [my research into capabilities of different FeliCa card types](https://github.com/kormax/felica-tool?tab=readme-ov-file#supported-felica-commands).

When attempting node discovery, the command will use the most effective available method for that particular card, attempted in the order of decreasing efficiency. It is also possible to specify a method to try using `--method` param:

- **RequestCodeList**  `request_code_list`:
  Returns up to 16 services and 16 areas per iteration. 
This is the most efficient method, but it is supported only by some Android mobile devices.

- **SearchServiceCode**  `search_service_code`:
  Returns one node at a time iteratively. 
It works on most card types, except first-generation export-market FeliCa, some Octopus variants, and FeliCa Lite.

- **RequestService**   `request_service`
  Can probe up to 32 node codes per request via the key version request mechanism. In practice, although this sounds efficient, the node space is still large enough that full discovery requires roughly 1,000 requests even with batching.

  Two important limitations apply to this method:
  - Some cards do not reliably support batches of 32 nodes. During testing for this PR, I found that there exist cards that support only 16-20 nodes (yet again, Octopus), so the current implementation uses 16.
  - This method does not return the node code itself, only allows to check if a particular one exists, so it cannot determine area end codes or reconstruct area nesting relationships.

  This method works on all cards that support standard DES/AES encryption, which includes most FeliCa variants except FeliCa Lite.

- **Read Without Encryption**  `read_without_encryption`
  Probes one node at a time and works only for read-enabled services. 
Although very inefficient, it is primarily intended for FeliCa Lite, which does not support `RequestService` or `SearchServiceCode` and instead relies on the fixed node code `0`.

### 1. RequestCodeList
```log
[usb] pm3 --> hf felica discnodes --method request_code_list
[?] Area and service codes are printed in network order.
[=] Node discovery method used: RequestCodeList
[=] ┌───────────────────────────────────────────────
[=] ├── AREA_0000FEFF (0-1023)
[=] ├── AREA_00103F17 (64-92)
[=] │   ├── AREA_01103F17 (64-92)
[=] │   │   ├── SVC_0810 (64)
< OUTPUT TRUNCATED >
[=] │   │   ├── SVC_8975 (470)
[=] └───┴───┴───────────────────────
[=] Node discovery duration: 318 ms
[+] Service code and area dump complete. Discovered 129 node(s): 19 area(s), 110 service(s).
```
This method discovered all 129 nodes (19 areas, 110 services) in 318ms.

### 2. SearchServiceCode
```log
[usb] pm3 --> hf felica discnodes --method search_service_code
[?] Area and service codes are printed in network order.
[=] Node discovery method used: SearchServiceCode
[=] ┌───────────────────────────────────────────────
[=] ├── AREA_0000FEFF (0-1023)
[=] ├── AREA_00103F17 (64-92)
[=] │   ├── AREA_01103F17 (64-92)
[=] │   │   ├── SVC_0810 (64)
< OUTPUT TRUNCATED >
[=] │   │   ├── SVC_8975 (470)
[=] └───┴───┴───────────────────────
[=] Node discovery duration: 3810 ms
[+] Service code and area dump complete. Discovered 129 node(s): 19 area(s), 110 service(s).
```
This method discovered all 129 nodes (19 areas, 110 services) but in 3810ms. Ten-fold increase from RequestCodeList, but still plenty fast.

### 3. RequestService
```log
[usb] pm3 --> hf felica discnodes --method request_service
[?] Area and service codes are printed in network order.
[=] Node discovery method used: RequestService
[=] ┌───────────────────────────────────────────────
[=] ├── AREA_0000 (0-?)
[=] ├── AREA_0010 (64-?)
[=] ├── AREA_0110 (64-?)
[=] ├── SVC_0810 (64)
[=] ├── SVC_0811 (68)
[=] ├── SVC_0A11 (68)
<OUTPUT TRUNCATED>
[=] └───────────────────────
[=] Node discovery duration: 78606 ms
[+] Service code and area dump complete. Discovered 129 node(s): 19 area(s), 110 service(s).
```
This method discovered all 129 nodes (19 areas, 110 services) but in 78606ms, a 20-fold increase over `SearchServiceCode`. 

Additionally, it wasn't able to determine area end codes, hence why no nesting is shown in the visualization.

### 4. ReadWithoutEncryption
```log
[usb] pm3 --> hf felica discnodes --method read_without_encryption
[?] Area and service codes are printed in network order.
[=] Node discovery method used: ReadWithoutEncryption
[=] ┌───────────────────────────────────────────────
[=] ├── SVC_0B11 (68)
[=] ├── SVC_1713 (76)
< OUTPUT TRUNCATED >
[=] └───────────────────────
[=] Node discovery duration: 241688 ms
[+] Service code and area dump complete. Discovered 42 node(s): 0 area(s), 42 service(s).
``` 
This method was only able to discover read-available nodes, no areas, and it took a whopping 241688 milliseconds to do so.

## Additional changes

- `hf felica scsvcode` was updated to use new discovery system, but only in `SearchServiceCode` mode to keep existing behavior. We may think about removing that command in the future if this one takes off.
- `hf felica dump` now uses the new discovery system. This now allows non-protected FeliCa lite cards to be dumped as a side-effect.
- All commands that rely on new node discovery system can now be interrupted by pressing ENTER.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added `hf felica discnodes` command (@kormax)
 - Added `hf mfp dump` command (@apply-science)
 - Added `hf felica seacinfo` command (@kormax)
 - Added `hf mfdes bruteisofid` and `hf mfdes selectisofid` commands (@kormax)
--- a/client/src/cmdhffelica.c
+++ b/client/src/cmdhffelica.c
@@ -47,6 +47,8 @@
 
 #define FELICA_DEFAULT_TIMEOUT_MS 2000U
 #define FELICA_DEFAULT_RETRY_COUNT 3U
+#define FELICA_DISCOVER_DEFAULT_RETRY_COUNT 5U
+#define FELICA_DISCOVERY_RETRY_BACKOFF_MS 1000U
 #define FELICA_PLATFORM_INFO_MAX_LEN 64U
 #define FELICA_PLATFORM_INFO_WITH_MAC_INFO_LEN 25U
 #define FELICA_PLATFORM_INFO_WITH_MAC_LEN 20U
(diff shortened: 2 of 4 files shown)


iceman1001:
Awesome!

PR #3119, opened 2026-03-09
mfc: dic: Vanderbilt ACT

jof:
Add discovered Mifare Classic Keys for the Vanderbilt ACT reader line


iceman1001:
Cool,    just to be sure,  these keys are static , ie,  used across many cards from the same system?

block number 00-FF  in the keys..   is that a & b keys?     Its almost like we should have a kdf for it instead

jof:
Given the symbolism, it is highly likely that `414354616374` is static.

`4D61071B7254` and `34016FAC127D` are observed on two different systems, but likely configured by the same integrator. I'm less confident in these, but they seem to be static across credentials.

The `4163636573`+`xx` style keys are as B keys. A keys are set to FFFFFFFFFFFF. +1 for a KDF -- that seems like a much more compact way to represent this. I'll have a look around the codebase for this.

PR #3128, opened 2026-03-11
Refine known AIDs; add Android HCE

kormax:
This PR fixes up some names and descriptions for entries in aidlist.json.

It also adds an extra entry for "Android HCE" - AID which is available on ALL Android devices, provided they have HCE feature turned on, as that AID SELECT is automatically handled by the system NFC service.

```log
[usb] pm3 --> hf 14a info --aidsearch

[+]  UID: 08 07 67 9B   ( RID - random ID )
[+] ATQA: 00 04
[+]  SAK: 20 [1]
[=] -------------------------- ATS ----------------------------------
< OUTPUT TRUNCATED >
[=] -------------------- AID Search --------------------
[+] AID : A000000476416E64726F6964484345 | Google | Android HCE
[+] AID : 4F53452E5641532E3031 | Apple, Google | Value-Added Services (OSE.VAS.01)
[+] AID : A000000476D0000111 | Google | Google SmartTap 2.0
[+] AID : A000000809434343444B467631 | Car Connectivity Consortium (CCC) | Digital Car Key Framework
[+] AID : A000000809434343444B417631 | Car Connectivity Consortium (CCC) | Digital Car Key
[+] AID : A000000909ACCE5502 | Connectivity Standards Alliance (CSA) | Aliro Step Up
[+] AID : A000000909ACCE5501 | Connectivity Standards Alliance (CSA) | Aliro Expedited
[=] ----------------------------------------------------
```

Taking this opportunity, I'd like to ask for an opinion on the following potential change:

Currently, both `hf 14a info` and `hf 14b info` attempt AID selections only with `--aidsearch` argument.

If we could make probing for "Android HCE", and "OSE.VAS" happen regardless of that argument, those 2 SELECTs would give us an opportunity to detect when a mobile phone is used with 99% certainty (in fact, just OSE.VAS would give us a big chunk, minus the Chinese phones), and help to adjust the potential `hf 14{x} info` display format based on that.

### DIFF
--- a/client/resources/aidlist.json
+++ b/client/resources/aidlist.json
@@ -1536,6 +1536,14 @@
         "Type": "EMV"
     },
     {
+        "AID": "A000000476416E64726F6964484345",
+        "Vendor": "Google",
+        "Country": "N/A",
+        "Name": "Android HCE",
+        "Description": "Available on AOSP-based android devices with active HCE",
+        "Type": ""
+    },
+    {
         "AID": "A000000476A010",
         "Vendor": "Google",
         "Country": "United States",
@@ -2275,8 +2283,8 @@
         "AID": "A0000008580102",
         "Vendor": "Apple",
         "Country": "",
-        "Name": "Apple Home Key Framework",
-        "Description": "Home Key configuration applet. Used for attestation exchange",
+        "Name": "Apple Home Key Step Up",
+        "Description": "Used for reading the attestation certificate",
         "Type": ""
     },
     {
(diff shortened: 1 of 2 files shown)


iceman1001:
the HCE automatic detection in both 14a / 14b `info`   go for it.   This is why we have reader vs info command. 
reader should to anticollision  and minimal detection,    info should give as much as possible as long as it not doing long run of searches,  etc.

iceman1001:
regarding your textual changes,   as long as we don't loose any information its good.   Now I feel we lost some interesting aspects of the descriptions

iceman1001:
Sorry,  miss clicked.

kormax:
I don’t think I removed any information?

For instance, “Aliro/Home” “configuration” turned out to not be “configuration”, but called “step up” officially, and not have any config features.

As for OSE.VAS - the issue here is, that this AID is used by both Google SmartTap and Apple VAS, so we can’t decide which one of those is it based solely on AID presence.
Ideally you can detect the “ApplePay” or “AndroidPay” returned from it, but it would require changes to the existing logic, which I may perform separately as you’re ok with HCE detection.

PR #3132, opened 2026-03-13
Add `hf gst` commands

kormax:
### Introduction

This PR adds support for the Google Smart Tap protocol, enabling reading of mobile loyalty, ticketing, and other generic pass types from Google Wallet.

The following flow is implemented:
* SELECT OSE.VAS.01
* (Optionally) SELECT SMART TAP 2
* NEGOTIATE SECURE CHANNEL
* GET DATA

To retrieve information about a device’s GST version without reading a specific pass, `hf gst info` command can be used:

```log
[usb] pm3 --> hf gst info
[=] ------------------------------ OSE Information (6F) ------------------------------
[=] Wallet type............... AndroidPay
[=] Features (C1)............. CC00000000008080
[=]   Transaction mode........ 11001100 (0xCC)
[=]    1 ...... - Payment enabled
[=]    .1 ..... - Payment requested
[=]    ..0 .... - Reserved/unknown bit clear
[=]    ...0 ... - Reserved/unknown bit clear
[=]    ....1 .. - Pass enabled
[=]    .....1 . - Pass requested
[=]    ......0  - Reserved/unknown bit clear
[=]    .......0 - Reserved/unknown bit clear
[=] Capabilities (DF62)....... 00000011 (0x03)
[=]    ......1  - VAS support advertised
[=]    .......1 - Skipping second select allowed
[=] Minimum version........... 0 (0.0)
[=] Maximum version........... 1 (1.0)
[=] Device nonce (C2)......... 73DFC75EA83C59D59C09EDC971762484E570B07A873957B0A2EE78C36F6D4C9D
[=] Device nonce (DF6E)....... not present
```

If the device is not equipped with Google Smart Tap - this and other commands will raise a warning:
```log
[usb] pm3 --> hf gst info
[=] ------------------------------ OSE Information (6F) ------------------------------
[=] Wallet type............... ApplePay
[!] ⚠️  Wallet type is not AndroidPay. This likely isn't Google Smart Tap.
```

For reading cards, the `hf gst read` command was added.

The following parameters are mandatory:

* `--cid` / `--collector-id` specifies the collector identifier to present to the device. This is a 32-bit value and may be given in decimal or hexadecimal form.
* `--rpk` / `--reader-private-key` specifies the reader private key used to prove authorization for that collector. It may be provided as a PEM key, PEM base64 body, DER hex, raw 32-byte scalar hex, or a file path.

The command also supports several optional parameters:

* `--key-version` selects the long-term key version of a collector key to advertise. The default is `1`. Is not always 1.
* `--mode` selects the reader mode. Supported values are `pass-only`, `payment-only`, `pass-and-payment`, and `pass-over-payment` (default). Does not seem to affect the behavior.
* `--select-smarttap2` controls whether the optional Smart Tap v2 applet select is performed. Supported values are `auto` (default), `yes`, and `no`.
* `--session-id`, `--reader-nonce`, and `--reader-ephemeral-private-key` override randomly generated values, which is mainly useful for testing and reproducibility.
* `--no-live-auth` disables live authentication and uses a zeroed handset nonce in the reader signature. Not supported by generally-available passes, as that feature is enabled by Google server-side for a particular pass.
* `--apdu` prints APDU requests and responses for debugging.

### Explainer

In Google Smart Tap, reader does not request particular cards directly. Instead, the protocol uses the concept of "collectors".

A collector is an entity destined to "collect" data.
Pass issuers configure whether a given pass is readable by a one or multiple collectors, and as a result, each collector may be eligible to read one or multiple types of passes.

To collect data, the reader presents a collector identifier and generates a transaction-bound proof to the mobile device, using one of the collector key pairs, proving that it is authorized to read the passes.

If verification succeeds, the mobile device assembles a list of "objects" eligible for that collector, which may or may not include data from multiple passes.

If all objects can be returned in a single request — for example, when the user has a single pass, or multiple non-conflicting passes — the transaction completes immediately. 

Otherwise, the user is shown a disambiguation screen and asked to choose which card to emulate.

### Examples

Attempt to read from a device that has multiple eligible passes:

```log
[usb] pm3 --> hf gst read --cid 20180608 --rpk gst.google.der
[=] 
[=] ------------------------------ OSE Information (6F) ------------------------------
[=] Wallet type............... AndroidPay
[=] Features (C1)............. CC00000000008080
[=]   Transaction mode........ 11001100 (0xCC)
[=]    1 ...... - Payment enabled
[=]    .1 ..... - Payment requested
[=]    ..0 .... - Reserved/unknown bit clear
[=]    ...0 ... - Reserved/unknown bit clear
[=]    ....1 .. - Pass enabled
[=]    .....1 . - Pass requested
[=]    ......0  - Reserved/unknown bit clear
[=]    .......0 - Reserved/unknown bit clear
[=] Capabilities (DF62)....... 00000011 (0x03)
[=]    ......1  - VAS support advertised
[=]    .......1 - Skipping second select allowed
[=] Minimum version........... 0 (0.0)
[=] Maximum version........... 1 (1.0)
[=] Device nonce (C2)......... 357171AD54A038F15CFF5B5BF348F6D82383C4D501B934BA457A462EA24C96A8
[=] Device nonce (DF6E)....... not present
[=] 
[=] ---------------------------- Negotiate secure channel ----------------------------
[=] Collector ID.............. 20180608 (0x0133EE80)
[=] Long-term key version..... 1 (0x00000001)
[=] Session ID................ 683F42AAB855F29E
[=] Reader nonce.............. CA70B9425AEDF2AA17D1F2F27C6D678157816746C2938CC5A9835F8B181D921A
[=] Device nonce.............. 357171AD54A038F15CFF5B5BF348F6D82383C4D501B934BA457A462EA24C96A8
[=] Reader ephemeral pubkey... 02E0B5D4694B0F673B393BBE15C12FF646E7AB8066B0C56465C01B4058F7A716BF
[=] Reader signature.......... 3045022020D06C2DD20C4F62AEB6549C5EBB8902143F96EBF984ED464453828F3EB85004022100E3026EF3DA640544EC85B374B6DDFA31DD1E2928EC449F43C696C42C20DB32B3
[=] Live authentication....... yes
[=] NEGOTIATE status.......... 9000 (OK)
[=] 
[=] ------------------------------------ Get data ------------------------------------
[=] GET DATA status........... 9302 (DISAMBIGUATION_SCREEN_SHOWN)
```

Provided that user had opened a particular pass on a screen, or chose a pass on a disambiguation screen, the transaction can be performed again, this time to the end: 

```log
[usb] pm3 --> hf gst read --cid 20180608 --rpk gst.google.der
[=] 
[=] ------------------------------ OSE Information (6F) ------------------------------
[=] Wallet type............... AndroidPay
[=] Features (C1)............. CC00000000008080
[=]   Transaction mode........ 11001100 (0xCC)
[=]    1 ...... - Payment enabled
[=]    .1 ..... - Payment requested
[=]    ..0 .... - Reserved/unknown bit clear
[=]    ...0 ... - Reserved/unknown bit clear
[=]    ....1 .. - Pass enabled
[=]    .....1 . - Pass requested
[=]    ......0  - Reserved/unknown bit clear
[=]    .......0 - Reserved/unknown bit clear
[=] Capabilities (DF62)....... 00000011 (0x03)
[=]    ......1  - VAS support advertised
[=]    .......1 - Skipping second select allowed
[=] Minimum version........... 0 (0.0)
[=] Maximum version........... 1 (1.0)
[=] Device nonce (C2)......... D8766BCDB827710EDB78D2B346B440D1DD7D2FFEF37417F507D621310928D3D3
[=] Device nonce (DF6E)....... not present
[=] 
[=] ---------------------------- Negotiate secure channel ----------------------------
[=] Collector ID.............. 20180608 (0x0133EE80)
[=] Long-term key version..... 1 (0x00000001)
[=] Session ID................ 83EE0687F14CC761
[=] Reader nonce.............. A6C24F453F930D98037E09AF2E59FBC8C429F6E9EC6C4E0E7CCDDFFEA4FDAEAB
[=] Device nonce.............. D8766BCDB827710EDB78D2B346B440D1DD7D2FFEF37417F507D621310928D3D3
[=] Reader ephemeral pubkey... 032D73395EB2CA8E94F30C9CB77504E866CCCF25984FC38BD44C2401A22707D4F8
[=] Reader signature.......... 3045022100CB75DDD831FCF6C6A96F472CF406919992755FAEFDAE3F3CA39D7BE88B2F701B0220421D43F49D0834F62F0E26F5AD0AE895C3ED2203ED371FAFD27C3347C8FCFA88
[=] Live authentication....... yes
[=] NEGOTIATE status.......... 9000 (OK)
[=] 
[=] ------------------------------------ Get data ------------------------------------
[=] GET DATA status........... 9000 (OK)
[=] Bundle flags............... 00000011 (0x03)
[=]    ......1  - Payload compressed
[=]    .......1 - Payload encrypted
[=] 
[=] -------------------------------- Service objects ---------------------------------
[=] 
[=] Object #1
[=]   kind.................... customer
[=]   issuer_type............. wallet (0x02)
[=]   issuer_id............... 71797971
[=]   customer_id............. 0400000000000000000000000000000000
[=]   language................ "en" (fmt=0x00, ASCII)
[=] 
[=] Object #2
[=]   kind.................... loyalty
[=]   issuer_type............. merchant (0x01)
[=]   issuer_id............... 00000000
[=]   object_id............... 04C1986A7134EDB1EF
[=]   service_number.......... "2018" (fmt=0x00, ASCII)
```

As a result of Google Smart Tap transaction - reader receives a list of data objects. Types of objects depend on type of the pass, "service_number" carries the custom payload.
There may be auxiliary objects sent alongside pass objects, like the "customer" object, indicating user language.

To make it easier to test the implementation, demo keys were added for the following pass issuers that share their pass samples for free online (`--help` argument provides examples with those keys):
* Google `gst.google.der`;
* Passkit `gst.passkit.der`.

### Notes

* To support decompression of pass payload for faster read speed, `zlib` was introduced into the project as an optional dependency. In case user does not have the dependency - PM3 reader will indicate to the mobile device that response data should not be compressed to avoid the need to decompress.
* Users can provide private EC key data for `--reader-private-key`  in multiple formats.
I've attempted for it to be as flexible as possible, with the intention of updating other commands `aliro`/`vas` to use flexible EC key parsing down the line:
    * As 32 byte HEX scalar of key data;
    * As full pem string with header;
    * As base64 pem string without header;
    * As DER hex;
    * As a file path to a file containing any of the above formats.
 * Following the topic of potential updates, down the line I'd like to move functions like `gst_print_big_header` to a common location, as there are similar implementations elsewhere (I.E, I added one for Aliro too, but I don't want to touch it in this PR).

### DIFF
--- a/client/src/cmdhfgst.c
+++ b/client/src/cmdhfgst.c
@@ -1560,6 +1560,116 @@ static int gst_decompress_payload_zlib(const uint8_t *compressed_payload, size_t
 }
 #endif
 
+static int gst_selftest_zlib(void) {
+#ifdef HAVE_ZLIB
+    // Captured from a compressed RECORD_BUNDLE payload produced by Google's demo pass.
+    static const uint8_t compressed_payload[] = {
+        0x78, 0xDA, 0x5D, 0xD4, 0xBD, 0x4A, 0x03, 0x41, 0x18, 0x85, 0xE1, 0xDD,
+        0xEC, 0x22, 0x58, 0x6A, 0x67, 0x99, 0xD4, 0x0A, 0xF9, 0x7E, 0xA2, 0xC6,
+        0xDA, 0x26, 0x76, 0xE2, 0x56, 0x82, 0x4A, 0x48, 0x02, 0x46, 0x82, 0x1A,
+        0x93, 0x08, 0x41, 0xEC, 0x2C, 0x2D, 0x2C, 0x82, 0x8D, 0xE0, 0x05, 0x88,
+        0xB5, 0xD7, 0xA0, 0x77, 0x21, 0xDE, 0x87, 0xE0, 0xA2, 0xCC, 0xD9, 0x39,
+        0x4E, 0xB5, 0xF3, 0xEE, 0x34, 0x0F, 0xF3, 0x73, 0x97, 0x25, 0x49, 0xFE,
+        0xDE, 0x9D, 0x5C, 0x2F, 0xD2, 0xA5, 0x61, 0x9E, 0x26, 0xE5, 0xD8, 0xAD,
+        0x95, 0xE9, 0x6D, 0x34, 0x5F, 0x64, 0xCB, 0x17, 0xC3, 0x7E, 0x5E, 0xFF,
+        0xD8, 0xDB, 0x4F, 0xBF, 0x8E, 0x5F, 0x3A, 0xE5, 0xCF, 0xFC, 0x39, 0x2D,
+        0xCE, 0x93, 0x9B, 0x46, 0xBF, 0x3B, 0xED, 0x6E, 0x34, 0x1B, 0x3B, 0xBF,
+        0x1F, 0x27, 0xA7, 0x83, 0xAB, 0x41, 0xB3, 0xB1, 0xFE, 0x57, 0x25, 0xAE,
+        0x12, 0xAA, 0xC6, 0x55, 0x43, 0xB5, 0xB8, 0x5A, 0xA8, 0x1E, 0x57, 0x0F,
+        0xB5, 0x15, 0xD7, 0x56, 0xA8, 0x9B, 0x71, 0xDD, 0x0C, 0x75, 0x2B, 0xAE,
+        0x5B, 0xA1, 0x6E, 0xC7, 0x75, 0x3B, 0xD4, 0x76, 0x5C, 0xDB, 0x50, 0x10,
+        0x4E, 0x2A, 0x1D, 0xF3, 0xE0, 0x13, 0x02, 0x0A, 0x84, 0x42, 0x44, 0x81,
+        0x51, 0x08, 0x29, 0x50, 0x0A, 0x31, 0x05, 0x4E, 0x21, 0xA8, 0x40, 0x2A,
+        0x44, 0x15, 0x58, 0x85, 0xB0, 0x02, 0xAD, 0x10, 0x57, 0xE0, 0x55, 0xF2,
+        0x2A, 0xBC, 0x4A, 0x5E, 0xAD, 0xF6, 0x93, 0x37, 0x14, 0x5E, 0x25, 0xAF,
+        0xC2, 0xAB, 0xE4, 0x55, 0x78, 0x95, 0xBC, 0x0A, 0xAF, 0x92, 0x57, 0xE1,
+        0x55, 0xF2, 0x2A, 0xBC, 0x4A, 0x5E, 0x85, 0x57, 0xC9, 0xAB, 0xF0, 0x1A,
+        0x79, 0x0D, 0x5E, 0x23, 0xAF, 0xC1, 0x6B, 0xE4, 0xB5, 0xEA, 0x04, 0xF3,
+        0x11, 0x86, 0xD7, 0xC8, 0x6B, 0xF0, 0x1A, 0x79, 0x0D, 0x5E, 0x23, 0xAF,
+        0xC1, 0x6B, 0xE4, 0x35, 0x78, 0x8D, 0xBC, 0x06, 0xAF, 0x91, 0xD7, 0xE0,
+        0x75, 0xF2, 0x3A, 0xBC, 0x4E, 0x5E, 0x87, 0xD7, 0xC9, 0xEB, 0xF0, 0x3A,
+        0x79, 0xBD, 0xBA, 0xB3, 0x7C, 0x69, 0xE1, 0x75, 0xF2, 0x3A, 0xBC, 0x4E,
+        0x5E, 0x87, 0xD7, 0xC9, 0xEB, 0xF0, 0x3A, 0x79, 0x1D, 0x5E, 0x27, 0x6F,
+        0x39, 0xBB, 0x2D, 0xB2, 0x4E, 0x78, 0xCD, 0x6A, 0xE3, 0xF9, 0x7C, 0x5C,
+        0x64, 0xED, 0xDE, 0x6C, 0xB2, 0xC8, 0x56, 0x7A, 0xE5, 0x5B, 0x96, 0xFC,
+        0x1B, 0x6B, 0x69, 0x96, 0x15, 0xBD, 0xCB, 0x51, 0x32, 0x38, 0x2F, 0xCA,
+        0x15, 0xB3, 0x69, 0x7E, 0x7F, 0x74, 0xF8, 0x50, 0x7F, 0x3C, 0x58, 0xFD,
+        0x1E, 0x7C, 0x8E, 0xC6, 0x67, 0x4F, 0xAF, 0x3F, 0xF7, 0x6A, 0x82, 0x30,
+    };
+    static const uint8_t loyalty_object_id[] = {0x04, 0x21, 0xC9, 0x4A, 0x51, 0x01, 0xE2, 0x5E, 0xAA};
+    static const uint8_t customer_id[] = {0x04, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+                                          0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+    static const size_t sample_decompressed_len = 1312;
+    static const size_t service_value_record_1_len = 1224;
+    static const size_t loyalty_record_len = 1206;
+    static const size_t service_value_record_2_len = 73;
+    static const size_t customer_record_len = 57;
+
+    uint8_t *decompressed_payload = NULL;
+    size_t decompressed_payload_len = 0;
+    int res = gst_decompress_payload_zlib(compressed_payload, sizeof(compressed_payload),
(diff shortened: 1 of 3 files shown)


client/src/cmdhfgst.c
```diff
@@ -0,0 +1,2467 @@
...
+    *offset += src_len;
+    return PM3_SUCCESS;
+}
+
+static void gst_u32_to_be(uint32_t value, uint8_t out[4]) {
+    out[0] = (uint8_t)(value >> 24);
+    out[1] = (uint8_t)(value >> 16);
+    out[2] = (uint8_t)(value >> 8);
+    out[3] = (uint8_t)(value);
+}
```
iceman1001: there should be in common/utils.c  functions for this already

client/src/cmdhfgst.c
```diff
@@ -0,0 +1,2467 @@
...
+static void gst_u64_to_be(uint64_t value, uint8_t out[8]) {
+    out[0] = (uint8_t)(value >> 56);
+    out[1] = (uint8_t)(value >> 48);
+    out[2] = (uint8_t)(value >> 40);
+    out[3] = (uint8_t)(value >> 32);
+    out[4] = (uint8_t)(value >> 24);
+    out[5] = (uint8_t)(value >> 16);
+    out[6] = (uint8_t)(value >> 8);
+    out[7] = (uint8_t)(value);
+}
```
iceman1001: there should be in common/utils.c  functions for this already

iceman1001:
Wow,   dunno how much code you have laying around :)    But this is awesome!

Its a zlib dependency added,  but it looks like system install checks,   and optional. 
Usually @doegox  domain,

kormax:
Update:
* Moved shared RNG context to `libpcrypto`, which can be re-used in Aliro later;
* Replaced use of redundant uint conversion methods with existing utilities;
* Added extra utilities, that can be later re-used in Aliro module:
   * `parse_uint32_hex_or_dec`;
   * `bytes_equal_not_null`;
   * `buffer_append_bytes_with_offset`.
 * Added ui helper for printing big headers `PrintAndLogInfoHeader`, also to be re-used in Aliro module later.
 * Declared object types separately before using them in object type list.


As for ZLIB inclusion: considering that I don't have an ability to test or envision every environment configuration for all users, I wanted to include ZLIB in a manner that wouldn't break any builds or installs in case a user does not have it for one reason or another, hence why it's optional. In case you make a decision on making it mandatory, or would like me to update make configuration, I'm all for it.

kormax:
As for other things I had in backlog: 

- There is a massive improvement for FeliCa planned regarding lots of new commands and card classification, but I don't want to move forward with it unless someone knowledgeable makes RF/FPGA perform somewhat comparably to other modulations, as it's still pretty bad. I've tried stuff, but to no avail.

- Additionally, I have had a partially done implementation of in-trace DESFire key detection and traffic decryption laying around for quite a while, akin to MFC and MFP. I think on finally finishing that after all of the time.

doegox:
Amazing PR!
* Zlib: it follows the same logic as for our other optional deps for Bluetooth, QT, gd,.. so it's fine with me.
* if you can update the doc/md instructions for the platforms you know about to add zlib + "you can skip..."
 and we'll figure it out for the other platforms.

FTR zlib was in the pm3 code until mid 2020 when we replaced it by Bzip2 and LZ4

kormax:
On both systems I've tested - macOS 26 and Ubuntu 25, ZLIB support was detected and included upon build from the get go.

This may be caused by my system installs not being "fresh" and having had acquired zlib previously.

Nevertheless, I've verified that removing `zlib1g-dev` on Ubuntu causes ZLIB support to be omitted - so I've added it to the docs and into a sample install command.

On macOS, with brew, removing `zlib` didn't cause my install to lack it, with it getting pulled from `xcode-select --install` (`/Library/Developer/CommandLineTools/SDKs/MacOSX26.sdk/usr/include`), which brew usually asks for anyway.
But to be sure, I've added `zlib` to brew install command example, as it's a valid package there and won't hurt.

To avoid guessing with platforms I'm unfamiliar with, no other docs were updated.

doegox:
Thanks! This makes me thinking of sth else: could you add a little self-contained test to tools/pm3_tests.sh that can be executed just with the client and no proxmark?
The goal is to automate compilation tests on the different platforms and make sure zlib was properly detected and usable. (for example the test with "hf waveshare load" checks that libgd is properly detected and used)

If not possible with your existing commands, it might require adding a "test" command, similar to hf mfdes test or hf cipurse test.

Thank you!

kormax:
Yes, I can add such a test.

When running through pm3_tests.sh, If ZLIB is missing, should that test FAIL, return OK, or SKIP somehow?

Update: pushed a test to decompress a real payload sample, if zlib missing from build, it returns "skipped" & PM3_ENOTIMPL.

doegox:
The test must fail.
Thanks

kormax:
The script returns fail if zlib is missing

Issue #2828, opened 2025-04-21
[idea] Implement ST25_TEAROFF in hf 14b wrbl


yashikada:
I see the commit 19789381df5b1e683b3c13f059c9e7972b421c5c, it looks interesting but I would like to try it with the command hf 14b wrbl, looking at the code it seems like it tries to write and then turns off the field after a few uS, so the standalone and flash-only functionality seem like absurd constraints to me.

@SecLabz could you implement this?

SecLabz:
You're right, integrating the tear-off directly into the command would likely be more useful for many cases. However, I honestly didn't feel comfortable enough with the wider codebase to implement it cleanly there myself – hence why I opted for the self-contained standalone mode.

If you feel up to the challenge of integrating it properly and submitting a PR, or even if you just want to experiment, I do have a separate, less polished implementation (https://github.com/SecLabz/proxmark3) that works more like what you described (as a command). 
It's functional (at least on my PM3 Easy), but it would definitely need a good amount of polishing to be ready for a clean merge, which honestly, I probably won't have the bandwidth to do myself.

yashikada:
Thank you for your suggestion and you great work. 
I tested on block 5 where in a sri4k I have the value 46 46 46 46 and I restored the value F5 FF FF FF, but after many tries I never reached FF FF FF FF which was my target.
I see many times FF FF FF FF as green but it never stop.
I tested with counter 6 where I have the value 00 00 00 00 but I see always bits at 0 never a change, is it normal?

I don't know if I can make a PR, but I'll evaluate what to do. 
I think the right way is to put a parameter to the wrbl command, move the tear-off functions into a dedicated c file and then the other parts of the code should use it.

JimmyLaurent:
Getting the counter to exactly 0xFFFFFFFF using the tearing method is tricky because that specific state tends to be unstable. Even if some reads momentarily show 0xFFFFFFFF, it likely won't remain reliably in that state due to how the tearing attack affects the tag's internal workings.

To achieve a stable state after resetting, values slightly lower are often necessary. For instance, reaching 0xFFFFFFFD or below usually involves subsequent operations (like two decrements) that help stabilize the tag's internal counters. The state 0xFFFFFFFF achieved directly via tearing hasn't undergone this stabilization.

[...]

PR #3134, opened 2026-03-14
hf 14b tearoff: speed up tear-off loop timings

xNovyz:
After taking a look at issue #2828 the patch from @yashikada I realized the delays were way too long.
The original optimization removed the BigBuf reinit, assuming DMA state was clean, now we:
- Reinitialize BigBuf memory before FpgaSetupSsc to reset PDC internal buffer pointers
- With properly reset DMA pointers, the field stabilization time can be safely reduced from 5.0ms down to 0.25ms (which matches the timing in the patch)
- Reduced the ST25TB charge capacitor drain wait time from 10ms to an aggressive 0.5ms (the tRST minimum specified in ST25TB datasheet)

I tested on a couple of SRI4K

### DIFF
--- a/armsrc/iso14443b.c
+++ b/armsrc/iso14443b.c
@@ -2653,32 +2653,34 @@ static int tearoff_rand(void) {
 //  - BigBuf_free + BigBuf_calloc (demod buffers persist)
 //  - 100ms field stabilization (tag only needs ~20ms to power up)
 static void tearoff_field_on(void) {
-    // Re-enable reader mode
+    // Reset the AT91 SSC PDC (DMA) receive buffer before each field-on
+    // by mirroring iso14443b_setup() memory allocations for Demod14b/Uart14b.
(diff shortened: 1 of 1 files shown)


iceman1001:
Excellent!

yashikada:
Great improvement.
Is it possible to reset counter 6 to the default value of 0xFFFFFFFF?
I've tried several times, but it never reaches that value.
With counter 5, I managed to reset it to 0xFFFFFFFE. 
With my patch, I also set the value to 0xFFFFFFFF in counter 5. I know this is against the datasheet, it was just a test.

iceman1001:
@yashikada   maybe you should add a PR for your  write counter command?

yashikada:
@iceman1001 My write counter command it's very similar to this PR.

However I managed to set 0xFFFFFFFF to counters 5 and 6, randomly stopping the execution of the 14b tearoff command.

PR #3139, opened 2026-03-17
Allow iclass encoding to SIM, unify wiegand and allow various ways to provide the binary, lotsa tests, etc.  2/2

cindersocket:
iCLASS emulator encode/decode cleanup, long PACS support, and sim stability fixes

This builds on feat-mifare-encode-hid.

This one pulls together the iCLASS emulator work and cleans up the rough edges so the workflow is actually solid.

Main changes:

- Added support for iCLASS DES/NONE encryption.
- Added direct hf iclass encode --emu support so credentials can be written straight to simulator memory.  This required removing legacy cardhopper support in iCLASS pathways.
-  iCLASS emulator encoding uses a bundled default iCLASS card image.
- Removed cardhelper from the iCLASS emulator crypto/decrypt path to stop the weird session instability and stray smartcard traffic.
- Fixed repeated hf iclass decrypt --emu / view --emu usage so the client/session stays alive.
- Added proper PM3-side break handling for iCLASS sim, so pm3 button or pressing <Enter> actually stops simulation cleanly.
- Made hf iclass encode [--emu] print next-step guidance:
    - hf iclass sim -t 3
    - hf iclass eview
    - hf iclass decrypt --emu --ns
- Fixed emulator app-area rewriting so shorter payloads don’t leave stale bits behind from previous writes.
- Reworked legacy iCLASS PACS packing to follow the RE container layout across blocks 7/8/9.  
- Added support for long legacy PACS payloads in iCLASS:
  - full container handling
  - sentinel-aware encode/decode
  - graceful decode behavior above 96 bits (up to 143)
- Kept shared packed Wiegand/HID handling capped where it belongs, but let long iCLASS --bin / --new use a plain binary path cleanly.
- Added a lot of online tests and fixed them when things broke
- Expanded emulator end-to-end coverage for iCLASS crypto:
      - none
      - des
      - 2k3des
- Added long-payload and rewrite regression coverage to the online test script.
- Updated docs, built-in help, cheatsheet notes, and emulator workflow examples.

Validation:

- make -j4 passed
- tools/pm3_online_tests.sh --pm3bin ./pm3 hf_emu_smoke passed

A couple of the fixes here came from chasing ugly emulator/session regressions, so this PR is a mix of feature work and “oye, let’s make this thing behave properly” cleanup.

Sorry for the size, getting iCLASS write to EMU meant a number of changes, at which point I might as well unify wiegand and test the crap out of things.  This is based on the other PR, so if it ends up getting pulled, the commit size will go way down.

### DIFF
--- a/client/src/wiegand_formatutils.c
+++ b/client/src/wiegand_formatutils.c
@@ -359,7 +359,7 @@ int wiegand_pack_from_plain_bin(const char *binstr, wiegand_input_t *input) {
         return res;
     }
 
-    if (input->bin_len > 96) {
+    if (input->bin_len > 84) {
         input->packed_valid = false;
         return PM3_SUCCESS;
     }
@@ -375,7 +375,7 @@ int wiegand_pack_from_new_pacs(const uint8_t *pacs, size_t pacs_len, wiegand_inp
         return res;
     }
 
-    if (input->bin_len > 96) {
+    if (input->bin_len > 84) {
         input->packed_valid = false;
         return PM3_SUCCESS;
     }
--- a/tools/pm3_online_tests.sh
+++ b/tools/pm3_online_tests.sh
@@ -151,7 +151,7 @@ C_FAIL='\xe2\x9d\x8c'
 
 # Check if file exists
 function CheckFileExist() {
-  printf "%-40s" "$1 "
+  printf "%-47s" "$1 "
   if [ -f "$2" ]; then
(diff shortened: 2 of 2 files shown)


iceman1001:
Monster PR's.....  

We do prefer smaller PR's so it is easy to follow the changes.   
This is most likely nice changes ,    

Cut it up in smaller ones, please.

PR #3140, opened 2026-03-17
Add `--timeout` to `hf 15 sim`

recursivenomad:
A simple PR which adds the optional `-t` or `--timeout` to `hf 15 sim`.

Adding this feature has enabled me to perform some automated testing with scripted payload changes, as it resolves the issue of needing to press the physical button after every iteration.

I am new to the codebase, so please let me know if I've missed anything with this contribution.


iceman1001:
Interesting,
looks like an all client side change,  why add a variable to the struct and send it to the device?

Keep a local variable instead and make it cleaner.   Also add parenthesises please 
```
if () {
   my = 0;
} 
```

recursivenomad:
Cool, this was the kind of guidance I was hoping for - I had just mimicked how the other options were written/used since I don't have deep familiarity yet.

Adapting now 👍

recursivenomad:
Confirmed working - Let me know if you prefer rebases.

iceman1001:
and if you look at  the arg_get_int_def,   you see you can set a default int value.   How about -1,  and remove the check / assign later?

recursivenomad:
My thinking was that for the end-user, passing `--timeout 0` or `-1` should make the device simulate (effectively) forever; and my assumption is that fixup would be necessary if the user passes 0, making that check needed regardless.

Would you still like -1 as the default?

iceman1001:
see,    the normal flow is for the simulation to run until key press.    You seem to want to change the default flow.   Which is a no go. 

`--timeout 0 or -1`  is not interesting and should not be used but since we accept it,  it should still just ignored and be the normal flow. 

your timeout is a exception flow.

recursivenomad:
Ah, I think I see what you're saying.  I was worried about passing `0` to `WaitForResponseTimeout` before.  Agreed.

Edits confirmed working again.

PR #3142, opened 2026-03-17
Correct return value type int in EmGetCmd

erland-ca:
The return false would trigger unwanted behaviour as emulation completed with button pressed.


iceman1001:
Good catch!

PR #3144, opened 2026-03-17
Unify Wiegand/HID input handling and expand coverage

cindersocket:
## Summary

  This PR refactors Wiegand/HID input handling around a shared normalization layer and uses it to broaden command support across LF HID and MIFARE Classic HID
  flows.

  ## What changed

  - add shared Wiegand utility helpers to normalize:
      - packed Wiegand messages
      - raw HID transport data
      - new PACS (--new) payloads
      - plain binary Wiegand payloads
  - update lf hid sim and lf hid clone to accept a consistent set of inputs (--raw, --bin, --new, or formatted --wiegand/--fc/--cn)
  - enforce transport-specific limits after normalization, with clearer error messages for oversized credentials
  - update hf mf encodehid to use the same normalized Wiegand payload path and support --raw and --new inputs
  - clean up Wiegand PACS decoding/printing so decoded output shows the logical payload bits instead of intermediate padded binary
  - centralize Wiegand bitstring rendering and PACS decoding helpers to reduce duplicated logic

  ## Tests

  - extend offline tests for Wiegand encode/decode, including a check that wiegand decode --new no longer prints padded binary
  - expand tools/pm3_online_tests.sh with:
      - LF HID clone/readback coverage
      - MIFARE Classic hf mf encodehid round-trip coverage
      - optional manual verification flow for external reader checks

  ## Why

The main goal is to stop each command from implementing its own partial Wiegand parsing and packing rules. Normalizing every input form into one shared representation makes the CLI more consistent, reduces duplicated conversion code, and makes it easier to validate edge cases and add test coverage.

This also lays the foundation for later work on mfc/iclass features and encoding

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 - Added `hf iclass blacktears` command to perform an automated tearoff of block 1 to set non-secure page mode(@antiklesys)
+- Changed wiegand encoding to use shared helpers and have unified parameters (--raw, --bin, --new, --wiegand, etc.) (@cindersocket)
 - Added `hf gst read` command (@kormax)
 - Added `hf gst info` command (@kormax)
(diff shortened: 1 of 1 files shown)


iceman1001:
A large one but it looks solid.   +1

PR #3146, opened 2026-03-18
[hf seos] Increase maximum ADF OID length

aaronjamt:
Some newer MOB/ICE OIDs have gotten quite large. Custom OIDs may also be even larger. I just tried to use my Proxmark with a 20-byte OID, but I've increased the limit to 32 to make sure it won't be a problem in the future.


iceman1001:
Might just as well make  a `#define` instead,    it helps understanding code better

PR #3148, opened 2026-03-18
Fixing SEOS Get Data for longer responses

evildaemond:
Fixing an issue with TLV parsing of the GET Data response, this popped up while someone was looking at other tags on the Zero keyed card by NVX, where the TLV was hard coding the position of the Len, and expecting it to be only 1 byte but it can be more.


iceman1001:
Nice one!

PR #3150, opened 2026-03-18
iclass: add 143-bit legacy PACS decode/encode and DES/PLAIN crypto support

cindersocket:
- Fix legacy PACS payload extraction to accept the full 143-bit width, which is the card’s maximum (oddly encoded
    but valid per iCLASS), so client-side decode (hf iclass view) no longer rejects max-width credentials as
    invalid/missing-sentinel.
- Add physical iCLASS roundtrip coverage that validates both:
    - block readback after write, and
    - client-side decode output.
- Exercise both standard and 143-bit vectors in the roundtrip path.
- Include coverage for DES and unencrypted iCLASS credentials to match actual card behavior.

Note: CodeQL warnings for insecure algorithms are expected here because these card formats can legitimately use DES credentials.

### DIFF
--- a/client/src/cmdhficlass.c
+++ b/client/src/cmdhficlass.c
@@ -6421,14 +6421,12 @@ static int CmdHFiClassEncode(const char *Cmd) {
     CLIParserContext *ctx;
     CLIParserInit(&ctx, "hf iclass encode",
                   "Encode binary wiegand to block 7,8,9\n"
-                  "Use exactly one of --bin, --raw, --new, or --wiegand/--fc/--cn/--issue\n"
-                  "When using emulator you have to first load a credential into emulator memory",
+                  "Use exactly one of --bin, --raw, --new, or --wiegand/--fc/--cn/--issue",
                   "hf iclass encode --bin 10001111100000001010100011 --ki 0            -> FC 31 CN 337 (H10301)\n"
-                  "hf iclass encode --raw 063E02A3 --ki 0                         -> Raw HID input example (H10301)\n"
-                  "hf iclass encode --new 068F80A8C0 --ki 0                         -> ASN.1 PACS input (H10301)\n"
+                  "hf iclass encode --raw 063E02A3 --ki 0                              -> Raw HID input example (H10301)\n"
+                  "hf iclass encode --new 068F80A8C0 --ki 0                            -> ASN.1 PACS input (H10301)\n"
                   "hf iclass encode -w H10301 --fc 31 --cn 337 --ki 0                  -> FC 31 CN 337 (H10301)\n"
-                  "hf iclass encode --bin 10001111100000001010100011 --ki 0 --elite    -> FC 31 CN 337 (H10301), writing w elite key\n"
-                  "hf iclass encode -w H10301 --fc 31 --cn 337 --emu                   -> Writes the ecoded data to emulator memory"
+                  "hf iclass encode --bin 10001111100000001010100011 --ki 0 --elite    -> FC 31 CN 337 (H10301), writing w elite key"
                  );
 
     enum {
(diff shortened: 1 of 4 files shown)


client/src/cmdhficlass.c
```diff
@@ -159,41 +169,83 @@
...
+    mbedtls_des_free(&ctx);
+}
+
+static void iclass_2k3des_block_transform(uint8_t *blk_data, const uint8_t *key, bool encrypt) {
+    mbedtls_des3_context ctx;
+    if (encrypt) {
+        mbedtls_des3_set2key_enc(&ctx, key);
+    } else {
+        mbedtls_des3_set2key_dec(&ctx, key);
+    }
```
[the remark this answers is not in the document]
cindersocket: Yep, iCLASS supports weak crypto.

cindersocket:
Sorry for the commit size.  Almost half of this is tests.

iceman1001:

cindersocket:
> I get a strong sense of AI coding here, is it?

https://github.com/RfidResearchGroup/proxmark3/pull/3155#issuecomment-4108364524

iceman1001:
I need to do proper testing across the board for all wiegand encode/decode,  and  I don't have all my cards available. 

I tried some cards recent and the output was all wrong.   Meaning this PR and others will be thoroughly tested

client/src/cmdhficlass.c
```diff
@@ -160,41 +170,83 @@ static int iclass_load_transport(uint8_t *key, uint8_t n) {
...
-            default:
-                continue;
+        if (i <= 6 || i > 9) {
+            continue;
         }
 
-        if (decrypted_block789) {
-            // Set the 2 last bits of block6 to 0 to mark the data as decrypted
-            dec_data[(6 * PICOPASS_BLOCK_SIZE) + 7] &= 0xFC;
+        if (iclass_apply_transport_mode_to_block(dec_data + idx, key, aa1_encryption, false) == PM3_SUCCESS &&
```
iceman1001, line 180: I think the logic should be reverse.   check aa1_encryption first ,  then do crypto operations

client/src/cmdhficlass.c
```diff
@@ -6326,215 +6456,234 @@ static int CmdHFiClassEncode(const char *Cmd) {
...
+        ENC_ARG_FC = 8,
+        ENC_ARG_CN = 9,
+        ENC_ARG_ISSUE = 10,
+        ENC_ARG_WIEGAND = 11,
+        ENC_ARG_ENC = 12,
+        ENC_ARG_ENCKEY = 13,
+        ENC_ARG_SHALLOW = 14,
+        ENC_ARG_VERBOSE = 15,
+    };
+
```
iceman1001, line 6484: this enum does not help in understanding and maintaining the parameter objects.  
don't use it. 
because you can re-sort or move it around ,  and the argparse doesnt know.

tools/pm3_tests.sh
```diff
@@ -259,6 +259,113 @@ function CheckExecute() {
...
+  local BLOCK6 BLOCK7 BLOCK8 BLOCK9
+  BLOCK6=$(printf "%s\n" "$RES" | sed -n 's/.*Block 6\/0x06 -> \([0-9A-F]*\).*/\1/p' | tail -n1)
+  BLOCK7=$(printf "%s\n" "$RES" | sed -n 's/.*Block 7\/0x07 -> \([0-9A-F]*\).*/\1/p' | tail -n1)
+  BLOCK8=$(printf "%s\n" "$RES" | sed -n 's/.*Block 8\/0x08 -> \([0-9A-F]*\).*/\1/p' | tail -n1)
+  BLOCK9=$(printf "%s\n" "$RES" | sed -n 's/.*Block 9\/0x09 -> \([0-9A-F]*\).*/\1/p' | tail -n1)
+
+  if [ -z "$BLOCK6" ] || [ -z "$BLOCK7" ] || [ -z "$BLOCK8" ] || [ -z "$BLOCK9" ]; then
+    echo "$RES"
+    return 1
+  fi
```
iceman1001, line 326: not sure all of these extra functions,  feels like overkill,   just use hard coded values for your blocks

iceman1001:
ok,   what is going on with both having blocks and binary array and binary strings?   Feels like overcomplication

client/src/cmdhficlass.c
cindersocket: Yes.  It was backwards.

client/src/cmdhficlass.c
cindersocket: Understood.

cindersocket:
> ok, what is going on with both having blocks and binary array and binary strings? Feels like overcomplication

I was trying to be easier to read, but you are write.  Too much abstraction.  I took time to get to know the SourceCode better and to try to do the same.

PR #3153, opened 2026-03-18
[hf 14a sim] Preliminary MFP_WRITEPERSO sim support

aaronjamt:
Building off #3152, adds initial support for fetching personalization keys written to the simulated tag. For now it just logs the values, since there's no storage or anything else implemented yet. Not sure if this is really ready to merge, but I'll leave the final decision up to you.

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -2270,6 +2270,49 @@ void SimulateIso14443aTagEx(uint8_t tagType, uint16_t flags, uint8_t *useruid, u
             EmSendCmd(cmd, sizeof(cmd));
             p_response = NULL;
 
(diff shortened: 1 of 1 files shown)


armsrc/iso14443a.c
```diff
@@ -2270,6 +2270,49 @@ void SimulateIso14443aTagEx(uint8_t tagType, uint16_t flags, uint8_t *useruid, u
...
+                        receivedCmd[10],
+                        receivedCmd[11],
+                        receivedCmd[12],
+                        receivedCmd[13],
+                        receivedCmd[14],
+                        receivedCmd[15],
+                        receivedCmd[16],
+                        receivedCmd[17],
+                        receivedCmd[18]
+                    );
```
iceman1001, line 2302: use the dbhexdump  when printing longer hex string

PR #3155, opened 2026-03-20
Fix long HID prox transport and add prox sim timeout

cindersocket:
There was a prox regression introduced in `c5b3443fe`.

https://github.com/RfidResearchGroup/proxmark3/commit/c5b3443fe63c25caf4f70e5adaa44a24925c2ca9

Before that commit, long HID simulation used an explicit long-format path. In practice, the old code treated long HID as a distinct transport mode and changed how it was sent.

`c5b3443fe` replaced that with a new inference rule for whether HID should be sent as long format:

- it stopped using the explicit long path
- it decided longFMT from packed.Mid > 0xFFF

This caused encodings over 37 bits to fail silently.  This is why I added the earlier gate to prevent their use.

This restores the old way of doing things, adds some testing, and adds a timeout to hf sim to support that testing.

### DIFF
--- a/client/src/cmdlfhid.c
+++ b/client/src/cmdlfhid.c
@@ -59,23 +59,13 @@ typedef struct {
     int new_pacs_len;
 } lf_hid_cli_input_t;
 
-// Enforce the narrower LF HID transport limits after the shared Wiegand layer has
-// normalized whichever user-facing input mode was selected.
+// Reject credentials that exceed the absolute 84-bit packed HID transport limit.
+// The shared Wiegand layer can normalize credentials that are wider than the LF HID
+// transport, so check packed_valid here rather than relying on the caller.
 static int lf_hid_validate_packed_transport(const wiegand_input_t *input, const char *command_name) {
-    (void)command_name;
-
-    // The shared Wiegand layer can normalize credentials that are wider than the LF HID
-    // transport. Reject only when this specific command needs a packed HID frame.
     if (input->packed_valid == false) {
-        PrintAndLogEx(ERR, "Credential encoded successfully, but %" PRIuMAX "-bit Wiegand data cannot be represented as a packed HID credential", (uintmax_t)input->bin_len);
-        PrintAndLogEx(ERR, "Packed HID transport supports up to 84 bits total");
+        PrintAndLogEx(ERR, "%s: %" PRIuMAX "-bit credential exceeds the 84-bit packed HID transport limit", command_name, (uintmax_t)input->bin_len);
         return PM3_EINVARG;
-    }
-
-    // Raw HID input already arrives in transport form and intentionally bypasses the
-    // packed Wiegand length check that applies to bin/new/formatted inputs.
-    if (input->packed.Length == 0) {
-        return PM3_SUCCESS;
     }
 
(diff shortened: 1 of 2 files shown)


iceman1001:

cindersocket:
> I get a strong sense of AI coding here, is it?

I understand why it might look AI-generated, but mostly no. Some AI was involved, yes, but not in the way people usually mean. My goal is to make every type of RFID fob work properly on the Proxmark3, and AI is not good enough for that kind of work. It makes too many mistakes, and you cannot trust it on its own.  It can answer questions and be a useful guide sometimes, especially when trying to answer questions like where certain code is, or asking about a new codebase to you.

I wrote firmware to connect a Raspberry Pi to Wiegand and USB, then hooked a HID reader up to it. I also had ChatGPT help test the firmware by working backward until it found where things were breaking, using the USB side to confirm whether it was still working.

I wrote the test tools for reader-to-PM3 and PM3-to-PM3 communication so I could verify that things actually worked. I also wrote the fix itself. My editor has Copilot, so I do get inline completion suggestions, and I used JetBrains AI Assistant to help suggest documentation text. English is not my first language, and my writing is not very strong, so I use AI for proofreading. I also check and correct what it gives me.

For MIFARE and iCLASS Legacy, I fuzz readers to see what breaks, then I investigate by hand, sniff traffic, and experiment to figure out how they work. I also reproduced the Heart of Darkness attack and used Ghidra, radare2, Binary Ninja, and IDA to help understand how Legacy works internally. New readers use the SAM for MIFARE Legacy, so the behavior I am targeting has to come from older systems.

There are a lot of hidden details in HID Prox: ASK, FSK, autodetection, built-in formats, OEM codes, and long versus short variants. I want all of them to work, every type. T55xx detection was broken. I fixed that and made encode and decode line up properly. I worked on making more things use the emulator to help with testing, to find more edge cases.

I care a lot about tests because I do not want fixes to break something else later. In my day job, if it is not tested, it is not really fixed. So I write tests for everything: one-way, two-way, back and forth.

I read every line of code. I write most of the code myself. I use AI to help with tests, to improve writing, and sometimes to speed up reverse engineering, but I do not trust it, and you should not either.

client/src/cmdlfhid.c
```diff
@@ -62,11 +62,13 @@ typedef struct {
 // Enforce the narrower LF HID transport limits after the shared Wiegand layer has
 // normalized whichever user-facing input mode was selected.
 static int lf_hid_validate_packed_transport(const wiegand_input_t *input, const char *command_name) {
+    (void)command_name;
```
cindersocket: This suppresses compiler warnings for an Unused Parameter.

iceman1001: if its unused,   why is it there to begin with?

cindersocket: It was used, but the limit wasn't 37 bit anymore.  I was peeling off pieces of an earlier too large commit.  Corrected.

iceman1001:
ok,    this PR has too much in one go.  
Split it up in 4  PR,s

one if for timeout sim
one for wiegand additions
one for 143 bit fixes 
one for new test harness

Issue #3041, opened 2025-12-03
Buglet in new QR code functionality


nneul:
Not a full bug report, just passing along a mini one since I tried it out after nerding out over your commit message when I pulled latest. 

**Describe the bug**

Invalid (non parsable but looks like a qr code) resulting image and no error report complaining about the data being invalid/wrong format/etc.: 

```data qrcode -d https://some.url.com ```

Yes, I know that isn't valid based on reading the usage afterwards, who RTFM's *before* trying stuff at random? :)

Seg fault:   
```data qrcode -d aa```

nneul:
Works fine with multiple hex bytes. (i.e. not giving garbage input) - but the single byte segfault was unexpected.

PR #3156, opened 2026-03-20
Fixes RfidResearchGroup/proxmark3#3041

cindersocket:
- Fixes RfidResearchGroup/proxmark3#3041
- Adds --ascii option to qr code generation to avoid the use of unicode for terminals that don't support it


cindersocket:

Ain't worth it, jaja.

iceman1001:

PR #3159, opened 2026-03-20
Seos improvements

kormax:
This PR adds an ability to specify a custom `--aid` for `hf seos` commands, which is useful when testing some of custom Seos implementations - E.G. HID partner provider HCE applications, custom Seos emulator apps, etc.

During the testing, I found that the current ADF SELECT response code does not properly validate the presence of all tags, which lead to memory-related crashes, so I also added validation of presence of the required data.

Additionally, I found that ADF SELECT command was issued with a truncated/missing LE, which caused some Seos implementations to think that the reader did not expect any response data, causing them to return `9000` and/or expect to return leftover data through chaining. This was fixed by ensuring that `le` isn't truncated.

After the changes, my `2K3DES_CBC_MODE + SHA-1` and `AES-128_CBC_MODE + SHA-256` cards programmed with a ZERO-ADF seem to work fine with all commands, including a credential which uses a custom AID, but I invite to verify that I didn't break anything, as I may not have enough test vectors.

### DIFF
--- a/client/src/cmdhfseos.c
+++ b/client/src/cmdhfseos.c
@@ -101,8 +101,8 @@ static const known_algo_t known_algorithm_map[] = {
 };
 
 static const char *known_seos_aids[] = {
-    "A0000004400001010001",
-    "A000000382002D000101",
+    "A0000004400001010001",   // STANDARD_SEOS
+    "A000000382002D000101",   // MOBILE_SEOS_ADMIN_CARD
 };
 
 static int seos_get_custom_aid(CLIParserContext *ctx, int arg_index, uint8_t *aid, int *aid_len) {
@@ -1592,10 +1592,6 @@ static int seos_load_keys(char *filename) {
     free(dump);
     PrintAndLogEx(SUCCESS, "Loaded" _GREEN_("%2zd") " keys from %s", i, filename);
     return PM3_SUCCESS;
-}
-
-int infoSeos(bool verbose) {
-    return seos_select(NULL, 0);
 }
 
 static int CmdHfSeosInfo(const char *Cmd) {
--- a/client/src/cmdhfseos.h
+++ b/client/src/cmdhfseos.h
@@ -22,7 +22,6 @@
 #include "common.h"
 #include "seos_cmd.h"
 
-int infoSeos(bool verbose);
 int CmdHFSeos(const char *Cmd);
 int seos_kdf(bool encryption, uint8_t *masterKey, uint8_t keyslot,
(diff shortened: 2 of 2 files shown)


iceman1001:
Cool,   two major PR's for seos,   and too little time.   @nvx    do you have time to look into this one?

nvx:
> Cool,   two major PR's for seos,   and too little time.   @nvx    do you have time to look into this one?  

Will have a squiz either tonight or tomorrow when I've finished catching up and unpacking

client/src/cmdhfseos.c
```diff
@@ -99,6 +100,19 @@ static const known_algo_t known_algorithm_map[] = {
     {9, "AES-128_CBC_MODE"},
 };
 
+static const char *known_seos_aids[] = {
+    "A0000004400001010001",
+    "A000000382002D000101",
```
nvx: These AIDs were labelled previously, would be worth throwing in a comment at least so the information about what each AID is retained

client/src/cmdhfseos.c
```diff
@@ -1526,7 +1595,7 @@ static int seos_load_keys(char *filename) {
 }
 
 int infoSeos(bool verbose) {
```
nvx: it doesn't look like this is still used anymore now that `CmdHfSeosInfo` calls `seos_select` directly to be able to pass in the custom aid

that said, I notice this function is exported while `seos_select` is not, if it does make sense for this to be exposed for eg standalone modes would it make more sense to expose `seos_select` instead? personally I'd be more inclined to just remove it and if/when it's needed on eg a standalone mode export it then but maybe @iceman1001 has opinions on this front

client/src/cmdhfseos.c
kormax: I've re-added names as comments.

client/src/cmdhfseos.c
kormax: Function was dropped altogether

iceman1001:
ok,  so I am a fan of not removing information,   like the desc or names of aid in a static hard coded list.  
rule # 1,  

If a standalone more needs something,  they can update and fix it when that day come.   The SEOS code was kind of rushed together,    so `hf 14a simaid`   and `hf seos`   has a bunch of code which should been refactored and nice made.

if something isn't used anymore,    usally with info is exposed, it is call from `hf search`  or somewhere else to bring information to other commands.    Just make a double check if its called,   or if it should be called  like for `hf search`...  

dont be too worried about the repo being modified,  its bleeding edge for a reason.   releases are stable,   lastest master is hacker place

iceman1001:
I also do like having a name/description  printed when doing things.   Looking at just an AID is after a couple of thousand times,  annoying to look up or memorise.

iceman1001:
Merging for now and then the rest of the suggestions can be in a new PR

PR #3162, opened 2026-03-22
Add more AID values to 'aidlist.json'

kormax:
This PR adds some new AID values encountered on mobile devices with HCE.

I'd like to add some more encountered AID values, I.E. some management AID values in reader configuration apps, but I am currently concerned with polluting the list too much, causing AID brute force and other users of this file to slow down.

As an option to allow expansion of the list, I'd like to suggest adding "tags" field, which would allow to specify additional markers for an AID entry:
- If it's available on `hce` capable devices or `android`.
- Protocol names (I.E. DESFire-related AIDs can be marked as `desfire`)
- If it's available over `contactless` or `contact`.
- If AID is not for a credential, but for a `management`/`technical` app;
- If AID is known, but not usable from outside, like the SEOS file system applet AID. 

I'm suggesting that those markers are to be used downstream to filter-out unneeded entries, refine brute-force lists, provide suggestions, etc.
Perhaps, those "tags" can be expressed as separate fields by themselves, but I'm throwing this idea in just in case, I'd be fine with any implementation.

Additionally, for `android` or `hce` AID values, it could be valuable to store application package names. I.E, that could allow to find out and show which apps the user has, or serve as a general store of info on what apps emulate what kind of credentials.

As a more general use field, we may add "sources" field, which would store both the app package names, but also attribution to web sites (If an AID was taken from a public source).

### DIFF
--- a/client/resources/aidlist.json
+++ b/client/resources/aidlist.json
@@ -2448,11 +2448,19 @@
         "Type": "access"
     },
     {
+        "AID": "A0000004400001010001000002",
+        "Vendor": "HID Global",
+        "Country": "",
+        "Name": "SEOS Mobile",
+        "Description": "Declared by some SEOS-compatible HID partner applications for HCE",
+        "Type": "access"
+    },
+    {
         "AID": "A00000054000060100010000FF",
         "Vendor": "HID Global",
         "Country": "",
         "Name": "SEOS Mobile",
-        "Description": "Declared by some SEOS-compatible applications for HCE",
+        "Description": "Declared by some SEOS-compatible HID partner applications for HCE",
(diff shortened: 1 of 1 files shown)


iceman1001:
expanding the data stored in each object,  sure,   all fields you recommend are good ones to have.
dont worry about the size of the list.   

But if we have more fields,  ww need a new `note`   for /docs   where we explain the purpose of the list,  and also give examples where and how to collect the correct data for each field.   

We have many issues with endianness and descriptions,  where multiple vendors use same AID.  

How the list is to be used,  is a later question.    Collecting all necessary data is the first focus

PR #3164, opened 2026-03-23
Add a simple doc for 'aidlist.json'

kormax:
To get a feel for potential aidlist doc, I'm PRing it separately, before expanding the format for AIDlist with newer fields. 

Please tell if there is anything else worth pointing out in the file, or if It's OK as-is.


iceman1001:
its a great start,  and when we add more fields  this is a great place for why and samples :)

PR #3165, opened 2026-03-23
Added script mfuev1_counter_reset.py

W0rthlessS0ul:
This PR adds script mfuev1_counter_reset.py - script bypasses the Anti-Tearing protection on MIFARE Ultralight EV1 monotonic counters and allows resetting the counter value
The script is based on the research published by [Quarkslab](https://blog.quarkslab.com/rfid-monotonic-counter-anti-tearing-defeated.html)
An example of how the script works is shown in the video below:

https://github.com/user-attachments/assets/4c9d91df-9e55-425b-a0aa-595cfe3d7d8e

### DIFF
--- a/client/pyscripts/mfuev1_counter_reset.py
+++ b/client/pyscripts/mfuev1_counter_reset.py
@@ -1,0 +1,224 @@
+#!/usr/bin/env python3
+
+#-----------------------------------------------------------------------------
+# Copyright (C) Proxmark3 contributors. See AUTHORS.md for details.
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


iceman1001:
Well done!

Now you can adapt  `hf mfu tear`  ->  its commented out code but should be enough skeleton for you to apply your technique and voilá  we have a native command  too :)

W0rthlessS0ul:
> Now you can adapt hf mfu tear -> its commented out code but should be enough skeleton for you to apply your technique and voilá we have a native command too :)

Ok, I'll try to do this

PR #3176, opened 2026-03-26
Added `hf iclass tagsim` command and `--live` `hf iclass lookup`

Antiklesys:
- Added `hf iclass simtag` command to quickly simulate an iclass tag based on facility code and card number
- Added `--live` option to `hf iclass lookup` command to perform a live recovery of the reader's key by simulating a tag and running the lookup command against both standard and elite dictionaries 
- Improved SIO Parsing in `hf iclass view`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Improved SIO parsing for `hf iclass view` based on Iceman's "Dismantling the SEOS Protocol" talk (@antiklesys)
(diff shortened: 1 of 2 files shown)


iceman1001:
Not sure why you don't use a large empty iclass dump file as base for all changes,  which you then eload up to the device once you applied your changes.

You can have an empty legacy,  legacy elite,  or  SR,   or SE ...

I think there are some in  `traces/iclass`

Antiklesys:
> Not sure why you don't use a large empty iclass dump file as base for all changes,  which you then eload up to the device once you applied your changes.
> 
> 
> 
> You can have an empty legacy,  legacy elite,  or  SR,   or SE ...
> 
> 
> 
> I think there are some in  `traces/iclass`   
> 
> 
> 
> 

Not sure on why would I do that?
Both these commands would only work on legacy anyway, what's the point of loading any different format and changing it?

Antiklesys:
@iceman1001 ready to merge

PR #3177, opened 2026-03-26
Remove unused ISO14A_COMMAND flags

kormax:
This PR removes legacy `ISO14A_COMMAND` flags that are not used anymore.

I've also noticed a weird `ISO14A_REQUEST_TRIGGER` flag that seems to do nothing useful but cause LED to turn on, but I haven't removed it, as I may not understand its deeper purpose... Ready to remove it too do if I'm correct here.


iceman1001:
request trigger comes from way back when there was more researchers using the device. 
you needed a trigger signal when to start capturing for your logic analyser or oscilloscope.

you have those kinds of things laying around in the device side.     Easy to hook to the led trace to get a trigger signal

Issue #3175, opened 2026-03-25
Null dereference in `mifare_classic_authex_cmd`


nffq:
**Describe the bug**
Dereferencing `uint32_t *ntencptr` without null checks

**To Reproduce**
Run `hf mf isen --addauth` on a Mifare Classic
should reveal `auth nested ... ntenc: ea000006`

**Expected behavior**
Should display the correct nested nt

**Screenshots**
<img width="739" height="72" alt="Image" src="https://github.com/user-attachments/assets/17b8affa-2116-4270-9a64-2cd220e9c1b5" />

<img width="1063" height="309" alt="Image" src="https://github.com/user-attachments/assets/6f2e9950-5f75-438a-8d5f-1386ceb8c983" />

**Desktop (please complete the following information):**
 - OS: Kali

```
[ Proxmark3 ]

 [ Client ]
  Iceman/master/v4.21128-suspect 2026-02-25 16:15:01 ddaba4e24
  Compiler.................. GCC 14.2.0
  Platform.................. Linux / x86_64
  Readline support.......... present
  QT GUI support............ absent
  Native BT support......... absent
  Python script support..... present ( 3.13.9 )
  Python SWIG support....... present
  Lua script support........ present ( 5.4.7 )
  Lua SWIG support.......... present

 [ Model ]
  Firmware.................. PM3 GENERIC

 [ ARM ]
  Bootrom.... Iceman/master/v4.21128-suspect 2026-02-25 16:15:01 ddaba4e24
  OS......... Iceman/master/v4.21128-suspect 2026-02-25 16:15:01 ddaba4e24
  Compiler... GCC 14.2.1 20241119

 [ FPGA ]
 fpga_pm3_hf.ncd image 2s30vq100 14-03-2026 01:39:46
 fpga_pm3_lf.ncd image 2s30vq100 14-03-2026 01:39:46
 fpga_pm3_felica.ncd image 2s30vq100 14-03-2026 01:39:46
 fpga_pm3_hf_15.ncd image 2s30vq100 14-03-2026 01:39:46

 [ Hardware ]
  --= uC: AT91SAM7S512 Rev A
  --= Embedded Processor: ARM7TDMI
  --= Internal SRAM size: 64K bytes
  --= Architecture identifier: AT91SAM7Sxx Series
  --= Embedded flash memory 512K bytes ( 71% used )
```

iceman1001:
[...]

PR #3180, opened 2026-03-27
Allow debug prints when `isNested == AUTH_NESTED` and `ntencptr == NULL`

nffq:
Extension of #3175


iceman1001:
@doegox   this is one for you

PR #3225, opened 2026-04-09
Polish the `hf mf acl` command

team-orangeBlue:


team-orangeBlue:
oops the enter key actually submits the request

Anyhow:
* user now must enter 3 bytes of ACLs
* ACLs like 7F0788 and 787788 are also recognized (you judge the names)

Obviously something like `00F0FF`/`00f87f` "killer" configs could be labelled but I don't see much of a point

iceman1001:
should have some generic check like "hf mf acl"  for invalid ACL.

the three of four bytes,   meh,    if four,   it should be printed as GBD (etc if its an indication for MAD etc) 

or like in your mfp acl,   it should indicate the encrypted/plain etc...

PR #3229, opened 2026-04-10
Add ST check to `hf mfp wrbl`, fix ST RO checks for 16-block sectors

team-orangeBlue:
Two relevant changes:

1. When using `hf mfp wrbl` and providing a block that is a sector trailer, its' ACL will be checked for validity. Code is purely copy-pasted from regular mifare + my previous commits.
2. The ST read-only checks performed by the function invoked in `hf mf wrbl` had a bug where providing them with a 16-block sector number would make them bug out and instead treat the sector as if it was 4 blocks. Now the errors display block ranges as explained in the ICS70 datasheet


client/src/cmdhfmfp.c
```diff
@@ -1180,6 +1180,49 @@ static int CmdHFMFPRdsc(const char *Cmd) {
...
+            else
+                PrintAndLogEx(WARNING, "Strict ReadOnly Access Conditions on blocks " _YELLOW_("%u-%u") " detected", blockno - bar*4 + 1 + foo*5, blockno - bar*4 + 1 + foo*5 + 4);
+            ro_detected = true;
+        }
+    }
+    if (ro_detected) {
+        if (force) {
+            PrintAndLogEx(WARNING, " --force override, continuing...");
+        } else {
+            PrintAndLogEx(INFO, "Exiting, please run `" _YELLOW_("hf mf acl -d %s") "` to understand", sprint_hex_inrow(&block[6], 3));
```
iceman1001, line 1215: hf mfp acl  ?

team-orangeBlue: The ST checker will print out the regular 3 bytes of MFC ACLs should there be an issue.

MFP and MFC process them the same. Asking the user to use the other variant won't change much.

I would put in `hf mfp acl`, but the only use case I can come up with here is if we did not get encrypted exchange support 3 years ago and if the user tried to make some blocks encrypted only. So I don't quite see the point in changing that.

PR #3237, opened 2026-04-13
Satisfy gst for cppcheck

kormax:
As requested by @doegox, I've added zero-init for a bunch of GST structs that caused warnings.

Additionally, I've fixed a mistake made in CHANGELOG when other contributors were resolving merge issues.

While on that point, I think it's worth discussing the merit of CHANGELOG file.
In my opinion, its existence, alongside the bot nagging contributors to update it, leads to unnecessary merge conflicts, which in turn cause mistakes and dropped lines when no care is taken when resolving the said conflicts.
The intended could be replicated with git history.


doegox:
Thank you!
the CHANGELOG is only to give a high level overview shared as release notes, no need to update it especially when fixing things not yet released.

PR #3241, opened 2026-04-16
Fix to improve/fix circular buffer behavior/issues

Antiklesys:


iceman1001:
See if this helps the sniffing

PR #3251, opened 2026-04-19
Fix overflow in 14a to send buffer

jrozner:
Currently there is no bounds checking on the to send buffer which results in larger requests overflowing the buffer and corrupting the CRC. This fixes that by adding bounds checking. It also increases DYNAMIC_MODULATION_BUFFER2_SIZE to support frames up to 256 bytes


iceman1001:
good catch!   

this might fix the emrtd problems we had

Issue #3250, opened 2026-04-19
Build fail


Dramelac:
**Describe the bug**
The code fail during build:
```
# make
[...]
In file included from /usr/lib/gcc/x86_64-linux-gnu/12/include/immintrin.h:49,
                 from src/loclass/cipher_bs_avx512.c:20:
In function ‘_mm512_andnot_si512’,
    inlined from ‘doMAC_brute_match512’ at src/loclass/cipher_bs_avx512.c:225:21:
/usr/lib/gcc/x86_64-linux-gnu/12/include/avx512fintrin.h:7845:20: error: ‘__Y’ is used uninitialized [-Werror=uninitialized]
 7845 |   return (__m512i) __builtin_ia32_pandnd512_mask ((__v16si) __A,
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 7846 |                                                   (__v16si) __B,
      |                                                   ~~~~~~~~~~~~~~
 7847 |                                                   (__v16si)
      |                                                   ~~~~~~~~~
 7848 |                                                   _mm512_undefined_epi32 (),
      |                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~
 7849 |                                                   (__mmask16) -1);
      |                                                   ~~~~~~~~~~~~~~~
/usr/lib/gcc/x86_64-linux-gnu/12/include/avx512fintrin.h: In function ‘doMAC_brute_match512’:
/usr/lib/gcc/x86_64-linux-gnu/12/include/avx512fintrin.h:206:11: note: ‘__Y’ was declared here
  206 |   __m512i __Y = __Y;
      |           ^~~
cc1: all warnings being treated as errors
make[2]: *** [Makefile:1112: obj/loclass/cipher_bs_avx512.o] Error 1
make[1]: *** [Makefile:185: client/all] Error 2
make[1]: Leaving directory '/tmp/proxmark3'
make: *** [Makefile:40: all] Error 1
```

**To Reproduce**
1. `git clone --depth 1 https://github.com/RfidResearchGroup/proxmark3.git`
2. `cd proxmark3`
3. `make clean && make`

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux Debian 12 AMD64

Dramelac:
For testing purpose, I rollback to an older version: 
```
git checkout d6c823bea38cab98e2ce983e4ec1db0ce6caf4bc
make all
[...]

PR #3253, opened 2026-04-20
Fix for issue https://github.com/RfidResearchGroup/proxmark3/issues/3250

Antiklesys:
The root cause: a GCC intrinsic macro uses an uninitialized variable by design Intel's AVX-512 intrinsics in GCC are implemented as inline wrappers around __builtin_ia32_* builtins. Some of those builtins take a "merge source" operand that the user-facing intrinsic doesn't expose. For _mm512_andnot_si512, GCC's implementation in avx512fintrin.h looks like:

return (__m512i) __builtin_ia32_pandnd512_mask ((__v16si) __A,
                                                 (__v16si) __B,
                                                 (__v16si) _mm512_undefined_epi32 (),
                                                 (__mmask16) -1);
The mask is -1 (all lanes written), so the "merge source" is discarded — that's why GCC passes _mm512_undefined_epi32(), which is literally defined as:

__m512i __Y = __Y;   // intentionally self-initialized, value is "don't care"
return __Y;

----------------

The line __m512i __Y = __Y; lives in GCC's own header:

/usr/lib/gcc/x86_64-linux-gnu/12/include/avx512fintrin.h That file belongs to the distro's gcc-12 package. Editing it would:

Only fix this one developer's machine.
Get silently reverted the next time apt upgrade runs. Cause divergence from every other build host / CI runner. Not help anyone else who clones the repo.
So patching the header in-place is a non-starter for a committed fix.

-----------------------
Implemented solution:

Silence only the false positive around the call — add a diagnostic pragma right at the intrinsic site so the rest of the file keeps full warnings.

### DIFF
--- a/client/src/loclass/cipher_bs_avx512.c
+++ b/client/src/loclass/cipher_bs_avx512.c
@@ -222,7 +222,14 @@ void doMAC_brute_match512(const uint64_t y_bits_bs[96 * BS512_WORDS],
     __m512i mac_match = BS_ONES;
     for (int tick = 0; tick < 32; tick++) {
         const __m512i diff = _mm512_xor_si512(r[2], _mm512_loadu_si512((const void *)&target_mac_bs[tick * BS512_WORDS]));
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wuninitialized"
+#endif
         mac_match = _mm512_andnot_si512(diff, mac_match);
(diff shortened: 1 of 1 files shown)


iceman1001:
```
mac_match = _mm512_andnot_si512(diff, mac_match);
```

maybe verify that two parameters is using initialised variables when calling?

iceman1001:
yeah,  the issue seems to stem with the end user's dev env.   

I will close this one since I dislike silence messages inlined in source code. 

I think the user needs to check there compile flags,  if they have overridden anything or try updating their gcc.

PR #3274, opened 2026-04-27
Update mifaredesfire.c to support DESELECT response

gentilkiwi:
Even if I don't see any reason to have the `GetVersion` command in `ARM` codebase (no time sensitive - can be moved in the client part), this PR fix the incorrect `HLTA` command sent during `Deselect` reponse

We had:

_`HLTA` command sent during `Deselect` reponse_

After the fix:

_Only `Deselect` response, as `HLTA` is not needed_

### DIFF
--- a/armsrc/mifaredesfire.c
+++ b/armsrc/mifaredesfire.c
@@ -756,11 +756,19 @@ size_t CreateAPDU(uint8_t *datain, size_t len, uint8_t *dataout) {
 // uint32_t crc = crc_finish(&desfire_crc32);
 
 void OnSuccess(void) {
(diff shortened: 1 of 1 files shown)


iceman1001:
that is something from like a decade ago...       don't be hard on my younger self,    the way and which / what is going to execute on device side or client side wasn't that crystal clear back then.

But its a great fix!

gentilkiwi:
Haha, if it's still have a beard, it's the same @iceman1001 ;)

Not a critiscim, more a 'why' I made this fix like this, even if I think the real one shoud be outside this part :)

PR #3276, opened 2026-04-28
DUOX-related cleanup

kormax:
This PR makes the following changes:
* Use public key stored in client resources instead of hard-coding it;
* Add and use `pcrypto_rng_fill_oneshot` which does not require to init and free `rng`;
* Other minor code adjustments.


iceman1001:
Not so sure we need that kind of random....   we prefer bad random numbers in this project :)

PR #3288, opened 2026-05-04
DESFire EV3C support

suut:
This is the first pull request to add EV3C support, it's initially marked as draft as it's going to be the target of other PRs to add the rest of the support

This first PR adds the command `hf mfdes makemfclicense`

### DIFF
--- a/client/src/cmdhfmfdes.c
+++ b/client/src/cmdhfmfdes.c
@@ -69,6 +69,10 @@
 #define MFDES_PC_MAX_ROUNDS             8U
 #define MFDES_PC_MAC_LEN                8U
 
+#define MFDES_EV3C_MFC_KILL_KEY                ((uint8_t)0x31U)
+#define MFDES_EV3C_ALLOWED_DATA_PERMISSIONS    ((uint8_t)0x11)
+#define MFDES_EV3C_ALLOWED_TRAILER_PERMISSIONS ((uint8_t)0x1F)
+
 // DUOX ISO Internal Authenticate
 #define DUOX_INTAUTH_CHALLENGE_LEN      16
 #define DUOX_INTAUTH_SIG_LEN            64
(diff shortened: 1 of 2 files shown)


client/src/cmdhfmfdes.c
```diff
@@ -70,6 +70,10 @@
 #define MFDES_PC_MAX_ROUNDS             8U
 #define MFDES_PC_MAC_LEN                8U
 
+#define MFDES_EV3C_MFC_KILL_KEY                ((uint8_t)0x31U)
+#define MFDES_EV3C_ALLOWED_DATA_PERMISSIONS    ((uint8_t)0x11)
+#define MFDES_EV3C_ALLOWED_TRAILER_PERMISSIONS ((uint8_t)0x1F)
```
iceman1001: any particular reason for typecasting a DEFINE?

iceman1001:
You have a strange `--force`  in the `hf mfdes auth`  where you are now mixing two commands.   One is for authentication and one is for configuration (aka kill MFC support).     Don't mix.

kormax:
> You have a strange `--force` in the `hf mfdes auth` where you are now mixing two commands. One is for authentication and one is for configuration (aka kill MFC support). Don't mix.

From what @suut discovered, just authenticating to MFC by itself is enough to disable functionality on the card. So they're adding a safeguard to auth command to prevent users from accidentally doing it.

client/src/cmdhfmfdes.c
suut: Ah just habit, I tend to write it like that to be safe but for integers it's not necessary indeed

suut:
> You have a strange `--force` in the `hf mfdes auth` where you are now mixing two commands. One is for authentication and one is for configuration (aka kill MFC support). Don't mix.

Yes so that's how you kill the MFC support on EV3C as weird as it sounds, you authenticate at PICC level with the AES key 0x31 with AuthenticateEV2First and that permanently disables the MFC functionality

iceman1001:
Current logic:     ` if (dctx.keyNum == MFDES_EV3C_MFC_KILL_KEY && !force)`

mentioned logic:  `you authenticate at PICC level with the AES key 0x31 with AuthenticateEV2First`  

I guess the implemented logic is a bit wider than what is actually needed?   PICC level and AuthenticateEV2First seems missing.

iceman1001:
since this is marked as wip,  you need to let me know when you are considering it ready

suut:
I marked it as ready, it's easier if we review and merge this first so I can create the other merge requests for the other commands

> I guess the implemented logic is a bit wider than what is actually needed? PICC level and AuthenticateEV2First seems missing.

Yeah indeed I think I will remove the check, authenticating at PICC level with EV2 secure messaging with key 0x31 is so specific that it won't happen by accident so I'll just remove it and just document the behavior

iceman1001:
Yeah,  document it is the way forward.   Make a note for the tag in `/docs`

Issue #2904, opened 2025-06-14
Failed to clone EM4102 to Hitag S/82xx tag after first successful write on F8310 tag


sleepwalkera:
**Describe the bug**
When cloning EM4102 to Hitag S/82xx (F8310) tags, the first write succeeds but subsequent writes to the same tag fail with "Something went wrong in step 0". The tag becomes effectively read-only after initial programming.
```[usb] pm3 --> lf em 410x clone --id 0A004EEC71 --hts
[+] Preparing to clone EM4102 to Hitag S/82xx tag with EM Tag ID 0A004EEC71 (RF/64)
[=] Encoded to FF 82 80 02 7B DC 3C 68 
[!] ⚠️  Something went wrong in step 0
```

**To Reproduce**
1. Place fresh F8310 tag on antenna
2. Run: `lf em 410x clone --id [SOME_ID] --hts` (success)
3. Run: `lf em 410x clone --id [ANY_NEW_ID] --hts` (fails)
4. Observe step 0 error on subsequent attempts

**Expected behavior**
The command should allow rewriting the same Hitag S/82xx tag multiple times with different EM4102 IDs, just like the initial successful write.

**Additional context**
Tested with latest v4.20469
Tag remains readable after failed writes (retains first-written ID)
```
[usb] pm3 --> lf search

[=] Note: False Positives ARE possible
[=] 
[=] Checking for known tags...
[=] 
[+] EM 410x ID 0F0368568B
[+] EM410x ( RF/64 )
[=] -------- Possible de-scramble patterns ---------
[+] Unique TAG ID      : F0C0166AD1
[=] HoneyWell IdentKey
[+]     DEZ 8          : 06837899
[+]     DEZ 10         : 0057169547
[+]     DEZ 5.5        : 00872.22155
[+]     DEZ 3.5A       : 015.22155
[+]     DEZ 3.5B       : 003.22155
[+]     DEZ 3.5C       : 104.22155
[+]     DEZ 14/IK2     : 00064481678987
[+]     DEZ 15/IK3     : 001034014845649
[+]     DEZ 20/ZK      : 15001200010606101301
[=] 
[+] Other              : 22155_104_06837899
[+] Pattern Paxton     : 259822731 [0xF7C948B]
[+] Pattern 1          : 9750181 [0x94C6A5]
[+] Pattern Sebury     : 22155 104 6837899  [0x568B 0x68 0x68568B]
[+] VD / ID            : 015 / 0057169547
[...]

PR #3290, opened 2026-05-06
Fix Hitag S / 82xx cloning and compatibility

sleepwalkera:
## Summary
Fixes an issue where cloning EM410x to Hitag S / 82xx tags (like F8310) would fail on subsequent attempts with "Something went wrong in step 0".

## Changes
- **Write sequence**: Reordered to write data (Pages 4/5) before configuration (Page 1). The current logic enables TTF mode first, and the resulting broadcast likely interferes with the remaining write commands.
- **Handshake**: Switched from FADV to ADV1. Some 82xx clones don't seem to respond to FADV once they've been programmed, while ADV1 appears more reliable.
- **Config header**: Adjusted RES/auth bits to produce a 0xDA header instead of 0xCA. This matches the "standard" configuration used by most standalone 82xx duplicators and fixes issues where tags written by PM3 weren't recognized by other readers.

## Test plan
Tested with F8310 tags. They are now successfully rewritable and work across different 82xx devices.

### DIFF
--- a/client/src/cmdlfem410x.c
+++ b/client/src/cmdlfem410x.c
@@ -712,14 +712,25 @@ static int CmdEM410xClone(const char *Cmd) {
 
             switch (step) {
                 case 0: {
+                    memcpy(packet.data, &data[HITAGS_PAGE_SIZE * 0], HITAGS_PAGE_SIZE);
+                    packet.page = 4;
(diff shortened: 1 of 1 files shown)


sleepwalkera:
Fixes #2904

PR #3292, opened 2026-05-07
Some improvements to a couple lua scripts for Legic. 

YoungJules:
Improved help, made the purpose (of the 'clone' script) clearer, added json dump support, better support for blank cards, look for KGH (Kaba Group Header) etc.


iceman1001:
Good good!

PR #3294, opened 2026-05-07
add native LEGIC clone workflow

YoungJules:
Port the useful LEGIC clone path into the client, including dump rewriting and optional tag restore, while keeping the existing Lua editor for advanced workflows. Fixed the legic sim bug, I think so! Also improved simulator exit handling and file path expansion for the new native commands.

### DIFF
--- a/client/deps/cliparser/cliparser.c
+++ b/client/deps/cliparser/cliparser.c
@@ -32,6 +32,26 @@
 // Option width set to 30 to allow option descriptions to align.  approx line 74
 // Example width set to 50 to allow help descriptions to align.  approx line 93
 
+static void expand_tilde_path(const char *src, char *dst, size_t dst_len) {
+    if (src == NULL || dst == NULL || dst_len == 0) {
+        return;
+    }
+
+    const char *home = getenv("HOME");
+#ifdef _WIN32
+    if (home == NULL) {
+        home = getenv("USERPROFILE");
+    }
+#endif
+
+    if (home != NULL && (strcmp(src, "~") == 0 || strncmp(src, "~/", 2) == 0 || strncmp(src, "~\\", 2) == 0)) {
+        snprintf(dst, dst_len, "%s%s", home, src + 1);
+        return;
+    }
+
+    snprintf(dst, dst_len, "%s", src);
+}
+
 int CLIParserInit(CLIParserContext **ctx, const char *vprogramName, const char *vprogramHint, const char *vprogramHelp) {
     *ctx = calloc(sizeof(CLIParserContext), sizeof(uint8_t));
     if (*ctx == NULL) {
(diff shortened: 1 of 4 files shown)


client/deps/cliparser/cliparser.c
```diff
@@ -32,6 +32,26 @@
...
+    }
+
+    const char *home = getenv("HOME");
+#ifdef _WIN32
+    if (home == NULL) {
+        home = getenv("USERPROFILE");
+    }
+#endif
+
+    if (home != NULL && (strcmp(src, "~") == 0 || strncmp(src, "~/", 2) == 0 || strncmp(src, "~\\", 2) == 0)) {
```
iceman1001: strcmp / strncmp  is a bit over kill if all you are doing is comparing the first or two first bytes in the string.

looking at it,  you should just be able to look at the first byte ,  since the snprintf ignores the second byte
```
if (home != NULL) && src[0] == '~') {  

}
```

iceman1001:
The `hf legic sim`   now takes a `-f` parameter.    the normal flow for pm3 client is,

```
hf legic eload -f
hf legic sim 
```

client/deps/cliparser/cliparser.c
YoungJules: Good point. I changed it to direct character checks to preserve the existing semantics for ~, ~/, and ~\, while avoiding the extra strcmp/strncmp calls.

YoungJules:
Good point also about the perceived change in workflow.
`hf legic sim -f` is meant as a convenience path for the common case where you just want to load a dump and start simulating immediately. The existing `hf legic eload -f` +` hf legic sim` flow still works and remains the more explicit option if you want to inspect or tweak emulator memory before simulating.

YoungJules:
Watch this space for a new commit...

YoungJules:
Added a fix for a vulnerability picked up by CodeQL in cmddata.c

client/src/cmddata.c
```diff
@@ -3058,7 +3058,7 @@ static int CmdDiff(const char *Cmd) {
 
     PrintAndLogEx(INFO, "");
     PrintAndLogEx(INFO, hdr1);
-    PrintAndLogEx(INFO, hdr0);
+    PrintAndLogEx(INFO, "%s", hdr0);
     PrintAndLogEx(INFO, hdr1);
```
iceman1001: you have the same issue above and below your fix....

YoungJules: Hdr1 is not unchecked (user pokeable?) content so it's ok.

iceman1001: ok,  follow the code style and add the string formatter...

YoungJules: OK, just added a new commit: Apply same percent-s formatting to PrintAndLogEx calls for hdr1. Not strictly needed from a security standpoint but makes the code prettier and next developer will not be left wondering why hdr0 is handled differently to hdr1 :-)

PR #3296, opened 2026-05-08
Calypso maintenance

kormax:
This PR introduces the following changes to Calypso module:
* Adds extraction of "data objects" in `hf calypso info`;
* For big data object entries in both `hf calypso info` and `hf calypso dump`, they're now split by line for improved readability;
* If GET DATA reports UNSUPPORTED when extracting a list of data objects - the client stops any further attempts to do that.

```log
[=] --- Get Data Objects ----------------------
[+]  0185 Traceability Information : 96 bytes
[=]       14 10 01 70 82 03 03 01 92 03 21 02 0A 01 00 00 14 01 01 1A 22 1B 04 3F 10 01 00 00 00 00 00 00 
[=]       00 04 00 71 B3 00 00 00 00 00 XX XX XX XX XX XX XX XX XX 30 98 44 00 1B A0 00 03 96 00 FF 00 00 
[+]  5F52 ATR historical bytes     : 5F 52 0F 80 5A 28 13 XX XX XX XX XX XX XX XX 82 90 00 
```

### DIFF
--- a/client/src/cmdhfcalypso.c
+++ b/client/src/cmdhfcalypso.c
@@ -1754,6 +1754,39 @@ static void calypso_print_ef_list_lids(const calypso_ef_list_t *ef_list) {
     PrintAndLogEx(SUCCESS, " EF List LIDs      : " _YELLOW_("%s") "%s", lids, ef_list->truncated ? " (truncated)" : "");
 }
 
+static void calypso_print_info_data_objects(void) {
+    bool printed_header = false;
+
+    for (size_t i = 0; i < ARRAYLEN(calypso_get_data_probes); i++) {
+        const calypso_get_data_probe_t *probe = &calypso_get_data_probes[i];
+        if (probe->tag != 0x0185 && probe->tag != 0x5F52) {
+            continue;
+        }
+
+        uint8_t response[APDU_RES_LEN] = {0};
+        size_t response_len = 0;
+        uint16_t sw = 0;
+        int res = calypso_get_data_object(probe->tag, response, sizeof(response), &response_len, &sw);
+        bool has_data = res == PM3_SUCCESS && calypso_read_sw_has_data(sw, response_len);
+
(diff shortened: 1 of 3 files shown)


client/src/cmdhfcalypso.c
```diff
@@ -1736,22 +1739,58 @@ static bool calypso_ef_list_contains_lid(const calypso_ef_list_t *ef_list, uint1
     return false;
 }
 
-static void calypso_print_ef_list_lids(const calypso_ef_list_t *ef_list) {
-    if (ef_list == NULL || ef_list->count == 0) {
+static void calypso_print_get_data_hex_block(const uint8_t *data, size_t data_len) {
+    for (size_t pos = 0; pos < data_len; pos += CALYPSO_GET_DATA_HEX_BREAK) {
+        size_t chunk_len = MIN(data_len - pos, CALYPSO_GET_DATA_HEX_BREAK);
+        PrintAndLogEx(INFO, CALYPSO_GET_DATA_HEX_INDENT _YELLOW_("%s"), sprint_hex(data + pos, chunk_len));
+    }
```
iceman1001: should be a print hex with break in util.c

kormax: Funny that you mention that.

I initially used that version, but it features no color-coding, and prints all data without spaces or indentation needed for the context, so I replaced it with a custom version.

I think one thing we can do - add a new helper function that does:
- Color coding
- Prefix (for spacing indentation)
- Custom length limiting

kormax: Internal helper removed, a more generic and broad one added to `util`.

While new helper is a superset of the existing one, I didn't touch the older one to be sure that nothing affects code external to this PR.

iceman1001: nice nice,   sadly enough,  I think there is many of those "helper functions"  which prints hex strings.  

I never got around to unify them,

PR #3310, opened 2026-05-16
Auto detect Makefile.platform

nemanjan00:
### Problem

Users currently must manually create `Makefile.platform` with the correct `PLATFORM` and `PLATFORM_EXTRAS` settings. Getting these wrong leads to firmware that doesn't match the hardware, or missing features. The firmware already reports nearly all of this information via `CMD_CAPABILITIES` and `CMD_VERSION`, but there's no way to turn that into a build config automatically.

### Solution

**New CLI command: `hw platformconfig`**

Connects to the device, queries capabilities and version info, and prints valid `Makefile.platform` content to stdout:

```
$ ./pm3 -c "hw platformconfig"
PLATFORM=PM3ICOPYX
PLATFORM_EXTRAS=FLASH
```

Only `PLATFORM` and `PLATFORM_EXTRAS` are auto-detected. Other settings (standalone mode, skip flags, LED order) are left for manual configuration.

Works with existing firmware (no flash needed) by using heuristics:
- FPGA bitstream names in the version string identify the platform (`fpga_icopyx_*` → PM3ICOPYX, `fpga_pm3_ult_*` → PM3ULTIMATE, `is_rdv4` → PM3RDV4, else PM3GENERIC)
- `compiled_with_*` capability flags reconstruct `PLATFORM_EXTRAS` (FLASH, SMARTCARD, BTADDON)

**New firmware command: `CMD_BUILD_CONFIG`**

Reports build configuration as structured data (`build_config_t`) with platform type and LED order. The client auto-detects support via a `has_build_config` bit in `capabilities_t` (uses existing padding — no struct size change). When available, the client uses this instead of heuristics for exact platform identification. On version mismatch, warns and falls back to heuristic detection.

**New file: `Makefile.platform.autodetect`**

Drop-in replacement for `Makefile.platform` that probes the device at build time:

```
cp Makefile.platform.autodetect Makefile.platform
make
```

Uses `./pm3` for auto-detection. Supports `PM3_PORT=` for multi-device setups. Includes a manual overrides section for standalone mode, skip flags, and other options.

### Files changed

- `include/pm3_cmd.h` — `CMD_BUILD_CONFIG` (0x011B), `build_config_t` struct, `has_build_config` capability bit
- `armsrc/appmain.c` — `SendBuildConfig()` handler + dispatch
- `client/src/cmdhw.c` — `CmdPlatformConfig()` with structured path + heuristic fallback
- `Makefile.platform.autodetect` — Auto-detecting Makefile.platform drop-in
- `.gitignore` — Ignore autodetect cache file
- `doc/md/Use_of_Proxmark/4_Advanced-compilation-parameters.md` — Documentation
- `CHANGELOG.md` — Changelog entry

### Testing

Tested against a PM3 Easy Max (PM3ICOPYX) with existing (pre-patch) firmware:
- `hw platformconfig` correctly outputs `PLATFORM=PM3ICOPYX` + `PLATFORM_EXTRAS=FLASH` via the heuristic fallback path
- `Makefile.platform.autodetect` correctly sets build variables at make time
- Output matches the device's actual `Makefile.platform`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ originalitysig.csv
 Makefile.platform
 Makefile.platform.*
 !Makefile.platform.sample
+!Makefile.platform.autodetect
 
 # Cache for detecting platform def changes
 .Makefile.options.cache
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ Makefile.platform
 Makefile.platform.*
 !Makefile.platform.sample
 !Makefile.platform.autodetect
+.Makefile.platform.autodetect.cache
 
 # Cache for detecting platform def changes
 .Makefile.options.cache
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added `hw platformconfig` command and `Makefile.platform.autodetect` to auto-detect platform config from a connected device (@nemanjan00)
 - Added secure channel operations to `hf iclass sam` and fixed/improved I2C bus operations (@antiklesys)
 - Added `hf calypso list` command (@kormax)
 - Added `hf mfd verifycert` command (@kormax)
--- a/Makefile.platform.autodetect
+++ b/Makefile.platform.autodetect
@@ -2,16 +2,23 @@
 # Copy this file as Makefile.platform to use auto-detection.
 # Requires a connected Proxmark3 device and a pre-built client.
 #
-# Set PM3_PORT to your device, e.g.:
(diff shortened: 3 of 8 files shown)


iceman1001:
1. You need to bump the config_t  minor version number also...  otherwise we will have issues.
2. does this overwrite an existing `Makefile.platform`   ?

nemanjan00:
1. I did not find "config_t minor" in project. You mean capabilities version?
2. No, it is alternative sample that auto-detects platform, but, there is an issue right now, it can only be used when client exists, so, it is pretty useless, not sure if it should be merged

nemanjan00:
Just to reiterate on issue I am attempting to solve...

I have built client already, I want to swap to different hardware... 

```
cp Makefile.platform.autodetect Makefile.platform
make
./pm3-flash-all
```

And it does not matter what hardware it is, it will be discovered

First build I have no solution for other than building client twice

nemanjan00:
I have realized I have no way of making this work, without bypassing capabilities version, which is not something I like as an idea

PR #3314, opened 2026-05-17
Add configurable card polling with smart HF/LF detection

cuijianzhuang:
Introduces continuous polling mode (-@) for hf search, lf search, and the auto command, with a persistent millisecond interval preference.

ui.h: add poll_interval_ms field to session_arg_t (default 500 ms)
preferences.c: save/load poll.interval_ms to JSON prefs file; add prefs set poll.interval --ms <N> and prefs get poll.interval commands so the interval can be changed at runtime without restart
cmdhf.c: add -@ flag to hf search for continuous HF-only polling
cmdlf.c: add -@ flag to lf search for continuous LF-only polling
cmdmain.c: add -@ flag to auto for smart polling — tries HF first; if an HF card is found the LF scan is skipped that iteration, so a card placed near the HF reader triggers only HF polling and a card placed near the LF reader triggers only LF polling

### DIFF
--- a/client/src/cmdhf.c
+++ b/client/src/cmdhf.c
@@ -21,6 +21,7 @@
 #include "cmdparser.h"      // command_t
 #include "cliparser.h"      // parse
 #include "comms.h"          // clearCommandBuffer
+#include "util_posix.h"     // msleep
 #include "lfdemod.h"        // computeSignalProperties
 #include "cmdhf14a.h"       // ISO14443-A
 #include "cmdhf14b.h"       // ISO14443-B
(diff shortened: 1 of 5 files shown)


iceman1001:
I repeat what I wrote in the other closed PR,  

The Proxmark3 isn't designed to be run in a looping mode since you will burn out the fpga flash memory which is cleared for 100k writes, now one loop of hf & lf search is 5 writes. 20 000 loops and you burnt it.

cuijianzhuang:
Ok

PR #3319, opened 2026-05-20
General fixes and changes to primary issues shown

C2Pwn:
- Gate SEN recovery on MIFARE Classic fingerprinting
- Check default keys before nonce candidate generation
- Add missing default keys and prioritize 0000-prefixed sector 32 key B candidates
- Fix SEN progress output alignment, inplace refresh behavior
- Populate quiet single sector fchk results to avoid null key propagation
- Recheck found keys for key reuse across sectors
- Minor fix on mifare host due to contract bug in mf_check_keys_fast_ex, where it reports success and has the key, but does not populate the output structure the caller is expected to read which would fail.


doegox:
cf https://github.com/RfidResearchGroup/proxmark3/commit/cc7b235a5ba5042eb97d910c2e7d031d026f91f8

PR #3323, opened 2026-05-22
python fixes

shuffle2:
now, one can do "script run mything.py" multiple times without
crashing, and exit codes are correct.

Pm3PyRun_SimpleFileNoExit:
* use XDECREF instead of DECREF (handles possible nullptr).
* don't double-free "er".
* print exit status correctly.

CmdScriptRun:
* allow invoking scripts multiple times.
* fix memleak of argtable contents if help path not triggered.
* configure_c_stdio should be 0 or 1, not < 0.
* py_conf: don't explicitly set options to their default values, it's confusing.
* Call Py_Finalize only when exiting pm3.

main_loop:
* cleanup cmdscript from main_loop. Important to call
  Py_Finalize from here in case Qt is used (in which case
  main_loop is on Qt thread).

ProxGuiQT::MainLoop:
* reduce timer delay to 0 - there isn't a need to wait here.

### DIFF
--- a/client/src/cmdscript.c
+++ b/client/src/cmdscript.c
@@ -52,6 +52,19 @@ extern int luaopen_pm3(lua_State *L);
 extern PyObject *PyInit__pm3(void);
 #endif // HAVE_PYTHON_SWIG
 
+static void Pm3Py_FlushStream(const char *stream_name) {
+    PyObject *flush_stream = PySys_GetObject(stream_name);
+    if (!flush_stream) {
+        return;
+    }
+    PyObject *fr = PyObject_CallMethod(flush_stream, "flush", NULL);
+    if (fr) {
+        Py_DECREF(fr);
+    } else {
+        PyErr_Clear();
+    }
+}
+
 // Partly ripped from PyRun_SimpleFileExFlags
 // but does not terminate client on sys.exit
 // and print exit code only if != 0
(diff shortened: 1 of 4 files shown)


iceman1001:
The GH Actions all broke...   You need to address that first

shuffle2:
fwiw the fix for test failures was making sure that python's stdout/stderr is flushed before pm3 exits. i fixed in 2 ways:
* make sure flush happens after each script is run (might help for interactive use)
* reinstate use of Py_Finalize before pm3 exits - ensures anything buffered is flushed before the app exits.

iceman1001:
Cool cool,    Now one more test,   the  libpm3 / SWIG 
Verify that these also works as expected 🙏🏼 

https://github.com/RfidResearchGroup/proxmark3/tree/master/client/experimental_client_with_swig

https://github.com/RfidResearchGroup/proxmark3/tree/master/client/experimental_lib

shuffle2:
> Cool cool, Now one more test, the libpm3 / SWIG Verify that these also works as expected 🙏🏼
> 
> https://github.com/RfidResearchGroup/proxmark3/tree/master/client/experimental_client_with_swig
> 
> https://github.com/RfidResearchGroup/proxmark3/tree/master/client/experimental_lib

Yes they pass

PR #3325, opened 2026-05-23
Fixing the issue where the word "Loading" is used instead of the COM …

boss-chifra:
…port

"[+] Using UART port Loading"
image


boss-chifra:
Using `-NoProfile` ensures that PowerShell starts without loading user profile scripts. This is important for automation/scanning scenarios, as a profile may print additional text to stdout (such as banners, prompt customizations, or diagnostic messages), which breaks machine-readable output and can interfere with subsequent parsing in shell scripts.

iceman1001:
Interesting catch,   

for consistency you need to apply same fix in the other calls to powershell  a bit down in the script

boss-chifra:
> for consistency you need to apply same fix in the other calls to powershell a bit down in the script

Yes, you are absolutely right — I did not think of that. However, I see that you have already taken care of everything. Thank you very much.

iceman1001:
well,   there could be the wsl path...

boss-chifra:
done!
https://github.com/RfidResearchGroup/proxmark3/pull/3329

PR #3331, opened 2026-05-24
Added Xerox write block function, plus remap for selection model/consumables

JeroenSteen:
Added the following:

- Write-commando **(0x31)** found from sniff trace to protocols.h
- Write block function, the tag response still needs to be handled
- JSON-data for printermodels/consumables (some partnumbers are duplicated or imperfect) named xerox.json.
- Remap function concept for selection model/consumables (gives error: Segmentation fault      $CLIENT "$@").


iceman1001:
Cool PR,

Some quick feedback, 
 - I notice more and more custom json handling everywhere in the client.  We trying to funnel all file operations into  `fileutils.c`
 - Custom spacing and long oneliners,   they are not worth doing since we run `make style` onto the source code.  
 
 If you are using an AI ,    please tell it to read `CONTRIBUTING.md`  and run  `make style`   before making a pull request.

iceman1001:
This is still marked as a draft. 
if you still haven't gotten any success , try throw some AI on your research.

JeroenSteen:
> This is still marked as a draft. if you still haven't gotten any success , try throw some AI on your research.

The writing and remap works, but not as smooth as expected because of sometimes there is no response from tag. I don't know what would be the best strategy for that. I also don't know how to write all the partnumbers if they're in different format.

PR #3334, opened 2026-05-25
Update memory allocation failure error message, fix bugs and missed DropField() checks

klks:
use the proper memory failed error message
increased buffer on a few variables
add DropField that were missed


client/src/cmdhffmcos.c
```diff
@@ -3440,6 +3546,9 @@ static int CmdHFFmcosOverdraft(const char *Cmd) {
 
             uint8_t tac_calc[4] = {0};
             if (!fmcos_des_mac(tac_buf, 23, tac_key, zero_iv, tac_calc, 4)) {
+                if (!keep) {
+                    DropField();
+                }
                 return PM3_EMALLOC;
```
iceman1001: PM3_EMALLOC?

iceman1001:
better,   still some dubious  emalloc returns when I figure it should be  esoft...

Issue #3346, opened 2026-06-02
MADDecode off-by-one: 2-byte stack OOB write on MAD2 cards/dumps


AlxCzl:
**Describe the bug**

`MADDecode()` in `client/src/mifare/mad.c` has an off-by-one that writes past the end of the `mad[]` stack array when a MAD version 2 card (or dump) is processed.

The array is declared as `uint16_t mad[7 + 8 + 8 + 8 + 8]` (= 39 elements, valid indices 0-38) at every call site (`cmdhfmf.c:6959`, `cmdhfmf.c:7044`, `mad.c:503`), but `MADDecode()` writes 16 (MAD1) + 1 (MAD2 marker) + 23 (MAD2 loop) = **40** entries. The last iteration of the MAD2 loop writes `mad[39]`, one element past the end, a 2-byte out-of-bounds stack write whose value comes from the card/dump data.

The sizing comment `// 7 + 8 + 8 + 8 + 8 == 39` does not account for the extra MAD2 marker entry (`0x0005`) inserted at `mad.c:273`.

The relevant code in `MADDecode()` (`mad.c:266–281`):

```c
for (int i = 1; i <= 16; i++) {                              // 16 writes: indices 0..15
    mad[*madlen] = madGetAID(sector0, swapmad, 1, i);
    (*madlen)++;
}
if (haveMAD2) {
    mad[*madlen] = 0x0005;                                   // index 16
    (*madlen)++;
    for (int i = 1; i < 24; i++) {                           // 23 writes: indices 17..39
        mad[*madlen] = madGetAID(sector16, swapmad, 2, i);  // mad[39] is OOB
        (*madlen)++;
    }
}
```

`haveMAD2` is set by `MADCheck()` when the dump's General Purpose Byte indicates MAD version 2, so this is reachable from any MAD2 card or dump file.

The write corrupts whatever sits after `mad[]` on the stack (adjacent locals, saved registers, etc.), and the value is attacker-influenced (it comes from the card's AID bytes). The default build has `-fstack-protector-strong` and `-D_FORTIFY_SOURCE=3` enabled (unless `NOHARDENING=1`), so in practice the canary will catch this and abort, but it still means every MAD2 card or dump silently triggers undefined behavior and a crash on canary-protected builds.

**To Reproduce**

No hardware required. Offline reproduction with a crafted MIFARE Classic 4K dump:

1. Generate the dump:
   ```
[...]

PR #3349, opened 2026-06-02
Fix MADDecode off-by-one (#3346): mad[] sized for 39 entries but 40 written when MAD2 present

AlxCzl:
This patch corrects the off-by-one vulnerability that was described in issue #3346.

### DIFF
--- a/client/src/cmdhfmf.c
+++ b/client/src/cmdhfmf.c
@@ -6956,7 +6956,7 @@ static int CmdHF14AMfMAD(const char *Cmd) {
         }
 
         if (aidlen == 2 || decodeholder) {
-            uint16_t mad[7 + 8 + 8 + 8 + 8] = {0};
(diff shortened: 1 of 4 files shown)


iceman1001:
There is another bug in MAD2 handling of memory which is about   index vs position of allocated memory. 

See if you can spot it with your AI/Fuzzer

AlxCzl:
Sure! Was looking for bugs using a few claude skills but testing the vulns is actually taking me longer than patching them :rofl:

AlxCzl:
Got 4 potential issues (NULL deref, Buffer over-read, and off-by-one) in MAD-related code. These are small patches, want 4 PRs or a detailed one?

iceman1001:
many small PR's  :)

AlxCzl:
Working on it :D

PR #3355, opened 2026-06-03
MAD revamp to harden the parser with typed structs and to fix memory safety bugs

AlxCzl:
## mad: harden MAD parser with typed structs and fix memory safety bugs

Referencing issue #3346, this patch aims to fix the root cause of the various bugs found in the Mifare MAD parser.

The original code relied on raw byte-offset arithmetic to access on-card MAD structures, which made bounds checking impractical and resulted in several memory safety bugs. \
This rewrite introduces typed structs that map directly to the card layout, so the compiler enforces correctness that was previously left to the programmer.
Every caller across `cmdhfmf`, `cmdhfmfp`, `cmdnfc`, and `cmdhfgallagher` now uses the new code.

#### Heap buffer over-read in `convert_mad_to_arr`

`convert_mad_to_arr` copied 64 bytes from the input without checking `ilen` and extracted NDEF sectors without validating offsets against the input buffer size; a crafted dump file could read past the allocated buffer. \
Added bounds checks on all input accesses and an `olen_max` parameter to cap output writes.

#### `NULL` deref in `print_aid_description`

`strlen(fmad)` on the return value of `mad_json_get_str()` with no NULL check; crashes if a JSON entry has no `"mad"` field.

#### Stack overflow via VLA in `print_aid_description`

`char result[result_len]` where `result_len` comes from JSON string lengths; a crafted `mad.json` with huge strings blows the stack. \
Replaced with a fixed-size buffer.

#### Card holder info read from wrong offset

`MADCardHolderInfoDecode` was called with the raw dump starting at byte `0` (manufacturer block) instead of the sectors whose AID matches `0x0004`. \
Now locates matching sectors through the MAD and extracts their data with bounds checking.

#### Missing `datalen` overflow check

`memcpy` into the card holder data buffer (both `cmdhfmf.c` and `cmdhfmfp.c` live-card paths) had no bounds check before appending sector data.

#### `ilen` truncation in `convert_mad_to_arr`

The parameter was `uint16_t`, quietly truncating files larger than 64 KB. \
Changed to `size_t`.

### Hardening

- Added `mad1_sector_t` and `mad2_sector_t` structs that map directly to the on-card MAD layout, with compile-time size assertions (`sizeof == MFBLOCK_SIZE * 4`). All raw byte-offset math (`sector[16 + 2 + (i-1)*2]`) is replaced with typed field access (`m->aid[i-1]`);
- Added `mad_sector_t` wrapper (pointer + length) so every MAD function carries the buffer size next to the data pointer. Bounds validation happens at every entry point through `mad_sector_valid()`;
- Replaced raw `uint16_t mad[]` / `size_t madlen` output arrays with a `mad_t` struct that has a fixed-size `entries[39]` array and a bounds-checked `count` field;
- `MADCardHolderInfoDecode` now validates TLV lengths against remaining data and rejects unknown type values instead of indexing out of bounds; and
- NULL checks on all public API entry points.

### API changes

- `MADComputeCRC(m)` macro: computes the MAD CRC directly from a `mad1_sector_t*` or `mad2_sector_t*`; Replaces the error-prone `CRC8Mad(&sector[16+1], 15+16)` pattern in `mad.c`, `cmdhfgallagher.c`, and `gallaghertest.c`;
- `mad_aid_get()` / `mad_aid_set()` inline helpers for reading and writing AIDs in on-card byte order; and
- Merged the split AID sub-arrays (`aid1[7]` + `aid2[8]`) into single `aid[15]` / `aid[23]` arrays. The memory layout is identical (PACKED), but indexing becomes a simple `m->aid[sector_no - 1]`.

### Files changed

| File | What changed |
|------|--------------|
| `client/src/mifare/mad.h` | New structs, macros, typed API signatures |
| `client/src/mifare/mad.c` | Rewrite of internals, all fixes listed above |
| `client/src/cmdhfmf.c` | Migrate MAD call sites to new API, fix card holder info |
| `client/src/cmdhfmfp.c` | Migrate MAD call sites, add datalen bounds check |
| `client/src/cmdnfc.c` | Update `convert_mad_to_arr` call to new signature |
| `client/src/cmdhfgallagher.c` | Use structs and `MADComputeCRC` instead of raw byte offsets |
| `client/src/mifare/gallaghertest.c` | Update CRC test to use `MADComputeCRC` |

### Testing

107 offline differential tests comparing the vanilla (master) client and the patched client across every offline MAD code path: `hf mf mad`, `hf mf view`, `nfc decode` with all flag combinations (`--force`, `-v`, `--be`, `--dch`, `--aid`), plus edge-case dumps (empty AIDs, invalid GPB, bad CRC, duplicate AIDs, HID/VIGIK detection, byte-swap, info byte boundaries). Hardware tests ran against a MIFARE Classic 1K with MAD+NDEF data. \
I did not include them here as I did not know if there is a test folder or a specific way to include tests.

Below is the output of my test script:

<details>
<summary>Tests output</summary>

```
================================================================
=== Running MAD comparison ===
================================================================

--- [A] hf mf mad -f <dump> --force ---
[SAME] A_mad1_all_ffff
[SAME] A_mad1_all_free
[SAME] A_mad1_allff
[SAME] A_mad1_bad_version
[SAME] A_mad1_cardholder
[SAME] A_mad1_cardholder_real
[SAME] A_mad1_continuation
[SAME] A_mad1_empty
[SAME] A_mad1_hid_pacs
[SAME] A_mad1_info_invalid
[SAME] A_mad1_info_max
[SAME] A_mad1_info_zero
[SAME] A_mad1_mixed
[SAME] A_mad1_multiapp
[SAME] A_mad1_ndef
[SAME] A_mad1_ndef_data
[SAME] A_mad1_no_da
[SAME] A_mad1_no_mad_key
[SAME] A_mad1_publisher
[SAME] A_mad1_swap_test
[SAME] A_mad1_vigik
[SAME] A_mad2_4k
[SAME] A_mad2_all_ndef
[SAME] A_mad2_empty_s16
[SAME] A_mad2_mixed
[SAME] A_mad2_s16_info_invalid
[SAME] A_mad2_s16_info_max
[SAME] A_mad2_s16_info_self
[SAME] A_mad2_s16_info_zero

--- [B] hf mf mad -f <dump> --force -v ---
[SAME] B_mad1_all_ffff
[SAME] B_mad1_all_free
[SAME] B_mad1_allff
[SAME] B_mad1_bad_version
[SAME] B_mad1_cardholder
[SAME] B_mad1_cardholder_real
[SAME] B_mad1_continuation
[SAME] B_mad1_empty
[SAME] B_mad1_hid_pacs
[SAME] B_mad1_info_invalid
[SAME] B_mad1_info_max
[SAME] B_mad1_info_zero
[SAME] B_mad1_mixed
[SAME] B_mad1_multiapp
[SAME] B_mad1_ndef
[SAME] B_mad1_ndef_data
[SAME] B_mad1_no_da
[SAME] B_mad1_no_mad_key
[SAME] B_mad1_publisher
[SAME] B_mad1_swap_test
[SAME] B_mad1_vigik
[SAME] B_mad2_4k
[SAME] B_mad2_all_ndef
[SAME] B_mad2_empty_s16
[SAME] B_mad2_mixed
[SAME] B_mad2_s16_info_invalid
[SAME] B_mad2_s16_info_max
[SAME] B_mad2_s16_info_self
[SAME] B_mad2_s16_info_zero

--- [C] hf mf mad -f <dump> --force --be ---
[SAME] C_mad1_swap_test
[SAME] C_mad1_mixed
[SAME] C_mad2_mixed
[SAME] C_mad1_ndef

--- [D] hf mf mad -f <dump> --force --dch ---
[SAME] D_mad1_cardholder
[SAME] D_mad1_mixed
[SAME] D_mad2_mixed
[SAME] D_mad1_ndef

--- [E] hf mf mad -f <dump> --force --aid <hex> ---
[SAME] E_mad1_mixed_aide103
[SAME] E_mad2_mixed_aide103
[SAME] E_mad1_mixed_aid0004
[SAME] E_mad2_mixed_aid0004
[SAME] E_mad1_mixed_aid484d
[SAME] E_mad2_mixed_aid484d
[SAME] E_mad1_mixed_aid4910
[SAME] E_mad2_mixed_aid4910
[SAME] E_mad1_mixed_aid1234
[SAME] E_mad2_mixed_aid1234
[SAME] E_mad1_mixed_aidffff
[SAME] E_mad2_mixed_aidffff
[SAME] E_mad1_mixed_aid0000
[SAME] E_mad2_mixed_aid0000

--- [F] hf mf mad -f <dump> --force --be --dch ---
[SAME] F_mad1_cardholder
[SAME] F_mad1_mixed

--- [G] hf mf mad -f <dump> (no --force, tests HasMADKey gate) ---
[SAME] G_mad1_no_mad_key
[SAME] G_mad1_ndef
[SAME] G_mad1_mixed
[SAME] G_mad2_mixed

--- [H] hf mf view -f <dump> ---
[SAME] H_mad1_vigik
[SAME] H_mad1_hid_pacs
[SAME] H_mad1_ndef
[SAME] H_mad1_mixed
[SAME] H_mad2_mixed
[SAME] H_mad1_no_mad_key

--- [I] hf mf view -f <dump> -v ---
[SAME] I_mad1_vigik
[SAME] I_mad1_hid_pacs
[SAME] I_mad1_mixed

--- [J] nfc decode -f <dump> ---
[SAME] J_mad1_ndef_data
[SAME] J_mad1_ndef
[SAME] J_mad1_no_mad_key
[SAME] J_mad2_all_ndef

--- [K] Repo trace files ---
[SAME] K_s50_empty_mad
[SAME] K_s50_empty_view
[SAME] K_s50_empty_nfc

================================================================
  Total tests: 173
  Identical:   173
  Different:   0
================================================================
ALL OUTPUTS IDENTICAL
================================================================
```
</details>

### DIFF
--- a/client/src/cmdnfc.c
+++ b/client/src/cmdnfc.c
@@ -138,12 +138,13 @@ static int CmdNfcDecode(const char *Cmd) {
         } else  {
 
             // convert from MFC dump file to a pure NDEF byte array
-            if (HasMADKey(tmp)) {
+            mad_sector_t s = {tmp, bytes_read};
+            if (HasMADKey(&s)) {
                 PrintAndLogEx(SUCCESS, "MFC dump file detected. Converting...");
                 uint8_t ndef[4096] = {0};
-                uint16_t ndeflen = 0;
+                size_t ndeflen = 0;
 
-                if (convert_mad_to_arr(tmp, bytes_read, ndef, &ndeflen, override) != PM3_SUCCESS) {
+                if (convert_mad_to_arr(tmp, bytes_read, ndef, &ndeflen, sizeof(ndef), override) != PM3_SUCCESS) {
                     PrintAndLogEx(FAILED, "Failed converting, aborting...");
                     free(dump);
                     return PM3_ESOFT;
--- a/client/src/cmdhfgallagher.c
+++ b/client/src/cmdhfgallagher.c
@@ -999,20 +999,19 @@ static int hfgal_update_mad(uint8_t cred_sector, uint8_t cad_sector,
         return res;
     }
 
+    mad1_sector_t *m = (mad1_sector_t *)sector0;
+
     // Set AID for credential sector (0x4812)
     if (cred_sector >= 1 && cred_sector <= 15) {
-        sector0[16 + 2 + (cred_sector - 1) * 2]     = CLASSIC_CRED_AID & 0xFF;
-        sector0[16 + 2 + (cred_sector - 1) * 2 + 1]  = (CLASSIC_CRED_AID >> 8) & 0xFF;
+        mad_aid_set(&m->aid[cred_sector - 1], CLASSIC_CRED_AID);
     }
 
     // Set AID for CAD sector (0x4811)
     if (cad_sector >= 1 && cad_sector <= 15) {
-        sector0[16 + 2 + (cad_sector - 1) * 2]     = CLASSIC_CAD_AID & 0xFF;
-        sector0[16 + 2 + (cad_sector - 1) * 2 + 1]  = (CLASSIC_CAD_AID >> 8) & 0xFF;
+        mad_aid_set(&m->aid[cad_sector - 1], CLASSIC_CAD_AID);
     }
 
-    // Recalculate CRC over bytes 17..47 (info byte + 15 AID entries)
-    sector0[16] = CRC8Mad(&sector0[16 + 1], 15 + 16);
+    m->crc = MADComputeCRC(m);
 
     // Write blocks 1 and 2 of sector 0 back (block 0 is manufacturer block, don't touch)
     res = mf_write_block(1, mad_key_type, mad_key, &sector0[MFBLOCK_SIZE]);
--- a/client/src/cmdhfmf.c
+++ b/client/src/cmdhfmf.c
@@ -703,7 +703,8 @@ void mf_print_blocks(uint16_t n, uint8_t *d, bool verbose) {
     }
 
     // MAD detection
-    if (HasMADKey(d)) {
+    mad_sector_t mad_sec = {d, n * MFBLOCK_SIZE};
+    if (HasMADKey(&mad_sec)) {
         PrintAndLogEx(HINT, "Hint: MAD key detected. Try `" _YELLOW_("hf mf mad") "` for more details");
     }
     PrintAndLogEx(NORMAL, "");
@@ -6902,7 +6903,8 @@ static int CmdHF14AMfMAD(const char *Cmd) {
         }
 
         // MAD detection
-        if ((HasMADKey(dump) == false) && (force == false)) {
+        mad_sector_t dump_sec = {dump, bytes_read};
+        if ((HasMADKey(&dump_sec) == false) && (force == false)) {
             PrintAndLogEx(FAILED, "No MAD key was detected in the dump file");
             free(dump);
             return PM3_ESOFT;
@@ -6910,9 +6912,9 @@ static int CmdHF14AMfMAD(const char *Cmd) {
 
         MADPrintHeader();
         bool haveMAD2 = false;
-        MAD1DecodeAndPrint(dump, swapmad, verbose, &haveMAD2);
+        MAD1DecodeAndPrint(&dump_sec, swapmad, verbose, &haveMAD2);
 
-        int sector = DetectHID(dump, 0x484d);
+        int sector = DetectHID(&dump_sec, 0x484d);
         if (sector > -1) {
 
             // decode it
@@ -6921,7 +6923,7 @@ static int CmdHF14AMfMAD(const char *Cmd) {
             PrintAndLogEx(INFO, _CYAN_("HID PACS detected"));
 
             uint8_t pacs_sector[MFBLOCK_SIZE * 3] = {0};
-            memcpy(pacs_sector, dump + (sector * 4 * 16), sizeof(pacs_sector));
+            memcpy(pacs_sector, dump + (sector * 4 * MFBLOCK_SIZE), sizeof(pacs_sector));
 
             if (pacs_sector[16] == 0x02) {
 
@@ -6944,7 +6946,7 @@ static int CmdHF14AMfMAD(const char *Cmd) {
             }
         }
 
-        sector = DetectHID(dump, 0x4910);
+        sector = DetectHID(&dump_sec, 0x4910);
         if (sector > -1) {
             // decode it
             PrintAndLogEx(INFO, "");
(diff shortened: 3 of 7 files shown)


client/src/cmdnfc.c
```diff
@@ -138,12 +138,13 @@ static int CmdNfcDecode(const char *Cmd) {
...
             // convert from MFC dump file to a pure NDEF byte array
-            if (HasMADKey(tmp)) {
+            mad_sector_t s = {tmp, bytes_read};
+            if (HasMADKey(&s)) {
                 PrintAndLogEx(SUCCESS, "MFC dump file detected. Converting...");
                 uint8_t ndef[4096] = {0};
-                uint16_t ndeflen = 0;
+                size_t ndeflen = 0;
 
-                if (convert_mad_to_arr(tmp, bytes_read, ndef, &ndeflen, override) != PM3_SUCCESS) {
```
iceman1001, line 146: to be fare... MFC can't have 64kb files since it only have the 4K (4096bytes)  cards.  
but MFDES is a different story,  still not breaking 64kb limit though.

iceman1001:
my old MAD parsing implementation was always a sore in my eye.   Thanks for sorting it out!

iceman1001:
And I am happy your AI is learning to understand what we want in this repo better :)

AlxCzl:
I wasn’t sure if you wanted a rewrite or just bugfixes haha, you should’ve told me sooner :D
Glad you like the patch!

-------- Original Message --------

iceman1001:
if you have an even better impl...   as long as it's  human verified.. 

We need a MAD implementation that can be used with the MIFARE different card technologies that supports it.
It would be great to have a  "READ"  and get a continuous array of memory out (regard less of card tech)
and a "WRITE"  that writes back the array to different card tech accordingly.
and a "VERIFY"  that verifies the written data.     As in functionality of the client behind the scenes. 

`hf mf mad`  is  how user interacts today with it.    Not sure how the client command set should look like yet.

AlxCzl:
I see, I’ll try to sketch out something.

-------- Original Message --------

AlxCzl:
So, for MAD1/2 I was thinking about adding `hf mf mad {read,write,verify}` and `hf mfp mad {read,write,verify}`, and we'd use them like this:
```
hf mf mad read   --aid E103 [-k key] [-b] [-o file] [-v]
hf mf mad write  --aid E103 [-k key] [-b] [-d hex | -f file] [-v]
hf mf mad verify --aid E103 [-k key] [-b] [-d hex | -f file] [-v]

hf mfp mad read   --aid E103 [-k key] [-b] [-o file] [-v]
hf mfp mad write  --aid E103 [-k key] [-b] [-d hex | -f file] [-v]
hf mfp mad verify --aid E103 [-k key] [-b] [-d hex | -f file] [-v]
```

The goal would be to have a new underlying mad operation structure like:
```c
 typedef struct {
     int (*read_sector)(uint8_t sector_no, uint8_t key_type,
                        const uint8_t *key, uint8_t *buf, bool verbose);
     int (*write_sector_data)(uint8_t sector_no, uint8_t key_type,
                              const uint8_t *key, const uint8_t *data, bool verbose);
     const uint8_t *mad_key;
     uint8_t mad_key_type;
     const uint8_t *app_key;
     uint8_t app_key_type;
     bool verbose;
 } mad_ops_t;
```

Which would be used with:

`mad_read_directory` (static helper):
 - Reads sector 0 into a stack-allocated `mad1_sector_t` via `ops->read_sector`, casting (`uint8_t *) &sector0`
 - Attempts to read sector 16 into a `mad2_sector_t` (non-fatal if it fails, just means no MAD2)
 - Calls existing `MADDecode(&sector0, &mad2, &mad_list, swapmad, override)`
 - Returns the populated `mad_entry_list_t`

`mad_app_read`:
 - Calls `mad_read_directory` to get the entry list
 - Iterates entries matching `aid`
 - For each of these entries: reads the corresponding sector into a buffer, then copies data blocks only and appends to output
 - Reuse `mfFirstBlockOfSector()`, `mfNumBlocksPerSector()` from `include/mifare.h`

`mad_app_write`:
 - Calls `mad_read_directory` to get the entry list
 - Calculates total capacity across matching sectors
 - For each matching sector: prepares a zero-padded data block buffer, calls `ops->write_sector_data`
 - Verbose output on each sector that was written

`mad_app_verify`:
 - Calls `mad_app_read` to get readback data
 - `memcmp` or the likes against the expected data, reports first differing offset on mismatch

We'd keep the code backward compatible though (i.e. adapting `hf mf mad`, `hf mf mad --aid e103 -k ffffffffffff -b`, `hf mf mad --dch` and others to the new implem).

Does that look good to you?

iceman1001:
Interesting but one issue. 

`hf mf mad` would be lost. 

how about we make  `mad`  its own head category,   since its going to be used with all Mifare card tech that supports it. 
just like we have `nfc` for  nfc/ndef.      and alias in the `hf mf* `  commands  just like `ndefread` etc.

iceman1001:
And this is not the place to have this dialogue.   this is a closed PR.

AlxCzl:
Sure, I've opened #3356.

PR #3358, opened 2026-06-05
LEGIC: clarify restore/clone flow and add explicit migration helper

YoungJules:
This PR clarifies the LEGIC Prime workflows and fixes a few issues in the clone/restore path:
- Fix hf legic clone --write so it no longer requires --mcc or an output filename.
- Fix LEGIC segment CRC recalculation for the target UID by starting at the correct decoded payload offset.
- Add an explicit hf legic migrate helper that clones a dump to a new tag and writes DCF only when explicit flag set.
- Clarify the difference between restore, clone, and migrate in the command help/docs.
- Keep DCF updates separate and explicit via wrbl, preserving the safety gate.
Behavior changes:
- hf legic restore is documented as restoring an exact dump back to the same card family.
- hf legic clone --write now handles UID-dependent segment CRCs correctly for the target tag.
- hf legic migrate provides a safer migration flow without making DCF updates automatic.

### DIFF
--- a/client/src/cmdhflegic.c
+++ b/client/src/cmdhflegic.c
@@ -65,7 +65,8 @@ static void legic_xor_with_crc(uint8_t *data, uint16_t cardsize, uint8_t crc) {
 }
 
 static bool legic_clone_update_segment_crcs(uint8_t *data, size_t bytes_read, const uint8_t uid[4]) {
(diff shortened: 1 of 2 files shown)


iceman1001:
Awesome!

PR #3361, opened 2026-06-06
PACK fix for the lua script hf_mf_ultimatecard

dangerous-tac0s:
TLDR, the PACK is stored on page 13. This corrects the script so that it reports and changes the PACK correctly. Testing can be [found here](https://forum.dangerousthings.com/t/ug4-and-the-pack/28230).

### DIFF
--- a/client/luascripts/hf_mf_ultimatecard.lua
+++ b/client/luascripts/hf_mf_ultimatecard.lua
@@ -263,7 +263,7 @@ local function read_config()
     if pwd == '04' then lib14a.disconnect(); return nil, "can't read configuration, "..err_lock end
     -- read PACK
     cpack = send("30F1"):sub(1,4)
-    pack = send("30E6"):sub(1,4)
+    pack = send("3013"):sub(1,4) -- *should* be E6, testing shows 13.
     -- read SIGNATURE
     signature1 = send('30F2'):sub(1,32)
     signature2 = send('30F6'):sub(1,32)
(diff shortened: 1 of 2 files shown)


dangerous-tac0s:
Now that I am digging a bit deeper, there are some bugs in the script unrelated to this... Would you all prefer I make another PR for those or just include them in this?

iceman1001:
hm,   so different generations of UMC  stores the pack page differently?   This means an ntag 215, 216 will have a faulty page 13 block when loading a dump upon them...

dangerous-tac0s:
> different generations of UMC stores the pack page differently?

I can't confirm this but it is implied from the magic card notes doc that the older version (which we have never had) featured a working shadow mode so, maybe? I added the version check out of an abundance of caution in not wanting to break something that was hopefully working.

dangerous-tac0s:
From the magic card notes doc:
👉 TODO In Mifare Ultralight / NTAG mode, the special writes (hf mfu restore option -s, -e, -r) do not apply. Use script run hf_mf_ultimatecard for UID and signature, and hf mfu wrbl for PWD and PACK.

So that part doesn't appear to have worked as expected--likely since it always defaulted to the 216 locations regardless of the flavor it emulated.

iceman1001:
Not sure if your comment made things any clearer.   If one specific tag always write to 213 location ,  this is something that should be notified to the user.   Since there is a high probability that memory dump will be wrong.

dangerous-tac0s:
If you want I can address the very old todos for hf mf gview and hf mfu restore? I had intended to in another PR since this one was only scoped for the lua  script. The updated script does inform the user of the appropriate PACK page with the -c switch : )

Does anyone have access to the old UMC or the Fuse Tool application?

iceman1001:
What I mean is the script should inform user that if they are setting pack (via uploading dump,  or setting pack direct)  that on some card if  configured with NTAG215,216 ,  this puts a wrong block in the memory.

And this should be mirrored into magic document and other relevant documents.

PR #3363, opened 2026-06-07
iclass: add 143-bit legacy PACS decode/encode and DES/PLAIN crypto support

cindersocket:
Updated version of Previous PR (#3150) to address comments from Iceman.  Smaller commit than before, reusing more things.

* Fix legacy PACS payload extraction to accept the full 143-bit width, which is the card’s maximum (oddly encoded
but valid per iCLASS), so client-side decode (hf iclass view) no longer rejects max-width credentials as
invalid/missing-sentinel.
* Add tests
* Add plain and des to decode, encode, tagsim
* Exercise both standard and 143-bit vectors in the roundtrip path.
Include coverage for DES and unencrypted iCLASS credentials to match actual card behavior.

CodeQL warnings for insecure algorithms are expected here because these card formats can legitimately use DES credentials.

Still There is more to do on iclass, including PIN and different sio locations and pages and books but this is most of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added support for multi-block/max-size (143-bit) iCLASS. (@cindersocket)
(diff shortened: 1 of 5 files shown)


iceman1001:
This is too much changes for one PR.

you have changes to device side not even commented why.
the removal of cardhelper,
the what looks like new subcategories of commands,
the 134 addition.
the des/3des 

All of these should be one PR  a piece. 

With all this said ,   your AI is a bit wild but productive.   I like it.

PR #3366, opened 2026-06-07
Preserve iCLASS emulator memory across HF_15 reloads

cindersocket:
Fixes an iCLASS emulator memory regression where a partial emulator write could clear the previously loaded dump if the FPGA bitstream cache had been invalidated.

Before this change, CMD_HF_ICLASS_EML_MEMSET ensured HF_15 was loaded through the normal FpgaDownloadAndGo() path. If a reload was needed, that path cleared BigBuf, which erased emulator memory before emlSet() wrote the requested block. The result was an emulator card with only the newly written block populated.

This adds a FpgaDownloadAndGo_keep_EM() variant that preserves emulator memory while ensuring the requested FPGA bitstream is loaded, and uses it for iCLASS emulator writes.

The online test now forces the failure state with:

```
hf iclass eload ...
hw fpgaoff
hf iclass esetblk ...
hf iclass eview -s 64
```

This verifies that the original CSN block remains present while the requested block update is applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 - Added `hf mfdes getversion` command (@kormax)
+- Fixed iCLASS emulator writes clearing previously loaded emulator memory after an FPGA reload (@cindersocket)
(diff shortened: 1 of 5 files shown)


iceman1001:
Hm,   I thought this one was fixed many years ago.   Good thing with a online test to actually test it!

PR #3367, opened 2026-06-08
Added signal passing to stop capture

yackermann:
Right now PM3 requires physical press to stop capturing.

This adjustment to CLI and firmware add proper signaling that lets CLI to send signal to Proxmark to stop capture, acting same as a physical button press.

This is really useful for automated testing flows.

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -1857,13 +1857,13 @@ static void PacketReceived(PacketCommandNG *packet) {
             break;
         }
         case CMD_HF_ISO14443A_SNIFF: {
-            SniffIso14443a(packet->data.asBytes[0]);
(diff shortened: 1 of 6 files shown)


iceman1001:
Not sure I like the client side implementation.  `--stop`    is counter predictive to the sniff command. 

The Pm3 client doesn't want to squeeze all functions in to parameters.   We have separate commands for things.

right now,  `hf 14a sniff`  triggers the device to start sniffing,  then wait until given instructions to not sniff. 
We already have  `hw break` which is breaking the loop.   Next step is for  you to download the trace file
which is  `hf 14a list`   or  `trace list -t 14a`  

I think you don't understand the pm3 client and how it works.

yackermann:
@iceman1001 you are fully right. Thanks a lot!

PR #3368, opened 2026-06-08
Support non-06 iCLASS SE AIA SIO blocks

cindersocket:
The 0x06 in the AIA indicates the SIO block number.  If you change it, readers try reading different blocks.  This commit adds support for having other SIO locations on a tag.


iceman1001:
Interesting,  how often does this happen?

iceman1001:
I see,  cindersocket is one of your sockpuppet accounts.

PR #3369, opened 2026-06-08
15693 inventory enhancements

dangerous-tac0s:
Added support for reading multiple ISO15693 transponders in the field. Also reduced the empty slot timeout.

Screenshot from 2026-06-08 14-14-44


iceman1001:
Pretty cool!

PR #3374, opened 2026-06-11
fix incorrect reading of MF Classic 4K

horrordash:
Reason: The function IsSectorTrailer() accepts an absolute block number, not a sector-based.


iceman1001:
Good catch!

PR #3376, opened 2026-06-12
Add `hf felica sim` command

kormax:
Having made some steps at adding this functionality in the past, I have been reluctant from merging it, given that FeliCa performance in PM3 is barely usable for reading, and even worse at emulating, but given recent interest in the topic, and potential new hardware improvements, there is value in this functionality now.

This PR adds `hf felica sim` command, allowing to emulate FeliCa targets based on dumps generated by `hf felica dump`.

At the current moment, the following commands are supported:
* `Polling` (No Request)
* `Polling` (System Code)
* `Polling` (Communication Performance)
* `RequestResponse`
* `RequestSystemCode`
* `RequestSpecificationVersion`
* `RequestProductInformation`
* `GetContainerIssueInformation`
* `SearchServiceCode`
* `RequestService`
* `RequestServiceV2`
* `ReadWithoutEncryption`

To make the development and implementation of emulation simpler, the current version, instead of trying to replicate each unique behavior for each particular chip version, mashes all "best" behavior together, trying to be as "accomodating" and "flexible" to any potential reader it could encounter.
As one of the consequences of that, that for any given dump, by default, the emulator attempts to respond to all commands, given sensible defaults, even if the card that the data was dumped originally from did not support them. I.E:
* For DES-only dumps,  `RequestServiceV2` will still respond, reporting AES version codes as missing, using a DES encryption identifier.
* `RequestResponse`/`RequestSystemCode`/`SearchServiceCode` work on all dumps, even if taken from FeliCa Lite.
* `Polling (Communication Performance)` declares support for 212kbps only;
* `ReadWithoutEncryption` supports up to 15 blocks in a request at a time regardless of the original card capabilities;

In the future, it may be worth exposing some behaviors, like command support, as arguments, with default `auto` value figuring it out based on the shape of the dump, or IC/ROM version of the chip extracted from IDM.

Note, that as stated prior, PM3 really struggles with emulation, so testing this functionality may prove hard.
For instance, I am unable to read the emulated on Pixel 9A, but am able to on iPhone 16 or Pixel 7 pro:

WAON

CJRC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added `hf felica sim` command (@kormax)
 - Added `mad read`, `mad write`, `mad verify`, `mad decode`, `mad encode` commands with typed struct MAD API (@AlxCzl)
 - Added `hf mfdes getversion` command (@kormax)
 - Fixed iCLASS emulator writes clearing previously loaded emulator memory after an FPGA reload (@cindersocket)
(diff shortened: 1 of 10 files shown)


iceman1001:
Maybe let your AI find the bugs with FeliCa device side?   @xianglin1998   told me about a spi transfer bug he found.  Which could be a reason

PR #3377, opened 2026-06-12
Add missing files to CMakeLists for experimental client lib

Sanduuz:
Add missing files to CMakeLists.txt for experimental client lib. This is the same problem than previously resolved in PRs #3304 #3075 and #3047.

It might be beneficial to implement CI/CD checks for this to automatically notice this in the future.


iceman1001:
Good idea!
You make one?

Sanduuz:
Sure, I can try at least! :)

PR #3378, opened 2026-06-12
Feature: CI check for building client experimental lib

Sanduuz:
Added a CI check that tries to build the client experimental lib. Added linker option `-Wl,--no-undefined` to fail at compile-time if there are undefined symbols in the experimental client lib. Also added better error message for a failed check to improve developer experience.


iceman1001:
Cool!

PR #3388, opened 2026-06-17
Add support for parsing Finnish Helsinki Regional Transport (HRT) travel cards

Sanduuz:
This feature adds support for parsing Finnish Helsinki region travel cards (HRT cards) that are MIFARE DESFire cards. Currently this version hooks the `hf mfdes info` command to parse the value stored on the card if it is recognized to be a valid HRT travel card.

The parser currently implements functionality to parse other information from the card as well, but as of now it is not hooked to any client commands. I would love to get some feedback on what would be the best way to call this functionality. Would it be to continue with hooking the `hf mfdes info` command or something else?

The current version still contains some TODOs, the biggest one being adding support for the older card version (AID EF2011). Hence, I'm marking this pull request as a draft. Ideas, contributions and discussions are greatly welcome!


iceman1001:
Cool!

But,   
you seem to have implemented your own  malloc /free  ,   and it feels redundant.     
another thing is `bytearray_t` struct ?   which now just confusing since all other places in pm3 source uses  `uint8_t*` 
which feels like even more "extras"  which we dont want introduced in the source code. 


We use  `calloc`  in order to make sure all memory allocations are verified zeroed before use.  We also use a unified failure message when allocation fails.     Follow that pattern please.

Sanduuz:
Thanks for the feedback!

I completely understand your point regarding adding "extras". I'll unify the code to match the codebase better and will keep that in mind when working with the new functionality.

kormax:
I'd like to chime in a little bit.

While it's useful to offer built-in decoding for various transit systems and card type formats, we should be mindful of the fact that there are orders of magnitude more ways to format and parse a card than there are card technologies overall.

To avoid overburdening the codebase, I'd like to suggest 2 improvements:
* Move all card-type-format-specific logic into a dedicated package. Separate parsing code from the code that handles card technologies themselves.
* Ideally, trigger custom decoding hooks only after specific conditions are met, before allowing them access to perform IO.
I.E. "list apps" executed by "info" command returned a known AID - we try parsers that know that value.
This won't work if a particular card type locks the app discovery out. So it may be acceptable while there is a little amount of such cases.

iceman1001:
yeah,  
I see we will have a bunch of `parsers`   in a folder,    which  we then have a look up / trigger for in relevant commands.
only run a parser once it has a positive identification.     Each parser should have a function "IsValid"  or something which means we can easily iterate all and call just the ones that are valid for the scenario at hand. 

Not sure if this made things clear.

Sanduuz:
For sure, thanks for the great suggestions!

I do agree that separating the parser calling code would be good to prevent e.g. the `cmdhfmfdes.c` bloating. Did I understand correctly that the idea would be to create e.g. `client/parsers` directory and move the parser calling code + actual parse logic under there? Then when calling `hf mfdes info`, check the returned AID list and if the AID list contains the correct AID, start parsing the application from under parsers.

So the direcory structure would look something like this?:
```
client/
├── src
│   ├── parsers
│   │   ├── parsehrt.c   <-- When `hf mfdes info` returns correct AID, call `parse_hrt` function from this file. 
│   │   ├── hrtparser
│   │   │   ├── hrtparser.c   <-- This file contains the actual parsing logic called from `parsehrt.c`
│   │   │   ├── hrtparser.h
```

We could "standardize" adding parsers in the future according to this structure. Then adding the `is_valid` function would be a good addition to keep the API simple and unified.

In case this is not what you were thinking, please help me understand what would be the preferred way. Otherwise I can start implementing these changes.

kormax:
Yeah looks fine by me.

Sanduuz:
Alright, I have now implemented everything I wanted to implement. For reference here's 2 images of parsing 1 new and 1 old card:

Old card:
hrt_parsed_old_redacted

New card:
hrt_parsed_new_redacted

I have redacted some information from the images since it contains information where and when I have moved around :)

I'd love to get some feedback if possible @kormax @iceman1001
Others can also give feedback and that'd be greatly appreciated as well!

I know there's quite a lot of changes to review, so I'd be happy to help by answering questions regarding the code/HRT cards in general.

iceman1001:
Looking good!  

We now have a parsers folder :)    

How much was AI?

Sanduuz:
Thanks!

I utilized AI mainly in reading and analyzing the codebase on where I should add the hooks and also discussed design choices etc. I asked for proposals on certain things, but read and implemented all of the code myself. No code was blindly accepted :)

PR #3390, opened 2026-06-18
hf felica auth1: refactor

q0jt:
Remove unnecessary code and change the success check for the auth1 command to use the response code instead of the presence of IDm.


iceman1001:
@kormax   for you to review

kormax:
I'd say, this PR is an improvement.

But from what I see, this command still needs some "love" to be given, potentially with subsequent PRs, before it can become fully useful:
- No way to provide GSK/USK, or at least the pre-derived card/reader subkeys;
- For GSK/USK approach, there would be need for a proper "KDF" - card/reader sub-key derivation, IDM XORing;
- Currently, there is no way to provide more than one authentication path and/or target node(s).

kormax (approved):

PR #3400, opened 2026-06-29
ASK CTS improvements

kormax:
This PR makes the following adjustments to the [ASK CTS protocol support to align it with available documentation](http://www.proxmark.org/files/Documents/13.56%20MHz%20-%20ASK%20CST/CTM512B-DS-V1.5.pdf):
* Increase field activation guard time;
* Handle ASK CTS tag responses without SOF/EOF by also detecting end of response when load modulation stops;
* Add probing for ASK CTS to `hf 14b info`:

```log
[usb] pm3 --> hf 14b info
[+] ASK C-Ticket
[+]            UID: AA BB CC DD
[+]   Product Code: 60
[+]  Facility Code: 02
```

With those changes, the observed success rate for detecting ASK CTS tags with either the `hf 14b reader` or `hf 14b info` has risen from ~10% all the way to 90%+. Not perfect, but more than usable for card type detection.

I've tested how NFC-B works for other card subtypes with the NO-SOF shortcut, and haven't noticed any noticeable regression with:
- ISO14443-4B;
- ST25TB;
- Innovatron.

### DIFF
--- a/armsrc/iso14443b.c
+++ b/armsrc/iso14443b.c
@@ -185,6 +185,8 @@
 # define ISO14B_BLOCK_SIZE  4
 #endif
 
+#define CTS512B_POWER_ON_DELAY_US 500
+
 // 4sample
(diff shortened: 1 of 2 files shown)


iceman1001:
Nice!

PR #3401, opened 2026-06-30
hf 14b dump: implement Standard ISO14443-B card dumping

thisiscamk:
## Summary

- The `ISO14B_STANDARD` branch of `CmdHF14BDump` previously printed *"Dumping Standard ISO14443-B tags is not implemented yet"* and returned immediately.
- This PR implements a two-phase read strategy and a new `print_std_blocks()` display function to close that gap.

## Approach

**Phase A — transparent READ BINARY**
Issues `00 B0 P1 P2 Le` with a 15-bit offset (ISO7816-4 §7.2.2), reading 58-byte chunks up to 16 KB. Covers simple ISO14443-B memory cards that have an implicit current EF.

**Phase B — SFI scan**
If Phase A yields nothing (e.g. the card returns SW=6986 "no current EF"), scans Short File Identifiers 1–30 using READ BINARY with P1 bit 7 set (ISO7816-4 §7.2.3). This reads each EF directly without requiring a prior `SELECT EF`, so the 6986 condition is bypassed. P2 carries the byte offset within the EF (0–255, the limit of SFI-form addressing).

Field state is tracked across phases so the WUPB+ATTRIB reconnect is paid only once.

**Fallback**
If both phases yield no data, actionable hints are printed pointing the user to `hf 14b apdu` or `hf 14b ndefread`.

## Display

New `print_std_blocks()` function renders the dump in 16-byte rows with offset (decimal/hex), hex, and ASCII columns, and supports `--dense` suppression of repeated rows — consistent with the existing SR block display style.

## Tested on

- Standard ISO14443-B card returning SW=6986 on direct READ BINARY — Phase B SFI scan produced a successful dump
- ISO14443-B card with no accessible EFs — fallback hints printed correctly

## Test plan

- [ ] Place a standard ISO14443-B memory card in field, run `hf 14b dump` — verify dump output and saved file
- [ ] Place a card that requires file selection (returns SW=6986 on READ BINARY), run `hf 14b dump` — verify SFI scan kicks in
- [ ] Run `hf 14b dump --ns` — verify output prints but no file is saved
- [ ] Run `hf 14b dump -z` — verify dense output collapses repeated rows
- [ ] Run `hf 14b dump -f myfile` — verify custom filename is used
- [ ] With no card in field — verify "no tag found" message

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added `hf 14b dump` support for Standard ISO14443-B tags: two-phase READ BINARY sweep (transparent EF, then SFI scan per ISO7816-4 §7.2.3) with a new `print_std_blocks` display (@thisiscamk)
 - Fixed `hf legic migrate` failing to parse the optional DCF argument as hex (@IdanHo)
 - Add support for parsing Finnish Helsinki Regional Transport (HRT) travel cards (@sanduuz)
 - Added standalone mode `HF_DOEGOX_COMMIT`: DESFire suspended commit without relay (@doegox)
--- a/client/src/cmdhf14b.c
+++ b/client/src/cmdhf14b.c
@@ -892,6 +892,48 @@ static void print_sr_blocks(uint8_t *data, size_t len, const uint8_t *uid, bool 
     print_footer();
 }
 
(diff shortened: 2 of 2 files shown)


iceman1001:
Looks like AI,    

This line makes me worried since we don't have a "hf emv scan".  

`OCBC EMV bank card (Phase B triggered; hint correctly directed user to hf emv scan)`

kormax:
Having no ability to test that this command implementation works ourselves, we'd like to know it if and how works for you, before we can accept it.

Please provide an example snippet demonstrating how the proposed new command works.
Feel free to mask the UID or sensitive data with 00/FF or any other nibble, if you wish so.

PR #3402, opened 2026-06-30
ASK CTS improvements

kormax:
This PR makes the following changes:
* Add command usage hints for ASK CTS and Innovatron -> Calypso;
* Update output format labels for ASK CTS tag type info;
* Add `hf 14b ctrdbl` and `hf 14b ctdump` commands for reading blocks and dumping the whole ASK CTS tag respectively.
 

```log
[usb] pm3 --> hf 14b ctdump

[=] --- ASK CTS / C-ticket -----------------------
[+]  UID          : AA BB CC DD
[+]  Product Code : 60 ( CTS512B )
[+]  Fab Code     : 02

[=] Reading 32 blocks

[=]  block#  | data  | ascii
[=] ---------+-------+------
[=]   0/0x00 | 00 00 | ..
[=]   1/0x01 | 00 00 | ..
[=]   2/0x02 | 00 00 | ..
[=]   3/0x03 | 00 00 | ..
[=]   4/0x04 | 00 00 | ..
* OUTPUT TRUNCATED *
```

```log
[usb] pm3 --> hf 14b

---------        ----------------------- General -----------------------
help             This help
config           Configure 14b settings (use with caution)
list             List ISO-14443-B history
---------        ----------------------- Operations -----------------------
apdu             Send ISO 14443-4 APDU to tag
dump             Read all memory pages of an ISO-14443-B tag, save to file
info             Tag information
* OUTPUT TRUNCATED *
---------        ------------------ ASK CTS / C-ticket ------------------
ctdump           Dump ASK CTS/C-ticket
ctrdbl           Read ASK CTS/C-ticket block
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added `hf 14b ctrdbl` and `hf 14b ctdump` commands for interacting with ASK CTS tags (@kormax)
 - Fixed `hf legic migrate` failing to parse the optional DCF argument as hex (@IdanHo)
 - Add support for parsing Finnish Helsinki Regional Transport (HRT) travel cards (@sanduuz)
 - Added standalone mode `HF_DOEGOX_COMMIT`: DESFire suspended commit without relay (@doegox)
--- a/armsrc/iso14443b.c
+++ b/armsrc/iso14443b.c
@@ -1818,7 +1818,7 @@ static int iso14443b_select_cts_card(iso14b_cts_card_select_t *card) {
     }
 
     if (card) {
-        // pc. fc  Product code, Facility code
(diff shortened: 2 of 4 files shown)


iceman1001:
We have a dump command improvement here.
https://github.com/RfidResearchGroup/proxmark3/pull/3401

I prefer for  generic commands instead of dedicated ones.   
ie:   `hf 14b dump`   should be able to autodetect what kind of card is on the device and act accordingly.

kormax:
I'd like to get clarification on what is the suggestion here.
Would you like me to wait for the first person to finish, and have this change applied after that, taking it in mind? Or do it before them, serving an example?

If it were up to me, I would have made CTS commands a whole separate command category altogether like it is for Calypso (and also would do the same for ST25TB). I'm not feeling entirely good about mashing all commands for various proprietary subtypes right in the `hf 14b` command set.

But I would agree that having a universal `hf 14b dump` command, which automatically dispatches to card-type-specific logic would be the best solution for user experience. It's worth doing regardless of how modules are split.

Please let me know how best to proceed.

iceman1001:
Is the other PR ok to merge?   You been playing with 14b/felica most of us.

PR #3413, opened 2026-07-11
fix CMD_HF_MIFARE_READBL not working in Lua scripts since 2024-09-24...

0xDRRB:
Hi,

The scripts `hf_mf_dump_luxeo.lua` (mine), `hf_mf_gen3_writer.lua`, `hf_mf_mini_dumpdecrypt.lua`, `hf_mf_tnp3_clone.lua`, `hf_mf_tnp3_dump.lua`, `hf_mf_tnp3_dump.lua`, and `hf_mf_tnp3_dump.lua` have not been working for some time now (since 96d462ace).

This is because `Command:sendNG()` and `Command:sendMIX()` in `client/lualibs/commands.lua` still parse the response header with the pre-2024 format string 'SSIsSLLL', reading status as a single signed 16-bit value.

My fix:
Unpack `status` and `reason` as two separate signed bytes (`cc`) instead of one signed short (`s`) and expose `Reason` in the table returned by `sendNG`.

The `hf_mf_dump_luxeo.lua` script (the only one I've tested) then works again without any changes needed.

### DIFF
--- a/client/lualibs/commands.lua
+++ b/client/lualibs/commands.lua
@@ -151,10 +151,10 @@ function Command:sendMIX( ignore_response, timeout, use_cmd_ack)
     arg1 = nil
     arg2 = nil
     arg3 = nil
-    local count, length, magic, status, crc, ng
+    local count, length, magic, status, reason, crc, ng
 
---         S    S       I      s       S    L     L     L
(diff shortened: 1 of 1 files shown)


iceman1001:
Good fix!

PR #3385, opened 2026-06-17
feat(emv): add EMV terminal emulator (lab research tool)

andrew867:
## ⚠️ FOR RESEARCH AND LAB USE ONLY — NO WARRANTY — PROVIDED AS-IS

This PR adds a complete **EMV payment terminal emulator** (`emv terminal`) to the Proxmark3 Iceman client. It is **not** a certified payment terminal and must only be used with **authorized EMV test cards** in a lab environment.

**Required reading before use:**
- [docs/emv-terminal-emulator/README.md](docs/emv-terminal-emulator/README.md) — legal disclaimer and overview
- [docs/emv-terminal-emulator/OPERATOR-GUIDE.md](docs/emv-terminal-emulator/OPERATOR-GUIDE.md) — day-to-day commands and workflows
- [docs/emv-terminal-emulator/SPEC-security-privacy.md](docs/emv-terminal-emulator/SPEC-security-privacy.md) — PIN handling, redaction, threat model

---

## Summary

Client-side terminal phase engine under `client/src/emv/terminal/` implementing EMV Book 3 phases: init → ODA → restrictions → CVM → TRM → TAA → CAA → online → complete. Includes host simulator (ARQC/ARPC), multi-scheme profiles, golden CI (no USB), session export/redaction, mock APDU replay, TCP mock acquirer, Lua bindings, PCAP trace export, and legal-use banner.

**Code location:** `client/src/emv/terminal/` (~40 modules)  
**Docs:** `docs/emv-terminal-emulator/` (specs, operator guide, test plans)  
**Fixtures:** `client/src/emv/test/fixtures/` (6 golden fixtures)

---

## MVP (M1–M6) — Terminal phase pipeline

| Feature | Command / module |
|---------|------------------|
| Full phase loop | `emv terminal run` |
| Single phase debug | `emv terminal step <phase>` |
| ODA (SDA/DDA/fDDA/CDA paths) | `phase_oda.c` |
| CVM — offline plain + enciphered PIN | `phase_cvm.c`, `emv terminal pin` |
| Restrictions + TRM + TAA | `phase_restrict.c`, `phase_trm.c`, `phase_taa.c` |
| Generate AC (CAA) | `phase_caa.c` |
| Online lab stub (EXTERNAL AUTH + AC2) | `emv terminal online`, `phase_online.c` |
| Issuer script tag 71 (pre-AC2) | `phase_scripts.c` |
| Session JSON export | `emv_term_session.c` |
[...]

PR #3417, opened 2026-07-18
docs(emv): EMV terminal emulator planning bundle (1/4)

andrew867:
Split of #3385. Docs only under `doc/planning/emv-terminal-emulator/`.
Merge order: 1 → 2 → 3a → 3b → 4.
RDV4 sim is 3.3V only (5V cards may not work).
Fork review draft: https://github.com/andrew867/proxmark3/pull/18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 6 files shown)


iceman1001:
Great,

However I fail to see how this is documentation of implementation ,  ie how to use.
It looks more of a set of AI generated documentation for how  it should implement the EMV terminal support,  rather than how to use it.

andrew867:
> Great,
> 
> However I fail to see how this is documentation of implementation , ie how to use. It looks more of a set of AI generated documentation for how it should implement the EMV terminal support, rather than how to use it.

:x ha well then, my sleepy eyes needed a break last night! That certainly was implementation docs and not user side docs, which pretty much should be of no use now haha

PR #3418, opened 2026-07-18
chore(emv): terminal profiles, scheme JSON, and golden fixtures (2/4)

andrew867:
Depends on PR 1. Public Interac test keys / synthetic fixtures only. Fork: https://github.com/andrew867/proxmark3/pull/19


iceman1001:
Great,

It look like a mix of resources for the EMV terminal and a bunch of AI generate instructions for itself on how to implement the EMV terminal.

Our documentation is usually about how to use a set of commands and how a standard might work.

PR #3427, opened 2026-07-24
Added support for emulating ST25TA tag (IKEA Rothult) with custom NDEF response

Sanduuz:
This allows emulating some ST25TA tags with custom NDEF responses (originally hardcoded response). This can be used with e.g. emulating IKEA Rothult lock ST25TA keycards. Similar logic was already available in standalone mode: https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/Standalone/hf_tcprst.c 

image


Sanduuz:
The failing check seems to be related to the jansson dependency, unrelated to the code changes presented here. Feel free to give feedback on the changes and I will update accordingly if need be. Otherwise this is ready for merge.

doegox:
Seems fine to me. Could you rebase it on current master for the CI?

iceman1001:
No,   the base  "hf 14a sim"  is  bad enough as it is.    We must have some sort of eload and the function needs to use the reserved EM part of bigbuf.

Sanduuz:
> No, the base "hf 14a sim" is bad enough as it is. We must have some sort of eload and the function needs to use the reserved EM part of bigbuf.

The most recent version implements `eload`, `eview` and `esave` commands similarly to others. This has been implemented in commit 0cb82f3f6d192f1a8af47680737295b4d5fa2dda.

I will rebase this later this evening to current master branch.

Sanduuz:
Alright, the commits have now been rebased and all checks seem to pass now. If there's something that still needs updating/changes, please do let me know.

doegox:
meanwhile, may I ask you to remove client/src/pm3line_vocabulary.h diff from the "make style" commit?
It's now unnecessary as the vocabulary is generated automatically.
This will solve the conflict.
Thanks!

iceman1001:
alrighty,  its time to merge.   there is the two commands.json / commands.md   which shouldn't be inside a PR since we do it separately using `make style` or `make commands`   but it's fine this time.

Only thing that got me worried was the dynamic allocation of memory.  it isn't a simple value to manipulate for different reasons.   With the latest rework of bigbuf / device side source it is less complicated and prone for stack smashing.

PR #3428, opened 2026-07-24
Allow mifare to recover after usb disconnection

mistial-dev:
If you fire up the mifare emulator, and turn the battery on the rdv4, it's possible to walk the proxmark3 to a target and continue the emulation.

When you return to the computer, USB won't enumerate.  This fix allows USB to re-enumerate so the client can connect.


mistial-dev:

Nah.  Not for something this small.

PR #3431, opened 2026-07-25
Legic - add wrbl repeat (fill) a byte until end of tag, fix segment parsing and improve bytes writing

YoungJules:
- Fix LEGIC clone segment parsing: the segment length high nibble uses 0x0F, not 0x07, so clone/update code reads the full 4-bit length field correctly.
- Replace the old one-shot write helper with a chunked legic_write_bytes_to_tag() helper that can write from any offset and reuse the same transport path.
- Make the write IV explicit at the call site instead of hard-coding 0x55 inside the helper.
- Add hf legic wrbl --repeat <byte> to fill from an offset to end-of-card without supplying a full data blob.
- Allow hf legic wrbl to accept either --data or --repeat, but not both.
- Update doc/commands.md so hf legic wrbl documents the new --repeat fill mode.

### DIFF
--- a/client/src/cmdhflegic.c
+++ b/client/src/cmdhflegic.c
@@ -65,12 +65,12 @@ static void legic_xor_with_crc(uint8_t *data, uint16_t cardsize, uint8_t crc) {
 }
 
 static bool legic_clone_update_segment_crcs(uint8_t *data, size_t bytes_read, const uint8_t uid[4]) {
-    // Decoded segmented payload starts at byte 22.
+    // Segment headers are parsed from offset 22 everywhere else in LEGIC decoding.
     size_t start = 22;
     bool found_segment = false;
 
     while (start + 5 <= bytes_read) {
-        uint16_t seg_len = (((uint16_t)data[start + 1] & 0x07) << 8) | data[start];
+        uint16_t seg_len = (((uint16_t)data[start + 1] & 0x0F) << 8) | data[start];
         if (seg_len < 5 || start + seg_len > bytes_read) {
             break;
         }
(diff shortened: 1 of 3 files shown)


YoungJules:
CI failed on a JSON library warning treated as an error on Windows. I pushed a small dependency fix to unblock the build.

client/src/cmdhflegic.c
```diff
@@ -850,12 +817,14 @@ static int CmdLegicWrbl(const char *Cmd) {
     CLIParserInit(&ctx, "hf legic wrbl",
                   "Write data to a LEGIC Prime tag. It autodetects tagsize to ensure proper write",
                   "hf legic wrbl -o 0 -d 11223344    -> Write 0x11223344 starting from offset 0)\n"
-                  "hf legic wrbl -o 10 -d DEADBEEF   -> Write 0xdeadbeef starting from offset 10");
+                  "hf legic wrbl -o 10 -d DEADBEEF   -> Write 0xdeadbeef starting from offset 10\n"
+                  "hf legic wrbl -o 35 --repeat 00   -> Repeat 0x00 from offset 35 to end");
```
iceman1001: Use "--fill"  instead

YoungJules: I thought of that first and decided on repeat to avoid confusion with -f already being used for files 🤠
No worries, I'll change it when I am back in range of my computer...

YoungJules: Done.

iceman1001:
I will merge but don't manually touch doc/command.*  files.  they are generated.

PR #3447, opened 2026-08-01
FeliCa SEAC improvements

kormax:
This PR makes the following changes related to FeliCa:
* Add command constants and descriptions for known SEAC commands;
* Improve stability of SEAC polling by configuring a proper timeslot parameter;
* Fold `hf felica seacinfo` into `hf felica info`;
* Add discovery of SEAC into `hf search`;
* Add SEAC command/response annotation to `hf felica list`.


iceman1001:
Nice,  but next time split things up.  The annotations is a typical its own PR,   addition of a command same.

PR #3449, opened 2026-08-01
Proxmark5 & Proxmark3 Hardware abstraction layer

xianglin1998:
Some important remarks for using and developing Proxmark5 have been documented in `doc/md/Development/Proxmark5.md`. This will prevent the loss of relevant information after the PR is merged and will make it easier to read.

If you encounter any problems, you can try to find solutions from remarks. or are trying to track the development progress of Proxmark5, you can get details progress and TODO from remarks.

[**VIEW REMARKS**](https://github.com/xianglin1998/proxmark3/blob/proxmark5/doc/md/Development/Proxmark5.md)


nieldk:
That’s one huge PR 😱

Build-breaking

bootrom/CMakeLists.txt 
lists version_pm3.c as a source but contains no rule to generate it (no mkversion/add_custom_command). Bootrom can’t compile. This matches your own note #16, confirmed, not just declared.

Defect

common/commonutil.c: the lookup table is defined as a non-static global:
`struct version_info_map_chip_type { int magic; uint32_t chiptype; } map[] = { ... };`

map gets external linkage and is exported from a utility TU that links into both firmware and client. The header only declares the two Check* functions, not map, so this is an unintended global export,  multiple-definition / symbol-collision risk at link. Should be static const. The struct should be static too.

armsrc/at32_unit_test.c is developer scratch: multiple #if 0 dead blocks, Chinese debug strings, hardcoded I2C addresses. Inside one dead block, pbuf can remain NULL and is then passed to I2C_BufferWrite(pbuf, sizeof(buf), ...), also a pbuf/sizeof(buf) size/pointer mismatch. Dead code today, but junk to merge.

xianglin1998:
> That’s one huge PR 😱
> 
> Build-breaking
> 
> bootrom/CMakeLists.txt lists version_pm3.c as a source but contains no rule to generate it (no mkversion/add_custom_command). Bootrom can’t compile. This matches your own note #16, confirmed, not just declared.
> 
> Defect
> 
> common/commonutil.c: the lookup table is defined as a non-static global: `struct version_info_map_chip_type { int magic; uint32_t chiptype; } map[] = { ... };`
> 
> map gets external linkage and is exported from a utility TU that links into both firmware and client. The header only declares the two Check* functions, not map, so this is an unintended global export, multiple-definition / symbol-collision risk at link. Should be static const. The struct should be static too.
> 
> armsrc/at32_unit_test.c is developer scratch: multiple #if 0 dead blocks, Chinese debug strings, hardcoded I2C addresses. Inside one dead block, pbuf can remain NULL and is then passed to I2C_BufferWrite(pbuf, sizeof(buf), ...), also a pbuf/sizeof(buf) size/pointer mismatch. Dead code today, but junk to merge.

Haha, I accidentally clicked the wrong thing.

* armsrc/at32_unit_test.c contains some test code left over from the development process, so I will set some unnecessary test logic to 'if 0' to disable it.
* Let me take a look at common/commonutil.c.
* We need to consider how to elegantly use CMake to manage armsrc/bootrom and fix the Makefile.

xianglin1998:
By now, you should be able to successfully compile armsrc and bootrom using CMake, which defaults to compiling for the RDV4 platform. If you need to compile for PM5, specify the following parameter for CMake: -DPLATFORM:STRING=PM5

The next steps: **to fix the makefile-based build and optimize makefile support for HAL.**

nieldk:
Sweet, will test / review some more this week. Looking forward to getting the actual device

doegox:
even bootrom with cmake doesn't use/find my arm-none-eabi-gcc and fails. (cd bootrom; mkdir build; cd build; cmake ..; make)

I also tried from the root path (mkdir build; cd build; cmake ..) and cmake fails on 
CMake Error at armsrc/CMakeLists.txt:217 (include):
  include could not find requested file:
    Standalone/StandaloneSource.cmake

What is the supposed way to compile bootrom and fw with cmake ?

nieldk:
> even bootrom with cmake doesn't use/find my arm-none-eabi-gcc and fails. (cd bootrom; mkdir build; cd build; cmake ..; make)
> 
> I also tried from the root path (mkdir build; cd build; cmake ..) and cmake fails on CMake Error at armsrc/CMakeLists.txt:217 (include): include could not find requested file: Standalone/StandaloneSource.cmake
> 
> What is the supposed way to compile bootrom and fw with cmake ?

Something like

cp common/default_version_pm3.c bootrom/version_pm3.c

cd bootrom
cmake -B build \
  -DCMAKE_C_COMPILER=arm-none-eabi-gcc \
  -DCMAKE_ASM_COMPILER=arm-none-eabi-gcc \
  -DPLATFORM=PM5 
cmake --build build

but it’s not complete yet as I see it

bootrom/CMakeLists.txt

needs something like

set(MKVERSION_OUTPUT_PATH ${CMAKE_CURRENT_SOURCE_DIR})
include(../tools/MKVersionScript.cmake)

xianglin1998:
> > even bootrom with cmake doesn't use/find my arm-none-eabi-gcc and fails. (cd bootrom; mkdir build; cd build; cmake ..; make)
> > I also tried from the root path (mkdir build; cd build; cmake ..) and cmake fails on CMake Error at armsrc/CMakeLists.txt:217 (include): include could not find requested file: Standalone/StandaloneSource.cmake
> > What is the supposed way to compile bootrom and fw with cmake ?
> 
> Something like
> 
> cp common/default_version_pm3.c bootrom/version_pm3.c
> 
> cd bootrom cmake -B build -DCMAKE_C_COMPILER=arm-none-eabi-gcc -DCMAKE_ASM_COMPILER=arm-none-eabi-gcc -DPLATFORM=PM5 cmake --build build
> 
> but it’s not complete yet as I see it
> 
> bootrom/CMakeLists.txt
> 
> needs something like
> 
> set(MKVERSION_OUTPUT_PATH ${CMAKE_CURRENT_SOURCE_DIR}) include(../tools/MKVersionScript.cmake)



> > even bootrom with cmake doesn't use/find my arm-none-eabi-gcc and fails. (cd bootrom; mkdir build; cd build; cmake ..; make)
> > I also tried from the root path (mkdir build; cd build; cmake ..) and cmake fails on CMake Error at armsrc/CMakeLists.txt:217 (include): include could not find requested file: Standalone/StandaloneSource.cmake
> > What is the supposed way to compile bootrom and fw with cmake ?
> 
> Something like
> 
> cp common/default_version_pm3.c bootrom/version_pm3.c
> 
> cd bootrom cmake -B build -DCMAKE_C_COMPILER=arm-none-eabi-gcc -DCMAKE_ASM_COMPILER=arm-none-eabi-gcc -DPLATFORM=PM5 cmake --build build
> 
> but it’s not complete yet as I see it
> 
> bootrom/CMakeLists.txt
> 
> needs something like
> 
> set(MKVERSION_OUTPUT_PATH ${CMAKE_CURRENT_SOURCE_DIR}) include(../tools/MKVersionScript.cmake)

* This variable(MKVERSION_OUTPUT_PATH) is no longer needed in the latest commit.
* I'm very sorry, I'm currently using CLion for CMake compilation and haven't tested command-line compilation. I'll test it now.
* It's definitely necessary to add an automatic discovery feature to the compiler; otherwise, it's very troublesome to manually pass it in every time. I'll try to add it.

xianglin1998:
> even bootrom with cmake doesn't use/find my arm-none-eabi-gcc and fails. (cd bootrom; mkdir build; cd build; cmake ..; make)
> 
> I also tried from the root path (mkdir build; cd build; cmake ..) and cmake fails on CMake Error at armsrc/CMakeLists.txt:217 (include): include could not find requested file: Standalone/StandaloneSource.cmake
> 
> What is the supposed way to compile bootrom and fw with cmake ?

### In the project root directory, use msys2 to compile the latest commit (test compilation of RDV4, armsrc & bootrom):

1. `mkdir build && cd build`

2. `cmake ..`  or  `cmake -DPLATFORM:STRING=PM5 ..`

3. `cmake --build .`
After compilation, the output file will be placed in `ProjectRoot/build/obj/`

### In the bootrom directory,  use msys2 to build bootrom:

1. `mkdir build && cd build`

2. `cmake -DCMAKE_TOOLCHAIN_FILE=../../tools/ToolchainForArm.cmake ..` or `cmake -DPLATFORM=PM5 -DCMAKE_TOOLCHAIN_FILE=../../tools/ToolchainForArm.cmake ..`

3. `cmake --build .`
After compilation, the output file will be placed in `ProjectRoot/bootrom/build/obj/`

### Build for armsrc
> Same with bootrom. change dir to armsrc to run cmd.

---

**The CMakeLists.txt file in the project root directory automatically includes ToolchainForArm.cmake, so `-DCMAKE_TOOLCHAIN_FILE=../../tools/ToolchainForArm.cmake` is not needed.**

---

### I'm not sure why, but the .gitignore file is ignoring *.cmake files. If there are any missing *.cmake files, I apologize and will commit them as soon as possible.

doegox:
@xianglin1998 nice progression :)
Could you migrate all standalone modes? Many still contain "fpgaloader.h" and "ticks.h" and might need some more changes.
See tools/build_all_firmwares.sh

xianglin1998:
ok, i will do that.🧐



---Original---
From: "Philippe ***@***.***&gt;
Date: Thu, Aug 6, 2026 20:43 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [RfidResearchGroup/proxmark3] Proxmark5 &amp; Proxmark3 Hardware abstraction layer (PR #3449)


doegox left a comment (RfidResearchGroup/proxmark3#3449)
 
@xianglin1998 nice progression :)
 Could you migrate all standalone modes? Many still contain "fpgaloader.h" and "ticks.h" and might need some more changes.
 See tools/build_all_firmwares.sh
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

xianglin1998:
@doegox All standalone modes have been migrated successfully.

doegox:
@xianglin1998 could you write all your remarks (your first message associated to this PR) into doc/md/Development/Proxmark5.md ? So we can keep track of them.

xianglin1998:
> @xianglin1998 could you write all your remarks (your first message associated to this PR) into doc/md/Development/Proxmark5.md ? So we can keep track of them.

OK.

Gobo42:
**Describe the bug**

On Proxmark 5 (AT32F435/437), with freshly built-and-flashed `PLATFORM=PM5` firmware (bootrom and OS versions confirmed matching each other and the just-built source), a Hitag S or Hitag µ reader command completely and permanently hangs the device — it stops responding to USB entirely and needs a physical power-cycle to recover. This happens via `lf search`/`auto` (which call into Hitag S/µ detection as part of their "auth LF and special cases" phase), but I've also isolated it down to a single, minimal repro: running `lf hitag hts reader` as the very first command after a fresh boot, with nothing else run before it, hangs the device immediately.

I spent a while bisecting this against the real hardware before filing, since the original symptom (`auto` producing a flood of `UART:: write time-out` / `timeout while waiting for reply`) touched both LF and HF search and I wanted to know how much of that was one bug versus several:

- 30 rapid-fire `hw status` calls back-to-back: clean. Rules out "too many commands too fast" as a general cause.
- Each of the 7 individual detectors `lf search`'s "auth LF and special cases" phase calls (EM4x05, T55xx, Hitag2, Hitag S, Hitag µ, EM4x50, EM4x70), run alone on a fresh boot: clean, *except* Hitag S and Hitag µ, which reliably hang the device the moment they're invoked — confirmed with `lf hitag hts reader` and `lf hitag htu reader` each run as the sole first command after a power-cycle.
- All 11 individual HF protocol readers `hf search` calls (ThinFilm, Topaz, LTO-CM, ISO14443-A, LEGIC, TEXKOM, Fuji/Xerox, ISO14443-B, ISO15693, iCLASS, FeliCa), run alone: all clean.
- `hf search` run entirely on its own (no `lf search` beforehand): completely clean, and it correctly read a real ISO14443-A/MIFARE Classic 1K tag near the antenna with no errors.

So this isolates to exactly two commands — Hitag S and Hitag µ reading — and HF is not independently affected at all. The HF-side timeouts visible in the original `auto` transcript below are downstream collateral damage: `auto` runs `lf search` first, which hits this hang, and everything that follows (including the entire `hf search` phase) is just a client timing out against an already-wedged device, not a separate bug. That also explains the spurious `Valid iCLASS tag / PicoPass tag found` in that same transcript — almost certainly a garbled/corrupted response from the wedged device being misread as valid, not a real detection (no iCLASS card was anywhere near the antenna).

I dug into why Hitag S/µ specifically hang, and found what looks like the actual root cause:

- `include/proxmark3_arm.h:25` unconditionally does `#include "at91sam7s512.h"`, regardless of `PLATFORM`. That header only really makes sense for AT91SAM7S-based hardware (RDV4 etc.), but because the include isn't gated, its register macros (`AT91C_BASE_TC0`, `AT91C_BASE_TC1`, ...) are available and compile cleanly even in a `PLATFORM=PM5` (AT32F435/437) build.
- `armsrc/hitagS.c:483` (inside `hts_send_receive()`, called by `hts_read_uid()` → `CMD_LF_HITAGS_UID`'s handler in `armsrc/appmain.c:1520-1522`) does:
  ```c
  while (AT91C_BASE_TC0->TC_CV < T0 * t_wait) {};
  ```
  a busy-wait directly against a raw AT91 Timer/Counter peripheral register, used for the precise bit-level timing Hitag protocols need. The same pattern appears again at `hitagS.c:435` and in `armsrc/hitagu.c:441`. On real AT91 silicon this is a legitimate hardware timer read. On AT32 silicon, that fixed physical address doesn't back a real, running Timer/Counter channel — it's not part of the AT32 HAL that was built out elsewhere in this branch (`common_arm/ticks/ticks_hw_at32.c` etc. exist for other subsystems, but this Hitag S/µ timing code was apparently never ported). If that address doesn't behave the way the loop expects (e.g. reads back stuck at a fixed value that never satisfies the condition), the loop never exits — a genuine infinite loop that blocks the whole firmware main loop, including USB servicing, which matches the total, power-cycle-required hang exactly.
- `armsrc/hitag2.c` has the same raw-register pattern (`AT91C_BASE_TC0`/`TC1` busy-waits, e.g. lines 157-172, 1235-1416) but `lf hitag reader` (Hitag2) did *not* reproduce a hang in my testing. I can't fully explain that difference without a JTAG/SWD debug probe on the board — my guess is Hitag2's reader path exits early via a different check when no tag is present, before reaching the timing-critical code, but I want to flag this rather than claim Hitag2 is safe — it shares the same underlying risky pattern and may fail under different conditions (e.g. with a tag actually present).

**To Reproduce**

1. Build with `PLATFORM=PM5` (`Makefile.platform` containing `PLATFORM=PM5`), `make clean && make -j`.
2. Flash bootrom + fullimage until `./pm3`'s version banner shows bootrom and OS both matching the freshly built source.
3. Connect: `./pm3`.
4. Run `lf hitag hts reader` (or `lf hitag htu reader`) as the very first command. No card needs to be present.
5. Observe `[!] timeout while waiting for reply`, and the client returns to the prompt.
6. Try any further command, e.g. `hw status` — it also times out, and stays timed out indefinitely. The device requires a physical power-cycle/replug to respond again.
7. (Equivalently: `lf search` or `auto` reproduce the same hang, since they call into the same Hitag S/µ detection internally, followed downstream by the HF phase inheriting the broken state — see the earlier transcript below.)

**Expected behavior**

Hitag S/µ reader commands should either work correctly on PM5 hardware or fail gracefully (the way they do on RDV4, and the way every other LF/HF protocol already does on PM5) without permanently hanging the device. More broadly, `include/proxmark3_arm.h` unconditionally pulling in AT91-specific hardware register definitions regardless of `PLATFORM` seems worth addressing structurally, since it lets AT91-only code silently compile (and appear to work, until it doesn't) on AT32 targets instead of failing at build time.

**Desktop:**
- OS: Debian (Linux 6.12)
- Branch: `proxmark5` @ `cab7ba19a` (`Iceman/proxmark5/v4.9237-13097-gcab7ba19a`), built with `PLATFORM=PM5`
- Version banner (bootrom and OS both freshly built, matching):

```
[ Proxmark3 ]

MCU....... Unknown
Memory.... 32 KB ( 971% used )
Target.... PM3 GENERIC

Client.... Iceman/proxmark5/v4.9237-13097-gcab7ba19a 2026-08-09 08:10:13
Bootrom... Iceman/proxmark5/v4.9237-13097-gcab7ba19a-suspect 2026-08-09 08:09:23 98f9b11a1
OS........ Iceman/proxmark5/v4.9237-13097-gcab7ba19a-suspect 2026-08-09 08:09:39 98f9b11a1
```

(`MCU`/`Memory`/`Target` showing `Unknown`/`971% used`/`PM3 GENERIC` instead of real AT32 values matches the project's own noted limitation that "client-side logic for identifying the chip capacity of the AT32F435 is not yet complete" — not separately reporting that here, just noting it for context.)

**Additional context**

Minimal isolated repro — `lf hitag hts reader` as the literal first command after a fresh power-cycle (sanitized: `user`/`host` substituted for my actual username/hostname):
```
(user@host)$ ./pm3 -c "lf hitag hts reader"
[=] Session log /home/user/.proxmark3/logs/log_20260810054314.txt
[+] loaded `/home/user/.proxmark3/preferences.json`
[+] execute command from commandline: lf hitag hts reader

[+] Using UART port /dev/ttyACM0
[+] Communicating with PM3 over USB-CDC
[usb|script] pm3 --> lf hitag hts reader
[!] timeout while waiting for reply
```
Every command after this (including plain `hw status`) times out until the board is physically power-cycled.

For contrast, `hf search` run completely on its own, fresh boot, no `lf search` involved at all — fully clean, correctly reads a real tag:
```
(user@host)$ ./pm3 -c "hf search"
...
[usb|script] pm3 --> hf search
...Searching for ISO14443-A tag...
[+]  UID: 6B DF 24 68   ( ONUID, re-used )
[+] ATQA: 00 04
[+]  SAK: 08 [2]
[+] Possible types:
[+]    MIFARE Classic 1K
...
[+] Valid ISO 14443-A tag found
...Searching for LEGIC tag...TEXKOM...Fuji/Xerox...ISO14443-B...ISO15693...iCLASS / PicoPass...FeliCa...
```
No errors, no timeouts, device fully responsive afterward.

Original `auto` transcript that first surfaced this (kept here since it shows the downstream HF-phase collateral damage described above — note the LF-phase timeouts appear first, before `hf search` even starts, and the spurious `Valid iCLASS tag / PicoPass tag found` despite no such card being present):
```
(08:11:44)user@host:[~/proxmark/proxmark5]$./pm3
[=] Waiting for Proxmark3 to appear...
[+] Communicating with PM3 over USB-CDC

[ Proxmark3 ]
Client.... Iceman/proxmark5/v4.9237-13097-gcab7ba19a 2026-08-09 08:10:13
Bootrom... Iceman/proxmark5/v4.9237-13097-gcab7ba19a-suspect 2026-08-09 08:09:23 98f9b11a1
OS........ Iceman/proxmark5/v4.9237-13097-gcab7ba19a-suspect 2026-08-09 08:09:39 98f9b11a1

[usb] pm3 --> auto
[=] lf search
[=] Checking for known tags...
[-] No known 125/134 kHz tags found!
[=] Searching for auth LF and special cases...
[!] timeout while waiting for reply
[!] timeout while waiting for reply
[!] (em4x50) timeout while waiting for reply
[!!] UART:: write time-out
[!] timeout while waiting for reply
[=] Couldn't identify a chipset
[=] hf search
[|] Searching for ThinFilm tag...[!!] UART:: write time-out
[!] timeout while waiting for reply
[-] Searching for Topaz tag...[!!] UART:: write time-out
[!!] UART:: write time-out
[!!] UART:: write time-out
[|] Searching for LTO-CM tag...[!!] UART:: write time-out
[!!] UART:: write time-out
[!!] UART:: write time-out
[!!] UART:: write time-out
[-] Searching for ISO14443-A tag...[!!] UART:: write time-out
[!!] UART:: write time-out
[!] Failed to get current device debug level
[|] Searching for LEGIC tag...[!!] UART:: write time-out
[-] Searching for TEXKOM tag...[!!] UART:: write time-out
[|] Searching for Fuji/Xerox tag...[!!] UART:: write time-out
[!!] UART:: write time-out
[-] Searching for ISO14443-B tag...[!!] UART:: write time-out
[!!] UART:: write time-out
[!!] UART:: write time-out
[!!] UART:: write time-out
[!!] UART:: write time-out
[!!] UART:: write time-out
[!] timeout while waiting for reply (first)
[!!] UART:: write time-out
[!] timeout while waiting for reply (first)
[!!] UART:: write time-out
[!] timeout while waiting for reply (first)
[|] Searching for ISO15693 tag...[!!] UART:: write time-out
[!!] UART:: write time-out
[-] Searching for iCLASS / PicoPass tag...[!!] UART:: write time-out

[+] Valid iCLASS tag / PicoPass tag found

[|] Searching for FeliCa tag...[!!] UART:: write time-out
[!!] UART:: write time-out
[!!] UART:: write time-out
[?] Hint: Try `hf iclass` commands

[usb] pm3 --> [!!] UART:: write time-out
[usb] pm3 --> exit
[!!] UART:: write time-out
```
(Device required a physical power-cycle to respond again after this.)

xianglin1998:
Note of [92b9a5e](https://github.com/RfidResearchGroup/proxmark3/pull/3449/commits/92b9a5ecd9ef08479d34221a6d2120206f76b118)

After two days of thought and testing, I attempted to implement a double-buffered capability similar to AT91 using EDMA + linked list transfer, but ultimately failed. I consulted with the FAE (Artery) and confirmed the following two points regarding linked list transfer:

1. DMA_SxM0ADDR is not updated by the linked list descriptor.
2. DMA_SxLLP is runtime immutable and does not change due to changes in the current transfer descriptor.

Therefore, for AT32, single buffering should be used instead of double buffering (AT32's processing speed is fast enough).

xianglin1998:
After two days of hard work, the HAL implementation for hitagU and hitagS was finally completed.
1. I moved all the timer-related operations that were originally in hitag_common to the ticks module and designed them as a universal interface, with different platforms implementing them separately.
2. A bug has been found in the FPGA firmware that causes cross-platform issues. This bug prevents the LoEdge module from detecting edge changes correctly. You can flash the following firmware to fix it.
  [PM5_FPGA_fix_loedge_bug.zip](https://github.com/user-attachments/files/31105593/PM5_FPGA_fix_loedge_bug.zip)
    > Download fpga bitstream via client: `hw fpga config -f PM5_FPGA_fix_loedge_bug.bin`
    > Again: The FPGA firmware needs to wait until the ARM firmware's PR is merged before it can be organized and released. Too busy hahaha.
3. Currently, only hitagS and hitagU have been fixed; the lf sea command may still freeze for other reasons.
4. To test hitagXXX:
    > hitagS: lf hitag hts reader 
    > hitagU: lf hitag htu reader

xianglin1998:
The lf sea command no longer causes devices and clients to freeze (although there may still be errors/warnings, it no longer affects usage).

The next step is to continue implementing HAL.

doegox:
The corresponding code has been merged into master, after some local rewrite of the history to avoid someone to end up with a bricked bootrom when doing a git bisect.
I'm closing the PR.

Issue #3468, opened 2026-08-14
hf mfdes chk pattern2b and dictionary key loading logic is broken


corollary-de:
**Describe the bug**
The logic behind loading keys for hf mfdes chk / CmdHF14aDesChk is broken.

**To Reproduce**
Use `hf mfdes chk -f <dictionary>` with a size of > 1024 lines.

**Expected behavior**
I am honestly not entirely sure what the intended behavior is so if someone could clarify that would be awesome, but the current behavior is certainly not intended.
The following code snippet is from the file `client/src/cmdhfmfdes.c` and the function CmdHF14aDesChk
```c
    for (uint32_t x = 0; x < app_ids_len / 3; x++) {

        ...

        if (dict_filenamelen) {
            if (verbose == false) {
                PrintAndLogEx(NORMAL, "d" NOLF);
            }

            uint32_t keycnt = 0;
            res = loadFileDICTIONARYEx((char *)dict_filename, deskeyList, sizeof(deskeyList), NULL, 16, &keycnt, endFilePosition, &endFilePosition, false);
            if (res == PM3_SUCCESS && endFilePosition) {
                deskeyListLen = keycnt;
            }

            keycnt = 0;
            res = loadFileDICTIONARYEx((char *)dict_filename, aeskeyList, sizeof(aeskeyList), NULL, 16, &keycnt, endFilePosition, &endFilePosition, false);
            if (res == PM3_SUCCESS && endFilePosition) {
                aeskeyListLen = keycnt;
            }

            keycnt = 0;
            res = loadFileDICTIONARYEx((char *)dict_filename, k3kkeyList, sizeof(k3kkeyList), NULL, 16, &keycnt, endFilePosition, &endFilePosition, false);
            if (res == PM3_SUCCESS && endFilePosition) {
                k3kkeyListLen = keycnt;
            }

            continue;
        }
    }
```
Two big issues stand out to me, the first is that every key type gets loaded with a size of 16 which is not correct, matching a few lines higher des should be 8 bytes, aes 16 bytes and k3k 24 bytes.
The second is that this advances the read position in the file for every key type *and* for every AID.
This doesn't seem like it would be intended behavior to me.
I would expect this type of program to try every key in the dictionary for every AID.
The current behavior is (loosely) to read 1024 des keys, then 1024 aes/2tdea keys, then 1024 3tdea keys for every AID,
so for a large dictionary file
- it will not be read in its entirety
- every AID gets a different portion of the dictionary
[...]

PR #3470, opened 2026-08-15
Fixes to key loading logic for mfdes chk

corollary-de:
This PR addresses Issue #3468.


corollary-de:
These are my proposed changes to the key loading logic from dictionary and pattern2b for the `hf mfdes chk` command. Refer to Issue #3468 for details.
Since this does change the behavior this is work in progress and i would appreciate feedback.

corollary-de:
The changes i am making are exclusively to `hf mfdes chk --pattern2b` and `hf mfdes chk -f` as follows:
### For the pattern2b:
- One minor fix where the 24 byte key was only filled with 20 bytes of the pattern.
- For every AID, the entire pattern space is now iterated over in blocks of `MAX_KEYS_LIST_LEN` (1024).

Considering that without a `--startp2b` the space is $2^{16}$, maybe a `--stopp2b` and potentially `--stepp2b` is warranted? I am unsure what the best solution is here, but i wanted to mirror the behavior of the dictionary logic for now.

### For the dictionary:
- One minor fix where every key was considered to be 24 bytes long, presumably resulting in skipping 8 and 16 byte keys. This did not pose an out-of-bounds write risk since the bounds were enforced in another way.
- For every AID, the entire dictionary is iterated over in separate blocks of `MAX_KEYS_LIST_LEN` for des, aes and k3k.
- If the end of file is reached for any type of key, that type is not loaded any further.
- If the end of file is reached for des keys, that means there are no more keys in the dictionary and it is considered complete.

From the logic of `loadFileDICTIONARYEx` it follows that longer keys are truncated to fit shorter keys, so a 24 byte key
in the dictionary will *also* be treated as a 16 byte and 8 byte key, and a 16 byte key will *also* be treated as a 8 byte key. This feels like unintended behavior but it is likely that people are already relying on this so i am not sure if it should be changed.

Currently i have a regression where it no longer prints the amount of keys loaded. However the previous print was inaccurate anyway, because it would show at most 1024. Adding this back in would be relatively straight forward and more of a design choice for when it should be printed.

corollary-de:
I am also not entirely sure what
```c
if (verbose == false) {
	PrintAndLogEx(NORMAL, "d" NOLF);
}
```
is aiming to achieve. Can this safely be removed?
And in `AuthCheckDesfire(..., (verbose == false));` the value of verbose is flipped. Is this intentional?

iceman1001:
> I am also not entirely sure what
> 
> ```c
> if (verbose == false) {
> 	PrintAndLogEx(NORMAL, "d" NOLF);
> }
> ```
> 
> is aiming to achieve. Can this safely be removed? And in `AuthCheckDesfire(..., (verbose == false));` the value of verbose is flipped. Is this intentional?

In the progress bar for `hf mfdes chk`  you see different statuses. ` -. +.  d` 
`d` would mean it was found a dictionary key,  i.e. a known key.
a generated key would have a `g` ?

iceman1001:
> The changes i am making are exclusively to `hf mfdes chk --pattern2b` and `hf mfdes chk -f` as follows:
> ### For the pattern2b:
> 
>     * One minor fix where the 24 byte key was only filled with 20 bytes of the pattern.
> 
>     * For every AID, the entire pattern space is now iterated over in blocks of `MAX_KEYS_LIST_LEN` (1024).
> 
> 
> Considering that without a `--startp2b` the space is 2 16 , maybe a `--stopp2b` and potentially `--stepp2b` is warranted? I am unsure what the best solution is here, but i wanted to mirror the behavior of the dictionary logic for now.
> ### For the dictionary:
> 
>     * One minor fix where every key was considered to be 24 bytes long, presumably resulting in skipping 8 and 16 byte keys. This did not pose an out-of-bounds write risk since the bounds were enforced in another way.
> 
>     * For every AID, the entire dictionary is iterated over in separate blocks of `MAX_KEYS_LIST_LEN` for des, aes and k3k.
> 
>     * If the end of file is reached for any type of key, that type is not loaded any further.
> 
>     * If the end of file is reached for des keys, that means there are no more keys in the dictionary and it is considered complete.
> 
> 
> From the logic of `loadFileDICTIONARYEx` it follows that longer keys are truncated to fit shorter keys, so a 24 byte key in the dictionary will _also_ be treated as a 16 byte and 8 byte key, and a 16 byte key will _also_ be treated as a 8 byte key. This feels like unintended behavior but it is likely that people are already relying on this so i am not sure if it should be changed.
> 
> Currently i have a regression where it no longer prints the amount of keys loaded. However the previous print was inaccurate anyway, because it would show at most 1024. Adding this back in would be relatively straight forward and more of a design choice for when it should be printed.

// pattern2b
1. 20 of 24 bytes only filled,   sounds like a bug
2. not sure what to say.

// dictionary
3.   we need 24 bytes fit all kinds of possible keys.   keylen is what determine actual length of key.
4.  not sure what to say
5.  not sure I follow ,  too little context

corollary-de:
> In the progress bar for `hf mfdes chk` you see different statuses. ` -. +. d` `d` would mean it was found a dictionary key, i.e. a known key. a generated key would have a `g` ?

Oh so it is supposed to print whenever it finds a key?
Currently there is no check for whether it found a key and prints every time, so I will consider that a bug and fix it too.

iceman1001:
the u32 vs size   usually you need to look at what the rest of the function call uses.   If it uses size_t ,  then the proper fix is to change the string formatter to  %zu

corollary-de:
thanks for pointing that out, u32 is correct in this case

iceman1001:
this one is still marked as WIP...

corollary-de:
I forgot i had put that in the title as well

PR #3471, opened 2026-08-17
Set the stringop-overflow flags before add_subdirectory(deps)

munzzyy:
`client/CMakeLists.txt` disables the `stringop-overflow` false positive at line 882, but
`add_subdirectory(deps)` runs at line 295. CMake copies the calling scope into a subdirectory
at the point of the call, so flags set further down never reach the bundled deps and jansson
still hits `-Werror=stringop-overflow` on its own.

Moving the block above the call fixes it. Counted on current master, from the generated
`flags.make` of every deps target:

```
before:  0 of 17 deps targets have -Wno-error=stringop-overflow
after:  17 of 17
```

Same edit in `experimental_lib/CMakeLists.txt`, same ordering. ubuntu-cmake and macos-cmake
were green either way, since their compilers do not emit this one.

The Windows linker workaround that this PR originally carried is gone: master already
downgrades the ARM linker in `windows.yml`.

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -292,6 +292,21 @@ if (NOT SKIPWHEREAMISYSTEM EQUAL 1)
     endif (WHEREAMI_INCLUDE_DIRS AND WHEREAMI_LIBRARIES)
 endif (NOT SKIPWHEREAMISYSTEM EQUAL 1)
 
+# These must be set BEFORE add_subdirectory(deps): CMake copies the
+# current variable scope into a subdirectory at that point, so flags set
+# further down never reach the bundled deps and jansson fails the
+# -Werror=stringop-overflow false positive on its own.
+# GCC 10 has issues with false positives on stringop-overflow,
+# let's disable them for now (cf https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92955, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94335)
+# beware these flags didn't exist for GCC < 7
+if(CMAKE_COMPILER_IS_GNUCXX)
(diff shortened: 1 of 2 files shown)


munzzyy:
Pushed a second commit. The pin works, and it uncovered a separate failure that was sitting behind it.

With the linker pinned, the proxspace job gets through the bootrom and everything after it ([run](https://github.com/RfidResearchGroup/proxmark3/actions/runs/32047495951)):

```
10. Pin ARM binutils to 2.46.1: success
12. Build:                      success
13. Test:                       success
15. Build btaddon:              success
16. Test btaddon:               success
20. Build cmake:                failure
```

Step 20 is a different bug, unreachable before because the build died at step 12:

```
client/deps/jansson/load.c:193:35: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
```

`client/CMakeLists.txt` disables that false positive at line 876, but `add_subdirectory(deps)` runs at line 295. CMake copies the calling scope into the subdirectory at that point, so jansson never sees it. Moving the block above the call fixes it. Counted which deps targets actually got the flag:

```
before: 0 of 17 deps targets have -Wno-error=stringop-overflow in flags.make
after: 17 of 17
```

Same edit in `experimental_lib/CMakeLists.txt`, same ordering. ubuntu-cmake and macos-cmake were green either way; their compilers do not emit this one.

Want the cmake ordering split into its own PR, or is one enough?

iceman1001:
Please update this one since its already handled in the windows.yml

munzzyy:
You are right, master already downgrades the ARM linker in windows.yml. Dropped that commit and rebased, so this is now only the cmake ordering change.

Re-counted on current master, from the generated flags.make of every deps target:

```
before:  0 of 17 deps targets have -Wno-error=stringop-overflow
after:  17 of 17
```

Same edit in experimental_lib.

PR #3472, opened 2026-08-17
fix(mifare): preserve readable Key B in sector trailer dumps

oSPANNERo:
## Problem

`hf mf dump` unconditionally replaces both Key A and Key B in sector
trailers with values from the supplied key file.

Key A is never readable from a MIFARE Classic sector trailer, so replacing
it with the known key is appropriate. Key B, however, may be readable
depending on the sector trailer access conditions.

When Key B is readable, the current behavior discards the value actually
returned by the card and replaces it with the value from the key file.

## Fix

Preserve the Key B bytes returned by the card when the sector trailer ACL
indicates that Key B is readable.

- Continue replacing Key A with the known value from the key file.
- Preserve card-returned Key B when it is readable.
- Continue substituting the known Key B from the key file when it is not
  readable.

## Testing

Tested against a MIFARE Classic 1K with sector trailer access bytes
FF 07 80 69.

Before this change, direct `hf mf rdbl` reads returned the actual readable
Key B values, while `hf mf dump --ns --keys` replaced those bytes with the
corresponding values from the supplied key file.

Seven affected sector trailers were tested.

After this change:

- all seven dumped Key B values matched the direct block reads;
- sector trailers where Key B was not readable retained the existing
  substitution behavior;
- a full client build completed successfully.

Test environment:

- Proxmark3 Easy
- AT91SAM7S512 / 512 KB
- `PM3GENERIC`
- Windows / MSYS2 UCRT64
- Based on Iceman `master` commit `2362cabcd`


oSPANNERo:
Additional regression testing completed against the patched `hf mf dump` implementation.

I created a purpose-built MIFARE Classic 1K test image with alternating sector-trailer configurations so both Key B cases could be exercised in a single dump.

### Test design

The card contains 16 sectors:

- **Even sectors** use trailer ACL `FF 07 80 69`
  - Key B is readable as data
  - Actual card-returned B value: `E0 E1 E2 E3 E4 xx`
- **Odd sectors** use trailer ACL `F7 8F 00 69`
  - Key B is protected and usable for authentication
  - Actual authentication key: `B0 B1 B2 B3 B4 xx`

The binary key file was deliberately constructed differently:

- All 16 Key A values are correct
- Odd-sector Key B entries are correct: `B0 B1 B2 B3 B4 xx`
- Even-sector Key B entries are intentionally WRONG: `D0 D1 D2 D3 D4 xx`

This makes the original bug immediately visible. The old behavior would overwrite the readable Key B bytes returned by the card with the bogus `D0...` values from the key file.

### Test command

```text
hf mf dump --ns --keys client/mfc1k_iceman_keyb_regression-key.bin
```

The test card was emulated using a Chameleon Ultra and read by a Proxmark3.

The complete 1K dump succeeded:

```text
[+] Succeeded in dumping all blocks
[+] time: 9 seconds
```

### Results

All readable Key B fields in even sectors preserved the actual bytes returned by the card:

| Sector | Block | Expected readable B | Dump result |
|-------:|------:|---------------------|-------------|
| 0  | 3  | `E0 E1 E2 E3 E4 00` | PASS |
| 2  | 11 | `E0 E1 E2 E3 E4 02` | PASS |
| 4  | 19 | `E0 E1 E2 E3 E4 04` | PASS |
| 6  | 27 | `E0 E1 E2 E3 E4 06` | PASS |
| 8  | 35 | `E0 E1 E2 E3 E4 08` | PASS |
| 10 | 43 | `E0 E1 E2 E3 E4 0A` | PASS |
| 12 | 51 | `E0 E1 E2 E3 E4 0C` | PASS |
| 14 | 59 | `E0 E1 E2 E3 E4 0E` | PASS |

All protected Key B fields in odd sectors were correctly populated from the supplied authentication key file:

| Sector | Block | Expected protected B | Dump result |
|-------:|------:|----------------------|-------------|
| 1  | 7  | `B0 B1 B2 B3 B4 01` | PASS |
| 3  | 15 | `B0 B1 B2 B3 B4 03` | PASS |
| 5  | 23 | `B0 B1 B2 B3 B4 05` | PASS |
| 7  | 31 | `B0 B1 B2 B3 B4 07` | PASS |
| 9  | 39 | `B0 B1 B2 B3 B4 09` | PASS |
| 11 | 47 | `B0 B1 B2 B3 B4 0B` | PASS |
| 13 | 55 | `B0 B1 B2 B3 B4 0D` | PASS |
| 15 | 63 | `B0 B1 B2 B3 B4 0F` | PASS |

None of the deliberately poisoned even-sector key-file values (`D0 D1 D2 D3 D4 xx`) appeared in the dump.

### Conclusion

This regression test exercises both branches of the fix:

- **Readable Key B** → preserve the bytes returned by the card
- **Protected Key B** → populate the trailer from the supplied authentication key

Both behaviors passed across all 16 sectors.

This is in addition to the earlier validation against a physical MIFARE Classic 1K card, where the patched client also preserved the known readable Key B values correctly.

oSPANNERo:
Follow-up negative/control testing completed using a second state-matrix key file against the same MIFARE Classic 1K test image.

This adds the two cases that were not covered by the earlier alternating readable/protected Key B regression.

### Additional control cases

#### Sector 2: Key A unavailable, Key B readable

Card configuration:

- Key A: `A0 A1 A2 A3 A4 02`
- ACL: `FF 07 80 69`
- Readable Key B field: `E0 E1 E2 E3 E4 02`

Supplied key file:

- Key A: `C0 C1 C2 C3 C4 02` intentionally wrong
- Key B: `D0 D1 D2 D3 D4 02` intentionally wrong

Observed behavior:

```text
[!] Trying with key B instead...

[-] Failed to read access rights for sector  2  ( fallback to default )
```

All four blocks in sector 2 then failed authentication/read:

```text
[-] Sector...  2 Block...  0 ( fail )
[-] Sector...  2 Block...  1 ( fail )
[-] Sector...  2 Block...  2 ( fail )
[-] Sector...  2 Block...  3 ( fail )
```

This is the expected result.

Even though the card's Key B field is configured as readable data, it cannot be used as an authentication key. Without a valid Key A, the trailer cannot be read and the readable B value cannot be obtained.

#### Sector 11: Key A known, Key B protected but supplied B is wrong

Card configuration:

- Key A: `A0 A1 A2 A3 A4 0B`
- ACL: `F7 8F 00 69`
- Protected Key B: `B0 B1 B2 B3 B4 0B`

Supplied key file:

- Key A: correct
- Key B: `D0 D1 D2 D3 D4 0B` intentionally wrong

The sector data was read successfully using Key A, and the reconstructed trailer was:

```text
A0 A1 A2 A3 A4 0B F7 8F 00 69 D0 D1 D2 D3 D4 0B
```

This is also the expected current behavior.

Because Key B is protected, its bytes cannot be read from the card. The `D0...` value shown in the reconstructed trailer comes from the supplied key file and should not be interpreted as a Key B value that was read from the card or successfully validated for authentication.

### Regression behavior remained correct

All other readable-B sectors continued to preserve the actual card-returned `E0...` bytes rather than the deliberately poisoned `D0...` values from the key file.

Examples:

```text
Sector  0 block  3  ... FF 07 80 69 E0 E1 E2 E3 E4 00
Sector  4 block 19  ... FF 07 80 69 E0 E1 E2 E3 E4 04
Sector  6 block 27  ... FF 07 80 69 E0 E1 E2 E3 E4 06
Sector  8 block 35  ... FF 07 80 69 E0 E1 E2 E3 E4 08
Sector 10 block 43  ... FF 07 80 69 E0 E1 E2 E3 E4 0A
Sector 12 block 51  ... FF 07 80 69 E0 E1 E2 E3 E4 0C
Sector 14 block 59  ... FF 07 80 69 E0 E1 E2 E3 E4 0E
```

This gives the regression test four distinct states:

| State | Result |
|---|---|
| Key A known + Key B readable | Actual card-returned B preserved |
| Key A known + protected Key B known | Supplied authentication B reconstructed |
| Key A unavailable + Key B readable | Sector cannot be read, readable B cannot be obtained |
| Key A known + protected Key B unknown/wrong | Sector readable via A, supplied B appears only as reconstructed trailer data |

All four behaved as expected with the patch.

oSPANNERo:
While running the state-matrix regression test, I noticed a separate behavior in `hf mf dump` that may be worth looking at independently of the readable Key B fix.

One sector was intentionally configured so that neither supplied key could authenticate.

For sector 2, the client correctly reported that it could not read the access rights:

```text
[!] Trying with key B instead...

[-] Failed to read access rights for sector  2  ( fallback to default )
```

During the dump, all four blocks in that sector also correctly failed:

```text
[-] Sector...  2 Block...  0 ( fail )
[-] Sector...  2 Block...  1 ( fail )
[-] Sector...  2 Block...  2 ( fail )
[-] Sector...  2 Block...  3 ( fail )
```

However, after the remainder of the card was processed, the command still reported:

```text
[+] Succeeded in dumping all blocks
```

The resulting displayed dump contained zero-filled data for the failed sector:

```text
[=]    2 |   8 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[=]      |   9 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[=]      |  10 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[=]      |  11 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```

So the per-block failure reporting appears correct, but the final success message does not seem to reflect that one or more blocks failed to dump.

It may be clearer for `hf mf dump` to either:

- report a partial/failed dump when any block read fails, or
- report something like `Completed dump with read failures` rather than `Succeeded in dumping all blocks`.

This appears unrelated to the readable Key B patch itself, but the controlled regression test made the behavior easy to reproduce.

PR #3473, opened 2026-08-19
Feature/add kgh tools

YoungJules:
Add a bespoke kgh_tools exe for Windows (you can also build it for Linux). 
The tool makes a couple of important Legic Prime operations available from a small, focused exe.  Target is the so-called KGH (KABA Group Header) cards.
See the doco in client/experimental_lib/KGH_TOOLS.md
This PR is based purely on publicly available information (with particular thanks to the lua scripts), careful observations, and some leaps of intuition!


doegox:
What are the changes in pm3line for ?

YoungJules:
> What are the changes in pm3line for ?

pm3line now uses fgets() instead of getline() when the client is built without readline/linenoise, because getline() was causing MinGW/ProxSpace build trouble in the static build. 
There's no functional change to the normal interactive path.

iceman1001:
Let's see. 
- the doc about SSH doesn't feel like it belongs to this PR. 

Not sure this is a Windows exe tool either?  I see a new sample source code for the experimental lib.
I think we added things via SWIG to make it automatic for exposure.

YoungJules:
> the doc about SSH doesn't feel like it belongs to this PR.

Yeah, it was just something I did as part of this and I wanted to document it so I could come back to it later. I can remove it and stick it on a separate PR?

YoungJules:
> Not sure this is a Windows exe tool either? I see a new sample source code for the experimental lib.

Again, you're probably right. It's really making a small and simple command line tool that just does 2 or 3 things, but fast and well. My goal was to make a Windows exe of the tool... so that's how the PR ended up named/described.  I had tried using the full pm3.bat, calling it from another program with dynamically-built scripts, but it didn't work well for several reasons... proxmark3 exe processes getting left behind after scripts failed, slow startup time of the full pm3 client, the need to parse text output that's designed for humans to read., having to build script files on disk..

iceman1001:
You can remove the irrelevant doc.   
the major thing is what Doegox pointed out with the getline vs fget.     Which needs to be verified on all different platform we use.

YoungJules:
I removed the doc. Do you want it on another PR or think it's not really relevant to proxmark at all?
Do you want me to do something more about the pm3line change?

doegox:
please propose the pm3line in a separate PR.
For the kgh_tool, I'd avoid mixing it with the examples of experimental_lib.
You can modify CMakeLists.txt for making static exe under mingw, but all kgh specific elements should be in a separate directory, separate CMakeLists.txt. If you find easier to be able to include some .cmake from experimental_lib, feel free to refactor the experimental_lib/CMakeLists.txt.
@iceman1001 what do you thing about sth like `/client/tools/khg_tools/` or even `/tools/khg_tools/`?

Also the name "kgh_tools" don't convey any useful meaning for the users. Try sth like kgh_legic_migrate

iceman1001:
or  "legic_" as prefix  since its kind of how we name things.    I do agree `tools/kgh`  as a seperate tool would be for the better.

YoungJules:
I have removed the pm3line change from this PR and will put it on a separate PR, as I still think it's a worthwhile portability change. I'm working on the suggestions for location and naming of the new tool... watch this space...

iceman1001:
We still waiting for the move to `tools/kgh` in this PR...

PR #3475, opened 2026-08-19
fix(pm5): run flash-mode check on HICK before raising PLL clock

nemanjan00:
The bootrom's check_goto_flash_mode() accesses the battery-powered-domain BPR registers (system_bpr_chk_clear -> at32_bpr_write_dt1) to read and clear the "enter flash mode" magic the OS sets on a software reset. Writing those registers stalls the APB when the CPU runs at the full PLL clock (288MHz, 144MHz APB) because the ERTC clock domain is effectively idle.

This path runs only on the warm software-reset flash entry (client "start flash"/reset), so the bootrom hung before usb_enable() and the device never re-enumerated into the bootloader until a physical USB replug (a cold power-on reset, where the magic-clear path does not run). Cold boot + holding the button always worked because it takes the button branch and never touches the BPR magic.

The OS writes the same magic on the slow HICK clock (system_simple_reset) and works. Fix: perform check_goto_flash_mode() before ConfigSystemClocks() so the bootrom does its BPR access on HICK too, then raise the PLL clock for USB.


xianglin1998:
This operation theoretically breaks compatibility with RDV4 because SpinDelayUs is called in check_goto_flash_mode, but the PWM clock is not enabled before ConfigSystemClocks is called.

See: https://github.com/RfidResearchGroup/proxmark3/pull/3449/changes/ee7be0e261f01c46af5d3c0961b047a0b71b88c4

Perhaps we need to find a better way to solve this problem.
:) needs to be compatible with both RDV4 and PM5.

nemanjan00:
I mean, it is just order of things? Macros?

xianglin1998:
Ideally, this should be solved within the HAL function.
This can reduce the use of some macros (if possible).

nemanjan00:
This is best I can offer right now

nemanjan00:
@xianglin1998, what about now?

xianglin1998:
> @xianglin1998, what about now?

I just woke up, let me take a look now.

xianglin1998:
This is a fantastic approach to fixing this bug, as it avoids requiring modifications to the bootrom.

That leaves only one operation we need to perform: expand this function into direct register operations (reducing the number of functions compiled into the bootrom and thus reducing the bootrom size).

common_arm/sys/sys_hw_at32.c
```diff
@@ -41,10 +41,34 @@ EMPTY_CALL(_fini)
...
+ * is write-once until a domain reset, so if it is already selected this is a
+ * no-op.
  */
 static void at32_bpr_write_dt1(uint32_t data) {
     CRM->apb1en_bit.pwcen = TRUE;
     PWC->ctrl_bit.bpwen = TRUE;
+
+    // Select and enable an ERTC clock so the write-protection register is
+    // writable at the full APB frequency.
+    crm_ertc_clock_select(CRM_ERTC_CLOCK_HEXT_DIV_20);
```
xianglin1998:
```suggestion
    CRM->cfg_bit.ertcdiv = ((CRM_ERTC_CLOCK_HEXT_DIV_20 & 0x1F0) >> 4);
    CRM->bpdc_bit.ertcsel = (CRM_ERTC_CLOCK_HEXT_DIV_20 & 0xF);
```

nemanjan00:
Could we merge this and leave that work to someone else? 😃

PR #3478, opened 2026-08-19
feat(pm5): `hf tune --rgb` / `lf tune --rgb` - antenna RGB as a tuning meter

nemanjan00:
Adds an opt-in `--rgb` flag to the continuous `hf tune` / `lf tune` commands that mirrors the antenna tuning level on the PM5 antenna RGB LED: blue = low, green = mid, red = high, tracking the on-screen bar so you can find coupling (e.g. an implant) by feel without watching the screen.

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -29,6 +29,7 @@
 #include "fpga_loader.h"
 #include "fpga_apis.h"
 #include "rssi_apis.h"
+#include "rgb_apis.h"
 #include "string.h"
 #include "printf.h"
 #include "legicrf.h"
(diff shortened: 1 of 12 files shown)


client/src/cmdhf.c
```diff
@@ -323,6 +352,7 @@ int CmdHFTune(const char *Cmd) {
     bool is_mix = arg_get_lit(ctx, 3);
     bool is_value = arg_get_lit(ctx, 4);
     bool verbose = arg_get_lit(ctx, 5);
+    bool use_rgb = arg_get_lit(ctx, 6);
```
doegox, line 344: Please add a check if (use_rgb && ! IfPm5) Print error & return.
Same for LF.

doegox: or warning and set use_rgb=false

client/src/cmdhf.c
```diff
@@ -298,6 +298,34 @@ int CmdHFSearch(const char *Cmd) {
     return res;
 }
 
+// Set the PM5 antenna RGB LED to a raw colour (no-op on non-PM5 devices).
+static void tune_rgb_raw(uint8_t r, uint8_t g, uint8_t b) {
```
doegox: it's globally useful, please make it a shared fct "set_rgb" in util.c

doegox (commented):
could you make the proposed changes? thanks!

armsrc/appmain.c
```diff
@@ -3617,6 +3618,13 @@ static void PacketReceived(PacketCommandNG *packet) {
             reply_ng(CMD_PM5_QC_TEST, QCTestPM5(&failed_item) ? PM3_SUCCESS : PM3_EFAILED, &failed_item, 1);
             break;
         }
+        case CMD_PM5_RGB_SET: {
+            // payload: r, g, b -> set the antenna RGB LED colour (used by tune --rgb).
+            const uint8_t *rgb = packet->data.asBytes;
```
iceman1001: We still prefer a stuct even if its just three bytes here.    It will help for readablity.

nemanjan00:
Done

doegox:
nice! thank you. tested on rdv4 & pm5

Issue #973, opened 2020-09-22
QT5 rabbithole: solve QWheelEvent version-related issues


doegox:
Tracking it in an issue because I feel it won't be solved in a snap...

QWheelEvent supports different methods depending on QT version.

Current mouse position:
* QT4: `x()`:OK `position().x()`:no such member
* QT5.12.8 (Appveyor WSL): `x()`:OK `position().x()`:no such member
* QT5.14.2 (my debian): `x()`:OK `position().x()`:OK
* QT5.15.0 (Appveyor PS): `x()`:deprecated `position().x()`:OK
* QT5.15.1 (Travis OSX):  `x()`:deprecated `position().x()`:OK

Wheel movement:
* QT4: `delta()`:OK `angleDelta().y()`:no such member
* QT5.12.8 (Appveyor WSL): `delta()`:status? `angleDelta().y()`: OK
* QT5.14.2 (my debian): `delta()`:OK `angleDelta().y()`:OK
* QT5.15.0 (Appveyor PS): `delta()`:deprecated `angleDelta()`:OK
* QT5.15.1 (Travis OSX):  `delta()`:deprecated `angleDelta()`:OK

doegox:
ok now that all versions have been collected thanks to https://github.com/RfidResearchGroup/proxmark3/commit/75dab301536cccfa4378c969867c8adc1ec806bb
a fix has been pushed https://github.com/RfidResearchGroup/proxmark3/commit/5322b0c4a18341dd2d17301ab45e12c7e0e398f7

iceman1001:
This is a nice improvement @doegox   !   Well done,   making the plot window better

PR #3479, opened 2026-08-19
fix(mifare): validate readable Key B before marking it found

oSPANNERo:
## Summary

`hf mf chk` could report six readable sector-trailer bytes as a recovered Key B without verifying that those bytes were actually usable as Key B.

This change validates readable Key B fields before marking them as found:

- firmware-side `chkKey_readb()` now reselects the tag, authenticates the extracted value as Key B, and requires a subsequent block read to succeed
- the client-side fallback now performs the same usable-Key-B check before setting `foundKey[1]`
- readable trailer bytes are still shown as `Data:` even when they are not a valid authentication key
- changelog updated

## Regression testing

Validated against three cases:

1. **Readable data-only Key B field**
   - trailer bytes: `E0E1E2E3E400`
   - direct Key B authentication fails
   - before: `hf mf chk` reported the value as found
   - after: Key A is found, Key B remains not found

2. **Readable and genuinely usable Key B field**
   - Key A: `7D45A5FB1913`
   - readable Key B: `A976669417CD`
   - Key B authentication and memory read succeed
   - after: both A and B are reported as found

3. **Ordinary protected Key B candidate**
   - Key A: `A0A1A2A3A401`
   - Key B: `B0B1B2B3B401`
   - existing protected-Key-B path still reports both keys correctly

Build tested with `make armsrc/all client/all` on PM3GENERIC / 512K hardware.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 ## [unreleased][unreleased]
 - Added Proxmark5 (PM5/AT32) reporting in `hw version` / `hw status`: correct MCU (AT32F437), flash size and `PM5` target, instead of the AT91-decoded "Unknown / 32 KB / PM3 GENERIC". The device now also reports its on-chip flash size, appended to the `CMD_VERSION` reply in a backward-compatible way (@nemanjan00)
 - Fixed `hf mf dump` preserving readable sector trailer Key B data instead of overwriting it with values from the supplied key file (@oSPANNERo)
+- Fixed `hf mf chk` validating readable sector trailer Key B data before reporting it as a found key (@oSPANNERo)
 - Changed `magic_cards_notes.md` - documented the USCUID-UL helper scripts (`hf_mfu_uscuid` / `hf_mf_uscuid_prog`) and how to set the tag signature, replacing the outdated "No implemented commands" note (@c-barron)
 - Fixed `hf_mf_uscuid_prog.lua` - corrected the script name shown in its usage text (@c-barron)
 - Added `hf felica seacauth1` command
--- a/armsrc/mifarecmd.c
+++ b/armsrc/mifarecmd.c
@@ -2008,6 +2008,25 @@ static uint8_t chkKey_readb(struct chk_t *c, uint8_t *keyb) {
         }
         mifare_classic_halt(c->pcs);
     }
+    if (res != 0) {
+        return res;
+    }
+
+    // A readable Key B field may contain data rather than a usable key.
+    // Verify the extracted value as Key B and require subsequent memory access.
+    if (iso14443a_fast_select_card(c->uid, c->cl) == 0) {
+        return 2;
+    }
+
+    uint64_t key = bytes_to_num(keyb, 6);
+    res = mifare_classic_authex(c->pcs, c->cuid, c->block, MF_KEY_B, key, AUTH_FIRST, NULL, NULL);
+    if (res != 0) {
+        return res;
+    }
+
+    res = mifare_classic_readblock(c->pcs, c->block, data);
+    mifare_classic_halt(c->pcs);
+
     return res;
 }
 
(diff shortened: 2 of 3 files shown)


oSPANNERo:
## Why I submitted a second Proxmark3 PR

While working on [ChameleonUltraGUI PR #973](https://github.com/GameTec-live/ChameleonUltraGUI/pull/973), I found an important MIFARE Classic edge case involving readable sector-trailer Key B fields.

When the access conditions make the Key B field readable, those six bytes are **not necessarily a usable authentication key**. They may simply be readable trailer data.

As part of the Chameleon work, I changed the logic so that a value extracted from the Key B field is only treated as a valid Key B if it can:

1. Authenticate as Key B
2. Successfully perform a subsequent memory operation

That investigation prompted me to look for the same assumption elsewhere.

My first Proxmark3 PR, [#3472](https://github.com/RfidResearchGroup/proxmark3/pull/3472), addressed a related but separate issue in `hf mf dump`, where readable Key B trailer data could be overwritten by values from the supplied key file.

While testing the Chameleon fix, however, I discovered another issue in Proxmark3.

`hf mf chk` could read six non-zero bytes from the Key B field and report them as a successfully recovered Key B **without first verifying that they actually worked as Key B**.

I reproduced that false positive using the same state-matrix test fixture and then found that Proxmark3 contained both firmware-side and client-side paths that made this assumption.

That led to the second Proxmark3 PR:

[#3479 - fix(mifare): validate readable Key B before marking it found](https://github.com/RfidResearchGroup/proxmark3/pull/3479)

The fix now validates a readable Key B value through actual Key B authentication and subsequent memory access before reporting it as a found key.

### In short

The Chameleon investigation exposed an important distinction:

> **Readable Key B field data is not necessarily a usable Key B.**

Once I identified that distinction in ChameleonUltraGUI, I was able to recognize and reproduce the same class of issue in Proxmark3, which resulted in PR #3479.

iceman1001:
I think the datasheet says if the "key B is readable then the key can not be used to authenticate"    
this will make any  authenticate using key B fail.

doegox:
"key B is readable then the key can not be used to authenticate" is truly respected only for NXP cards and a few clones, other manufacturers don't care.
IMHO a KeyB must be considered as valid if you manage to use it for auth, independently of any read success.
Also you were trying to read a single block while ACL could allow KeyB to read some blocks and not others.
Best is just to test if the auth is possible with KeyB or not.

iceman1001:
> "key B is readable then the key can not be used to authenticate" is truly respected only for NXP cards and a few clones, other manufacturers don't care. IMHO a KeyB must be considered as valid if you manage to use it for auth, independently of any read success. Also you were trying to read a single block while ACL could allow KeyB to read some blocks and not others. Best is just to test if the auth is possible with KeyB or not.

I would suggest to broaden it then,   if read out of keyB works,  good enough.   It doesn't matter if it Auths or not.

oSPANNERo:
> I think the datasheet says if the "key B is readable then the key can not be used to authenticate" this will make any authenticate using key B fail.

> "key B is readable then the key can not be used to authenticate" is truly respected only for NXP cards and a few clones, other manufacturers don't care.

This was the source of much consternation yesterday when working through my Chameleon PR. Glad to see it isn't just me being unlucky.

oSPANNERo:
> I would suggest to broaden it then, if read out of keyB works, good enough. It doesn't matter if it Auths or not.

Just to make sure I am on the same script: The thought is that differentiating between B being a key or data is not meaningfully useful for the end user. So as long as the string is read, stored, and displayed correctly no additional information is necessary?

Not being obstinate, just making sure I am working toward the right goal and not wasting folks' time with unnecessary re-reviews.

doegox:
I'd say like Iceman.
If you extract KeyB info, that's valid info to get and to complete a dump image.
Is it really a key or not, it's up to the user to figure it out, depending on the ACL, depending on the actual chip,...
I don't see the need *in mf chk* to make that distinction.

oSPANNERo:
Thanks for the clarification. That answers the question for me.

If `mf chk` is intended to report recovered Key B field contents without distinguishing whether those bytes are usable for authentication, then this PR is solving a problem that does not need to be solved.

I'll close it rather than add unnecessary complexity.

Appreciate the review and explanation.

iceman1001:
Thank you for the effort.  Keep going with the improvements!

PR #3481, opened 2026-08-20
doc(pm5): run `hf 14a read` before `hw fpga config` to power on the FPGA

nemanjan00:
`hw fpga config` uploads to the FPGA over JTAG, which needs the FPGA powered on. Add an `hf 14a read` before the config to bring the FPGA up (the existing read after flashing is kept to confirm the new image).


barefootdeveloper:
Thank you for this, I was confused why my FPGA patch was failing.

iceman1001:
this sounds more of a coping than a good solution.   The real fix would make `hw fpga config` always load it at start

nemanjan00:
On it, but, please, consider untested until I convert it back to PR

nemanjan00:
Can someone confirm this is not problem on older models?

nemanjan00:
Tested on pm5. I assume it is not issue on older platforms

Issue #3463, opened 2026-08-14
Fudan recovery for 7b UID tags


Lostfanman:
Fudan recovery (hf mf sen and fm11rf08s_recovery.py) not working correctly for 7b UID tags. 

[=] Session log C:\ProxSpace\pm3/.proxmark3/logs/log_20260723110933.txt
[+] loaded `C:\ProxSpace\pm3/.proxmark3/preferences.json`
[+] Using UART port COM3
[+] Communicating with PM3 over USB-CDC

  8888888b.  888b     d888  .d8888b.
  888   Y88b 8888b   d8888 d88P  Y88b
  888    888 88888b.d88888      .d88P
  888   d88P 888Y88888P888     8888"
  8888888P"  888 Y888P 888      "Y8b.
  888        888  Y8P  888 888    888
  888        888   "   888 Y88b  d88P
  888        888       888  "Y8888P"

  [ And so I watch, I wait!  ]

  [ Proxmark3 ]

    MCU....... AT91SAM7S512 Rev B
    Memory.... 512 KB ( 75% used )
    Target.... PM3 GENERIC

    Client.... Iceman/master/v4.21611-548-g005f4abb1 2026-07-08 13:15:43
    Bootrom... Iceman/master/v4.21611-548-g005f4abb1-suspect 2026-07-08 13:08:51 0a4cd9dcb
    OS........ Iceman/master/v4.21611-548-g005f4abb1-suspect 2026-07-08 13:09:47 0a4cd9dcb

[usb] pm3 --> hf mf info

[=] --- ISO14443-a Information -----------------------------
[+]  UID: 1D F9 44 E1 00 00 03
[+] ATQA: 00 44
[+]  SAK: 08 [1]

[=] --- Keys Information
[+] loaded 2 user keys
[+] loaded 63 hardcoded keys
[+] Sector 0 key A... FFFFFFFFFFFF
[+] Sector 0 key B... FFFFFFFFFFFF
[+] Backdoor key..... A396EFA4E24F
[+] Block 0.......... 1DF944E1000003001075D9F16EF37590 | .u..n.u.

[=] --- Fingerprint
[+] Fudan FM11RF08S-7B 1090

[=] --- Magic Tag Information
[=] <n/a>

[=] --- PRNG Information
[+] Prng....... weak
[+] Static enc nonce... yes
[?] Hint: Try `hf mf sen`

[usb] pm3 --> script run fm11rf08s_recovery.py
[...]

PR #3483, opened 2026-08-20
fix: `hf mf sen` init wrong 4 bytes of UID for authentication

nffq:
Issue: `hf mf sen` does not work for a FM11RF08S with 7-byte UID because it incorrectly stores the first 4 bytes of UID for crypto1 authentication.

Fix: Store the last 4 bytes of UID.

### DIFF
--- a/client/src/cmdhfmfsen.c
+++ b/client/src/cmdhfmfsen.c
@@ -2893,7 +2893,13 @@ int HFMFSENRecover(bool keep_nonces, bool no_oob, bool reader_mode, bool offline
         active_key = active_key_bytes;
     }
 
-    uint32_t uid = bytes_to_num(card.uid + card.uidlen - 4, 4);
+    uint32_t uid;
+    if (card.uidlen == 7) {
+        uid = bytes_to_num(card.uid + 3, 4);
+    } else {
+        uid = bytes_to_num(card.uid, 4);
+    }
+
     uint32_t nonce_count = FM11RF08S_SECTORS * 2;
     char activity[80];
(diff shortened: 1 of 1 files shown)


nffq:
Addresses #3463 
Hopefully the last PR for sen tool..

client/src/cmdhfmfsen.c
```diff
@@ -2893,7 +2893,7 @@ int HFMFSENRecover(bool keep_nonces, bool no_oob, bool reader_mode, bool offline
         active_key = active_key_bytes;
     }
 
-    uint32_t uid = bytes_to_num(card.uid, 4);
```
iceman1001, line 2896: meh..  that is confusing to see. 
make it readable with a if-statement.

nffq:
```
uint32_t uid;
if (card.uidlen == 7) {
    uid = bytes_to_num(card.uid + 3, 4);
} else {
    uid = bytes_to_num(card.uid, 4);
}
```

Is this good?

iceman1001: Excellent!  Thank you!

nffq: Done, feel free to merge

iceman1001:
and if you are  up for it ,  this is the simplest form.   Since its assigned only once it doesn't matter that we always assign one value regardless.   Simple and elegant.

```
    uint32_t uid = bytes_to_num(card.uid, 4);
    if (card.uidlen == 7) {
        uid = bytes_to_num(card.uid + 3, 4);
    }
```

PR #3495, opened 2026-08-22
Use HAL in pcf7931.c/SendCmdPCF7931()

xianglin1998:
It reuses the ticks functions encapsulated during hitag* abstraction; the abstraction is fantastic!!!
@iceman1001


iceman1001:
now we actually need to make this source code part work for real  ;)   I don't think this code part has gotten much needed love for years.   I just don't have the right tags / reader for it.    Maybe you have?

PR #3496, opened 2026-08-22
lfadc HAL is completed, some remark and unused code can be deleted.

xianglin1998:
1. The timer is not depended upon by the lfadc module, so it can be safely removed.

2. While the lfadc init function contains `StopTicks`, all places referencing the lfadc module themselves include `StartTicks`, so this can also be safely removed.

3. `StartTicks` is executed within the `lf_finalize` function. After checking, I found that this function is generally called at the end of instruction execution, so starting ticks is pointless.

4. The `lf_finalize` function is referenced by many modules, but `lf_init` is not called. I feel it would be more appropriate to move this function from the `lfadc` module to another location.

### DIFF
--- a/armsrc/lfadc.c
+++ b/armsrc/lfadc.c
@@ -214,8 +214,6 @@ void lf_wait_periods(size_t periods) {
 }
 
 void lf_init(lf_adc_init_mode_t init_mode, lf_adc_edge_mode_t edge_mode, bool ledcontrol) {
-    StopTicks();
-
     g_init_mode = init_mode;
     g_edge_mode = edge_mode;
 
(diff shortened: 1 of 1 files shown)


iceman1001:
stopticks in the beginning usually indicates that the function is gonna use a different clock with different configuration hence the start ticks in the end of the function.

doegox:
@xianglin1998 do we need to change the `SpinDelay(2)` into `SpinDelay(3)` ?

xianglin1998:
> @xianglin1998 do we need to change the `SpinDelay(2)` into `SpinDelay(3)` ?

Yes, in the previous code, that was 3ms. Because I previously refactored hitagS and hitagU using lfadc, during the refactoring process I discovered that if the delay was 3ms, hitagU would never be able to select a card, because hitagU enters TTF mode 3ms after power-on. So change to 2ms.
See commit(old): https://github.com/xianglin1998/proxmark3/commit/f7f0605207b28f6fe8d59897f0598b08886e6747

---

Why was it changed back to 3ms? Because before merging the HAL PR, I pulled the latest commit at the time for merging. During the merging process, I found some code changes in hittagS and hittagU that caused conflicts that could not be resolved temporarily. Therefore, I abandoned the refactoring of hittagS and hittagU using lfadc, and thus reset the delay to 3ms.

doegox:
wait why is this merged as such @iceman1001 ? I thought removing StartTicks(); will break all LF fcts relying on it.

iceman1001:
oo, I thought it was mergeable ?    We can revert it

doegox:
cf https://github.com/RfidResearchGroup/proxmark3/pull/3523

PR #3493, opened 2026-08-22
Improve `hf mf gdm*` commands

0x6r1an0y:
Fix #3458.

- Every `hf mf gdm*` command now probes the card and picks a working access path. Passing `--gdm/--gen1a/--wupa` explicitly keeps the previous single-attempt behavior.

- `gdmwipe` could flip the wakeup style in the middle of the command, causing every subsequent block write to fail with `wupGDM1 error`.

- `gdmwipe` never checked block write status, so a wipe that wrote nothing still reported "Wipe completed!". Failures are now counted and reported.

- `--wakestyle <gdm|gen1a|off>` in `gdmsetcfg` is now the single backdoor control, and the overlapping `--gen1 <on|off>` is removed. This also clears up the naming confusion between `--gen1` and `--gen1a`. The help now makes the distinction clear with the `access:` prefix for wakeup flags and `config:` for stored config flags.

Tested all 3 wakeup methods × all 9 commands (27 combinations). All good.

PR #3497, opened 2026-08-22
Update USCUID section in magic_cards_notes

0x6r1an0y:
Depends on #3493.

Docs only, no code changes.

- Cleans up the MIFARE Classic USCUID section of `magic_cards_notes.md`, which had misleading terminology and stale command names. 

- Document the full `gdm*` command set in a table, matching the client's current commands.

- `gdmcfg` → `gdmgetcfg`, `Alt Magic Wakeup` →`Gen4 Magic Wakeup`

### DIFF
--- a/doc/magic_cards_notes.md
+++ b/doc/magic_cards_notes.md
@@ -1017,7 +1017,7 @@ hf 14a raw -s -c 90FD111100
 
 ^[Top](#top)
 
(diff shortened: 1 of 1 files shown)


iceman1001:
Hopefully this matches the newly merged changes to magic commands

PR #3498, opened 2026-08-22
Correct NLEN log message and empty record example

0x6r1an0y:
`hf 14a ndefwrite` targets a NFC Type 4 Tag. The NDEF file there starts with NLEN (a two byte big endian message length), not a TLV container, so 0x03 is not needed.

The log message said otherwise, and the first help example (`0300FE`) was a Type 2 TLV empty record. On a Type 4 tag that parses as NLEN = 0x0300, i.e. 768 bytes. 

Fixed both, added a note.


iceman1001:
That is a AI comment which this PR doesn't need.    This is a minor simple fix.

0x6r1an0y:
fixed in 910f6ba

PR #3502, opened 2026-08-23
PM5: add hw bwmsetcap to provision BWM fuel-gauge design capacity

nieldk:
Follow-up to the BWM hw status support. That work exposed that the BQ27427 fuel gauge ships with a default (~1000+ mAh) profile, so Remaining capacity in hw status reads roughly 2× wrong for the fitted cell (e.g. ~1048 mAh reported on a 500 mAh pack) until the gauge's Design Capacity is programmed. This adds a command to do that.

hw bwmsetcap [--cap <mAh>] programs the BQ27427 Design Capacity (and matching Design Energy) for the fitted cell, defaulting to 500 mAh (the reference VXE 502540 pack). It writes via the gauge's CONFIG_UPDATE sequence, enter config-update, 32-byte data-memory block write, checksum, soft-reset out, per the BQ27427 TRM (SLUUCD5).

hw bwmsetcap                      # set to default 500 mAh
hw bwmsetcap --cap 500     # explicit

Design Capacity location is per the BQ27427 TRM: State subclass 82 (0x52), offset 6 → block addr 0x46, big-endian. (Differs from the older BQ27421's offset 10.

Assumes the gauge is unsealed (BQ27427 factory default). A sealed unit would need an unseal step before SET_CFGUPDATE; not included since the reference hardware ships unsealed.

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -171,7 +171,7 @@ static void print_pm5_battery_status(void) {
     } else {
         Dbprintf("  Charger MainCtl..... 0x%02x  %s", mainctl,
                  (mainctl & BWM_CHG_FET_DIS) ? _RED_("FET_DIS (VMIX off / shipping)")
-                                             : _GREEN_("power path enabled"));
+                 : _GREEN_("power path enabled"));
 
         I2C_BufferReadRaw(&f1,    1, BWM_CHG_REG_FAULT, BWM_CHG_ADDR); // 1st = latched history
         I2C_BufferReadRaw(&fault, 1, BWM_CHG_REG_FAULT, BWM_CHG_ADDR); // 2nd = current state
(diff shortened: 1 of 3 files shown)


client/src/cmdhw.c
```diff
@@ -1965,6 +2008,7 @@ static command_t CommandTable[] = {
     {"setmux", CmdSetMux, IfPm3Present, "Set the ADC mux to a specific value"},
     {"standalone", CmdStandalone, IfPm3Present, "Start installed standalone mode on device"},
     {"tia", CmdTia, IfPm3Present, "Trigger a Timing Interval Acquisition to re-adjust the RealTimeCounter divider"},
+    {"bwmsetcap", CmdBwmSetCap, IfPm3Present, "Set BWM fuel-gauge design capacity (PM5, run once after battery change)"},
```
iceman1001: shouldn't it be IfPm5present?

armsrc/appmain.c
```diff
@@ -220,6 +220,80 @@ static void print_pm5_battery_status(void) {
...
+    uint16_t cur = 0;
+    if (bq_read_design_cap(&cur) && cur == cap_mah) {
+        return true;    // already correct - do NOT run another CFGUPDATE cycle
+    }
+
+    uint16_t energy_mwh = (uint16_t)(((uint32_t)cap_mah * 37) / 10);   // ~3.7 V nominal
+
+    if (!bq_control(0x0013)) return false;                    // SET_CFGUPDATE
+    uint16_t flags = 0; int tries = 0;
+    do { WaitMS(50); if (!bq_flags(&flags)) return false; }
```
iceman1001: and tell your AI to make nice style formatted code by following  https://github.com/RfidResearchGroup/proxmark3/blob/master/CONTRIBUTING.md

nieldk: Right, thanks !!

client/src/cmdhw.c
nieldk: yes indeed!

PR #3505, opened 2026-08-23
fix: add workFlags check in `hf mf cload`

### LINKED ISSUE #3073: [idea] skip GDM magic block write when using hf mf cload --4k
**Is your feature request related to a problem? Please describe.**
I’m using a 4K card that supports the magic commands, but it is not GDM or USCUID.
Because of this, I can no longer use the hf mf cload --4k command — it always writes the CDM magic block, which breaks my workflow.

I was previously using Proxmark ver 4.18 without any issues, but after updating to the latest version I noticed this problem.

**Describe the solution you’d like**
I’d like an option that allows users to skip the GDM magic block writing step when running hf mf cload --4k.
Something like --skip-conf (or similar) would be great.

**Additional context**
Card type: 4K card supporting magic commands (not GDM / not USCUID)
Device: RDV4
Firmware: previously 4.18, issue appeared after updating to the latest build.

Image

0x6r1an0y:
Fix #3073


iceman1001:
the prefered style for checking flags is:  `((FLAG & myflag) == myflag)` 

If the intended addition is to limit the write to ONLY GDM,  I suggest add a text comment above since its not quite clear why this was added

0x6r1an0y:
Roger that.

0x6r1an0y:
Sorry, the problem is more complex than I initially thought. I’ll close this PR and open an issue to discuss it with you.

PR #3515, opened 2026-08-25
Added documents related to PM5.

xianglin1998:
The mainboard's technical manual, as well as the technical manuals for the antenna controller, button controller, and RGB controller, have all been completed.
@doegox @iceman1001


doegox:
Great, thanks!!

PR #3517, opened 2026-08-25
Rename ePassport launcher to ePassport.py to avoid case collision

xianglin1998:
The launcher script tools/ePassport/ePassport and the package directory tools/ePassport/epassport/ differ only in case, so on case-insensitive filesystems (Windows) they occupy the same path and a fresh clone reports the tracked script as deleted. Rename the launcher to ePassport.py and update the Makefile run target, the README, and the python3 -m epassport shim docstring.


xianglin1998:
@doegox @iceman1001

PR #3524, opened 2026-08-25
hw status: add battery health (BWM)

nieldk:
Adds full charge capacity and a battery health percentage to the PM5 battery status output, alongside the existing SoC and voltage. Health is the gauge's full-charge capacity as a percentage of the programmed design capacity.

The output shows the design capacity in use the gauge should be provisioned with _hw bwmsetcap_ .

> [#] Battery / BWM
> [#]   Charger MainCtl..... 0xc0  power path enabled
> [#]   Charger fault....... 0x00 (none)
> [#]   Charge status....... charging, power good
> [#]   Input limit......... 500 mA, VIN_DPM 4.52 V
> [#]   Charge enable....... enabled
> [#]   Charge current...... 256 mA
> [#]   Battery SoC......... 56 %
> [#]   Battery voltage..... 3929 mV
> [#]   Battery current..... 192 mA (charging)
> [#]   Remaining capacity.. 266 mAh
> [#]   Temp (gauge)........ 54.6 C
> [#]   Full charge cap..... 480 mAh (design 500)
> **_[#]   Battery health...... 96 % (needs a full cycle to be accurate)_**

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -142,6 +142,8 @@ static bool g_bwm_present = false;
 #define BWM_GAUGE_TEMP       0x02   // 0.1 K
 #define BWM_GAUGE_VOLTAGE    0x04   // mV
 #define BWM_GAUGE_REMCAP     0x0C   // mAh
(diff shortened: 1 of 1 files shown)


iceman1001:
time to clean up the AI comments in the source ....

PR #3528, opened 2026-08-26
Proxmark5 BLE support over the BWM

nieldk:
Adds wireless host <-> PM5 comms over BWM (Battery Wireless Module.)

Client (Linux): native ble: transport  `pm3 -p ble:<MAC>`. Built only when bluez is present.
Client: `--wait` on a tcp: port now listens/accepts instead of connecting out.

Tested: _hw status_, _lf search_ commands over BLE.

Linux-only, connect-by-address (find the MAC via bluetoothctl). BWM UART must be 460800.


client/src/comms.h
doegox, line 1: please don't let AI removing existing comments

common_arm/Makefile.hal
```diff
@@ -180,6 +180,10 @@ else ifeq ($(PLATFORM),PM5)
     SKIP_COMPRESSION = true # Skip data section compress. The new mcu has enough flash space.
     PLATFORM_DEFS = -DWITH_FLASH -DPM5 -DCHIP_AT32F435_37 -DWITH_BWM_STATUS -DWITH_PM5_PWR_LED -DWITH_PM5_AUTOOFF # TODO 暂时不要编译i2c -DWITH_SMARTCARD
                     # -DWITH_BWM_LOWBATT_BEEP ===== if you want audible alert ====
+    # BLE via BWM: AT32 UART4 (PA0/PA1, 460800 8N1) carries framed app_com to the
+    # ESP32. WITH_BWM_FORWARD builds the AT32 UART4 driver + framing shim. Do NOT
+    # enable WITH_FPC_USART_HOST here - that pulls in the AT91 usart.c and bricks USB.
+    PLATFORM_DEFS += -DWITH_BWM_FORWARD
```
doegox: is the code safe without BWM? else it should be a PLATFORM_EXTRAS=BWM in Makefile.platform

client/CMakeLists.txt
doegox, line 1: please update experimental_lib/CMakeLists.txt as well

nieldk: > please update experimental_lib/CMakeLists.txt as well

done :)

common_arm/Makefile.hal
nieldk: > is the code safe without BWM? else it should be a PLATFORM_EXTRAS=BWM in Makefile.platform

It's safe without the BWM, nothing waits on the module. 

That said, you're right it shouldn't be default-on. So I've gated it behind PLATFORM_EXTRAS=BWM (PM5-only), matching the BTADDON/FPC_USART_DEV pattern. Build BLE with make PLATFORM=PM5 PLATFORM_EXTRAS=BWM.

PR #3530, opened 2026-08-26
Update README.md / Clarify iCopy-X

quantum-x:
Tweaked README.md to reflect the updated iCopy-X situation.  Devices running the Open Source firmware are 100% compatible with latest iceman firmware / no DRM / etc.
@doegox should be familiar with the project.

Thanks kindly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,8 @@ We define generic Proxmark3 platforms as following devices.
   - Proxmark3 X
     - **Note**: Community tested
     - **Note**: unknown device hw
+  - iCopy-X
+    - **Note** Compatible after installation of [iCopy-X Open Source firmware](https://github.com/lab-401/icopy-x)
 
 **Not supported**
  - ⚠  Proxmark Evolution (EVO) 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,10 +114,6 @@ We define generic Proxmark3 platforms as following devices.
    - **Note**: Company have disappeared, leaving their customers in the dark.
 
 **Experimental support**
- - ⚠  iCopy-X
```
doegox, line 115: Keep that section with "iCopy-X with original firmware"

README.md
```diff
@@ -103,6 +103,8 @@ We define generic Proxmark3 platforms as following devices.
   - Proxmark3 X
     - **Note**: Community tested
     - **Note**: unknown device hw
+  - iCopy-X
+    - **Note** Compatible after installation of [iCopy-X Open Source firmware](https://github.com/lab-401/icopy-x)
```
doegox: Insist "Compatible **only** after installation...

quantum-x:
Hi @doegox Added clarity, thanks for the suggestion.

iceman1001:
Does this mean the open source version doesn't use DRM anymore?

doegox:
Yes, see "DRM bypass" in the code and for the firmware device binding itself, it's jailbroken and firmwares are not linked anymore to serials.
(code is compatible with DRM-protected tags and normal tags, so no need to get DRM-protected tags but it's still compatible)

PR #3531, opened 2026-08-26
Add my found NFC keys

ajabep:
I sorted a bit the keys of the cards I collect since years, so, if it may help anyone, let's share the keys :D

Note: The [OV-Chipkaart](https://en.wikipedia.org/wiki/OV-chipkaart) keys should be unique per card. However, I don't have any evidence about this claim. The card corresponding to these keys is no longer in use.


doegox:
Thanks but it's counterproductive to add diversified keys in the repo dictionary.

PR #3539, opened 2026-08-27
client: build the tab completion vocabulary at runtime from the command tree

Msprg:
## Problem

Tab completion in the client uses a generated table, `client/src/pm3line_vocabulary.h`, refreshed by hand with `make commands`. It drifts from the real command tables:

- new commands are missing until someone regenerates (currently e.g. `hw bwm*`, `hf mfu ndefwrite`, `nfc encode`, `smart pps`), removed ones linger (`hf waveshare` on a build without GD),
- the `offline` flag is a snapshot of `IsAvailable()` taken on the machine that ran the generator, so it depends on that machine's platform — `IfPm5()` returns true offline on `PLATFORM=PM5` builds, so `hw qc_pm5` flips between `0` and `1` depending on who last regenerated,
- when a device is connected, completion offers *everything* (`hw lcd` on a PM5, `hw ant_pm5` on an RDV4), while `help` correctly hides what the device does not support.

## Change

Build the vocabulary at runtime from the live command tree instead, so it can never diverge from `help`:

- **cmdparser**: new `walkCommandsRecursive()`, a tree walk using a fourth internal sentinel (`XX_internal_command_walk_XX`) next to the three dump ones. It hands every leaf to a visitor as its `command_t` chain (ancestor categories + leaf). A dispatch counter detects entries that look like a category but have their own parser (`reveng`) and reports them as leaves.
- **pm3line_vocabulary.c** (new): dynamic vocabulary holding the `IsAvailable` predicates of each command *and* its ancestors, evaluated live at completion time — exactly the rule `CmdsHelp()` applies. `script run <relpath>` entries come from the same directories `script list` scans, sorted, including subdirectories with the path `script run` actually resolves (`script run examples/example_cmdline.lua`; the old table offered a bare basename that could not run).
- **pm3line.c**: readline and linenoise completers consume the live vocabulary. The walk runs with output disabled so category handlers stay silent, and it is skipped in builds without a line editor.
- `pm3_help2list.py` and the header regeneration in `make commands` are removed. `doc/commands.md` / `commands.json` generation is unchanged.

Net: `13 files changed, 387 insertions(+), 1379 deletions(-)` — the 1150-line generated table is gone.

### Behaviour change to be aware of

Completion now matches `help` in both directions. Entries whose category is hidden by `help` are no longer offered (offline: `mem help`, `mem spiffs help`, `usart help` — they still run if typed, as before), and when connected, commands the attached device cannot run are no longer offered. Everything the old table offered offline is still offered, and the missing commands above now appear.

## Testing

- Full sweep over all 98 categories comparing `help` output to `<cat> <TAB><TAB>` via a pty: offline 95/98 identical (the 3 are the intentional `mem`/`usart` cases), connected to a Proxmark5 97/98 (`usart`).
- Live device swap in one session: PM3 Easy connected (22 `hw` entries) → link lost (4) → `hw connect` to a PM5 (29); completion followed each change without restarting.
- Builds: readline (make), linenoise (`SKIPREADLINE=1`), CMake client, `experimental_lib` (LIBPM3), `PLATFORM=PM5` and default RDV4 platform; `-Wall -Wextra -Wshadow -Wconversion` and `gcc -fanalyzer` clean; ASAN build clean on all completion paths; `make style` clean; `make client/check` passes (except `hf waveshare`, which needs libgd locally).
- `--fulltext` / `-m` dumps unchanged; `client.exe.delay` is not paid per walked category (walk sentinel is handled before the delay).

Not tested here: Windows/ProxSpace and macOS builds — the new file only uses `scandir`/`alphasort` (already used by `fileutils.c` with the bundled `scandir.h` shim) and `PATHSEP`.


doegox:
nice! thank you

doegox:
merged manually

Msprg:
Thanks a bunch. Github didn't send me mail notifications about this issue so I got scared to see it suddenly vanish lol. Thx for the merge ❤️

doegox:
yeah there were merge conflicts so it's easier for me to handle them locally

Msprg:
@doegox Hello, I am planning a followup to this pr that should enhance the tab completion further, but the code changes are relatively large-scoped, and it's mostly written with ai assisted development. I still want to extensively test it by daily driving it, but I want to ask whether there would be interest in getting this ai slop in, or you'd want me to get the code to higher quality first?

doegox:
using AI is not a problem in itself, but it should be well tested and the code (and comment!) in line with the rest of the code style. large scope when it's somehow self-contained is easier than when the AI touches many files and existing functions at once... You're trying to get autocompletion on the options?

Msprg:
That's right, I am trying to get actual proper autocompletion on all the subcommands. Right now, the autocompletion tends to end pretty early when it just begins to list files in the directory.

One of the hardest things is to nail down the behavior / UX - getting what pressing tab does in line with what a user would expect the tab to do, while feeling consistent. Feel free to test it out from my fork https://github.com/Msprg/proxmark3/tree/enhanced-tab-autocomplete

PR #3546, opened 2026-08-29
PM5: prevent HF field-related lockups and stale field state

nieldk:
Three fixes for Proxmark5 (AT32 has no watchdog, so any stalled SSC wait hard-locks the device until a power cycle):

armsrc/felica.c: bound the FeliCa reader SSC TX/RX waits with an FPGA-independent (GetTickCount) timeout; a stalled ISO18092 mode now bails gracefully with a debug reason instead of wedging.
armsrc/iso14443b.c, armsrc/iso15693.c: same backstop on the 14b and 15693 TX waits.
armsrc/appmain.c: force the field off after FPGA config finish, so flashing no longer leaves RF energized (no manual hf 14a reader --drop needed).


iceman1001:
You have removed some of my ongoing changes.   Please don't. 

And the fixes for felica is this directed to Pm5?  then it should be guarded by  ispm5 until  pm5 gets an update for their felica fpga mode.

PR #3549, opened 2026-08-29
Complete `hf mfu format` command

0x6r1an0y:
Formats a MIFARE Ultralight / NTAG tag for NDEF: writes the Capability Container and an empty
NDEF message. Needed before `hf mfu ndefwrite` will touch a tag. `nfc type2 format` alias added.

```
hf mfu format
hf mfu format -v
hf mfu format --erase
hf mfu format -k FFFFFFFF
hf mfu format -d E1101200 --force
```

Per-type values and data sheet citations: `doc/mfu_ndef_format_notes.md`.


iceman1001:
Nice,  
how ever this shouldn't be in "hf mfu"   it should only be in  "nfc type2"    since format would give the wrong impression you are wiping or preparing a tag normally.      Not to mention "hf mfu" is starting to have too many commands.

0x6r1an0y:
You're right about the naming. Would `hf mfu ndefformat` work instead? That matches what the other chip trees already do  `hf 14a ndefformat`, `hf mf ndefformat`, `hf mfp ndefformat` and `hf mfu` is currently the only one with `ndefread`/`ndefwrite` but no `ndefformat`. The `ndef` prefix makes the scope obvious without relying on the tree name for context.

`nfc type2 format` stays either way. If you'd still rather have it only there, I'll drop the `hf mfu` side.

iceman1001:
nay you are right.  In the name of current consistency  if you can rename it  `ndefformat`  I will merge it. 

then if you are up to it,    all  ndef functions  should be under `nfc`  in its own source code files.

0x6r1an0y:
Renamed to `ndefformat`, pushed.

About moving everything under `nfc`, happy to do it, but it's bigger than this PR. Should I do just `hf mfu`, or all of them(`hf 14a`/`hf mf`/`hf mfp`) ? It can be a separate PR so this one isn't held up.

PR #3550, opened 2026-08-30
Fix ePassport viewer segfault on startup

pkilar:
## Problem

`tools/ePassport` dies with SIGSEGV before drawing anything, on every entry point — `--help` included. No error, no traceback, no window.

The minimum window size is pre-seeded into Kivy's `Config`:

```python
Config.set("graphics", "minimum_width", "760")
Config.set("graphics", "minimum_height", "520")
```

`WindowBase.__init__` is the only place Kivy reads those, so they are applied from inside `WindowSDL.create_window()` while `Window.initialized` is still `False`. If the window SDL has just made is smaller than the minimum, SDL resizes it on the spot; the resize comes back through the SDL event filter into `EventLoop.idle()` (`kivy/core/window/window_sdl2.py:280`), which runs the clock and re-enters `create_window()`. Still not initialised, that call runs `setup_window()` again and resizes again — unbounded recursion until the C stack is exhausted.

A scaled HiDPI session is enough to get there. With sdl2-compat on SDL3 under Wayland sizes come back in logical points, so the 800x600 inherited from `~/.kivy/config.ini` arrives as a 500x375 window, under the 760x520 minimum. Confirmed either side of the threshold: a 500x375 minimum starts fine, 501x376 segfaults.

Reproduced on Arch, Python 3.14.7, Kivy 2.3.1, sdl2-compat 2.32.70 / SDL 3.4.14, Wayland at 1.6x scale.

## Change

`epassport/app.py`:

- Drop the two `Config.set(...)` calls and the now-unused `kivy.config` import.
- Add `MIN_WINDOW_SIZE` and `apply_minimum_window_size()`, called from `_restore_window_size()` once the window exists. Applying it after creation is the fix — post-init the same resize takes Kivy's cheap already-initialised branch instead of re-running `setup_window()`. Growing the window to the minimum first only avoids a visible jump.
- `MIN_WINDOW_SIZE` also replaces the duplicated `760`/`520` literals in `_restore_window_size()`.
- `_install_kivy_logging()` now writes to `sys.__stderr__`.

That last one is a second fault on the same path, exposed by fixing the first. `kivy.logger` replaces `sys.stderr` with a stream that feeds whatever is written to it back in as a warning, and cautions about exactly this (`kivy/logger.py:629` — *"If any logging handlers output to sys.stderr they should be configured BEFORE this reconfiguration is done to avoid loops."*). The app attached a bare `logging.StreamHandler()` to Kivy's own logger, so any warning `_KivyNoise` did not drop answered itself until the recursion limit stopped it — with stderr as the broken part, so nothing legible came out.

Not theoretical: this machine logs `MTD: Unable to open device "/dev/input/event11"` at startup, which killed the app as soon as the segfault was out of the way. Verified directly against the old handler.

`tests/test_startup.py` (new): the suite never imported the app or built a window, which is why 254 tests passed against a tool that could not start. The new checks run in a subprocess or read source deliberately — an in-process import takes the whole session down with a segfault rather than failing a test.

## Behaviour change

- A window that opens smaller than 760x520 is now grown to it. Previously the constraint was imposed while the window was being built (and crashed); for a setup where it worked, the end state is unchanged.
- One Kivy warning is now emitted at startup and filtered: setting `Window.minimum_width` and `minimum_height` in sequence makes Kivy complain while only half the pair is set. Added to `_KivyNoise`, so the console is as quiet as before.
- Kivy warnings now reach the console instead of crashing the app. On a machine that emits any, output that was previously absent will appear.

## Testing

- `python3 -m pytest tests/` — 257 passed, 2 skipped (was 254 passed, 2 skipped).
- `./ePassport.py --help` exits 0 (was SIGSEGV).
- Plain `./ePassport.py`, `--dump` against a generated sample, and `-v` all start and stay up.
- `black --check` clean.
- Net diff: +160/−15 over two files, one of them new.

Not tested here: Windows/ProxSpace and macOS, and X11 rather than Wayland. Low risk — the change is Python-only, swapping two `Config.set` calls for two property assignments on the same Kivy object and `StreamHandler()` for `StreamHandler(sys.__stderr__)`. Neither is platform-specific, and the `sys.__stderr__ is None` case (pythonw, some frozen builds) is handled explicitly.

Not tested: hardware. Nothing here reaches the `pm3` path.

No CHANGELOG entry, since `tools/ePassport` was itself added in the current `[unreleased]` cycle — this fixes something that has not shipped. Happy to add one if you would rather have it.

## PR checklist

- [ ] Builds warning-free with gcc; also tried with clang — n/a, Python only, no compiled code touched
- [ ] `client/Makefile`, `client/CMakeLists.txt`, `client/experimental_lib/CMakeLists.txt` all updated — n/a, no client sources changed
- [ ] ARM firmware builds clean for RDV4 and PM5 — n/a, no `armsrc`/`bootrom`/`common_arm` changes
- [x] Platform-sensitive code reasoned about — tested on Linux/Wayland; untested platforms and why they are low risk are in Testing
- [x] Existing comments in touched files preserved — the one dropped described the `Config.set` calls this PR removes
- [ ] `doc/commands.md` / `doc/commands.json` regenerated — n/a, no client commands changed

### DIFF
--- a/tools/ePassport/epassport/app.py
+++ b/tools/ePassport/epassport/app.py
@@ -32,17 +32,6 @@ if not _VERBOSE:
 os.environ.setdefault("OPENCV_LOG_LEVEL", "DEBUG" if _VERBOSE else "ERROR")
 os.environ.setdefault("OPENCV_VIDEOIO_DEBUG", "1" if _VERBOSE else "0")
 
-from kivy.config import Config  # noqa: E402
-
-# Set the minimum size here rather than on Window: assigning Window.minimum_*
-# warns while the other half of the pair is still zero, whichever order it is
-# done in.
-# Screen space is worth being careful with: the passport page scales as a
-# unit, so the window can go a long way down before anything stops being
-# readable.
-Config.set("graphics", "minimum_width", "760")
-Config.set("graphics", "minimum_height", "520")
-
 from kivy.app import App  # noqa: E402
 from kivy.clock import Clock  # noqa: E402
 from kivy.core.window import Window  # noqa: E402
@@ -70,9 +59,27 @@ log = logging.getLogger("epassport")
 #: What the window manager shows.
 APP_TITLE = "ePassport"
 
+#: The smallest window the passport page stays readable in.
+MIN_WINDOW_SIZE = (760, 520)
+
 #: How many files a full dump writes, for the progress bar's denominator when
 #: EF_COM has not been read yet.
 DEFAULT_EXPECTED_FILES = 9
+
+
+def apply_minimum_window_size(window) -> None:
+    """Constrain the window, once it exists.
+
+    Setting this through Config instead makes Kivy apply it inside
+    create_window(), where the resize recurses until the stack gives out.
+    """
+    minimum_width, minimum_height = MIN_WINDOW_SIZE
+    window.size = (
+        max(window.width, minimum_width),
+        max(window.height, minimum_height),
+    )
+    window.minimum_width = minimum_width
+    window.minimum_height = minimum_height
 
 
 class Root(BoxLayout):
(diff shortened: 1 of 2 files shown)


iceman1001:
Nice catch,   remove the AI comments and its all good to go

pkilar:
Done.

PR #3554, opened 2026-08-30
fix(eMRTD): move EF_SOD parser buffers off the stack

actuallysparky:
## Problem

On macOS ARM, `hf emrtd info` can reach and parse EF_COM successfully, then crash while parsing EF_SOD with `SIGBUS` / `Thread stack size exceeded` in `emrtd_parse_ef_sod_hashes`. The parser nests several `EMRTD_MAX_FILE_SIZE` (70 KB) stack buffers in a 512 KB worker thread.

## Change

Allocate the EF_SOD signature-extraction and hash-list scratch buffers as contiguous heap allocations, release them on every exit path, and report allocation failure using `PM3_EMALLOC`.

## Behaviour change

Reading an eMRTD no longer exhausts the macOS worker-thread stack during EF_SOD parsing. Allocation failure now returns an error instead of crashing.

## Testing

- `make client -j4`
- `make client/check`
- Manual macOS Apple Silicon read of an ISO/IEC 14443-B eMRTD through EF_SOD and data groups after BAC authentication

Not tested: Linux, Windows, or non-eMRTD hardware paths.

### DIFF
--- a/client/src/cmdhfemrtd.c
+++ b/client/src/cmdhfemrtd.c
@@ -1738,15 +1738,22 @@ static int emrtd_print_ef_dg12_info(uint8_t *data, size_t datalen) {
 }
 
 static int emrtd_ef_sod_extract_signatures(uint8_t *data, size_t datalen, uint8_t *dataout, size_t *dataoutlen) {
-    uint8_t top[EMRTD_MAX_FILE_SIZE] = { 0x00 };
-    uint8_t signeddata[EMRTD_MAX_FILE_SIZE] = { 0x00 };
-    uint8_t emrtdsigcontainer[EMRTD_MAX_FILE_SIZE] = { 0x00 };
-    uint8_t emrtdsig[EMRTD_MAX_FILE_SIZE] = { 0x00 };
-    uint8_t emrtdsigtext[EMRTD_MAX_FILE_SIZE] = { 0x00 };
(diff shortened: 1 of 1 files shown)


iceman1001:
Seems to be problem with the device if it can't handle 70kb stack...     Which device are  you running on?

actuallysparky:
I’m currently using a generic 512 KB PM3 clone and eagerly waiting for my legitimate PM5 to arrive in the next week or two.

This isn’t a 70 KB stack limit on the Proxmark device. The crash happens in the macOS client process while parsing EF_SOD.

EMRTD_MAX_FILE_SIZE is 70,000 bytes, and several buffers of that size were stack-allocated in nested parser calls. The caller has one buffer, emrtd_parse_ef_sod_hashes() has two, and its helper has five more. At peak, this can require approximately 560 KB of scratch storage on the same client thread stack, exceeding the 544 KiB stack allocated for the Qt worker thread by default on OS X.

To resolve this issue, those EF_SOD parser scratch buffers were changed to heap allocations with error handling. Alternatively, you could configure a larger stack for the Qt worker using QThread::setStackSize() before it starts.

iceman1001:
It was unclear but I was referring to your host OS.      So there is a 500kb limit in QT for threads.  Good to know,. 

Merged :)

PR #3557, opened 2026-08-30
Bugfix: stale ISO14443A scheduler delay

Antiklesys:
Internal Card-API code restarts the SSP clock before every RF APDU, resetting it to zero. ISO14443A retains NextTransferTime as an absolute timestamp from the prior clock instance, so each next transmit waits for stale time.


iceman1001:
minimize  those AI comments..  rewrite them to one , two lines max

iceman1001:
great catch!

PR #3566, opened 2026-08-31
Fix PM5 auto-off leaving the board unable to power back on

nemanjan00:
bwm_autooff_check() released the power latch on the first VUSB-absent sample, ~250ms after VBUS was pulled. With a BWM fitted that lands while the AW32001E is still transitioning into Battery Discharge Mode, where BATFET is only turned fully on as part of that transition. The board powered down mid-transition and could then not be brought back up by button or by USB insertion - it looked dead until the cell was physically re-inserted.

Require the absence to persist for 5s (20 consecutive 250ms samples) before acting, resetting the count whenever USB reappears. This is the grace period the function's own comment already described, and it also means a single spurious low reading on PC2 - a floating input, configured GPIO_PULL_NONE - can no longer power down a working device.

The 5s figure is empirical rather than derived from a datasheet settling time. Other users report auto-off working on their boards, so this makes a marginal path robust rather than fixing something broken everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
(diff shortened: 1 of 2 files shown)


nemanjan00:
Om, this was only partial fix for me. If it stays off longer than a min-two minutes, I need to unplug battery to have it power on again

Issue #3394, opened 2026-06-24
Fails to compile in Raspberry Pi 5 (aarch64)


pyllyukko:
Proxmark (version 8fbec33) fails to compile in Raspberry Pi 5 running Debian 13.5.

It fails with the following error:

```
[-] CC src/cmdmad.c
In function ‘parse_sector_ranges’,
    inlined from ‘CmdMADEncode’ at src/cmdmad.c:526:24:
src/cmdmad.c:470:30: error: writing 16 bytes into a region of size 8 [-Werror=stringop-overflow=]
  470 |             sectors[count++] = (uint8_t)s;
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
src/cmdmad.c: In function ‘CmdMADEncode’:
src/cmdmad.c:525:17: note: at offset [32, 40] into destination object ‘sectors’ of size 40
  525 |         uint8_t sectors[40] = {0};
      |                 ^~~~~~~
cc1: all warnings being treated as errors
make[2]: *** [Makefile:1166: obj/cmdmad.o] Error 1
make[1]: *** [Makefile:186: client/all] Error 2
make[1]: Leaving directory '/home/pyllyukko/git/proxmark3'
make: *** [Makefile:41: all] Error 1
```

bjornjulian00:
Alright, I just figured out why this is happening:

1. GCC 14+ versions now error out with some warnings where old versions would keep going.
2. ARM64 hosts cause this warning because GCC tries to vectorize the problematic function writes while x86 doesn't.
3. Thus, ARM64 hosts running GCC 14+ error out.

This is really easy to fix, just add "NOERROR=1" to your build flags, so:

`make NOERROR=1`

Or, if you don't want to blindly suppress all warnings, you can put this on line 547 (the parse_sector_ranges function) around the problematic bit:

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-overflow"
           sectors[count++] = (uint8_t)s;
#pragma GCC diagnostic pop
```

iceman1001:
sounds like we need a proper fix instead of using `#pragma`  to silence the issue

DidierA:
in `CmdMADEncode()`, replacing the line 
```
        uint8_t sectors[40] = {0};
```
with 
```
        uint8_t sectors[48] = {0};
```
is sufficient to suppress the warning (tested on ODroid-HC4 with cortex A55 inside (aarch64)).

[...]

PR #3571, opened 2026-09-01
fixes compile fail on aarch64 (issue #3394)

DidierA:
proposed fix for issue #3394


iceman1001:
Still feels like this is a workaround for a compiler issue.

DidierA:
I agree, but I lack the skill and motivation to dig into how gcc processes this and file a bug to them 😅 I tried to extract a minimum code base to reproduce it but did not succeed.

PR #3573, opened 2026-09-01
Decode ePassport EF_DG13, and read Poland's PESEL from it

pkilar:
## Problem

A Polish ePassport carries `EF_DG13` and no `EF_DG11`. The PERSONAL tab therefore had nothing to show, and the data page fell back to `not on chip` for the personal number — even though the document does carry one, in DG13.

## Change

`parse_dg13()` reads DG13 as tagged values and names none of them. ICAO calls DG13 "Optional Detail(s)" and assigns it no meaning, so nothing is inferred from a tag on its own. The `5C` tag list is skipped, since it enumerates what follows rather than being a field.

Poland is the one case worth naming. It puts the **PESEL**, its national identity number, in tag `5F70`. `is_pesel()` checks the length and the check digit — DG13 being issuer-defined, the checksum is what separates a PESEL from eleven digits that happen to share a tag. That interpretation is applied **only when the MRZ nationality is `POL`**, so the same tag on another state's document is read and carried, but not claimed as a personal number.

It slots in behind the two sources `personal_number` already had — a placement its own docstring anticipated:

```
DG11 → MRZ optional-data field → DG13
```

`personal_number_source` reports `DG13`, so the PERSONAL tab (`Personal number (DG13)`) and the data page caption say where the value came from instead of implying DG11.

## Behaviour change

- A document with DG13 but no DG11 now shows a personal number where it previously showed nothing, or `not on chip` on the data page.
- DG13 is decoded for every document; only `POL` gets `5F70` interpreted.
- Nothing changes for documents that carry the number in DG11 or the MRZ.

## Testing

- `python3 -m pytest tests/` — 323 passed, 2 skipped (was 312 passed, 2 skipped).
- Confirmed against a real Polish passport: the 11 digits validate as a PESEL, their first six match the MRZ date of birth under PESEL's century-offset month encoding, and the gender digit agrees with the MRZ sex field.
- Two US documents are unaffected — they keep taking the number from the MRZ — and an older Polish one with none of the three sources shows no row, as before.
- The test fixture is a synthetic checksum-valid number, not anybody's.
- `black --check` clean.

The caption omits a "from" deliberately: the column is dp(190), which holds the 167px of `Personal number (DG13)` but not the 203px the longer wording needs, and a caption that wraps drops the source onto a line of its own.

One thing worth a reviewer's eye: the PERSONAL tab drops the value when it is the `not on chip` placeholder, because every other row in that tab omits absent fields rather than labelling them. Without that, switching to the resolved property made the placeholder appear as a row where previously there was none.

Not tested here: Windows/ProxSpace and macOS. Python only, no platform-specific calls.

## PR checklist

- [ ] Builds warning-free with gcc; also tried with clang — n/a, Python only
- [ ] `client/Makefile`, `client/CMakeLists.txt`, `client/experimental_lib/CMakeLists.txt` all updated — n/a, no client sources changed
- [ ] ARM firmware builds clean for RDV4 and PM5 — n/a, no firmware changes
- [x] Platform-sensitive code reasoned about — none; parsing and UI only
- [ ] `doc/commands.md` / `doc/commands.json` regenerated — n/a, no commands changed
- [x] CHANGELOG entry added under `[unreleased]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Added `tools/ePassport` - EF_DG13 is decoded, and a Polish document's PESEL is shown as the personal number when it carries no EF_DG11 (@pkilar)
 - Fixed `hf 14b` - a card asking for a waiting time extension of 4 or more left the timeout at zero, so any ISO14443-B card needing time to answer looked like it had stopped responding (@pkilar)
 - Fixed `hf mfdes chk` - now handles `-k` user supplied keys properly (@iceman1001)
 - Fixed `BigBuf_Clear_keep_EM()` - now also clear the trace length set (@iceman1001)
--- a/tools/ePassport/epassport/emrtd/dg.py
+++ b/tools/ePassport/epassport/emrtd/dg.py
@@ -15,9 +15,11 @@ from .model import (
     DocumentDetails,
     DumpFile,
     FileState,
+    OptionalDetails,
     PassportRecord,
     PersonalDetails,
     SecurityInfo,
+    is_pesel,  # noqa: F401  - re-exported, DG13 is where it is used
 )
 from .mrz import MrzError, parse as parse_mrz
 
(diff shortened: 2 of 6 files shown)


pkilar:
Let's wait a bit on that one, I found some layout issues I'm working on. Small, cosmetic change.

iceman1001:
just make a new one...

pkilar:
I did, the fix is in #3575 - the text wrapping issue.

PR #3576, opened 2026-09-01
Fix the ePassport personal number read from the MRZ optional-data field

pkilar:
## Problem

Two faults in the personal number the viewer takes from the MRZ optional-data field, both found while checking the decoder against ICAO 9303 Part 4.

### A filler was printed where a space belonged

Part 4, positions 29 to 42:

> Any special characters, including spaces, in the personal identification number … shall be replaced by the filler character (`<`). The number shall be followed by the filler character (`<`) repeated up to position 42.

The field therefore holds two different kinds of filler. The trailing run is padding, which was already stripped. But a filler *inside* the number stands for a space or special character substituted on the way in, and those were printed as they came — a US passport showed its personal number as `123456789<0123`.

### A long document number leaked into the same field

ICAO spills a document number longer than nine characters into the optional-data field: positions 1–9 hold the first nine, position 10 a filler in place of the check digit, and the optional field opens with the rest of the number followed by the check digit for the whole of it.

`_extended_document_number()` already puts that back together and returns the optional field with the overflow removed. `_parse_td3()` bound that fourth return value and then ignored it, re-slicing `l2[28:42]` raw:

```python
number, ndigit, number_ok, optional = _extended_document_number(
    l2[0:9], l2[9], l2[28:42]
)
...
optional_data=Checked(
    l2[28:42].rstrip(FILLER), l2[42], verify(l2[28:42], l2[42])   # optional unused
),
```

Since the personal number falls back to that field when DG11 carries none, the tail of the document number was shown as the holder's personal number. With a 12-character number `AB1234567890`, `personal_number` returned `8904`.

`_parse_td1()` has always used the return value, which is where the intended shape comes from. `_parse_td2()` does not call the helper at all, so TD2 carries no support for long numbers — a gap rather than this bug, and left alone.

## Change

- Strip the padding, then put a space back for any filler that remains. It is the closest recovery available: the original could have been a hyphen, and the MRZ does not keep which.
- Use the optional-data value `_extended_document_number()` already returns. Its check digit still answers for the field as transmitted, overflow included, since that is what the MRZ carries.

`optional_data` itself stays a faithful copy of the field. Nothing else reads it, and it is the personal number being rendered.

## Behaviour change

- A personal number carrying a space or special character shows a space instead of `<`.
- A document longer than nine characters no longer shows part of that number as the personal number, and its optional-data field reads empty unless the State put something else there.
- Numbers of nine characters or fewer with no internal filler are untouched.

## Testing

- `python3 -m pytest tests/` — 331 passed, 2 skipped (was 323 passed, 2 skipped). Eight new tests, quoting the spec rule they pin.
- Checked against four real TD3 documents — the generated sample, two US passports and two Polish ones. Both US documents carry an internal filler and now render a space; the Polish ones take their number from DG13 and are unaffected.
- `black --check` clean.

## The rest of the 9303 check

Nothing else came out of it:

- The **composite check digit** covers lower-line positions 1–10, 14–20 and 22–43, as Part 4 requires. Verified behaviourally by mutating all 43 covered positions one at a time: nationality (11–13) and sex (21) are correctly ignored, every other position is caught.
- The **7-3-1 digits** agree with an independent implementation of the rule across all four documents, for the document number, both dates, the optional field and the composite.
- **Field slicing, name splitting** (`<<` between identifiers, `<` between components, including multi-part surnames) and **date handling** all match.

One apparent failure was inherent to the algorithm rather than the decoder: substituting every legal character at position 1 caught 33 and missed 3, each missed one differing in value by exactly ±10 or ±20. A mod-10 digit cannot see that.

## PR checklist

- [ ] Builds warning-free with gcc — n/a, Python only
- [ ] `client/*` build files updated — n/a, no client sources changed
- [ ] ARM firmware builds clean — n/a, no firmware changes
- [x] Platform-sensitive code reasoned about — string handling only
- [ ] `doc/commands.md` / `doc/commands.json` regenerated — n/a
- [x] CHANGELOG entry added under `[unreleased]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Fixed `tools/ePassport` - a document number longer than nine characters left its overflow in the MRZ optional-data field, where it was reported as the holder's personal number (@pkilar)
 - Added `tools/ePassport` - EF_DG13 is decoded, and a Polish document's PESEL is shown as the personal number when it carries no EF_DG11 (@pkilar)
 - Fixed `tools/ePassport` - a disabled button drew Kivy's banded disabled texture, which left its label unreadable (@pkilar)
 - Fixed `hf 14b` - a card asking for a waiting time extension of 4 or more left the timeout at zero, so any ISO14443-B card needing time to answer looked like it had stopped responding (@pkilar)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
-- Fixed `tools/ePassport` - a document number longer than nine characters left its overflow in the MRZ optional-data field, where it was reported as the holder's personal number (@pkilar)
+- Fixed `tools/ePassport` - the personal number read from the MRZ optional-data field showed a filler character where a space belonged, and a document number longer than nine characters left its overflow there (@pkilar)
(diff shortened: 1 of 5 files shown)


iceman1001:
ok to merge?

PR #3577, opened 2026-09-01
lf t55xx: correctly report the psk2/psk3 ambiguity

mfcarroll:
Two commits. The first fixes what detect claims about psk2/psk3 and narrows block 0 using observable properties of the tag. The second removes the psk3 candidates from detect, which can only ever misfire. They are independent.

## Problem

Per the ATA5577C table psk3 is *"phase change on rising edge of input"*. The client demodulates psk2 and psk3 with the same `PSKDemod()` + `psk1TOpsk2()` pair, and `psk1TOpsk2()` marks every phase transition — which against a psk3 waveform *are* the rising edges. So what comes back is the rising-edge image of the tag's word: **every run of 1's collapses to its leading 1.**

The psk3 modulation field is `0b00011`, and collapsing that `11` to `10` gives `0b00010`, psk2. A psk3 tag therefore _always_ presents as psk2, one bit out:

```
lf t55xx write -b 0 -d 00083040
lf t55xx detect
  Modulation........ PSK2
  Block0............ 00082040 (auto detect)     # written 00083040, xor = bit 12
```

`t55xx_psk3_block0_candidates()` already enumerates the psk3 words a read is consistent with. Two things around it overstated the result:

1. **The modulation line named psk2 as fact** whenever that candidate set was non-empty — exactly when detect structurally cannot tell the two apart from the waveform alone. And `printConfiguration()` ran before the probe, so a tag the probe went on to resolve was printed as psk2 and contradicted three lines later.

2. **`t55xx_psk3_probe()` concluded psk3 from vacuous evidence.** It accepted when no readable data block held cyclically adjacent ones — but a block of zeroes satisfies that for free and reads back identically either way. A freshly wiped, genuinely psk2 tag was reported as psk3. Reproduced on hardware and on 2 of the 21 psk2 configs in a full sweep.

## Change

All in `client/src/cmdlft55xx.c` plus one struct field. No demodulation code is touched.

- **Weight the probe's evidence.** A clean block argues for psk3 only in proportion to the one bits it carries. Using P(a 32-bit word with k ones has no cyclically adjacent pair) = `(32/(32-k))·C(32-k,k)/C(32,k)`, require the product across probed blocks below 1%. A zero block scores 1.0 and contributes nothing.
- **Sample page 1 as well as page 0.** Three more blocks an adjacent pair of 1's can turn up in, providing additional opportunities to disprove psk3.
- **Apply the subcarrier already being measured.** The filter checked bit rate against the observed clock but only rejected the reserved value for the PSK carrier field, so candidates survived despite naming a subcarrier the tag demonstrably is not using. `GetPskCarrier()` already exists and `try_detect_modulation()` already calls it. Takes the list from 8 to 4 on the test tag.
- **Narrow by the broadcast period.** Regular-read cycles blocks 1..MAXBLK, so the stream repeats every MAXBLK x 32 bits — and MAXBLK lives in block 0. That constrains *every* reading of block 0, the psk2 one included, so where the psk2 reading's MAXBLK cannot produce the measured period, psk2 is **excluded outright** rather than weighed.
- **Rule out the sequence terminator where the period allows it.** ST inserts four bit periods at the start of each regular-read cycle, so a tag with it set cannot produce a period that is a whole number of 32-bit blocks. Having measured one at all therefore excludes ST, and candidates carrying that bit are dropped. Note: this leans on the measurement, not on `config.ST`, which the psk paths hardcode to `false` rather than detect — pruning on that would only confirm an assumption.
- **Rule out the password bit.** Password mode gates direct access, and direct access is how detect reads block 0 — so a detect that got there without a password says the tag is not in password mode, and one that needed a password says it is. Candidates disagreeing with whichever succeeded are dropped. Verified on hardware in both directions: with PWD set, a passwordless detect fails outright rather than returning something wrong, so the inference does not inherit the misfire risk the other heuristics carry. This one costs nothing, being a property of the detect that already happened.
- **Report the answer, not the working.** `Block0` with `(auto detect)` is already an inference from observed properties; once further properties rule that inference out it should stop being presented as the config word. `Block0` now reports the words it could be: one value with `(auto detect)` where the evidence pins it, otherwise `ambiguous, one of:` and an indented list, tagged `( psk2 )` / `( psk3 )` where the possibilities disagree about the modulation. Once psk3 is settled, the demodulated value is not printed at all — it is the collapsed image the candidates were derived from, which the tag cannot be holding, so it is working rather than result. This also retires the separate `psk3 ambiguity....` line, which existed only to carry what the `Block0` field now says itself. Detect prints once, after the card-side narrowing, so the list is final when it appears.
- **The probe no longer names the modulation.** It weighs rather than rules out, so it reports a leaning and leaves the reading ambiguous; only the broadcast exclusion sets psk3. Previously a heuristic could assert `PSK3` while `Block0` still showed the psk2 reading — two claims that cannot both be true.

## Behaviour change

- A psk2-classified tag with a legal psk3 pre-image prints `PSK2 or PSK3 ( ambiguous )`. Words containing adjacent ones have no pre-image and are unaffected, so this appears only where the ambiguity is real.
- `Block0` can now print several values, or a word that was never demodulated directly — the one the evidence pins the tag to. Where psk3 is settled but not pinned it still holds the collapsed image, which is not a claim about the config word but the observation the candidate list is derived from: printConfiguration() recomputes that list from it on every call, so a later lf t55xx config needs it. Nothing outside this file reads the field.
- The probe declines on empty or very sparse data blocks where it previously answered psk3, and no longer sets `config.modulation` at all.
- Detect prints its config block once, after narrowing. Consequence: where detection finds several candidate modulations it still reports the count and which was selected, but no longer prints a table per candidate. That path did not occur once in ~350 detects across the sweeps.
- Extra card reads on tags where the ambiguity path runs: the period measurement retries up to four times, and the probe reads up to nine blocks with up to three tries each. The subcarrier and password constraints cost nothing, both being properties of work already done. The path runs only when detect landed on psk2 *and* a psk3 pre-image exists, so nothing else pays.

## Second commit: dropping the psk3 candidates from detect

Both psk3 entries — one in `t55xxTryDetectModulationEx()`, one as the fourth variant in `t55xx_psk_coherent()` — run the same demodulation as the psk2 entry beside them and differ only in the constant handed to `test()`. `test()` accepts a word only when that word's own modulation field matches the constant, so the psk3 entry needs a recovered word whose field reads 3.

Field 3 is `00011`, so it needs two adjacent set bits. What this demodulation recovers is the data's rising edges, and a rising edge needs a 0 before the 1, so no two of them are ever adjacent. The words it produces can never carry field 3, and so the psk3 entries can never match the case they were written for.

What they can do is match on a demodulation error, and then name psk3 plus a block 0 the tag does not hold. The coherent variant also inverted before `psk1TOpsk2()`, and inverting a stream does not move its transitions, so it barely differed from the psk2 variant before it.

**The cost, honestly:** a tag that only these branches matched now comes out as psk2, or as undetected where no offset yields a plausible psk2 word. The narrowing in the first commit cannot rescue that — it engages only once detect has already landed on psk2. Four runs of `00082840`, the recurring misfire, gave psk2 three times and nothing once.

## Limitations

- **psk3 is not uniquely decodable.** Verified on hardware: with a psk3 config, writing `00000003` and `00000002` to block 1 both read back as the same word. So "implement psk3 demodulation properly" is not available as a fix.
- **The adjacent-ones probe can only disprove psk3, never confirm it.** Finding an adjacent pair is proof the tag is not psk3; not finding one is weak evidence, always — hence a threshold rather than a verdict. The broadcast test is the other way round, which is why it is tried first and short-circuits the probe.
- **It stays a heuristic where it is used.** A psk2 tag whose blocks hold `AAAAAAAA` still reads as favouring psk3, because such a tag genuinely is indistinguishable from a psk3 tag holding a pre-image of it.
- **ST is not detected on the psk path, and this does not fix that.** The psk paths set `config.ST = false` unconditionally, so `Seq. terminator... No` is an unimplemented default presented as an observation — on a tag written with ST set it is simply wrong. Detecting it properly is not a small job: measured on device, a psk3 tag with ST enabled has no periodic structure in its demodulated stream at any stride from 32 to 250 bits, against a 0.7%-vs-25% margin on the same tag with ST off. ST switches modulation off for two of its four bit periods, which for psk is a subcarrier dropout that destroys the phase reference rather than merely displacing the data — which is presumably what the datasheet means by ST being "not suitable" for psk. Worth its own issue rather than this PR.
- **Block 0 is not always pinned.** Where the broadcast period does not measure, or measures a divisor that rules nothing out, the list stays as several words. Nothing here writes to the tag to resolve that, and the constraints that do apply are only as good as the measurements behind them.

## Testing

Generic Proxmark3, `PLATFORM=PM3GENERIC`, T5577, macOS/clang. Firmware rebuilt and flashed from this tree; the branch touches no `armsrc` code, so the device runs what is behaviourally master's firmware for both sides of the comparison.

Every bit-rate and subcarrier config word for ten modulations — 119 words — swept on device against the stock client and the patched one: wipe, write four data blocks, write the config, detect, read the blocks back.

**The metric changed with the output.** Detect no longer always prints a single `Block0` value, so "block 0 read back equals what was written" no longer applies directly. The comparable question is whether the written word is **among the words detect says block 0 could be** — which is what stock's `psk3 ambiguity` list was already claiming, so it scores both sides on the same basis.

| modulation | stock | patched |
|---|---|---|
| ASK | 8/8 | 7/8 |
| Biphase | 7/8 | 7/8 |
| DIRECT/NRZ | 7/8 | 7/8 |
| FSK1 | 5/8 | 5/8 |
| FSK2 | 6/8 | 6/8 |
| FSK1a | 6/8 | 6/8 |
| FSK2a | 6/8 | 6/8 |
| PSK1 | 16/21 | 17/21 |
| PSK2 | 16/21 | 15/21 |
| PSK3 | 6/21 | 5/21 |

Every difference is ±1. A previously measured noise floor on this bench — the same sweep run twice against an unchanged binary — was 7 differing configs in 111, so these are inside it. Rows failing on both sides are the RF/8 bit rate, which fails
for every modulation here.

**What the patch is for.** Configs where detect claimed psk3 on the strength of the data-block probe:

| true modulation | stock | patched |
|---|---|---|
| PSK2 · 21 configs | 3 | **0** |
| PSK3 · 21 configs | 5 | 3 |

The psk2 row is the false-positive fault. The psk3 row falling is the intended conservatism — those configs are swept with mostly-zero data blocks, exactly the case with no evidence. A later repeat of the psk sweep showed 1 rather than 0 in the psk2 row, on `00142040`, a config whose block 0 read is corrupt in that run; repeating it four times on a settled tag gives a correct read and no verdict every time.

Configs where detect *named* psk3 outright, which the second commit removes:

| | stock | patched |
|---|---|---|
| configs detect names psk3 | `00041040`, `00082840` | **none** |

Worth noting `00041040` is a **psk1** config. The misfire was never confined to the psk2/psk3 pair.

**And what the narrowing is worth.** Measured with the tag held fixed and only the build varied, since a sweep average is dominated by how often the period measurement fires rather than by the constraints themselves. Tag holding `00083060`:

| after | words offered for block 0 |
|---|---|
| stock | 8 |
| subcarrier constraint | 4 |
| broadcast period | 3 |
| sequence terminator | 2 |
| password bit | **1** |

That last one is the true word. The two the terminator left, `00083060` and `00083070`, differ only in the password bit, and the detect that read block 0 did so without a password — which under password mode it could not have. Across the psk sweep 3 of the 10 detected psk3 configs come down to a single word this way.

The period measurement is the least dependable part, and every failure mode is a refusal to narrow rather than a wrong narrowing: it needs the acquisition to succeed (hence retries), it returns a divisor of MAXBLK where neighbouring blocks hold the same word, two full periods must fit in the demodulated stream so MAXBLK 6 or 7 is out of reach, and a one-block period is reported as unmeasured because it rules nothing out. It is believed only once two attempts agree, up to four attempts. On the test tag it measured on 9 of 10 consecutive detects; the tenth simply did not narrow. So the same tag can give slightly different output between runs, always in the direction of claiming less.

`make client/check` green. `cmdlft55xx.c` compiles clean under Apple clang and Homebrew gcc 16 with the project's `-Werror` set. `astyle` with the repo's flags is a no-op on the diff.

**Not tested:** Linux, Windows/ProxSpace, the CMake client and `experimental_lib` builds, a full client link under gcc. The change adds no files, symbols or build inputs, so the three client build systems have nothing to keep in sync.

**Note on the in-tree LF suite:** `client/luascripts/tests/lf_t55xx_*.lua` and `client/lualibs/t55xx_config.lua` call `core.console(cmd, true, true)` expecting the output back as a string, but `l_CmdConsole()` in `client/src/scripting.c` takes only the command and returns no values, so they abort immediately. Looks like a gap left by 2dbd4836c, unrelated to this PR, but it is why the sweeps were driven externally.

## Checklist

- [x] Builds warning-free with clang; **not** linked with gcc here
- [x] No client build-file changes needed (no files or deps added)
- [x] No `armsrc`/`bootrom`/`common_arm` changes
- [ ] Platform-sensitive code — n/a, no serial/filesystem/OS-conditional code touched
- [x] No command definitions changed, `doc/commands.*` unaffected

Claude was used extensively in researching this issue, and drafting of the code and PR. Everything has been human reviewed and edited.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Fixed `lf t55xx detect` - a psk3 tag read back as psk2 with a block 0 one bit out, and a freshly wiped psk2 tag was reported as psk3 on no evidence; the two are now reported as ambiguous where the waveform cannot separate them, and block 0 is narrowed using the subcarrier, the broadcast period and whether password mode is in use (@mfcarroll)
 - Added doc/md/PM5_Start_Here/Getting_Started_With_PM5.md now exists as a starting point for those who have their PM5 (@innocentbystanderproxmark)
 - Fixed `tools/ePassport` - the personal number read from the MRZ optional-data field showed a filler character where a space belonged, and a document number longer than nine characters left its overflow there (@pkilar)
 - Fixed `tools/ePassport` - the detail tabs named the files their data came from as fixed text, so a document without EF_DG11 was told otherwise, and the personal-number caption wrapped (@pkilar)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+- Fixed `lf t55xx detect` - dropped the psk3 candidates, which run the same demodulation as the psk2 one beside them and so can never match a psk3 tag, only fire on demodulation errors (@mfcarroll)
 - Fixed `lf t55xx detect` - a psk3 tag read back as psk2 with a block 0 one bit out, and a freshly wiped psk2 tag was reported as psk3 on no evidence; the two are now reported as ambiguous where the waveform cannot separate them, and block 0 is narrowed using the subcarrier, the broadcast period and whether password mode is in use (@mfcarroll)
 - Added doc/md/PM5_Start_Here/Getting_Started_With_PM5.md now exists as a starting point for those who have their PM5 (@innocentbystanderproxmark)
 - Fixed `tools/ePassport` - the personal number read from the MRZ optional-data field showed a filler character where a space belonged, and a document number longer than nine characters left its overflow there (@pkilar)
--- a/client/src/cmdlft55xx.c
+++ b/client/src/cmdlft55xx.c
@@ -1202,7 +1202,9 @@ static void t55xx_psk_coherent(int fitclk, uint8_t clk, t55xx_conf_block_t *test
 
     if (n >= 32) {
 
-        for (int variant = 0; variant < 4 && *hits == before; variant++) {
+        // three variants, not four: inverting does not move a stream's
+        // transitions, so a fourth barely differed from variant 2
+        for (int variant = 0; variant < 3 && *hits == before; variant++) {
 
             const bool inverted = ((variant & 1) != 0);
 
@@ -1217,7 +1219,7 @@ static void t55xx_psk_coherent(int fitclk, uint8_t clk, t55xx_conf_block_t *test
             setDemodBuff(work, n, 0);
             setClockGrid((uint32_t)(got_clk + 0.5), got_phase);
 
-            static const uint8_t modes[4] = { DEMOD_PSK1, DEMOD_PSK1, DEMOD_PSK2, DEMOD_PSK3 };
+            static const uint8_t modes[3] = { DEMOD_PSK1, DEMOD_PSK1, DEMOD_PSK2 };
 
             int bitRate = 0;
             if (test(modes[variant], &tests[*hits].offset, &bitRate, clk, &tests[*hits].Q5) == false) {
@@ -1776,20 +1778,13 @@ bool t55xxTryDetectModulationEx(uint8_t downlink_mode, bool print_config, uint32
                     ++hits;
                 }
             } // inverse waves does not affect this demod
-            // PSK3 - needs a call to psk1TOpsk2.
-            if (PSKDemod(0, 0, 6, false) == PM3_SUCCESS) {
-                psk1TOpsk2(g_DemodBuffer, g_DemodBufferLen);
-                if (test(DEMOD_PSK3, &tests[hits].offset, &bitRate, clk, &tests[hits].Q5)) {
-                    tests[hits].modulation = DEMOD_PSK3;
-                    tests[hits].psk_carrier = t55xx_observed_psk_carrier();
-                    tests[hits].bitrate = bitRate;
-                    tests[hits].inverted = false;
-                    tests[hits].block0 = PackBits(tests[hits].offset, 32, g_DemodBuffer);
-                    tests[hits].ST = false;
-                    tests[hits].downlink_mode = downlink_mode;
-                    ++hits;
-                }
-            } // inverse waves does not affect this demod
+
+            // no psk3 candidate here on purpose: it would demodulate the same
+            // as psk2 and only differ in the test() constant, which wants two
+            // adjacent ones - and this demod recovers rising edges, which are
+            // never adjacent.  psk3 is reached by ruling psk2 out instead, see
+            // t55xx_psk3_resolve()
+
             //undo trim samples
             restore_bufferS32(saveState, g_GraphBuffer);
             g_GridOffset = saveState.offset;
--- a/client/src/cmdlft55xx.c
+++ b/client/src/cmdlft55xx.c
@@ -55,9 +55,12 @@
 
 #define T55XX_PSK3_MAX_CAND 32
 
-static size_t t55xx_psk3_block0_candidates(uint32_t observed, uint8_t clk, uint32_t *out, size_t max);
+static size_t t55xx_psk3_block0_candidates(uint32_t observed, uint8_t clk, uint8_t carrier, uint32_t *out, size_t max);
 static bool t55xx_config_psk3_ambiguous(void);
 static bool t55xx_psk3_probe(bool usepwd, uint32_t password, uint8_t downlink_mode);
+static uint8_t t55xx_measure_broadcast_blocks(bool usepwd, uint32_t password, uint8_t downlink_mode);
+static uint8_t t55xx_measure_broadcast_blocks_once(bool usepwd, uint32_t password, uint8_t downlink_mode);
+static size_t t55xx_psk3_resolve(uint8_t nblk, uint32_t *only);
 
 // Default configuration
 static t55xx_conf_block_t config = {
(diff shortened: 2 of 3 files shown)


iceman1001:
Interesting, aim to fix psk3.   One thing, do go through your files and manually edit all the code comments to be short and direct. 
No AI walls of text please. 

See:   https://github.com/RfidResearchGroup/proxmark3/blob/master/AGENTS.md

mfcarroll:
Thanks. I'll work on the code comments.

mfcarroll:
@iceman1001 I've gone through the comments. Sorry about that, I did a thorough edit on the PR and commit messages but didn't give the same treatment to the code comments. (I do agree with you completely, and it might also be worth adding something on comment policy to the agents file, as there's nothing in there on length currently, just an instruction to preserve existing comments.)

In terms of the feature, a true "fix" for psk3 isn't really possible because of the nature of the collapse of consecutive 1's, but I think this approach gets as close as is possible, and does successfully produce the correct block 0 word in many cases, and I believe is honest about the ambiguity the rest of the time.

iceman1001:
No worries ,  since we never had PSK3  support in the pm3,   something that always annoyed me.
I let Claude had a go to improve PSK since it was crappy.   It became better.  There are a battery of tests in `client/luascripts/tests/`   which targets t5577 and the pm3 modulation.      

The goal is to get the best ASK/FSK/PSK   and even more in use LF modulations.    Just keep improving it.   Never trust the AI when it says "honest take,  it can't be done"...  

Get us 100% in those tests!

iceman1001:
And yes,  if you can improve the `AGENTS.md`  to make it write comments like you do then that would be great

mfcarroll:
> Never trust the AI when it says "honest take, it can't be done"

Absolutely. A couple of months ago both Gemini and Claude told me it was impossible to use a Flipper Zero to direct read a T5577 (like `lf t55xx dump`) because the Flipper rf front end lacks a FPGA. I now have it reading a variety of modulations/rates. Ironically, with the help of those same tools that told me it couldn't be done. 🤣

PR #3582, opened 2026-09-02
PM5 low-batt shutdown: gate power-off on VUSB pin, not charger PG bit

Msprg:
The PM5 critical-battery shutdown in bwm_lowbatt_check() decided the board was on battery from the AW32001 "power good" status bit. On this hardware that bit reads "power fail" even while the board is USB-powered under HF load (hw status shows power fail plus ~400 mA discharge on PC USB), so a low pack could latch the power-off (Gpio_ARM_Power_ON_Low() then while(1)) while USB keeps the MCU alive, leaving it spinning.

This gates the power-off on the dedicated VUSB sense pin (Gpio_VUSB_Read, the same source WITH_PM5_AUTOOFF uses): only power off when the cable is really out, otherwise clear the streak and keep running. It also prints the trigger voltage/SoC before cut-off and fixes a brace/formatting glitch on the trigger condition.

Tested on PM5 hardware.

Companion to #3581; both touch bwm_lowbatt_check() so whichever lands second needs a trivial rebase. They are otherwise independent.

### DIFF
--- a/armsrc/bwm_charger.c
+++ b/armsrc/bwm_charger.c
@@ -471,6 +471,7 @@ void bwm_lowbatt_check(void) {
     static uint32_t last_tick = 0;
 #ifdef WITH_PM5_LOWBATT_SHUTDOWN
     static uint8_t crit = 0;
+    static bool s_vusb_setup = false;
 #endif
 
     if ((last_tick != 0) && (GetTickCountDelta(last_tick) < BWM_LOWBATT_PERIOD_MS)) {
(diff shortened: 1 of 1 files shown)


iceman1001:
please go through source code changes manually and edit those AI comments to short direct comments.

Msprg:
@iceman1001  Comments reduced as requested, take a look or this is how it looks in my code editor:

obrázok

(The larger comment block at the top isn't my change :) )

Edit: wait I just realized that this should propably be co-authored and it isn't... I mean I thought it was, because my other commits were co-autored before 🤔

iceman1001:
yeah,   it is very much AI assisted code changes now.   But that doesn't mean we can let it loose freely.

PR #3583, opened 2026-09-02
Improve BLE negitiation and corrected wifi status

nieldk:
hw bwmwifi --status retruned errourness message when no wifi was connected, but BWM was seated.
Changed more 'true' UART speed negotiation


iceman1001:
Don't remove the angle brackets.   They are good to have in there.

PR #3585, opened 2026-09-02
armsrc: resync BWM link baud after an AT32-only reset

Msprg:
## Problem

On a PM5 with the BWM, the AT32<->ESP UART link stops working after any reset that restarts only the AT32: a bootloader round trip (`--flash`, `hw bootloader`), a firmware flash, or `hw reset`. `hw status` then reports `BWM link baud 460800` and every BWM command fails (`hw bwmwifi --status` -> "could not query BWM WiFi status"), while the ESP itself keeps running and its WiFi/TCP stays reachable. It only recovers once the BWM loses power (unplug USB, wake on battery).

Cause: the boot negotiation moves the ESP from 460800 to 921600, but that setting is runtime-only on the ESP and it is not reset when the AT32 restarts. The AT32 comes back at 460800, its set-baud request is garbage to an ESP listening at 921600, no ack arrives, and the AT32 stays at 460800 with the two ends mismatched.

## Change

`armsrc/bwm_forward.c`: `bwm_fwd_negotiate_baud()` now locates the ESP first by sending `GET_UART_BAUD` alternately at the boot baud and the target baud (4 rounds, 150 ms each). If the ESP answers at the target it is adopted as is; if it answers at the boot baud the existing set-baud handshake runs unchanged; if nothing answers the link stays at the boot baud as before. The verify/fallback tail of the handshake is factored into two small helpers (`bwm_probe_at`, `bwm_link_reset`) so the new path and the old one share it. No ESP firmware change needed.

`armsrc/bwm_forward.h`: doc comment updated.

## Behaviour change

- After an AT32-only reset the link comes back at 921600 instead of silently dropping to a mismatched 460800.
- Boot time with no BWM attached grows from ~0.9 s (3 x 300 ms set attempts) to ~1.2 s (4 x 2 x 150 ms probes). With a BWM present the first probe answers within a few ms.

## Testing

- `make fullimage` with `PLATFORM=PM5 PLATFORM_EXTRAS=BWM` and with `PLATFORM=PM3RDV4`: both warning-free (gcc). No client code touched, so clang/CMake/experimental_lib were not exercised.
- On hardware (PM5 + BWM, WiFi forward configured): reproduced the failure on current master, then with this change did a flash plus four bootloader round trips; `hw status` showed 921600 and `hw bwmwifi --status` answered after each. A cold power-up (ESP back at 460800) was also checked and negotiates up as before.
- Not tested: a BWM running firmware without the baud commands (falls into the "nothing answers" branch, same outcome as today).

## Checklist

- [x] Builds warning-free with gcc; clang not applicable (firmware only)
- [x] Client build files unchanged (no client sources touched)
- [x] ARM firmware builds clean for RDV4 and PM5 platform configs
- [x] Platform-sensitive host code: none touched
- [x] Existing comments in touched files preserved (two verbose blocks in the rewritten handshake tail were condensed)
- [x] No command changes, `doc/commands.*` untouched

### DIFF
--- a/armsrc/bwm_forward.c
+++ b/armsrc/bwm_forward.c
@@ -334,11 +334,15 @@ uint32_t bwm_read_ng(uint8_t *data, size_t len) {
 // send the request at the current baud, wait for that ack, then switch our own
 // UART4 to match. On timeout (old ESP without the command, or a lost ack) we
 // leave the link at the boot baud - it keeps working, just slower.
+//
+// The ESP keeps its baud across an AT32-only reset (bootloader, hw reset,
+// flash), so probe both rates first and adopt the one it answers at.
 // ---------------------------------------------------------------------------
 #define BWM_BAUD_ACK_WAIT_MS   300   // per-attempt wait for the ESP ack
 #define BWM_BAUD_ATTEMPTS      3     // resend attempts before giving up
 #define BWM_BAUD_VERIFY_MS     150   // per-probe wait for the GET_BAUD reply
 #define BWM_BAUD_VERIFY_TRIES  5     // GET_BAUD probes before declaring the switch failed
+#define BWM_BAUD_PROBE_ROUNDS  4     // boot/target probe pairs
 
 // Probe the ESP at the CURRENT baud with GET_UART_BAUD; true only if it answers,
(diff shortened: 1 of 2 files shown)


doegox:
Seems to work fine, thanks!

PR #3590, opened 2026-09-05
QC tester behavior

xianglin1998:
- The factory tester will automatically test whether each I/O is normal.
- Based on power consumption testing, it is currently known that the PM5 consumes approximately 600mA when the radio frequency and various LEDs such as RGB are enabled.
- For BWM, the current measured power consumption is approximately 110mA.


xianglin1998:
After this PR was merged, the device firmware used by the factory tester was essentially fixed.

xianglin1998:
@doegox This PR has been completed and can be merged.

PR #3596, opened 2026-09-07
Added possibility to get some device info

YoungJules:
even if the firmware does not match the client.
Reports bootloader, OS, fpga versions and the capabilities.


iceman1001:
It feels like we are adding more confusion with more printing of details.
The core of the issue is that there is a firmware mismatch and you need to reflash it.  

I will close this PR.

PR #3597, opened 2026-09-07
Add ISO14443-3 Type A timeslot support to 'hf 14a info' and 'hf 14a reader'

kormax:
This PR adds basic support for detection and activation for NFC-A cards that utilize the timeslot-based anticollision approach.

```log
[usb] pm3 --> hf 14a info

[=] ---------- ISO14443-A Timeslot Information ----------
[+]    UID: F0 00 00 00 FF FF FF FF
[+] ATQA_T: 44
[+]  SAK_T: 42
```

```log
[usb] pm3 --> hf 14a reader
[+]    UID: F0 00 00 00 FF FF FF FF
[+] ATQA_T: 44
[+]  SAK_T: 42
```

Special thanks to @mxcdoam for sharing their findings on these tag types existing in the wild, which made this possible.

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -607,7 +607,8 @@ void Demod14aInit(uint8_t *d, uint16_t n, uint8_t *par) {
 }
 
 // use parameter non_real_time to provide a timestamp. Set to 0 if the decoder should measure real time
-RAMFUNC int ManchesterDecoding(uint8_t bit, uint16_t offset, uint32_t non_real_time) {
+static RAMFUNC int ManchesterDecodingEx(uint8_t bit, uint16_t offset, uint32_t non_real_time, bool no_parity) {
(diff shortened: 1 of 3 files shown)


iceman1001:
reminds me of the `hf topaz`  commands with "14a using CRC-B"

mxcdoam:
Topaz commands "as is" wont work, unfortunately.

iceman1001:
hm,   I better check that then.  I luckily I have some Topaz fobs laying around

iceman1001:
image

For me topaz commands works great ,  you would need to give some context to your claim of "as is" and where it doesn't work for you.

mxcdoam:
HF Topaz commands do not work with ISO 14A timeslot cards. This pull request adds support for timeslot-based cards—those are the specific type I was referring to when I said "don't work as is".

PR #3602, opened 2026-09-08
Esp32 app descriptor check

nieldk:
Add app descriptor check to further prevent flashing a non-application image
Checks for 0xABCD5432 at offset 0x20

### DIFF
--- a/client/src/cmdhw.c
+++ b/client/src/cmdhw.c
@@ -2545,15 +2545,18 @@ static int CmdBWMUpgrade(const char *Cmd) {
         free(fw);
         return PM3_EFILE;
     }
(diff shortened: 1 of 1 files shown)


client/src/cmdhw.c
```diff
@@ -2550,6 +2551,13 @@ static int CmdBWMUpgrade(const char *Cmd) {
         free(fw);
         return PM3_EFILE;
     }
+
+    uint32_t app_sign = (uint32_t)(fw[0x20] | (fw[0x21] << 8) | (fw[0x22] << 16) | ((uint32_t)fw[0x23] << 24));
```
iceman1001: we have helper functions to convert from bytearray to numbers.   common_util.c

nieldk: thanks, made use of that

PR #3607, opened 2026-09-10
running make style

innocentbystanderproxmark:
My understanding is that `make style` needs to be run before PRs, clean master, ran make style, this is the output produced.


iceman1001:
you don't do that,  we maintainers do that...

PR #3611, opened 2026-09-10
Adding documentation to more clearly and explicitly document the usag…

innocentbystanderproxmark:
…e or lack thereof, of the make style command

An attempt to make it easier and clearer for newbies who want to contribute on when to (never) or when not to (always) run `make style` to avoid extra work by the maintainers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,11 @@
 
 :+1::tada: First off, thanks for taking the time to contribute! :tada::+1:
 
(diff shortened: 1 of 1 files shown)


iceman1001:
I am doubtful since  A,   this removes the make style reference everywhere and B,  it's irrelevant instruction to have a section for the reason of the command.  

Like @doegox   commented in the other PR the idea is for contributors to run make style but only commit their changes.   then checkout the other files in order to get their local repo in a clean state again.

iceman1001:
lets see if we can do a better rewrite.  closing meanwhile

PR #3612, opened 2026-09-10
Clarify instructions on make style and make commands usage

innocentbystanderproxmark:
Searched for `make style` and `make commands` because of a discord comment from @doegox and hopefully the documentation is clearer for new/newer contributors on not using `make style` or `make commands`

### DIFF
--- a/.github/ISSUE_TEMPLATE/checklist-for-release.md
+++ b/.github/ISSUE_TEMPLATE/checklist-for-release.md
@@ -10,6 +10,7 @@ assignees: doegox, iceman1001
 # Checklist
 
 - [ ] `make style`
+- [ ] `make commands`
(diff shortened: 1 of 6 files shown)


doegox:
well the point is that ppl can use make style or make commands, but they should commit only their own work, not the effect of make style on foreign code. 
and the addition of make commands in the release checklist is unnecessary, it's subset of make style iirc

innocentbystanderproxmark:
I was under the understanding (after iceman got mad at one of my PRs, and follow up in discord chat) that only maintainers should run `make style` and `make commands`

I want other newbies to not have the same issue I had, let me try to rework the wording to better reflect what you've said.

iceman1001:
lets see if we can do a better rewrite.  closing meanwhile

PR #3618, opened 2026-09-11
fix: correct encrypted nonce length broken in 3762130c4

TheArchitect0880:
Updates the payload length calculation to account for the 9-byte structure of encrypted nonces (two 4-byte nonces + 1 parity byte). 

This prevents payload truncation which previously caused the client to read incomplete data and fail with errors like:
`[-] ⛔ No match for the First_Byte_Sum (105), is the card a genuine MFC Ev1?`

### DIFF
--- a/armsrc/mifarecmd.c
+++ b/armsrc/mifarecmd.c
@@ -1235,7 +1235,7 @@ void MifareAcquireEncryptedNonces(const mf_acquire_nonces_t *payload) {
     mf_nonces_resp_t *response = (mf_nonces_resp_t *)respbuf;
     response->cuid = cuid;
     response->num_nonces = num_nonces;
-    uint16_t noncelen = MIN((uint16_t)(num_nonces * 4), (uint16_t)(MFC_MAX_NONCES * 4));
+    uint16_t noncelen = (num_nonces / 2) * 9;
     memcpy(response->nonces, buf, noncelen);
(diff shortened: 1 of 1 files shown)


iceman1001:
Current layout
```
// Reply for the same three commands. isOK travels in the NG status field.
typedef struct {
    uint32_t cuid;
    uint16_t num_nonces;
    uint8_t nonces[];
} PACKED mf_nonces_resp_t;

// most 4 byte nonces that fit alongside the reply header in one frame
#define MFC_MAX_NONCES ((PM3_CMD_DATA_SIZE - sizeof(mf_nonces_resp_t)) / 4)

uint16_t noncelen = MIN((uint16_t)(num_nonces * 4), (uint16_t)(MFC_MAX_NONCES * 4));
memcpy(response->nonces, buf, noncelen);
reply_ng(CMD_HF_MIFARE_ACQ_ENCRYPTED_NONCES, isOK, respbuf, sizeof(mf_nonces_resp_t) + noncelen);
```

Supposed fix: 
```
response->num_nonces = num_nonces;
uint16_t noncelen = (num_nonces / 2) * 9;
``` 

I am not following how the suggested fix will actually solve the issue?   if the nonces[] consists of 9 bytes items,  it still doesn't handle the reply_ng cutting of nonces when sending.

iceman1001:
ok,   I need to make some improvements to this one and push.    Will thank you in the commit/changelog.

iceman1001:
New fix pushed,   should handle it better now.   Thank you!

PR #3624, opened 2026-09-13
PM5 add option to connect by BLE name

nieldk:
Add option for connecting by name on Proxmark5 BWM as an addition to MAC adress

`pm3 -p ble:Proxmark5`


iceman1001:
that time usage needs to be addressed to use our posix time calls instead. 
and  the return codes should be PM3_E..  styled

iceman1001:
```
#include <string.h>
#include <strings.h>   // strcasecmp
```

hm,   why two... and don't we already have a support functions for this in utils.c
and tell your AI to follow the instructions in CONTRIBUTING.md and  AGENTS.md


This looks strange.
```
        if (s < 0) { if (errno == EINTR) break; break; }   // Ctrl-C or error -> give up cleanly
        if (s == 0) break;                                  // timeout
```

iceman1001:
```
ifeq ($(platform),Linux)
	    $(Q)$(INSTALLSUDO) setcap 'cap_net_raw,cap_net_admin+eip' $(DESTDIR)$(PREFIX)$(PATHSEP)$(INSTALLBINRELPATH)$(PATHSEP)$(INSTALLBIN) || echo "[!] setcap failed - BLE connect-by-name (-p ble:<name>) will need sudo; connect-by-MAC is unaffected"
endif
```

is this something the two CMakefile.txt  files needs to handle too?

iceman1001:
you removed an entry in the CMakefile.txt  and the experimental_lib/CMakefile.txt changes is missing

iceman1001:
the experimental_lib/CMakefile.txt changes is still missing

nieldk:
> the experimental_lib/CMakefile.txt changes is still missing

What are you missing here ?

PR #3626, opened 2026-09-15
PM5: idle power save, BWM power/WiFi/BLE settings, idle auto power-off, byte-window flow control

Msprg:
> [!TIP]
> Ready to merge!
---

## Problem

- A PM5 idles at 288 MHz with the PLL on, a 1.3 V core and a busy-polling main loop: 283 mA at the USB port with a BWM fitted, and it runs warm doing nothing.
- The BWM (ESP32-C2) module has no power management either, and nothing on the PM5 could change how it behaves: not its power saving, not its WiFi, not its BLE radio. The module ships with BLE open (anyone in range can connect and run commands) although its firmware implements bonding with a passkey.
- On battery the only automatic power-off is "USB was unplugged". A PM5 woken by the button and forgotten, or left idle after the phone disconnected, drains the cell.
- The BWM link's flow control counts frames (16), but a frame is 30 bytes to 2.1 KB and the module buffers un-acked frames in a 4 KB ring: a radio stall of ~45 ms at 921600 baud overflowed it and the module dropped PM3 bytes.

## Change

Seven commits, PM5 only. Companion module firmware PR: RfidResearchGroup/Proxmark5_BWM_esp32#5.

1. **`doc: regenerate commands`** - `hf ntag424 gettt` / `setconfig` are missing from the generated docs on master. Kept separate so the commits after it only carry their own doc changes; drop it if you regenerate yourselves.
2. **`pm5: reduce idle power`** - port of the "reduce idle power" feature from the Fantasi firmware (soeinova/Fantasi @ 9671309 by @soeinova and @noproto, plus its later ACC-SOF fix). Between commands the core drops from 288 MHz to the 48 MHz internal oscillator with the PLL off, the LDO at 1.1 V, the FPGA clock parked and the CPU in WFI until the next interrupt or a 1 ms tick. Commands and BWM housekeeping hold a refcounted 288 MHz boost, so command timing and throughput are unchanged; a held field / emulation blocks the downclock (`FpgaIsOff()`). USB moves to the crystal-less HICK-48 clock (ACC-trimmed against SOF) in the OS image only; the bootrom keeps the PLL path, so flashing is unchanged. New `hw powersave on|off` (default on); `hw status` shows the idle-clock and WFI fractions. Also adds the wake preamble on the BWM UART (four 0x55 + 10 ms, on the first write and after 1 s without TX) that the companion module firmware needs for light sleep; harmless to older module firmware.
3. **`pm5: hw bwm powersave`** - show/set the module's persisted power-save switch (`CMD_PM5_BWM_POWERSAVE` 0x0181, module 1019/1020). `hw status` prints it next to the BWM firmware version; both module queries there are time-bounded.
4. **`pm5: hw bwm wifipower`** - `off` turns WiFi fully off (same path as `hw bwm wifi stop`); `none|min|max` set the module's persisted WiFi modem power-save type (`CMD_PM5_BWM_WIFI_PS` 0x0182, module 2052/2053).
5. **`pm5: auto power-off`** - `hw bwm autooff [on|off] [--idle <sec>] [--unplug on|off]`. New opt-in idle trigger (off by default): on battery, no command over any transport, no button press and no BLE/WiFi client for `--idle` seconds powers the board off. The module reports its clients with a new LINK_STATE broadcast; the PM5 asks the module once more right before going off, so neither an older module (no broadcast) nor a lost broadcast powers it off under a connected client. `--unplug off` makes an unplug only restart the idle clock. All three settings are stored on the module (host value slots 1021/1022), because runtime settings reset at exactly the boot they matter for. Unplug detection is now edge-based.
6. **`pm5: BWM link flow control counts bytes`** - a FIFO of un-acked frame lengths against a byte budget sized to the module's UART ring (12 KB in the companion PR) minus 1 KB. The 50 ms ack timeout is unchanged: when acks stall that long on a full window the gate still sends on.
7. **`pm5: hw bwm ble`** - `status | on | off | pairing on|off [-k <6 digits>] | forget -i N|--all | txpower -a/-c <dBm>` (`CMD_PM5_BWM_BLE` 0x0183). `on`/`off` is the module's new persisted radio switch. `pairing on` requires the passkey (LE Secure Connections, MITM; the SPP characteristic then needs the encrypted link) and flags the factory key 123456. A pairing change restarts the module's stack, so the firmware replies first.

## Behaviour change

- PM5 idles at 48 MHz / WFI by default; `hw powersave off` restores the always-288 MHz loop at runtime. PM5 USB (OS image) runs off HICK-48 instead of the PLL.
- `hw bwm autooff` without arguments now shows the state instead of printing help; its setting persists on the module instead of resetting at boot. Defaults are unchanged (on, unplug trigger only).
- `hw status` on a PM5 has a Power section and, with a BWM, its power-save state, live BLE state and the auto power-off setting (at debug level also the client state the PM5 tracks from the module's broadcasts).
- New command ids 0x0180-0x0183; `CMD_PM5_BWM_AUTOOFF` takes a longer payload (a 1-byte payload from an older client still works).
- Nothing changes for non-PM5 targets: every hunk is behind `PM5` / `WITH_BWM_*` / AT32 guards, and `armsrc/cmd.c` is untouched.

### Which features need the new module firmware

| Feature | With module firmware before the companion PR |
|---|---|
| idle power save, wake preamble | works; the old module drops the preamble as noise |
| `hw bwm powersave`, `wifipower` none/min/max, `ble on/off` | "BWM did not answer" / switch shown as unknown, no hang |
| `hw bwm ble status/pairing/forget/txpower` | work (existing module commands) |
| auto power-off settings | apply for the session, reported as "not stored" |
| idle power-off vs. a silent BLE client | the BLE state query before power-off still protects it |
| byte-window flow control | works, but sized for the new 12 KB ring: pair it with the companion build |

The other direction (new module firmware, PM5 firmware from before this PR) is handled in the module: it only light-sleeps once it has seen a wake preamble, so an older PM5 firmware just does not get that saving.

## Measured

Idle current at the USB port (Power-Z, 5.1 V, 2 min traces):

| Configuration | Idle current |
|---|---|
| stock PM5 + stock BWM firmware | 283 mA |
| this PR, stock BWM firmware | 207 mA |
| this PR + companion BWM PR, both power saves on | 171 mA (-39 %) |
| the same on the final commits of both PRs (2026-09-19, 5.06 V) | 175 mA |

stock PM5 + stock BWM: ~280 mA, spiky

both power saves on: ~172 mA flat

(The advertising ticks in the second trace are 3 s apart; the final module build advertises once a second after a 30 s fast phase. The final-commit trace is just as flat, 889 mW against 880 mW: the ~1 % is supply voltage and run-to-run variation, the faster advertising accounts for about 0.1 mA. For scale, a module that does not light-sleep sits at 186-188 mA.)

On battery, over BLE from a phone, with the new `hw status` lines: ~200 mA drain.

Screenshot_2026-09-15-04-42-39-704_com termux-edit

## Testing

Builds, gcc, warning-free, every commit on its own: `PLATFORM=PM5 PLATFORM_EXTRAS=BWM`, plain `PLATFORM=PM5`, `PM3RDV4`, `PM3GENERIC` (armsrc; bootrom at the tip), default client, PM5 client; `make commands` output matches the committed docs at every commit. Linux (WSL) only.

On a PM5 with BWM:
- MIFARE Classic 4K: hw version/status/ping/tune, hf search, hf 14a info/reader across idle gaps, hf mf info/chk/rdbl, lf search/read, held field blocks the downclock, hw reset + re-enumerate, hw powersave off/on.
- `hw bwm powersave` off/on and persistence across a module reboot; `hw status` after 4 to 60 s idle with the module light-sleeping.
- `hw bwm wifipower`: default reads min with WiFi off; none/max/min set and read back; off tears WiFi down; max survives a module reboot.
- Auto power-off: idle power-off on battery, `--unplug off` with a cable pull, settings surviving a reboot of module and PM5; switching it off on USB, unplugging and switching it on again over BLE leaves the board running.
- Over BLE from a phone (Termux client via a TCP bridge): `hw status -m 10000` moved 866 KB in 10 s at 86.6 KB/s (423 frames, no truncation, no resync), about 94 % of what 921600 baud carries; `lf t55xx detect` clean three times in a row; `hw ping -l 2048`. Repeated on the final commits of both PRs: phone link, `hw ping -l 2048`, idle power-off 30 s after the phone disconnected.
- `hw bwm ble` over USB: status, off/on (off survives a module reboot), pairing on/off with the stack restart, passkey validation, forget argument checks, txpower set and read back.

**Not tested:** clang, Windows/ProxSpace, macOS (client changes are two command files using only CLIParser and `SendCommandNG`, nothing platform-specific); CMake builds (one firmware file added to both `armsrc/Makefile` and `armsrc/CMakeLists.txt`, no client file added); `hw bwm ble pairing` from a phone; WiFi power-save types with WiFi associated (no credentials on the bench).

Net diff: 22 files, +1850 / -127.

## Known limits, tracked separately

- #3648: `bwm_cmd()` discards forwarded data and broadcasts that arrive while it waits for its reply (upstream behaviour). With this PR it can eat a LINK_STATE broadcast when a module command itself changes the link state (`hw bwm wifi stop` with a TCP client attached).
- #3649: a forward frame the module cannot deliver is answered with CMD_ERROR, not an ack, and only leaves the flow-control window through the 50 ms timeout.

## Checklist

- [x] Builds warning-free with gcc; clang not tried (see above)
- [x] No client source files added; `armsrc/Makefile` and `armsrc/CMakeLists.txt` both list the new firmware file
- [x] ARM firmware builds clean for RDV4, GENERIC and PM5 (with and without BWM)
- [ ] Platform-sensitive code: Linux only, see above
- [x] New comments short and direct
- [x] `doc/commands.md` / `doc/commands.json` regenerated (default client build)

## Test build

A prebuilt module image is attached to the pre-release on my fork: [proxmark5_ble_wifi_module-idle-power-024a8a2.bin](https://github.com/Msprg/Proxmark5_BWM_esp32/releases/download/bwm-idle-power-024a8a2/proxmark5_ble_wifi_module-idle-power-024a8a2.bin). Flash it over the PM5 link, no header needed:

```
hw bwm upgrade -f proxmark5_ble_wifi_module-idle-power-024a8a2.bin
```


armsrc/appmain.c
```diff
@@ -4352,11 +4440,15 @@ void __attribute__((noreturn)) AppMain(void) {
 
         // Check if there is a packet available
         PacketCommandNG rx;
-        memset(&rx.data, 0, sizeof(rx.data));
-
```
iceman1001, line 4356: why remove this?

iceman1001: And I founded the answer further down the changes.

iceman1001:
Cool PR,   suffers little bit of AI comment happiness.     When it comes to Pm5 parts I usually don't care since that code base is so unknown and undocumented.

Msprg:
Force-pushed 39bbf20cc: rebased onto master (339db44cf: the `IfBwm` gate for the `hw bwm` table and capabilities v12) plus one functional change, folded into the flow-control commit rather than added as an eighth:

- The 50 ms ack timeout used to forget the whole window. That made the gate open loop for the rest of the command (the acks of the forgotten frames released younger ones early), so after the first eaten ack (#3648, `hw status` queries the module while its own status lines are still in flight) the byte budget protected the module's ring only until the first timeout. It now waits 200 ms and forgets just enough of the oldest frames to send the current one; a lost ack costs one wait and heals, a real radio stall leaks one frame per timeout instead of a window. A `CMD_ERROR` naming a forward frame releases its slot like the ack. `hw status` at debug level shows the window counters.
- With an honest window the module's late ack (sent after the last BLE chunk was accepted by its stack) cost about 10 % over BLE, so the companion PR now acks a forward frame when it leaves the module's UART ring, which is what the window budgets.

Measured on the final tips of both PRs over BLE from a phone: 85 to 86 KB/s bulk (against 88 KB/s for the old open-loop gate on the same link), a 20 s transfer with the phone scanning for Bluetooth meanwhile finishes clean, `hw ping -l 2048` 109 to 148 ms.

Build matrix as before at every commit (PM5 with and without BWM, RDV4, GENERIC, PM5 bootrom, default and PM5 client, docs regenerated). Net diff now 22 files, +1926 / -138. #3649 is closed as no longer applicable; the description above still describes commit 6 with the old timeout wording.

PR #3628, opened 2026-09-15
Add dfc-core DESFire-compatible emulator

cindersocket:
Adds dfc-core emulator support, an emulator for DESFire ev1/ev2/ev3 cards and technologies.

DFC-core uses an DER-based binary format (dfcb).  It has a text version (dfc).  Both files are supported with the DFC flipper app, and the DFC Chameleon Ultra pull request.  json is still supported.

Some issues were found during testing with USB buffer sizes.  Those fixes are also in this PR, but can be broken out if needed.

Current diff from upstream: 74 files, 16,236 insertions and 664
deletions. Excluding the vendored core: 38 files, 2,418 insertions and 664
deletions. Most added lines are the unchanged, revision-pinned upstream library.
The adapter replaces the old DESFire firmware handlers; generated command
documentation and regression fixtures account for part of the integration diff.

The vendored core could be done as a gitmodule.

Codex assisted with commits and testing, and writing the rest of this PR describing the steps taken and testing done.

## Changes

- Vendor a revision-pinned dfc-core with a source-sync check. Use the existing
  mbedTLS crypto implementation through a small platform adapter.
- Integrate the engine with the existing ISO14443-A simulator. Allocate working
  state from BigBuf and clear/release it on shutdown or unrelated commands.
- Extend existing DESFire emulator load/save commands for JSON, DFC and DFCB,
  with SPIFFS import/export and direct APDU/frame testing commands.
- Add HF_DFCSIM using one SPIFFS credential. Keep PM3's full EV1 default;
  use full EV3 with proximity checking and full core capacities on PM5.
  Expose profile and storage-capacity options in Make and CMake.
- Keep USB buffered-command handling and the CMake single-block data compression
  fix in their own commits, independently testable from DESFire functionality.

## Behaviour change

- DESFire emulation uses dfc-core rather than the previous command handlers.
- Emulator working memory is temporary; the existing 8 KiB emulator image stays
  available for RAM save/load until reused by another command.
- Standalone loads `hf_dfcsim.dfcb` and saves changes on USB or button exit.
  If absent, create and persist a factory credential with no applications or
  files, the default master key, and a generated UID. Do not replace an invalid
  existing file.
- PM3 retains 28 application slots, 28 file slots, a 256-byte file pool and a
  512-byte key pool. PM5 defaults to 28 applications, 32 files, an 8,192-byte
  file pool and a 4,096-byte key pool. These remain transient allocations.
- PM5 standalone entry uses `hw standalone`; the platform's existing disabled
  button-entry path is unchanged.
- Randomness uses the existing tick/PRNG facilities for development. Explicit
  random injection can make APDU tests deterministic.
- On both PM3 and PM5, USB commands already buffered by the parser no longer
  wait for another packet. Empty terminating OUT packets are acknowledged so
  they cannot block subsequent commands.
- Invalid DESFire RF frames are discarded before protocol dispatch. ISO-DEP
  chaining retains command fragments, preserves retry responses, and keeps the
  next command's block number synchronized.
- SPIFFS load/save options reject firmware without flash support immediately.
  HF_DFCSIM remains SPIFFS-only; both build systems reject unsupported targets.
- JSON preserves the complete communication-settings byte in `CommModeRaw`.
  Unknown legacy names without a raw value are rejected rather than changed.
- CMake compresses the initialized data section without FPGA interleaving padding,
  preventing a startup decompression failure for affected section sizes.

## Testing

I tested on macOS with physical PM3 RDV4 and Proxmark5 hardware.
RF checks used an ACS ACR1552 PC/SC reader. I also validated the EV1
field credential on a HID multiCLASS SE R10. Direct emulator tests used the client
transport on the actual device, without RF. 

### Build and offline checks

- After rebasing onto `a528d5ac1`, `make check`, the PM5 Make build and the
  RDV4 and PM5 `STANDALONE=HF_DFCSIM` Make builds passed. Earlier CMake and RF results
  below apply to the tested implementation before this rebase; hardware was
  not reflashed for the history-only consolidation or the two upstream updates.

- Host client: Make with Apple Clang, CMake client, and the CMake experimental
  shared-library build passed after the client changes.
- Firmware: ARM GCC builds passed for RDV4 and PM5 using Make and CMake.
  Standalone variants included full EV1 on RDV4 and PM5 during development,
  followed by PM5's final full-EV3 default with proximity checking enabled.
- Fresh CMake caches verified the final platform split: PM5 resolves to
  EV3/proximity-on and 28/32/8192/4096 capacities; RDV4 still resolves to
  EV1/proximity-off and 28/28/256/512. Both fresh builds passed.
- The PM5 USB receive change compiled in both application and bootrom builds.
  The bootrom path does not depend on the application-only receive buffer.
- Generic firmware without SPIFFS built with both Make and CMake. Selecting
  HF_DFCSIM without SPIFFS was rejected by both build systems.
- `make check` and the full client offline suite passed. Optional slow tests
  were not run.
- Client checks covered unknown communication modes, raw communication-byte
  preservation, and rejection of conflicting `view --emu`/file sources.
- Command documentation was regenerated. Whitespace checks passed.

### USB regression, independently of DESFire

- Reproduced the PM5 failure using ordinary pings, with no DFC session active:
  63- and 65-byte commands passed, an exact 64-byte command replied, and the
  next ordinary ping timed out.
- Reproduced the same stall with the isolated two-write DESFire sequence on
  unchanged firmware after reboot. This separated the transport problem from
  an assumed file-pool or application-count limitation.
- After the receive fix, the exact failing 64-byte boundary sequence passed,
  as did a 128-byte boundary followed by a normal ping.
- `tools/usb_ping_test.py --rounds 100` passed **1,800 pings on RDV4** and
  **1,800 pings on PM5**, including commands coalesced into one host write.
  Payload lengths were 0, 1, 53, 54, 55, 117, 118, 119 and 512 bytes. Each reply
  was checked for command, length, status and exact payload content.
- PM5 bootloader communication and flashing were exercised after recovery:
  the bootrom transfer and subsequent application transfer completed. The
  application then answered version/status queries and boundary pings.
- DFU recovery was backed by a complete 1 MiB MCU-flash backup. All 383,640
  bytes of the generated recovery image were read back and compared exactly.
  Recovery verification is not a claim of power-loss fault-injection testing.

### Direct emulator protocol checks on hardware

- **86/86 vectors passed on the final full-default PM5 firmware**, with no
  exclusions. The formerly hanging two-write case passed in this run.
- Coverage included EV1 authentication and access rights; D40, ISO and AES
  secure messaging; EV2 first/non-first authentication and secure messaging;
  key changes and key-set lifecycle; delegated applications; virtual-card
  behavior; EV3 secure dynamic messaging, transaction MAC, transaction timer
  and both proximity-check scenarios.
- File operations included creation/deletion, duplicate and invalid identifiers,
  capacity boundaries, read/write bounds, transactional backup/value/record
  behavior, commit/abort, and application enumeration.
- Protocol checks included native and ISO APDU dialects, ISO selection,
  response chaining, unexpected commands during chaining, field-off/reset,
  deterministic random input and random-byte underflow checks.
- The hardware adapter checked ordered responses and masked expected bytes.
  Internal engine-state assertions were checked separately in native tests;
  they were not exposed or independently checked by the hardware adapter.
- The generic no-SPIFFS device image also passed all 86 direct scenarios in
  the earlier full-profile run. This is separate from SPIFFS standalone tests.
- Direct frame tests exercised invalid sequence rejection, chained SELECT,
  lost-acknowledgement retries and correct sequencing of the following command.

### Credential formats and client workflows

- Native tests passed all **36 malformed-input rejection cases** with the
  expected error classifications, under both crypto backends.
- All **three credential format pairs** passed text-to-text, binary-to-binary,
  text-to-binary, binary-to-text and compile-on-load checks byte-for-byte.
- Hardware JSON tests covered export, reload and subsequent emulator use.
  A raw communication-settings value of `0x43` survived a byte-identical
  round-trip in its targeted check. Unknown legacy mode names without a raw
  value were rejected rather than silently converted to plain communication.
- On PM5, exported a reader-created credential as JSON, reloaded it into RAM,
  and verified its application and `DEADBEEF` file contents through the ACS
  reader. This establishes logical data preservation; the reencoded DFCB
  length differed, so this particular check is not called byte-identical.
- `view --emu` and the `eview` compatibility alias produced equivalent
  credential output on hardware, apart from command echo/allocation messages.
- SPIFFS `eload` and `esave` were exercised. Their unsupported-capability paths
  returned immediately on no-SPIFFS firmware, with subsequent ping still working.

### RF and standalone persistence

- RDV4: exercised RF activation, all three GetVersion frames and reader
  reconnection through ACS ACR1552.
- RDV4 factory path: preserved the original SPIFFS file, started with the
  standalone filename absent, and verified automatic creation of an EV1 card
  with zero applications/files and default key settings.
- Through RF, created an application and standard data file, wrote/read
  `DEADBEEF`, stopped/saved, restarted standalone and read the same data.
  The generated UID persisted across restart.
- Restored the original RDV4 credential and verified it byte-for-byte against
  its backup. The generated test credential was retained separately on the host.
- RDV4 physical button save/exit was exercised, as was USB-triggered save/exit.
- PM5: with initially empty SPIFFS, verified factory creation, application/file
  creation and write/read through ACS; standalone save/restart retained the
  UID and file contents. Normal RAM emulation loaded from SPIFFS also passed.
- After the final PM5 USB-fix recovery, both normal emulation and client-started
  standalone again passed ACS application selection and persisted file reads.
  Standalone then saved and exited on a USB ping, which received a valid reply.
- PM5 button-triggered standalone entry was not tested or enabled; the existing
  platform disables that entry path. PM5 tests used `hw standalone`.

### RF retry and authentication diagnostics

- Final PM5 normal-emulator checks loaded both D40 factory and EV1 field DFC
  credentials through `hf mfdes eload`. D40 completed mutual authentication,
  an exact 52-byte protected read, four-byte MAC verification and final `9100`
  over ACS PC/SC.
- EV1 completed ISO authentication with a 16-byte key and an exact 52-byte
  protected read. The PC/SC helper collected the `91AF` response, sent
  `90 AF 00 00 00`, reassembled the result and verified its eight-byte CMAC
  using the derived session key and command IV, ending with `9100`.
- I validated the EV1 field credential on the HID multiCLASS SE R10.
  The final ACS verification separately exercised the PC/SC path.
  The emulator then stopped cleanly through the client.
- These latest checks were normal emulation, not a new standalone stress run.
  Intermittent ACS no-card reports occurred during repositioning and client
  restarts; the successful rerun does not establish their root cause.

- Host tests cover malformed frames, sequence errors, duplicate frames,
  input/output chaining, overflow/capacity checks and session reset behavior.
- Temporary RF fault injection corrupted an authentication CRC and a later
  authentication parity bit. Reader retry recovery and successful authentication
  were observed in the targeted checks. Diagnostic injections were removed
  before the clean firmware was flashed again.
- RDV4 clean-image authentication runs included **1,000/1,000 in normal mode**
  and **1,000/1,000 in standalone mode**. Separate reader-transaction experiments
  passed 5,000/5,000 with a transaction lock and 2,000/2,000 without one.
- A later diagnostic rolling-trace run passed 5,000/5,000 unlocked
  authentications. That trace instrumentation was removed afterward; this
  result is not represented as a clean-image stress run.
- An earlier standalone run failed at authentication 693 with a two-byte
  continuation response. Its status bytes were not retained and the trace had
  already filled. Later passes do not establish the cause of that event.
- A PM5 rapid connect/select/read experiment failed at cycle 13 with PC/SC
  status `0x80100016`. It is not counted as a passing stress test, nor claimed
  fixed by the USB receive change. The final RF checks above are functional
  checks, not equivalent long-duration reliability coverage.

### Memory ownership and cleanup

- Relocating idle working state recovered **2,036 bytes** of shared static RAM.
- Exercised repeated emulator/standalone initialization, teardown and reuse by
  unrelated commands. Session state is cleared and its BigBuf workspace released.
- RDV4 returned to the full **32,904-byte** shared pool after image reuse;
  a recorded standalone run used a maximum stack of 4,472/7,160 bytes.
- Final PM5 checks reported **433,296 bytes total / 425,104 bytes available**
  after exit: the difference is exactly the intentionally retained 8,192-byte
  emulator image, not a lingering DFC workspace. Ping and status remained usable.
- These are observed allocation/cleanup checks, not a claim that every possible
  application/file combination fits into the retained serialized-image buffer.

### Core integration checks

- Both selectable crypto-backend host suites passed.
- Against the exact core source vendored for firmware, both mbedTLS and tiny
  backends passed 86/86 native vectors, all 36 rejection cases and all three
  format pairs, including internal-state assertions.
- Reduced-profile checks also passed: minimal EV1 59 supported / 27 omitted;
  full EV1 63 / 23; full EV2 76 / 10; minimal EV3 61 / 25. Those omissions are
  declared feature exclusions in separate profile tests, not skipped cases in
  the full PM5 run.
- Eleven focused core tests passed AddressSanitizer/UndefinedBehaviorSanitizer,
  including chaining limits, retransmission and resets.
- Factory initialization and legacy starter-card behavior were checked after
  reusing the factory helper, including the tiny backend.
- The vendored source and revision match public dfc-core 1.0.1 commit
  `52336109601ed29e43b9573ccbf092930f2d56e0`. The source-sync check passes.
- The CID chaining regression failed against the previous implementation and
  passed after the fix. Focused tests also verify CID matching/echo and rejection
  of unadvertised NAD. Both crypto-backend suites and reduced profiles passed
  after those changes. These core tests live in dfc-core.

## Review scope

The local series keeps the shared PM3/PM5 USB receive fix in one commit and the
CMake compression fix in another, separate from core vendoring, firmware
integration, client workflows and standalone support. All six existing commits
have valid CinderSocket signatures and the requested contribution attribution.

The series is six linear commits on `a528d5ac1` (upstream/master at review):
USB transport, core vendoring, firmware integration, client workflows, standalone,
and CMake compression. Follow-up changes were folded into their owning commits.
The consolidation preserved the complete tree before the upstream rebase, which
completed without conflicts. No diagnostic firmware injections are included.


iceman1001:
Interesting,  and still as in last spring I mentioned gigantic Pull requests is not good. 

Things has changed and I have now started with implementing a desfire simulation myself.   This PR will not be merged.

PR #3637, opened 2026-09-16
hf mfdes etest: drive the DESFire simulation from the host, with injectable randomness

mistial-dev:
## Problem

The DESFire simulation (`hf mfdes sim`) can only be exercised over the air, and its card-side random values are fixed: RndB is a constant and the random UID comes from the tick counter. A conformance suite that has recorded a genuine card's session (RndB pinned by decrypting the card's first authentication frame with the test key) cannot replay that session against the simulation, because there is no way to hand it the same RndB and no way to reach it without a reader in the loop.

Two symptoms while wiring that up:

- `hf mfdes eload` sends its `CMD_HF_MIFARE_EML_MEMSET` chunks without acknowledgement, and a small command packet sent right behind that burst was lost about 1 in 24 times on an RDV4 (the client then times out). This is the receive stall already diagnosed in #3628: a zero-length OUT packet is never acknowledged, so a receive bank stays selected and data in the other bank stalls; a command already buffered by `usb_read_ng` is also not noticed until another packet arrives.
- With a fixed RndB the simulator's first D40 authentication frame can be compared with silicon, but nothing downstream can be, and the random-UID activation path cannot be tested at all.

## Change

Four commits, smallest first.

**`common_arm/usb/usb_cdc_at91.c`, `usb_cdc_at32.c`** (author @cindersocket, cherry-picked from #3628 commit 495bc201d, unchanged): `usb_poll_validate_length()` reports a command already buffered by `usb_read_ng`, and acknowledges a zero-length OUT packet by reusing each platform's receive cleanup. Frame parsing is untouched. Kept as its own commit so it can be reviewed and reverted independently.

**`armsrc/desfiresim.c`, `desfiresim.h`, `appmain.c`, `include/desfire.h`, `include/pm3_cmd.h`**: `CMD_HF_DESFIRE_SIM_TEST` (0x0733) carries one operation per packet into the existing simulator state machine, no FPGA involved: BEGIN, SCAN, APDU, RANDOM, FIELDOFF, STATE, END. SCAN runs the same path RF activation does (`desfire_sim_init()`, `desfire_sim_reset()`, random id draw, `desfire_sim_set_fsd()`), and APDU goes through `desfire_sim_apdu()`, so native and ISO 7816 wrapped commands answer exactly as they do on air. Randomness goes through a 64-byte host-fed queue: RndB and the random id draw from it first and fall back to the existing constant or tick counter when it is short, raising a sticky underflow flag. `SimulateDesfireTag()` clears the queue on entry so a host session cannot leak into a reader session. The op enum and wire structs live in `include/desfire.h`, shared by ARM and client.

**`client/src/cmdhfmfdes.c`**: one command, `hf mfdes etest`, one action per call (`--begin`, `--scan`, `--apdu <hex>`, `--random <hex>`, `--fieldoff`, `--state`, `--end`), text output by default and `-j` for one line of JSON a script can read back. Reply lengths are checked before use; failures are reported distinctly (no image, not activated, queue full, no answer). `doc/commands.md` / `doc/commands.json` regenerated, with only the `etest` hunk kept (regenerating on an arm64 Mac also rewrites unrelated `--ie NEON` help, which is not this PR's business).

**`CHANGELOG.md`, `doc/desfire.md`**: two changelog lines and a short "How to drive the simulation from the host" section in the file's one-line-per-command style.

Why a host hook rather than a second reader: the conformance vectors are byte-exact captures from genuine EV1 4K silicon (PC/SC). Replaying them through USB removes RF timing and reader firmware from the comparison, so a mismatch is the simulator's and nothing else's. The vectors and the runner stay in their own repository; this PR only adds the hooks.

## Behaviour change

- `hf mfdes sim` behaviour is unchanged when nothing has been queued: same fixed RndB, same tick-derived random id. If a host queued bytes with `etest --random` and never ran `--end`, `sim` now discards them on start instead of using them on air.
- New device command 0x0733 and new client command `hf mfdes etest`. A firmware built with `SKIP_DESFIRE_SIM=1`, or older than this change, does not answer it; the client reports that after a 2 s timeout.
- USB: a transfer that ends on an exact endpoint-size boundary no longer stalls the next command, and a command that arrived in the same USB packet as the previous one is processed without waiting for another packet. Observable only as commands that used to time out now answering.

## Testing

Host: macOS 15 / arm64, Apple clang toolchain for the client, arm-none-eabi-gcc 13 for firmware.

- `make -j all` with `PLATFORM=PM3RDV4`: zero compiler warnings (the only linker note is the pre-existing macOS `-bind_at_load` deprecation).
- `make -j PLATFORM=PM5 fullimage`: zero warnings.
- `make -j client CC=clang CXX=clang++ LD=clang++`: zero warnings.
- `make check`: green.
- astyle with the project flags over the touched `.c`/`.h`: no changes.
- On an RDV4 (flashed with this firmware): 24 consecutive `eload` then `etest --begin` cycles with no lost reply, where the same loop lost 1 in 24 before the USB commit. `hf mfdes etest --scan`, `--random`, `--apdu`, `--state`, `--fieldoff`, `--end` exercised by hand, including the error paths (APDU before `--scan`, no image, overlong `--random`). The first D40 authentication frame answered with `--random` matches the byte string captured from genuine DESFire EV1 4K silicon.
- The external conformance suite (86 vectors, EV1/D40/EV2/EV3) runs end to end through `hf mfdes etest -j`: 22 pass, 33 fail, 31 skipped as out of scope for an EV1 simulator. The failures are simulator conformance findings (D40 final-frame encryption, ISO 7816 SELECT, duplicate DF-name checks, a handful of status codes) and are not touched here; they are the follow-up work this hook exists to find.

Not tested here:

- Linux and Windows (ProxSpace/WSL) hosts. The client change is plain `cliparser`/`PrintAndLogEx` code with no OS-specific paths; the USB change is device side.
- PM5 hardware. The AT32 half of the USB commit compiled but only the AT91 half ran on a device. It is the same fix per platform and was reported tested on both in #3628.
- `SKIPREADLINE=1`, CMake client, `experimental_lib`. No source files were added or removed, so no build file changed.
- `hf mfdes sim` over the air after this change. The RndB path is unchanged when nothing is queued, but a real-reader authentication would confirm it.

Net diff: 12 files, +542/-11.

## PR checklist

- [x] `client/Makefile`, `client/CMakeLists.txt`, and `client/experimental_lib/CMakeLists.txt` all updated if client sources changed (no files added, nothing to update)
- [x] ARM firmware builds clean for RDV4 and PM5 platform configs
- [ ] Platform-sensitive code reasoned about / tested on Linux, Windows (ProxSpace/WSL), and macOS: macOS tested, Linux and Windows reasoned about only (see above)
- [x] `doc/commands.md` / `doc/commands.json` regenerated if commands changed

Co-Authored-By: Claude Fable 5.1 (claude-fable-5-1) <noreply@anthropic.com>


doegox:
already merged. Damn these stacked PRs...

Issue #3466, opened 2026-08-14
hf mfu info can return with no output at all (silent MFU_TT_UL_ERROR)


mcfly777:
Suggested fixes:
- re-enable the commented-out message so the wrong-tag-type case is not silent
- print something at NORMAL level when the select fails
- optionally add a retry loop in `ul_select()` analogous to `cmdhf14a.c`

**To Reproduce**

1. Use a tag with weak coupling (e.g. an NTAG216 implant) or, for the wrong-type path, a MIFARE Classic card (SAK 08).
2. Run `hf 14a reader -@` — the tag is read reliably.
3. Press Enter to stop, leave the tag in exactly the same position.
4. Run `hf mfu info`.
5. The command returns immediately with no output whatsoever.
6. Run `data setdebugmode -1` and repeat step 4 — now the select failure becomes visible.

**Expected behavior**

`hf mfu info` should print a message explaining why it stopped — either "tag is not an Ultralight/NTAG" or "select failed" — instead of returning silently.

**Screenshots**

n/a

**Desktop**

- OS: macOS 27.0 (Apple Silicon)
- Client: Iceman/proxmark5/v4.9237-13102-g6468d2f1d
- Hardware: Proxmark5 (PM5), built from the `proxmark5` branch of `xianglin1998/proxmark3`. Note: the `cmdhfmfu.c` code path in question is unchanged from RfidResearchGroup master.

[usb] pm3 --> hw status
[#] Memory
[#]   BigBuf_size............. 498028
[#]   Available memory........ 495720
[#] Tracing
[#]   tracing ................ 1
[#]   traceLen ............... 1044
[#] Current FPGA image
[#]   mode.................... All-In-One
[#] Flash memory
[#]   Baudrate................ 24 MHz
[#]   Init.................... ok
[#]   Mfr ID / Dev ID......... EF / 14
[#]   JEDEC Mfr ID / Dev ID... EF / 4015
[#]   Memory size............. 2048 Kb ( 32 pages * 64k )
[#]   Unique ID (be).......... 0xE6656038E31B9423
[#] LF Sampling config
[#]   [q] divisor............. 95 ( 125.00 kHz )
[#]   [b] bits per sample..... 8
[#]   [d] decimation.......... 1
[#]   [a] averaging........... yes
[#]   [t] trigger threshold... 0
[#]   [s] samples to skip..... 0 
[#] 
[#] LF T55XX config
[...]

PR #3644, opened 2026-09-18
BigBuf: widen BIGBUF_ALIGN_MASK to 32-bit for PM5

Axi-ohm:
_Supersedes #3643, which could not be reopened after its branch was updated._

One-line, firmware-only fix (no client files touched). Base `master`. Built for PM5 and PM3RDV4,
and **confirmed on PM5 hardware** — see Testing.

---

## Problem

Commit `5b90940` (2026-09-14, "BigBuf: stop narrowing sizes") widened `BigBuf_max_traceLen()`,
`BigBuf_malloc()` and `BigBuf_calloc()` to `uint32_t` so a device with more than 64 KB of BigBuf
would stop having its buffer sizes truncated. On a PM5, BigBuf is ~498 KB (`hw status` reports
`BigBuf_size 498028`, see issue #3466), which is exactly the case that commit set out to fix.

The widening is incomplete. Every size still passes through:

```c
#define BIGBUF_ALIGN_MASK  (0xFFFF + 1 - BIGBUF_ALIGN_BYTES)   // = 0x0000FFFC
```

`0x0000FFFC` is a 16-bit mask, so the top bits are dropped the moment a size is aligned through it.
The uint32_t return types are correct but they carry an already-truncated value:

| call | today (broken mask) | should be |
|------|--------------------:|----------:|
| `BigBuf_max_traceLen()` with `s_bigbuf_hi = 498028` | `498028 & 0xFFFC` = **39276** | 498028 |
| `BigBuf_malloc(70000)` round-up | `70003 & 0xFFFC` = **4464** | 70000 |
| `BigBuf_malloc(65536)` round-up | `65539 & 0xFFFC` = **0** -> returns NULL | 65536 |

So on a PM5 today:

- **LF sampling** is handed ~39 KB of the ~498 KB buffer that is actually free
  (`initSampleBufferEx()`, `SampleLF()`, `doCotagAcquisition()`, `pcf7931.c`, `lfops.c` `ReadTItag`).
- Any `BigBuf_malloc()`/`BigBuf_calloc()` request of 64 KB or more that commit `5b90940` was
  meant to make work still silently returns a short buffer or NULL. The CHANGELOG already
  claims both of these are fixed.

## Change

`armsrc/BigBuf.c` — one line, plus a CHANGELOG entry.

```c
#define BIGBUF_ALIGN_MASK  (~(uint32_t)(BIGBUF_ALIGN_BYTES - 1))   // = 0xFFFFFFFC
```

The standard "align down" mask for a power-of-two `BIGBUF_ALIGN_BYTES`, spanning the whole word.
`(chunksize + BIGBUF_ALIGN_BYTES - 1) & BIGBUF_ALIGN_MASK` is unchanged as the round-up idiom;
only the mask width changes. No call sites change.

## Behaviour change

- **LF sample / trace buffers on any device with BigBuf > 64 KB (i.e. PM5) grow to the real free
  size.** This is the intended effect. Sizes at or below 65532 bytes are bit-for-bit identical
  under the old and new mask, so **AT91 platforms (RDV4/generic, BigBuf 33272) are unaffected.**
- **`BigBuf_malloc()`/`BigBuf_calloc()` requests of 64 KB and up now succeed** (subject to the
  normal free-space guard) instead of wrapping. This is the behaviour `5b90940`'s own guard
  (`chunksize > s_bigbuf_hi`) was written for but never saw, because the wrap happened first.

### Known limitation left in place (not introduced by this PR)

Raising the device-side trace cap means a PM5 can now *record* a trace larger than 64 KB. Two
consumers are still 16-bit and will truncate such a trace; both predate PM5 and are out of scope
for this one-line fix, but should be called out:

- `client/src/cmdtrace.c`: `gs_traceLen` and `download_trace_ex()`'s local `traceLen` are
  `uint16_t`, and the device already sends the length as `uint32_t` in `download_done_t.extra`.
  A trace over 65535 bytes would be truncated on download / display.
- `armsrc/hfsnoop.c`: `HfSniff()` takes `uint16_t *len`, so `hf sniff` stays capped at 64 KB
  (no worse than today — it truncated the same way before this fix).

These are latent limits the truncating mask happened to hide. LF sampling and the `BigBuf_malloc`
paths this PR targets do not go through them. I can follow up with a separate PR widening the
client trace path if wanted.

## Testing

- **ARM firmware build matrix, run as base-vs-fix on both platforms** (arm-none-eabi-gcc 12.2,
  `make -j bootrom fullimage`, from commit `90c41c6d`; `BigBuf.c`/`.h` are byte-identical on today's
  master so this applies unchanged):

  | build | `PLATFORM=PM5` | `PLATFORM=PM3RDV4` |
  |-------|:--------------:|:------------------:|
  | base (unmodified) | exit 0, 0 warnings, elf built | exit 0, 0 warnings, elf built |
  | fix (this PR)     | exit 0, 0 warnings, elf built | exit 0, 0 warnings, elf built |

  All four link `armsrc/obj/fullimage.elf` and `bootrom/obj/bootrom.elf`. Zero compiler warnings in
  every build, so the change adds none. `BigBuf.c` recompiles clean; the RDV4 result confirms the
  AT91 path is unaffected, as the arithmetic predicted.
- **Arithmetic proven in compiled C** (`gcc:13`, `-std=c11 -Wall -Wextra -Werror -O2`): old mask
  `0x0000FFFC`, new mask `0xFFFFFFFC`; the PM5 numbers above corrected; **all sizes 1..65532 identical
  under both masks**; the 4-byte alignment property holds for the new mask across 0..100000.
- **On-device confirmed on a PM5** (AT32F437; `hw status` reports `BigBuf_size 433232`). Built `armsrc`
  `fullimage` with the fix and flashed it with `./pm3-flash-fullimage`; `hw version` shows the patched OS.
  On stock firmware, LF captures were capped at **40016** samples (= `433232 & 0xFFFC`, exactly the
  truncating mask). With the fix, bounded reads return the full requested count well past that ceiling:
  `lf read -s 50000` -> `Got 50000 samples`, `-s 70000` -> `70000`, `-s 100000` -> `100000`. The PM5 now
  uses BigBuf above 64 KB as intended.
- A bare `lf read` (which requests the entire ~433232-sample buffer) overruns the client's default
  command-execution timeout on PM5; bounded `-s N` reads and realtime `lf read -@` are unaffected. This is
  a pre-existing client timeout, not introduced or changed by this firmware fix.
- Client (`make client`) was not built — this PR touches no client files.
- Not tested: Windows/ProxSpace and macOS host builds (firmware object code is host-OS-independent;
  the change is a preprocessor constant).

## PR checklist

- [x] Builds warning-free with gcc (arm-none-eabi 12.2) for both PM5 and PM3RDV4 firmware configs,
      base and fix, exit 0 / 0 warnings / elf linked in all four
- [x] No client build files touched (firmware-only, one define)
- [x] Existing comments preserved; no new comment added
- [x] No generated docs affected (`doc/commands.*` unchanged)
- [x] AI co-author trailer on the commit
- [x] Verified on PM5 hardware: LF buffer now exceeds the old 40016-sample cap (`-s 50000/70000/100000` all honored)


iceman1001:
```
armsrc/hfsnoop.c: HfSniff() takes uint16_t *len, so hf sniff stays capped at 64 KB
(no worse than today — it truncated the same way before this fix).
```

This doesn't depend on bigbuf in that sense.  It depends on the actual extra ram size left over on the fpga.  On an older rdv4, pm3 easy  it was about 3kb.      We don't know how much RAM is available for now.   @xianglin1998

Axi-ohm:
Good points, and agreed on hf sniff — HfSniff() takes uint16_t *len, so it's capped at 64 KB independently of this. Widening that (len + callers to uint32_t) would be a separate follow-up, not part of this PR.

The mask itself is really an allocator correctness bug rather than a RAM question: BigBuf_malloc() and BigBuf_max_traceLen() round a uint32_t size through a 16-bit BIGBUF_ALIGN_MASK (0xFFFC), so any size above 64 KB loses its top bits — regardless of platform or how much RAM is actually present.

It surfaces on PM5 because BigBuf is large. hw status reports BigBuf_size ≈ 433 KB on my unit; with the old mask, BigBuf_max_traceLen() keeps only the low 16 bits (~40 K), and BigBuf_malloc(70000) rounds down to 4464. Hardware-confirmed:

stock: lf read -s 50000 → clamped to 40016 samples
with the fix: lf read -s 50000 → 50000 samples
The demo uses lf read because it goes through the sample buffer / max_traceLen, not the uint16_t-capped sniff path — that's the part the truncation was hitting. @iceman1001

iceman1001:
It is a valid fix.   
I am just correcting one of its assumptions in here so it doesn't get followed up incorrect. 

A bit too much AI comments for the fix,    tell your AI to be less verbose  or just remove the comments and shorten the changelog line to something like:  "fixed the pm5 bigbuff allocation"

PR #3650, opened 2026-09-19
BWM: add ability to flash over wifi

wiillk3:
## Problem

PM5 over BWM Wi-Fi can talk to the OS image, but `--flash` could not. Bootrom only spoke USB: UART4 is one byte with no DMA, so the OS BWM path (DMA / `GetTickCount`) cannot run there. A 512-byte `PacketCommandOLD` also spans multiple `DATA_FORWARD` frames. The client then waited forever for bootloader replies, and `CMD_FINISH_WRITE` always appended into the erase unit, so a retried or split wireless packet landed in the wrong place.

## Change

Shared app_com de-framer moved to `common_arm/bwm/bwm_frame.*` so bootrom and OS use the same unwrap. That is a move, not a rewrite: the parser, CRC, FIFO, and Niel's comments are the existing `bwm_forward.c` code, relocated because bootrom cannot compile the OS file (DMA / `GetTickCount`). OS still owns TX, flow control, and baud in `bwm_forward.c`. Bootrom gets a new polled UART4 path in `bootrom/bwm_boot.*` (that part is new; it cannot share the OS UART). Gated on `PLATFORM_EXTRAS=BWM` / `WITH_BWM_FORWARD` (Makefile and CMake).

Bootrom `CMD_FINISH_WRITE` on a BWM packet places at `arg[2]` (byte offset in the unit) instead of appending. USB still appends. `CMD_DEVICE_INFO` sets `DEVICE_INFO_FLAG_UNDERSTANDS_BWM_STREAM` and reports the probed baud.

Client: timeouts and stale-ACK discard in the flasher; `arg[2]` carried on `CMD_FINISH_WRITE` (old bootroms ignore it); wireless ports refused if the bootrom lacks the BWM flag; `flash_pm3` compiled into libpm3 with a thin `pm3_flash()` next to `pm3_open`; if the image path is already a regular file, do not let `searchFile` append `.elf`.

No new client `.c` files, so Makefile / CMake / experimental_lib lists are unchanged. No new user commands.

## Behaviour change

- `proxmark3 tcp:host:port --flash --image …` can write ARM images on a BWM PM5 after a BWM-capable bootrom is on the device (first bootrom still needs USB).
- Bootloader `CMD_DEVICE_INFO` no longer waits forever: 3×3 s, then an error. Wireless text points at a BWM bootrom; USB stays generic.
- `wait_for_ack` times out at 15 s and drops non-ACK/NACK junk (lost-byte `0x0000`). USB flashing used to wait indefinitely.
- On ACK timeout, a flash block is retried (up to 4). Hard errors still abort.
- libpm3 exports `pm3_flash()`. `flash_pm3` is no longer `static` / CLI-only.
- `flash_pm3 /absolute/path/foo.elf` uses that file as-is.
- BWM bootrom: unknown cmds NACK instead of `Fatal()`. Non-BWM bootrom unchanged, including the 10 ms flash-mode debounce.
- RDV4 / non-BWM builds do not compile `bwm_boot` / `bwm_frame` into bootrom.

## Testing

Net +1001 / −332, 19 files. Most of the `bwm_forward.c` deletion is the de-framer moving to `common_arm/bwm/`, not a rewrite of that logic.

Built and Wi-Fi flashed a PM5 (`PLATFORM=PM5` `PLATFORM_EXTRAS=BWM`) on:

- macOS 26.5
- Windows: WSL 1, Ubuntu 24.04
- Linux: Ubuntu 25.10

Hardware: PM5 + BWM, `tcp:` `--flash` (bootrom + fullimage as used).

RDV4: firmware compiled and device runs fine (USB; BWM is PM5-only).

Clang host client on Linux: warning-free.

Not tested here:

- BLE flashing (out of scope; console-only on BLE for now)
- Native Windows / ProxSpace (WSL only)
- `experimental_lib` / CMake client as a separate build

Path skip for existing files uses `path_is_regular_file()` already used in `fileutils.c` on Windows and POSIX.

## Checklist

- [x] Builds warning-free with gcc (Linux / WSL) and clang (Linux client; macOS host client is clang)
- [x] `client/Makefile`, `client/CMakeLists.txt`, and `client/experimental_lib/CMakeLists.txt` — no new client sources; existing TUs already listed
- [x] ARM firmware built for PM5 + BWM (Wi-Fi flashed) and RDV4 (compiled and runs on hardware)
- [x] Platform-sensitive flash/path code run on macOS, Linux, and Windows (WSL 1). Not ProxSpace
- [x] `doc/commands.md` / `doc/commands.json` — no command list change


xianglin1998:
Personally, I don't think `common_arm` should include too many modules unrelated to "common" functionality; common_arm dir should really be reserved for cases where it's worth implementing a cross-platform HAL and reusing code. Clearly, `bwm` is currently used only for PM5 (and RGB).

wiillk3:
I think that's a good point, adjusted thanks.

TomHarkness:
Epic!

PR #3653, opened 2026-09-21
Improve `hf 15 csetuid` command

0x6r1an0y:
## Problem

`hf 15 csetuid` wrote before knowing what the tag was. On a normal tag the Gen1 sequence overwrote blocks 0x38/0x39/0x3E/0x3F, which is real user data on any tag with 64 blocks or more, and `--v3` wrote the UID blocks without checking for the V3 signature. A failed attempt left the tag half written.

Magic V3 (c2cd37025) was also implemented entirely client side, unlike every other magic family (`CMD_HF_MIFARE_CIDENT`, `CMD_HF_MIFARE_CSETBL`, `CMD_HF_MIFARE_GEN3UID`, `CMD_HF_MIFARE_G4_WRBL`, `CMD_HF_MIFARE_G4_GDM_WRBL`), so the client had to reason about card state it couldn't see.

## Change

Identification, writing and failure handling now run on the device in one field session. `PM3_SUCCESS` means the UID reads back as requested.

- `armsrc/iso15693.c`: `SetTag15693Uid()` probes 0x38/0x39/0x3E/0x3F and refuses to write if any of them answers with data, then verifies the UID and blanks the four blocks again if it didn't change. `SetTag15693Uid_v3()` and `FinalizeTag15693_v3()` are new and check the 0x14/0x15 signature before writing.
- `client/src/cmdhf15.c` (-76 lines): `csetuid` and `cfinalize` only pick a command and print the status. The V3 helpers and constants are deleted, which restores the original read helper with its CALLOC message and the `ISO15_ERROR_HANDLING_*` macros, and drops the TEAROFF handling a magic card doesn't need.
- `include/pm3_cmd.h`, `armsrc/iso15693.h`, `armsrc/appmain.c`: two new commands, `CMD_HF_ISO15693_CSETUID_V3` (0x0334) and `CMD_HF_ISO15693_CFINALIZE_V3` (0x0335). Additions only.

## Behaviour change

- `csetuid` on a normal tag aborts and names the block that held data instead of destroying it. `--v3` and `cfinalize` abort on a tag without the V3 signature.
- `cfinalize` without `-y` no longer energizes the tag and no longer reports configuration mode first, since the check is now on the device. Use `csetuid --v3` to confirm a V3 tag non-destructively, it's repeatable.
- Limitation: `00000000` counts as blank and 0x3F holding `69960000` counts as leftover from an earlier attempt, so a normal tag whose four blocks are all zero is still written to. There is no reliable passive way to identify a Gen1 tag.
- Needs matching firmware, a version mismatch misreports Gen1 status.

## Testing

Built warning-free: client (ProxSpace/MinGW gcc 13.2, `SKIPQT=1`), ARM RDV4, ARM `PLATFORM=PM5`. 
Not tested: clang, Linux, macOS — plain C using helpers already used elsewhere in `armsrc/iso15693.c`.

On RDV4 hardware: V1 tag `csetuid` ok; SLIX2 with data in 0x38 refused, names the block; SLIX2 with all four blocks blank proceeds, write fails, rolled back (the limitation above); V2 tag with `csetuid` fails and leaves the blocks blank; V2 with `--v2` ok; V3 with `--v3` ok; SLIX2 with `--v3` refused.

**`cfinalize` is untested.** It's irreversible and only one un-finalized V3 tag was available, so that path is compile-verified only. It reuses the same probe and write helpers as `csetuid --v3`, which is tested.

## Checklist

- [x] Warning-free with gcc, not tried with clang
- [x] No new client sources, so no build file changes needed
- [x] ARM builds clean for RDV4 and PM5
- [ ] Tested on Windows/ProxSpace only
- [x] New comments short and direct
- [ ] `doc/commands.md` / `doc/commands.json` regenerated


client/src/cmdhf15.c
```diff
@@ -3340,13 +3340,12 @@ static const uint8_t iso15_magic_v3_sig_b[4] = {0x21, 0xAE, 0x93, 0x00};
...
 
-// Read one 4-byte block from a magic V3 tag in unaddressed mode.
-static int hf15_magic_v3_read_blk(uint8_t blockno, uint8_t out[4]) {
+// Unaddressed 4-byte block read, silent because failed probes are expected
+static int hf15_magic_read_blk(uint8_t blockno, uint8_t out[4]) {
 
     uint16_t approxlen = 2 + 1 + 2;
     iso15_raw_cmd_t *packet = (iso15_raw_cmd_t *)calloc(1, sizeof(iso15_raw_cmd_t) + approxlen);
     if (packet == NULL) {
-        PrintAndLogEx(WARNING, "Failed to allocate memory");
```
iceman1001, line 3349: Don't touch CALLOC  messages.

client/src/cmdhf15.c
```diff
@@ -3364,39 +3363,141 @@ static int hf15_magic_v3_read_blk(uint8_t blockno, uint8_t out[4]) {
...
         return PM3_ETIMEOUT;
     }
 
-    ISO15_ERROR_HANDLING_RESPONSE
+    if (resp.status == PM3_ETEAROFF) {
+        return resp.status;
+    }
+    if (resp.length < 2) {
+        return PM3_EWRONGANSWER;
+    }
```
iceman1001: Not sure why we would need to handle  TEAROFF events on a magic card...    
more than code should look alike

client/src/cmdhf15.c
```diff
@@ -3364,39 +3363,141 @@ static int hf15_magic_v3_read_blk(uint8_t blockno, uint8_t out[4]) {
...
-    ISO15_ERROR_HANDLING_CARD_RESPONSE(d, resp.length)
+    if (check_crc(CRC_15693, d, resp.length) == false) {
+        return PM3_ECRC;
+    }
+    if ((d[0] & ISO15_RES_ERROR) == ISO15_RES_ERROR) {
+        if (d[1] == 0x0F || d[1] == 0x10) {
+            return PM3_EOUTOFBOUND;
+        }
+        return PM3_EFAILED;
+    }
```
iceman1001: Doesn't the old macro handle this?

iceman1001:
I don't know about this PR.    All logic should be on device side to handle identification of magic and failed attempts.

0x6r1an0y:
You're right, I'll move the Gen1 logic to the device side.

`SetTag15693Uid()` will do the probe, the four writes, the UID verify and the rollback in one field session, and return a status that the client just prints.

That also undoes all three of your inline comments: the read helper goes back to its original form with the ISO15_ERROR_HANDLING_* macros, the CALLOC message and no TEAROFF handling. I had only changed it because the client side needed a silent probe, which is no longer the case.

One question while I rewrite it: V3 (`hf 15 csetuid --v3` / `hf 15 cfinalize`) is already implemented client side in master, and this PR only adds a signature check on top of it. Do you want V3 moved to the device as well, or should it stay as it is for now?

iceman1001:
move it all.

iceman1001:
changing these values which are sensitive in a sense,  it's better to do it in one session and as close to the source as possible.
doing it on the client side is..  well, easy and lazy but we need to handle so much more because of it since we don't really know the state of the card.   

The idea is if the device reports back `PM3_SUCCESS`   we can assume the card is completely written according our wishes.

0x6r1an0y:
@iceman1001 I have a quick question about the codebase architecture to ensure I am on the right track. How exactly do we divide tasks between the device side and the client side? My understanding is that the device side should be kept as minimal as possible, and we only put operations there if they have strict timing requirements. Is this correct?

(got reply on discord. thanks)

iceman1001:
yes and no.  
Since proxmark3  has limited resources we had to divide it.  
With proxmark5  we don't have those limits yet.   Follow established code patterns and you should be able to fit in much more.

But if it's focused for Proxmark3 it is a problem.   We try to make good efficient source for device side.  

For client side we don't care at all.   Just don't break stack limits on MacOS

PR #3654, opened 2026-09-21
Add `analyse card`, identify an unreadable card as LF or HF

nemanjan00:
Identifies a card that won't answer `lf search` / `hf search` as LF, HF or
dual-frequency, and measures its LF coil resonance. A dead chip still has a
coil, and the coil loads the antenna: sweep the LF response clear, sweep it
again with the card on, subtract, and the card's resonance is left as a notch.

Adds `CMD_MEASURE_ANTENNA_TUNING_SWEEP` — full mV instead of `mV >> 9`,
selectable band, oversampling. The existing 0.5 V quantization buried the
difference. HF stays single-point; `ck_1356meg` is a fixed oscillator pin.

A card must clear three criteria: notch depth, localization, and reactive
lift (voltage pushed *above* baseline — a resonant tank returns energy, metal
only absorbs). Measured on a 25 V antenna:

|               | depth       | local.    | lift       |
|---------------|-------------|-----------|------------|
| empty         | 47–69 mV    | 1.9–2.7   | 58–81 mV   |
| bag of screws | 220–244 mV  | 3.3–3.4   | 58 mV      |
| real cards    | 808–2079 mV | 16.9–29.1 | 302–604 mV |

Resonance is interpolated through the notch (divisor grid is ~1.3 kHz at
125 kHz); 0.5 kHz repeatability. `--live` re-sweeps a narrow band for coil
tuning.

Tested on PM5: empty ×2, screws ×2, LF card, HF card, dual card — all correct,
≥2× margin on every threshold. LF only; HF loaded-Q needs
`MeasureAntennaTuningHfDecay` ported past its `#ifndef PM5`.


iceman1001:
LF is easier to handle since card usually starts speaking when entering a reader field  (TTF). 

HF tags waits for a reader command usually (RTF)    you would only see a power drop is something HF is in the reader field.

nemanjan00:
Just putting metal into field changes measured voltage, because it detunes antenna.

And especially cool part, on LF, changing frequency changes the V, because, while you sweep frequency you encounter freq LF coil is resonant on and you see a peak.

The problematic thing about HF is you can measure field changed, but, not really in what way. Since you can not sweep. But card had a property a hunk of metal does not, its current draw changes over time. I will try to play with that too.

iceman1001:
you can make a simple "send the exact first command"  in each of our known HF protocols and see if tag answers....

nemanjan00:
> you can make a simple "send the exact first command" in each of our known HF protocols and see if tag answers....

Is that not already in auto? I wanted to build something based only on physics, for identification of fully unknown cards

iceman1001:
I think `auto` and `hf search`   does a full anti collision

PR #3657, opened 2026-09-21
fix: preserve UTF-8 in generated command docs

0x6r1an0y:
## Problem
On MinGW, the client emits CP850 while `make commands` writes UTF-8 documentation, corrupting characters such as `µ`. Several Hitag strings also used escaped UTF-8 Greek mu bytes, creating mixed-encoding output.

螢幕擷取畫面 2026-09-22 032124
螢幕擷取畫面 2026-09-22 032113

## Change
Convert MinGW client output from CP850 to UTF-8, force the JSON generator to read UTF-8, and replace remaining `\xce\xbc` strings with `µ`. Other platforms pass output through unchanged.

## Behaviour change
`make commands` now generates valid UTF-8 Markdown and JSON on MinGW. Hitag descriptions consistently use `µ` (U+00B5).

## Testing

Environment:

- Windows (`MINGW64_NT-10.0-26100`), x86_64
- MSYS 3.4.9
- GCC 13.2.0 (`cc.exe`, MSYS2 Rev2)
- Python 3.11.5
- Preferred locale encoding: CP950
- Python stdin/stdout encoding: UTF-8

Tested:

- Ran `make commands` under ProxSpace/MSYS2.
- Confirmed the client rebuilt successfully with GCC and no new warnings.
- Confirmed `doc/commands.md` and `doc/commands.json` decode as UTF-8.
- Confirmed generated output contains the correct micro sign (`µ`) and none of the observed corrupted forms (`�`).

Not tested:

- Full `make clean && make all`.
- Clang build.
- Linux and macOS execution. Their documentation pipeline remains unchanged apart from a pass-through `cat`.
- CMake and `experimental_lib`; no source files, dependencies, or build lists were added or removed.
- ARM firmware targets; no firmware code was changed.
- Full offline test suite.

## Checklist
- [x] Proper style; no unrelated reformatting
- [ ] Full GCC and Clang build matrix
- [x] No client build-list changes required
- [x] No ARM-side changes
- [x] Windows/ProxSpace tested
- [x] Generated command documentation verified

Co-Authored-By: OpenAI Codex [noreply@openai.com](mailto:noreply@openai.com)


Makefile
```diff
@@ -354,13 +354,20 @@ style: commands
 		--style=google --unpad-paren --pad-header \
 		--align-pointer=name {} \;
 
+COMMANDS_UTF8_FILTER := cat
+COMMANDS_PYTHON := python3
```
iceman1001: Don't we have a DEFINE for python3 somewhere else already in the makefiles?

iceman1001:
This has been an issue for many years.

Makefile
0x6r1an0y: I checked the makefiles. The only top-level definition I found is PYTHON3_PKGCONFIG in Makefile.defs, but that is the pkg-config package name used to locate Python headers/libraries, not the Python executable. The commands target currently invokes python3 directly. Did you have another variable in mind?

iceman1001: It should be added to the `Makefile.defs` instead of here.

0x6r1an0y: done

